PR #17, opened 2017-09-20
Minor fix for rewriter

dsgouda:
Should have checked if the original node is of QualifierNameSyntax type too

### DIFF
--- a/src/simplifier/AddUsingsRewriter.cs
+++ b/src/simplifier/AddUsingsRewriter.cs
@@ -36,7 +36,7 @@ namespace AutoRest.Simplify
             );
             
 
-            if(node.Name!=null)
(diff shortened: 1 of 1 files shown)


src/simplifier/AddUsingsRewriter.cs
```diff
@@ -36,7 +36,7 @@ public override SyntaxNode VisitNamespaceDeclaration(NamespaceDeclarationSyntax
             );
             
 
-            if(node.Name!=null)
+            if(node.Name!=null && node.Name.GetType().Name=="QualifiedNameSyntax")
```
olydis: what about `node.Name?.GetType() == typeof(QualifiedNameSyntax)`? Little cleaner, no magic string constants involved :)

dsgouda: Huh, why did I not think of that. I might as well clean up some more code then!

PR #19, opened 2017-09-22
Adding Extensible enums support to generated code

dsgouda:
Addressing https://github.com/Azure/autorest/issues/1591
- Adding extensible enums support
- Is activated only when `extensible-enums` is passed as an argument to the plugin
- Test case validates compilation
- Will be helpful if we get a usage example from @brjohnstmsft / @mattchr so we can create appropriate usage test cases here
- Code template is derived from example [here](https://github.com/Azure/azure-sdk-for-net/blob/AutoRest/src/Search/Microsoft.Azure.Search/Customizations/Indexes/Models/DataType.cs)


src/vanilla/Templates/Rest/Common/EnumTemplate.cshtml
```diff
@@ -9,7 +9,51 @@ namespace @(Settings.Namespace).@(Settings.ModelsName)
...
+        @foreach (var t in Model.Values)
+        {
+            if (t.Description != null)
+            {<text>
+        /// <summary>
+        @WrapComment("/// ", t.Description)
+        /// </summary>
+        </text>
+        }
+        string ctor = "new "+Model.ClassName+"(\""+t.Name+"\")";
```
olydis: not used anywhere

src/vanilla/Templates/Rest/Common/EnumTemplate.cshtml
```diff
@@ -9,7 +9,51 @@ namespace @(Settings.Namespace).@(Settings.ModelsName)
...
+        @:public static readonly @Model.ClassName @t.Name = new @(Model.ClassName)("@(t.Name)");
```
olydis: this can't really use `new` or `new` has to be smarter: If you have a predefined "red" value here, which is `new`ed, then this will not end up in the `Dictionary` of the base class (correct?). Hence, if later someone does `Create`/`Lookup` with "red", you get a *new* instance that is not referentially equal to the predefined one

src/vanilla/Templates/Rest/Common/EnumTemplate.cshtml
```diff
@@ -9,7 +9,50 @@ namespace @(Settings.Namespace).@(Settings.ModelsName)
...
+        @foreach (var t in Model.Values)
+        {
+            if (t.Description != null)
+            {<text>
+        /// <summary>
+        @WrapComment("/// ", t.Description)
+        /// </summary>
+        </text>
+        }
+        @:public static readonly @Model.ClassName @t.Name = new @(Model.ClassName)("@(t.Name)");
```
olydis: not sure about `Name` here, but you can't have the same for key and value. If you look further down you'll see
```
t.SerializedName
t.MemberName
```
getting used.

src/vanilla/Templates/Rest/Common/EnumTemplate.cshtml
```diff
@@ -9,7 +9,50 @@ namespace @(Settings.Namespace).@(Settings.ModelsName)
...
+    }
+    @EmptyLine
+        @:private @(Model.ClassName)(string typeName) : base(typeName)
```
olydis: while we're at it, and since this is actually a highly potential data plane candidate and on our TODO list: we should also have support for non-string enums

brjohnstmsft:
@dsgouda Here is a usage example from the Search SDK: https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SDKs/Search/DataPlane/Microsoft.Azure.Search/Customizations/Indexes/Models/AnalyzerName.cs

It is the most complex example we have. Note that the values are exposed as separate constants rather than hard-coded in the constructor invocations. This is so we can use them in other places like custom attributes, where parameter values have to be compile-time constants. Please make sure the generated code follows this same pattern.

src/vanilla/Templates/Rest/Common/EnumTemplate.cshtml
dsgouda: This is a static readonly field, I guess `CreateValueMap` would populate this particular field as expected.
However, I do see that we should be adding the new type to our map in `Create`, which also makes me think that it maybe should be a part of the base class itself.
Thoughts?

brjohnstmsft: The map should not be modified after it is created. The goal is to "intern" known values, not necessarily every value that is passed to `Create`.

Please refer to the usage example I gave in my comment below (it's [here](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SDKs/Search/DataPlane/Microsoft.Azure.Search/Customizations/Indexes/Models/AnalyzerName.cs)).

olydis:
please also add a `Sample` so people can see the usage and outcome at a glance. That sample's Swagger would ideally contain 2 enums, one with `modelAsString` true, on with false; then, it would once be generated with the `extensible-enums`, once without. That way, we showcase the full matrix of what is possible and we'll detect regressions in any direction.

olydis:
# publish job
~~~ Haskell
> waiting for PR to get merged or closed
~~~

olydis:
# publish job
~~~ Haskell
> waiting for PR to get merged or closed
~~~

olydis:
# AutoRest automatic publish job
~~~ Haskell
> will publish once/if PR gets merged
~~~

olydis:
# 🤖 AutoRest automatic publish job 🤖
~~~ Haskell
> will publish once PR gets merged
~~~

olydis:
# 🤖 AutoRest automatic publish job 🤖
~~~ Haskell
> will publish once PR gets merged
~~~

olydis:
# 🤖 AutoRest automatic publish job 🤖
~~~ Haskell
> will publish once PR gets merged
~~~

src/vanilla/Templates/Rest/Common/EnumTemplate.cshtml
```diff
@@ -9,7 +9,52 @@ namespace @(Settings.Namespace).@(Settings.ModelsName)
     /// <summary>
     @WrapComment("/// ", "Defines values for " + Model.Name + ".")
     /// </summary>
-    @if (!Model.ModelAsString)
+    @if(Settings.CustomSettings.ContainsKey("ExtensibleEnums") && (bool)Settings.CustomSettings["ExtensibleEnums"])
+    {
+    var underlyingType = (Model.UnderlyingType == null)? "string" : Model.UnderlyingType.ToString();
```
amarzavery: @dsgouda - The fix in this [PR]((https://github.com/Azure/autorest.modeler/pull/22/files#diff-1f2c985b9819bf215e2376529fb1ba78R63)) should help you with underlyingType.

dsgouda: Yup, I realized that, and I am kinda waiting on it.

amarzavery: It is merged. You should be good to go with a dependency on newer version of the modeler

olydis:
# 🤖 AutoRest automatic publish job 🤖
~~~ Haskell
> will publish once PR gets merged
~~~

src/autorest.csharp.csproj
```diff
@@ -65,9 +65,8 @@
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="2.3.1" />
     <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.4" />
     <PackageReference Include="Microsoft.Rest.ClientRuntime.Azure" Version="3.3.4" />
-    <!--<PackageReference Include="autorest.common" Version="2.2.27" />-->
```
fearthecowboy: You don't want to commit this change here!

test/vanilla/Swagger/swagger-extensible-enums.json
```diff
@@ -0,0 +1,115 @@
...
+     "x-ms-enum": {
+      "modelAsString": true,
+      "name": "IntEnum",
+      "values": [
+       {
+        "value": "1",
+        "description": "one",
+        "name": "1",
+        "allowedValues": [
+         "1.1",
```
fearthecowboy: not generating this yet?

src/vanilla/Model/CompositeTypeCs.cs
```diff
@@ -159,13 +159,29 @@ public string Signature
                     foreach (PropertyCs property in Parameters.Where(p => !p.UnderlyingProperty.IsConstant).Select(p => p.UnderlyingProperty))
                     {
                         string format = (property.IsRequired ? "{0} {1}" : "{0} {1} = default({0})");
```
olydis: why does this change so much? I'd have expected signatures to look identical to what we did for `modelAsString: false` before

test/azure/Expected/AcceptanceTests/AzureCompositeModelClient/IPolymorphismOperations.cs
```diff
@@ -92,6 +92,42 @@ public partial interface IPolymorphismOperations
         /// </exception>
         Task<AzureOperationResponse> PutValidWithHttpMessagesAsync(Fish complexBody, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
         /// <summary>
+        /// Get complex types that are polymorphic, but not at the root of the
```
olydis: why are my testcases to polymorphic discrimination in here? :o

test/vanilla/Expected/AcceptanceTests/ExtensibleEnums/Models/DaysOfWeekExtensibleEnum.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        /// Create method used by deserializer to create instances of DaysOfWeekExtensibleEnum type
+        /// </summary>
+        /// <returns>The value as a DaysOfWeekExtensibleEnum.</returns>
+        public static DaysOfWeekExtensibleEnum Create(string value)
+        {
+            if(value == null)
+            {
+                return null;
```
olydis: why?

dsgouda: Would it make sense to generate an ExtensibleEnum object for a null underlying value?

fearthecowboy: Because `null ` indicates the absence of a value, not a distinct value.

If `null` is a possible value, then the field isn't required.

src/vanilla/Model/CompositeTypeCs.cs
dsgouda: For extensible enums we are generating the parameter type based on the information in `EnumType`, `property.ModelTypeName` does not match up with the name of the extensilbe enum class generated hence the change.
Also, we are using extensible enums only if modelAsString is also set to true

test/vanilla/Expected/AcceptanceTests/ExtensibleEnums/Models/DaysOfWeekExtensibleEnum.cs
dsgouda: Yup, `x-nullable` would be the right thing to do

olydis: Right, so having this check essentially makes the assumption that stuff is always nullable (whether via `required: false` or `x-nullable: true`).
Furthermore, when `null` actually *is* one of the enum values (perfectly valid Swagger), I'd somewhat expect it to end up in an `ExtensibleEnum` instance. In the *absence* of a value on the other hand, why would one even call `Create`? This check essentially shuts the door on an (arguably not common so far?) use case *but* at not visible (to me) benefit.

brjohnstmsft: I'm with @olydis on this one. `ExtensibleEnum` is a class, so it is already possible to have a `null` instance. Why allow `null` for `value` at all?

olydis:
# 🤖 AutoRest automatic publish job 🤖
~~~ Haskell
> will publish once PR gets merged
~~~

dsgouda:
Closing in favor of https://github.com/Azure/autorest.csharp/pull/45

PR #22, opened 2017-09-30
Made ExceptionTemplate use a partial class so exceptions can be extended

csuich2:
The model classes inside of a generated exception are already partial which is helpful for extending the error model, but the exception itself isn't partial. I've update ExceptionTemplate.cshtml to be a partial class so that exceptions can be extended.

For example, in my use case, I'd like to extend the exception's ToString() to print out a user friendly version of the error.

I tried looking through the tests to see how I could write a test for this, but couldn't find an existing analogous test. If a test is required for this change, can someone point me in the direction of how such a test would be written? For example, do we want an example that adds a partial exception and want to confirm that it compiles as expected, or is there some test that can check whether the class definition itself is marked as partial.


csuich2:
I'm going to keep researching the tests and will open another PR later.

PR #23, opened 2017-10-02
Chsuich/partial exceptions

csuich2:
The model classes inside of a generated exception are already partial which is helpful for extending the error model, but the exception itself isn't partial. I've update ExceptionTemplate.cshtml to be a partial class so that exceptions can be extended.

For example, in my use case, I'd like to extend the exception's ToString() to print out a user friendly version of the error.

So, I have:
- Update the template to be partial
- Regenerated the tests so that all the generated exceptions are now partial
- Added a partial implementation of the test/vanilla/AcceptanceTests/Http/Models/ErrorException.cs which simply adds a const string
- Added a test to AcceptanceTests.cs which references the const string so that compilation of the acceptance test will fail if the partials can't be compiled


olydis:
# publish job

PR #24, opened 2017-10-02
fix weird overwrite of partially constant properties

olydis:
The removed part is pretty much guaranteed to not make sense: It emits initialization for properties that are `IsRequired` (ignore all other filters). But required stuff is also provided via constructor (as required arg), so why do anything else than take what was passed to the constructor?


olydis:
# publish job

PR #25, opened 2017-10-03
Added 'ExcludeFromInterface' flag to totally hide methods

csuich2:
The 'hidden' flag in the code model is useful for hiding the default implementation of a method, but in our case, we'd like to the default method signature to not show up in the interface and to not create extension methods with that signature. Instead, we'd like there to be an internal default implementation and then we can create our own set of methods in the interface and extensions with the method signature of our choosing.

So, I have:
- Added an 'ExcludeFromInterface' code model field for methods
- Created a test configuration file (so that it is easier to have a directive which adds the $.hidden $.excludeFromInterface fields
- Updated regeneration.iced to regenerate this test config based on the configuration markdown file
- Added an AcceptanceTest which checks to make sure that the default method doesn't appear in the operation interface and invokes the custom signature extension method

Open questions:
- Should setting ExcludeFromInterface=true inherently set Hidden=true as well?
- Is this setup with the configuration file OK?


test/vanilla/Configurations/hidden-methods.md
```diff
@@ -0,0 +1,24 @@
...
+base-folder: ../../../
+input-file: node_modules/@microsoft.azure/autorest.testserver/swagger/body-complex.json
+payload-flattening-threshold: 1
+clear-output-folder: true
+directive:
+  - from: code-model-v1
+    where: $.operations[*].methods[?(@.serializedName == 'basic_putValid')]
+    transform: >-
+      $.hidden = true;
+      $.excludeFromInterface = true;
```
fearthecowboy: The fact that you added the member to the language-specific implementation `Method` (ie, `MethodCs.cs` ) and then simply used a `directive: transform`  to enable it is exactly how we wanted to approach this type of feature.

image

src/vanilla/Templates/Rest/Client/MethodGroupInterfaceTemplate.cshtml
```diff
@@ -19,6 +19,7 @@ namespace @Settings.Namespace
     {
     @foreach(MethodCs method in Model.Methods)
     {
+        if (method.ExcludeFromInterface) continue;
```
fearthecowboy: Quick nit: can you change the use of if to use the `{ }`  so that it fits in our style? (and the other uses)
```  csharp
if (method.ExcludeFromInterface) 
{ 
   continue;
}
```

test/vanilla/Configurations/hidden-methods.md
csuich2: Excellent! I'm glad this was the right approach.

src/vanilla/Templates/Rest/Client/MethodGroupInterfaceTemplate.cshtml
csuich2: Done.

fearthecowboy:
@csuich2 pull the latest and re-run `regenerate` to clear the conflicts...

fearthecowboy:
@csuich2  -- FYI, @olydis  is nearly done a feature for being able to declare new `directive`s in configuration so instead of having to use a `transform`, you can expose the feature with some really nice syntax.

csuich2:
Excellent. I'd be happy to open another PR to switch to that syntax once it is available.

olydis:
# publish job
~~~ Haskell
> waiting for PR to get merged or closed
~~~

olydis:
# publish job
~~~ Haskell
> waiting for PR to get merged or closed
~~~

olydis:
@fearthecowboy Did you wanna make this a declared directive or take it in as is?

fearthecowboy:
the declared directive stuff is merged right? Might as well tweak this to add it in now. It'd serve as a nice example.

olydis:
@fearthecowboy whoops just updated original comment

olydis:
Note that we also already have an example of a completely self-contained directive with method removal. It's also defined in terms of other declared directive, demonstrating the allowed recursiveness.

PR #36, opened 2017-10-24
Updated directive for operation hiding to use new directive format

csuich2:
I thought about adding this as a stock directive for CSharp, but it looks like CSharp doesn't have its own default configuration file yet. Would y'all prefer to see this as a canned directive for CSharp?

### DIFF
--- a/test/vanilla/Configurations/hidden-methods.md
+++ b/test/vanilla/Configurations/hidden-methods.md
@@ -11,7 +11,7 @@ payload-flattening-threshold: 1
 clear-output-folder: true
 directive:
   - from: code-model-v1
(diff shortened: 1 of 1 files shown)


olydis:
# 🤖 AutoRest automatic publish job 🤖
~~~ Haskell
> will publish once PR gets merged
~~~

olydis:
@csuich2 for canned responses, I'm always looking for demand/volume - I think it's okay for now, at least there are some other candidate directives I see used more often than method hiding. Note however that `readme.md` of the C# extension is what get's loaded, so directives can be declared in there. 😉

PR #90, opened 2018-04-06
Resource / Inner / Enum updates for c# fluent generator

jianghaolu:
Had to declare a few things virtual in the vanilla gen to make it work.


weshaggard:
@jianghaolu I assume it is safe to close this PR now as we don't need it any longer. If that isn't the case please reopen.

PR #91, opened 2018-04-06
Priortizing customization layer to initialize ContractsResolver

shahabhijeet:
Allowing customization layer to initialized ContractResolvers before a default ContractResolver is initialized.

### DIFF
--- a/src/vanilla/Templates/Rest/Client/ServiceClientTemplate.cshtml
+++ b/src/vanilla/Templates/Rest/Client/ServiceClientTemplate.cshtml
@@ -86,14 +86,23 @@ namespace @Settings.Namespace
             @:SerializationSettings.Converters.Add(new Microsoft.Rest.Serialization.PolymorphicSerializeJsonConverter<@(polymorphicType.Name)>("@(polymorphicType.PolymorphicDiscriminator)"));
             @:DeserializationSettings.Converters.Add(new  Microsoft.Rest.Serialization.PolymorphicDeserializeJsonConverter<@(polymorphicType.Name)>("@(polymorphicType.PolymorphicDiscriminator)"));
             }
-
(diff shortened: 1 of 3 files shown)


src/vanilla/Templates/Rest/Client/ServiceClientTemplate.cshtml
```diff
@@ -86,14 +86,23 @@ namespace @Settings.Namespace
             @:SerializationSettings.Converters.Add(new Microsoft.Rest.Serialization.PolymorphicSerializeJsonConverter<@(polymorphicType.Name)>("@(polymorphicType.PolymorphicDiscriminator)"));
             @:DeserializationSettings.Converters.Add(new  Microsoft.Rest.Serialization.PolymorphicDeserializeJsonConverter<@(polymorphicType.Name)>("@(polymorphicType.PolymorphicDiscriminator)"));
             }
-
+            @EmptyLine
             CustomInitialize();
+            @EmptyLine
+            if(SerializationSettings.ContractResolver == null)
```
fearthecowboy: You would need to remove line `62` and `78` as well (so that it's not setting it before CustomInitialize())

shahabhijeet: You know I had the most fun time razor template files.
Only the higher power is my witness how many times I had to checkout that file as for some odd reason one of the closing curly braces was not getting emitted and I could never get the code generated right especially the closing curly brace.
So in doing so, I might have missed this in my final commit as I constantly kept checkout files to start all over with my little change.
But yes that would dumb not to delete those two lines :)
Thank you, fixing it

weshaggard:
@shahabhijeet I assume it is safe to close this PR now, if not please reopen.

PR #105, opened 2018-11-16
Sample data & pseudo code for bug

MariusVolkhart:
Provide example Swagger configuration and psuedo code for fix in order to handle form parameters that aren't strings.

I was unable to get the project building on my computer, and don't know how to best approach solving this bug, so I hope that this information is useful as-is. This relates to Azure/autorest#3049


MiYanni:
For now, closing PRs for v2 changes. If there is business justification for this PR, we can reopen.

PR #111, opened 2019-10-08
Update version of modeler to latest 4.0.20

weshaggard:
Also add the generated/ folder to gitignore

PTAL @MiYanni


MiYanni:
@weshaggard Won't that ignore the Generated folder in the project (src\AutoRest.CSharp.V3\Pipeline\Generated)?

weshaggard:
LOL - It sure will. We should either come up with another output convention for our defaults or rename `pipelines\generated` folder.

What is the general output folder you have for your test runs?

weshaggard:
Updated the PR to just be the version bump for now.

PR #115, opened 2019-10-14
Basic plugin infrastructure

MiYanni:
Also, started the type resolution system.


.gitignore
```diff
@@ -351,5 +351,8 @@ MigrationBackup/
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
 
-# Ignore default autorest generated folder in teh root only.
-/generated
\ No newline at end of file
+# Ignores generated files from samples
+samples/**/*.yaml
```
weshaggard: Do we really want to ignore yaml files?

.gitignore
```diff
@@ -351,5 +351,8 @@ MigrationBackup/
...
+!samples/Xkcd/xkcd.yaml
+samples/**/*.cs
```
weshaggard: What settings do you use to have it generate under the samples folder? Or does that just happen when you use the readme.md as the configuration?

readme.md
```diff
@@ -12,10 +12,12 @@ pipeline:
     input: openapi-document/multi-api/identity
   modelerfour/new-transform:
     input: modelerfour
-  csharp-v3:
+  type-identifier:
     input: modelerfour/new-transform
-  csharp-v3/emitter:
-    input: csharp-v3
+  model-creator:
```
weshaggard: I think it might still be good to have these be csharp specific because in isolation I have no idea this is generating csharp code.

.gitignore
weshaggard: Oh I see the output-folder being defined in the readmes now. Cool that works.

samples/Xkcd/sort.ps1
```diff
@@ -0,0 +1,2 @@
+yml-sorter --input CodeModel-xkcd.yaml
```
weshaggard: Did you intend to commit this script?

src/AutoRest.CSharp.V3/CodeGen/BaseCSharpWriter.cs
```diff
@@ -6,30 +6,27 @@
...
-        private HashSet<string> _usings = new HashSet<string>();
+        private readonly HashSet<string> _usings = new HashSet<string>();
         public BaseCSharpWriter()
         {
         }
 
         public class CodeBlock : IDisposable
         {
-            public Action? _endBlockAction;
+            public Action? EndBlockAction;
```
weshaggard: Sorry this should have been a private field not public. In which case using the "_" made sense.

src/AutoRest.CSharp.V3/CodeGen/BaseCSharpWriter.cs
```diff
@@ -94,7 +91,7 @@ public CodeBlock Method(string modifiers, string type, string name, params strin
         public void DocSummary(string summary)
         {
             Line("/// <summary>");
-            string[] summaryLines = summary.Split(Environment.NewLine);
+            var summaryLines = summary.Split(Environment.NewLine);
```
weshaggard: Why switch to "var"? I generally only use var when the name of the type is on the other side of the assignment. That is the coding guidelines we use on the .NET team (https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) as well on our Azure .NET guidelines.

.gitignore
MiYanni: Yes, because they get generated as part of testing the pipeline.

readme.md
MiYanni: Once I have actual plugins clearly defined, that is the plan. For now, they are just temporary names.

src/AutoRest.CSharp.V3/Configuration.cs
```diff
@@ -0,0 +1,21 @@
...
+using AutoRest.CSharp.V3.JsonRpc;
+
+namespace AutoRest.CSharp.V3
+{
+    internal class Configuration
+    {
+        public string OutputPath { get; private set; }
+
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
+        private Configuration() { }
```
weshaggard: Instead of disabling the warning I would simply pass in the value to the constructor in your Create method below.

samples/Xkcd/sort.ps1
MiYanni: It was used to test comparing code model files prior to writing the order numbering algorithm. But, it didn't work that well anyway. So, I'll likely delete it.

src/AutoRest.CSharp.V3/CodeGen/BaseCSharpWriter.cs
MiYanni: I'll be cleaning this stuff up later anyway as there are several things that need to be worked on/fixed. I'm not worried about it at the moment.

src/AutoRest.CSharp.V3/PluginProcessor.cs
```diff
@@ -1,18 +1,36 @@
...
+        //public static Dictionary<string, Func<IPlugin>> Plugins = new Dictionary<string, Func<IPlugin>>
+        //{
+        //    { SerializeTester.PluginName, () => new SerializeTester() },
+        //    { ModelCreator.PluginName, () => new ModelCreator() },
+        //    { SerializeTester.PluginName, () => new SerializeTester() },
+        //    { SerializeTester.PluginName, () => new SerializeTester() },
+        //};
+
+        //https://stackoverflow.com/a/26750/294804
+        private static readonly Type[] PluginTypes = Assembly.GetExecutingAssembly().GetTypes()
```
weshaggard: Do we really think we will have enough plugins to warrant the need for dynamic discovery vs just a hard-coded list? Personally i don't think we will.

.gitignore
weshaggard: That means that we will have to force add any new samples that are yaml instead of json. Can we instead add a sub-folder convention to the output-path readmes? Instead of `.` have `.\output` or something we can then wildcard on?

src/AutoRest.CSharp.V3/CodeGen/BaseCSharpWriter.cs
MiYanni: My programming style uses `var` in every situation it can be used. I only use explicit types when they are needed, because it makes it obvious that the type on the left-hand side is intentionally being declared. Additionally, all refactoring on every project I've worked on was easier to do if all variables are declared as `var`. Additionally, I tend to make small, well defined methods, so excessive variable declaration doesn't generally happen.

I often use variables as 'logical separation points' and less as 'reusable values'. This helps readability for humans. As for the type names on the left-hand side, as long as the variables are named appropriately, I generally think that it actually helps readability to have them all as `var`, because you just understand it to be 'variable declaration' and are not particularly worried about the 'type of the container' (the type the variable is declared as).

Here is a small bit of coding conventions for using `var`. Again, I don't follow it, but if I was to follow something, this one seems to be the most straightforward: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#implicitly-typed-local-variables

src/AutoRest.CSharp.V3/Configuration.cs
MiYanni: This file is intended to be a grab-bag of stuff from the configuration that would be useful during code generation. All of the properties will be like this, so doing on the constructor means I can just add properties without worrying about it. However, I would like to be explicit about what is nullable versus non-nullable.

src/AutoRest.CSharp.V3/PluginProcessor.cs
MiYanni: It certainly makes it easier for me to work on in the short term. It is just something else I don't need to worry about until I know how this system will be designed.

.gitignore
MiYanni: For now, this is temporary. You had asked before how I was testing changes locally. So, I figured it would be good to add what I already have to the repo. None of these files are intended to remain long-term.

src/AutoRest.CSharp.V3/CodeGen/BaseCSharpWriter.cs
weshaggard: I recommend we use our teams coding guidelines which follows that of the .NET team which is to only use var when the type is obvious. 

I can tell you from my many years on the .NET team you read code much more then you write it and you cannot rely on tools to help you resolve types, for example doing code reviews on github, you cannot easily determine the types which makes things a lot harder to review.

src/AutoRest.CSharp.V3/Configuration.cs
weshaggard: Why is adding them to the constructor so much harder then setting them in the block below?

src/AutoRest.CSharp.V3/PluginProcessor.cs
weshaggard: Having worked on the MEF (Managed Extensibility Framework) team I can tell you debugging dynamically loaded things is a lot harder to and so I generally only do it if I need to dynamically discover things from different assemblies. When they are all in the same assembly I would recommend statically referencing them.

PR #250, opened 2019-11-21
Add .editorconfig and StyleCop rules, cleanup project files

pakrym:


pakrym:
cc @MiYanni

package.json
```diff
@@ -4,8 +4,8 @@
   "name": "@autorest/csharp-v3",
   "description": "See readme.md for instructions",
   "scripts": {
-    "start": "dotnet ./bin/AutoRest.CSharp.V3.dll --server",
-    "debug": "dotnet ./bin/AutoRest.CSharp.V3.dll --server --launch-debugger"
+    "start": "dotnet ./artifacts/bin/AutoRest.CSharp.V3/Debug/netcoreapp3.0/AutoRest.CSharp.V3.dll  --server",
+    "debug": "dotnet ./artifacts/bin/AutoRest.CSharp.V3/Debug/netcoreapp3.0/AutoRest.CSharp.V3.dll --server --launch-debugger"
```
MiYanni: This locks us into always running the debug version of the package. I intentionally made everything build into `bin` flat because:
- When shipping this extension, you can build it as Release and it will still work.
- If we update to netcoreappX.X, this also wouldn't need to change. The framework is basically irrelevant for the extension.
- AFAIK, there aren't multiple `start`s. So, we can't have a `start-debug` or `start-release` or something like that. Otherwise, you'd have to ship a different package.json than you are using for development.

Other than it being the standard convention to output to `[buildName]/[framework]`, let me know if there are other reasons for this. I just wanted to keep it simple for now.

src/AutoRest.CSharp.V3/JsonRpc/Message.cs
```diff
@@ -14,118 +16,4 @@ internal class Message : IMessage
...
-        Fatal,
-        // Hint messages offer guidance or support without forcing action.
-        Hint,
-        // File represents a file output from an extension. Details are a Artifact and are required.
-        File,
-        // Content represents an update/creation of a configuration file. The final URI will be in the same folder as the primary config file.
-        Configuration,
-        // Protect is a path to not remove during a clear-output-folder.
-        Protect
-    }
```
MiYanni: It is fine to separate these into different files. In development, it was much easier to have everything in the same file since these are basically POCOs with 1 method. So, they are all related to each other. You do lose that cohesiveness when going to separate files. Maybe put them into the same namespace?

package.json
pakrym: This `package.json` is not in a publishable state anyway and we would have to fix that in the future. For now I want to bring the repo structure to what we have in other repos.

src/AutoRest.CSharp.V3/JsonRpc/Message.cs
pakrym: They are in the same `AutoRest.CSharp.V3.JsonRpc` namespace

MiYanni: What I meant is... everything was in `Message.cs` which was already in `AutoRest.CSharp.V3.JsonRpc`. So, to preserve intent of that, the models that were in `Message.cs` could to be put into a nested namespace like `AutoRest.CSharp.V3.JsonRpc.MessageModels` (keeping them together and not mixed in with all other classes in `AutoRest.CSharp.V3.JsonRpc`).

package.json
MiYanni: That's fine. When we do get to publishing, it would likely make sense to keep it flatter as we likely won't publish multiple DLLs of different frameworks into the same package. But, we're not at the point of worrying about that anyway.

pakrym: Yes, but your concerns seem to be about things that would only matter when publishing.

MiYanni: I'm saying it is fine. I don't mind if we nest it like that for now. I was just pointing out why I didn't nest it originally, and was wondering what benefit there was to nesting it.

src/AutoRest.CSharp.V3/JsonRpc/Message.cs
pakrym: Done

PR #251, opened 2019-11-21
Operation codegen

MiYanni:
Multitude of changes. 
- Pulls in Azure.Core required internal types and processes them. Includes them in both the generator project and the generated output.
- Wrote basic serialization for model types
- Wrote basic logic for operations
- Added more samples, including a couple from the test server
  - Note, the BodyString sample won't work until I implement ConstantSchema handling
- Uses latest from AutoRest and ModelerFour
- Several bug fixes and enhancements
- Added a couple example projects for testing/debugging/designing


src/examples/Dns/Generated/Operations/RecordSetsOperations.cs
```diff
@@ -0,0 +1,166 @@
...
+            //using ClientDiagnostics scope = pipeline.CreateScope("Azure.Dns.Operations.V20180501.Update");
+            //scope.AddAttribute("key", name)
+            //scope.Start()
+            try
+            {
+                var request = pipeline.CreateRequest();
+                request.Method = RequestMethod.Patch;
+                //request.Uri.Reset();
+                if (ifMatch != null) { request.Headers.SetValue("If-Match", ifMatch.ToString() ?? string.Empty); }
+                request.Uri.AppendQuery("api-version", apiVersion.ToString() ?? string.Empty);
```
pakrym: `apiVersion.ToString() ?? string.Empty` expression doesn't make sense

src/examples/Dns/Generated/Operations/RecordSetsOperations.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+            catch
+            {
+                //scope.Failed(e);
+                throw;
+            }
+            throw new Exception();
```
pakrym: This methods seems to always throw.

src/examples/Dns/Generated/Operations/RecordSetsOperations.cs
MiYanni: `ToString()` returns `string?`. You pointed that out to me, remember?
https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Object.cs#L39

MiYanni: File removed.

src/examples/Dns/Generated/Operations/RecordSetsOperations.cs
MiYanni: File removed.

PR #268, opened 2019-12-03
Cleanup tests, generate url.json with workaround add couple of tests

pakrym:


test/AutoRest.TestServer.Tests/TestServerTestBase.cs
```diff
@@ -1,13 +1,48 @@
...
+            var server = TestServerSession.Start(scenario);
+            try
+            {
+                await test(server.Host);
+            }
+            catch (Exception ex)
+            {
+                try
+                {
+                    await server.DisposeAsync();
```
pakrym: This is so we don't hide test exceptions behind the scenario checks.

Issue #242, opened 2019-11-19
Operator: Handle escaped/not-escaped request variables


MiYanni:
Right now, it is assumed everything is escaped. Handle when they are not.

PR #270, opened 2019-12-04
Add ability to control escaping for paths and queries

pakrym:
Re-enable some tests.

requires https://github.com/Azure/autorest.test-server/pull/18


test/AutoRest.TestServer.Tests/url/Generated/Operations/PathsOperations.cs
```diff
@@ -45,7 +45,7 @@ public static async ValueTask<Response> GetBooleanFalseAsync(ClientDiagnostics c
                 request.Method = RequestMethod.Get;
                 request.Uri.Reset(new Uri($"{host}"));
                 request.Uri.AppendPath("/paths/bool/false/", false);
-                request.Uri.AppendPath("false", false);
+                request.Uri.AppendPath("false", true);
```
pakrym: Eventually we would be doing escaping for constants as part of generation not yet tho.

src/AutoRest.CSharp.V3/Plugins/Modeler.cs
```diff
@@ -190,6 +190,22 @@ private ConstantOrParameter[] ToParts(string httpRequestUri, Dictionary<string,
...
+        {
+            List<PathSegment> host = new List<PathSegment>();
+            foreach ((string text, bool isLiteral) in GetPathParts(httpRequestUri))
+            {
+                // WORKAROUND FOR https://github.com/Azure/autorest.modelerfour/issues/58
+                if (!parameters.ContainsKey(text))
+                {
+                    parameters[text] = StringConstant(text);
+                }
+                host.Add(new PathSegment(isLiteral ? StringConstant(text) : parameters[text], !isLiteral));
```
MiYanni: This is duplicate code of the method above it with 1 line difference.

pakrym: Cleaned up.

pakrym:
> Is this PR for #242?

There are more features around escaping we need to implement (x-ms-skipescaping-...) but this is a start.

pakrym:
I'll take this after the TestServer v1 PR goes in to verify I didn't break any tests.

PR #273, opened 2019-12-04
Add initial header tests

pakrym:


test/AutoRest.TestServer.Tests/TestServerTestBase.cs
```diff
@@ -15,17 +18,54 @@ namespace AutoRest.TestServer.Tests
...
+                if (missingScenarios.Any())
+                {
+                    Assert.Fail("Expected scenarios " + string.Join(Environment.NewLine, missingScenarios) + " not defined");
+                }
+            }
+        }
+
+        public Task TestStatus(Func<string, HttpPipeline, Task<Response>> test)
+        {
+            return TestStatus(TestContext.CurrentContext.Test.Name, test);
```
pakrym: Method name can now be used as intended scenario.

PR #297, opened 2019-12-05
Bug fix: collection of constants

MiYanni:
The TODOs I added, I'll end up removing when I remove nullability today.


src/AutoRest.CSharp.V3/CodeGen/CsProjWriter.cs
```diff
@@ -17,8 +17,7 @@ public bool WriteCsProj(Configuration configuration)
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <OutputPath>bin</OutputPath>
     <PublishDir>$(OutputPath)</PublishDir>
-    <!-- Some methods are marked async and don't have an await in them -->
-    <NoWarn>1998</NoWarn>
+    <NoWarn>SA1649</NoWarn>
```
pakrym: Why do we need this?

pakrym:
We need a test for this.

src/AutoRest.CSharp.V3/CodeGen/CsProjWriter.cs
MiYanni: Because we make files not named the encompassing type. Aka, the `Model.Serialiation.cs` files. Since this is by-design, we can ignore the warning.

MiYanni: And, we'll likely make more files like this in the future. For example, XML might be a different file versus JSON serialization.

MiYanni:
AppConfiguration sample is what I used to test this. So, we have a test. Also, this isn't a permanent fix, as indicated by the comments above the code fix.

pakrym:
If it doesn't fail CI build when the fix is undone it doesn't count :)

eng/Generate.ps1
```diff
@@ -17,28 +17,45 @@ function Invoke-Block([scriptblock]$cmd) {
...
-$testServerTestProject = "$repoRoot\test\AutoRest.TestServer.Tests"
-$testConfiguration = "$testServerTestProject\readme.md"
-$testServerSwaggerPath = "$repoRoot\node_modules\@microsoft.azure\autorest.testserver\swagger"
-$paths = 'url', 'body-string', 'body-complex', 'custom-baseUrl', 'custom-baseUrl-more-options'
-if ($name)
-{
-    $paths = $name
+function Invoke-AutoRest($debugFlags, $testConfiguration, $outputFolder, $inputFile, $name, $namespace, $repoRoot) {
+    Invoke-Block {
+        $outputFlag = if($outputFolder) { "--output-folder=$outputFolder" } else { '' }
```
pakrym: Can we just forward all the non-matched argumetns?

MiYanni: I'll have this cleaned up in the next PR. This PR simply allows the check to run. I'll make everything run the same way in the next one.

PR #298, opened 2019-12-06
Test infra restructure

MiYanni:
- Consolidated test project settings so that they use a `readme.tests.md` to control generalized configuration (output-folder, etc)
  - This is the same format used by the Sample projects
- Made test projects build independently
  - The appropriate source of these projects are still referenced in the overall AutoRest.TestServer.Tests project
  - This allows us to also use Sample source in the same manner, if we want to make tests for the samples
- Fixed a few bugs with pathing when running autorest-beta from different directories than the directory of the readme itself
- Fixed path issue with AutRest.CoreShared
- Added missing `header` project to Generate.ps1 script
- Fixed bug where Uri wasn't using `Type()` call, so the namespace of it was not being acknowledged


src/AutoRest.CSharp.V3/CodeGen/CsProjWriter.cs
```diff
@@ -17,7 +17,7 @@ public bool WriteCsProj(Configuration configuration)
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <OutputPath>bin</OutputPath>
     <PublishDir>$(OutputPath)</PublishDir>
-    <NoWarn>SA1649</NoWarn>
+    <NoWarn>SA1649,CS8618</NoWarn>
```
MiYanni: Since we currently do not have nullable references removed, this is a workaround to allow projects to not error about non-null properties in models. I attempted, at first, to make all of them nullable, but that causes more issues because the caller complains with another warning about "This value may be null". Instead, ignoring this warning for now is appropriate, until we remove nullable references for the development of the project.

src/AutoRest.CoreShared/Program.cs
```diff
@@ -46,7 +44,6 @@ private static void Main()
                 .Replace("options.GetType().Namespace", "options.GetType().Namespace!");
             File.WriteAllText($"../../{Path}/ClientDiagnostics.cs", cleanFile);
             File.WriteAllText($"../../AutoRest.CSharp.V3/Azure.Core.Shared/ClientDiagnostics.cs", cleanFile);
-            File.WriteAllText($"../../../test/AutoRest.TestServer.Tests/Azure.Core.Shared/ClientDiagnostics.cs", cleanFile);
```
MiYanni: These no longer exist since the test project uses references to the source from the `assets` folder.

test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj
```diff
@@ -16,6 +16,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="../../src/assets/**/*.cs"/>
+    <Folder Include="Tests" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="../../src/assets/**/*.cs" />
+    <Compile Include="../TestServerProjects/**/*.cs" Exclude="../TestServerProjects/**/Azure.Core.Shared/*.cs" LinkBase="TestServerProjects" />
```
MiYanni: This allows us to compile the files as native source to the test project. We can do the same with Samples if we want to write tests for them.

MiYanni:
There aren't actually a lot of changes, but I moved the test server project files to their own directory, so it looks like a lot more than it actually is.

eng/Generate.ps1
```diff
@@ -17,13 +17,14 @@ function Invoke-Block([scriptblock]$cmd) {
     }
 }
 
-function Invoke-AutoRest($debugFlags, $testConfiguration, $outputFolder, $inputFile, $name, $namespace, $repoRoot) {
+function Invoke-AutoRest($debugFlags, $requireFile, $outputFolder, $inputFile, $title, $namespace, $repoRoot) {
```
pakrym: Can you change this method to take just an array of arguments or string with all arguments?

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -38,14 +38,14 @@ public bool WriteClient(ServiceClient operationGroup)
...
-                        WriteOperation(method, cs.Namespace);
+                        WriteOperation(method, cs.Namespace, rootNamespace);
                     }
                 }
             }
             return true;
         }
 
-        private void WriteOperation(ClientMethod operation, CSharpNamespace? @namespace)
+        private void WriteOperation(ClientMethod operation, CSharpNamespace? @namespace, string rootNamespace)
```
pakrym: undo

src/AutoRest.CSharp.V3/CodeGen/CsProjWriter.cs
pakrym: Set nullable project property to annotation: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj#L6

src/AutoRest.CSharp.V3/Plugins/Modeler.cs
```diff
@@ -46,7 +46,7 @@ public async Task<bool> Execute(IAutoRestInterface autoRest, CodeModel codeModel
             foreach (var client in clients)
             {
                 var writer = new ClientWriter(typeFactory);
-                writer.WriteClient(client);
+                writer.WriteClient(client, configuration.Namespace);
```
pakrym: undo

test/AutoRest.TestServer.Tests/Tests/body-complex.cs
```diff
@@ -6,7 +6,7 @@
 using body_complex.Models.V20160229;
 using NUnit.Framework;
 
-namespace AutoRest.TestServer.Tests
+namespace AutoRest.TestServer.Tests.Tests
```
pakrym: This is a strange namespace.

eng/Generate.ps1
MiYanni: Not all calls to this function currently use all arguments. These just become named arguments. If I remove the named arguments, all arguments are accessible by $args, but become position-bound (meaning, you have to know which index into $args represents which data). Even with named arguments, you can still access all arguments via $args.

However, I can try to clean this up. I just didn't want to spent a lot of time on it.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs

src/AutoRest.CSharp.V3/Plugins/Modeler.cs
MiYanni: Fixed.

test/AutoRest.TestServer.Tests/Tests/body-complex.cs
MiYanni: Yup. The tests are in a Tests namespace. But, the project also has 'Tests' in the name of it. It made sense to put the tests into their own namespace since we'll end up with many of them. Then, the files at root are test infrastructure only. I could rename the namespace to 'UnitTests' if that is less weird.

pakrym: Can we just not moving tests into `Tests` subdirectory? It's very clear they are tests.

pakrym: I would rather move infrastructure into subfolder,

MiYanni:
@pakrym Removed the copies of the shared source files.

.gitignore
```diff
@@ -358,4 +358,5 @@ samples/**/*.cs
 samples/**/*.txt
 samples/**/*.csproj
 samples/**/*.sln
-test/**/*.yaml
\ No newline at end of file
+test/**/*.yaml
+test/TestServerProjects/**/Generated/Azure.Core.Shared/*.cs
```
pakrym: We can use Directory.Props to include shared file into project without having them copied around locally.

pakrym: You would also not need the exclude `Exclude="../TestServerProjects/**/Azure.Core.Shared/*.cs"`

test/AutoRest.TestServer.Tests/Tests/body-complex.cs
MiYanni: Done.

.gitignore
MiYanni: Done..

src/AutoRest.CSharp.V3/CodeGen/CsProjWriter.cs
MiYanni: Done.

eng/Generate.ps1
MiYanni: Done. Made Invoke-AutoRest take in a single $autoRestArguments parameter.

MiYanni:
I believe I've addressed all the comments now. 🤞

src/AutoRest.CSharp.V3/CodeGen/CsProjWriter.cs
pakrym: With annotations you don't need the CS8618 skipped anymore.

PR #308, opened 2019-12-06
Get generated projects closer to shipping SDKs

pakrym:
Trim project files, target netstandard
Keep only one copy of shared files and download it as part of the code check
Make operations internal


eng/CodeChecks.ps1
```diff
@@ -46,6 +46,11 @@ function Invoke-Block([scriptblock]$cmd) {
 }
 
 try {
+    Write-Host "Downloading files"
+    Invoke-Block {
+        & $PSScriptRoot\DownloadSharedSource.ps1 @script:PSBoundParameters
```
MiYanni: To run scripts, you don't need the call operator (used for external executables). Instead, you just source the script using the dot operator.
```suggestion
        . $PSScriptRoot\DownloadSharedSource.ps1 @script:PSBoundParameters
```

Same for the same generate.ps1 call below.

pakrym: But I explicitly don't want the side-effects of those scripts execution.

Issue #301, opened 2019-12-06
Implement numeric formats


MiYanni:
Test: PutComplexPrimitiveLong, PutComplexPrimitiveFloat
Description: Put complex types with long properties

Currently, only certain numeric properties are created as certain types. We need these to change type depending on the format of the numeric, and match the right format appropriately. 
- `long` becomes `int` 
- `float` becomes `double`

MiYanni:
Implemented.

PR #318, opened 2019-12-07
Numeric fixes

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/301

Properly handles numeric formats for short, int, long, float, double, and decimal. Enables a couple more tests to pass in body-complex.

### DIFF
--- a/src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
+++ b/src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
@@ -27,9 +27,9 @@ namespace AutoRest.CSharp.V3.Pipeline
             AllSchemaTypes.Dictionary => null,
             AllSchemaTypes.Duration => typeof(TimeSpan),
             AllSchemaTypes.Flag => null,
-            AllSchemaTypes.Integer => typeof(int),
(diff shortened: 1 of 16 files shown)


src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
```diff
@@ -27,9 +27,9 @@ internal static class Extensions
             AllSchemaTypes.Dictionary => null,
             AllSchemaTypes.Duration => typeof(TimeSpan),
             AllSchemaTypes.Flag => null,
-            AllSchemaTypes.Integer => typeof(int),
+            AllSchemaTypes.Integer => null, // Handled in ToFrameworkType for NumberSchema
```
pakrym: remove them from here?

pakrym:
This might be a good time to start adding tests that verify model structure. In this case we can verify that property types are what we expect.

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
MiYanni: This isn't a final design. I'm starting to get a handle of how we need to layer this stuff. Most of these mappings aren't going to hold up.

Issue #293, opened 2019-12-05
Implement date url parameter encoding


pakrym:
Includes unix-time etc.

PR #319, opened 2019-12-08
Fix URL/Query formatting

pakrym:
Fixes https://github.com/Azure/autorest.csharp/issues/293
Fixes https://github.com/Azure/autorest.csharp/issues/291

### DIFF
--- a/src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
+++ b/src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
@@ -89,53 +89,18 @@ namespace AutoRest.CSharp.V3.CodeGen
 
                     foreach (var segment in operation.Request.PathSegments)
                     {
-                        var value = segment.Value;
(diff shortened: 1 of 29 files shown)


src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -172,19 +137,107 @@ private void WriteOperation(ClientMethod operation, CSharpNamespace? @namespace)
...
+                Literal(formatSpecifier);
+            }
+        }
+
+        private void WriteQueryParameter(QueryParameter queryParameter)
+        {
+            ConstantOrParameter value = queryParameter.Value;
+            if (value.IsConstant)
+            {
+                Append("request.Uri.AppendQuery(");
```
pakrym: I know that the code between WriteQuery/Header/Path looks slightly duplicated but I think we need a bit of code-writer refactoring to enable deduplicating in cleanly.

PR #321, opened 2019-12-09
Update Autorest

pakrym:
Re-enable some tests and add a bit of enum-related model tests.


src/AutoRest.CodeModel/Program.cs
```diff
@@ -19,10 +19,10 @@ internal static class Program
         private static void Main()
         {
             using var webClient = new WebClient();
-            webClient.DownloadFile(@"https://raw.githubusercontent.com/Azure/perks/master/codemodel/.resources/all-in-one/json/code-model.json", "../../../../../src/AutoRest.CodeModel/code-model.json");
+            var schemaJson = webClient.DownloadString(@"https://raw.githubusercontent.com/Azure/perks/master/codemodel/.resources/all-in-one/json/code-model.json");
 
-            var schemaJsonLines = File.ReadAllLines("../../../../../src/AutoRest.CodeModel/code-model.json");
-            var schemaJson = String.Join(Environment.NewLine, schemaJsonLines)
```
MiYanni: The code-model schema was being cached into the repo so that we could keep track of the file we generated the CodeModel.cs from. That file has no versioning or anything, so this was just a simple mechanism to track that and see changes to the schema when they happen.

eng/Generate.ps1
```diff
@@ -1,6 +1,8 @@
-param($name, [switch]$noDebug)
+param($name, [switch]$noDebug, [switch]$NoReset)
 $ErrorActionPreference = 'Stop'
 
+$AutorestCoreVersion = "3.0.6162"
```
MiYanni: This should be part of the readme.md of the extension (file in root). Then, the extension itself depends on a specific version of the core, which is what we want when someone uses our extension.

src/AutoRest.CodeModel/Program.cs
pakrym: Fixed

eng/Generate.ps1
pakrym: Fixed

PR #327, opened 2019-12-10
Eng script fixes

MiYanni:
- Cleaned script logic to use simplified calling (no call operator, only single Invoke-Expression)
- Made the main parts into functions
- Dot sourced the functions into CodeChecks.ps1
- Allowed Generate.ps1 to run (for development) but be dot-sourceable for CodeChecks.ps1
- Both drivers (code ran when sourced) for Generate.ps1 and CodeChecks.ps1 are simply wrapped in a try/catch
  - Outputs exception on failure and instantly exits with non-zero exit code
- Added intentional failure to show CI handling (will be removed before merge)


eng/CodeChecks.ps1
```diff
@@ -1,82 +1,28 @@
-#requires -version 5
+#Requires -Version 6.0
```
MiYanni: Since we use Join-Path with more than 2 parameters, that version of the cmdlet is only supported in 6.0 and up.

eng/DownloadSharedSource.ps1
```diff
@@ -1,11 +1,17 @@
...
-$baseUrl = "https://raw.githubusercontent.com/Azure/azure-sdk-for-net/master/sdk/core/Azure.Core/src/Shared/"
-
-foreach ($file in $files)
+function Invoke-DownloadSharedSource
 {
-    $text = Invoke-WebRequest -Uri "$baseUrl/$file";
-    $text.Content.Replace("#nullable enable", "#pragma warning disable CS8600, CS8604, CS8605").Trim() | Out-File "$downloadPath/$file"
-}
+    $currentProgressPreference = $ProgressPreference
+    $ProgressPreference = 'SilentlyContinue'
```
MiYanni: Made the progress silent, as it has been a traditionally slow process in PowerShell.

eng/CodeChecks.ps1
```diff
@@ -1,82 +1,28 @@
...
-        & $PSScriptRoot\DownloadSharedSource.ps1 @script:PSBoundParameters
-    }
-
-    Write-Host "Generate test clients"
-    Invoke-Block {
-        & $PSScriptRoot\Generate.ps1 @script:PSBoundParameters
+        Invoke-AutoRest '--reset'
     }
 
-    Write-Host "git diff"
```
pakrym: What verifies that code is up-to-date now that you removed this part?

eng/CodeChecks.ps1
```diff
@@ -1,82 +1,28 @@
...
-    [ValidateNotNullOrEmpty()]
-    [string] $ServiceDirectory
-)
+param([switch]$noReset)
 
 $ErrorActionPreference = 'Stop'
 Set-StrictMode -Version 1
 
-$repoRoot = Resolve-Path "$PSScriptRoot/.."
+. (Join-Path $PSScriptRoot 'DownloadSharedSource.ps1')
```
pakrym: Why dot source?

eng/DownloadSharedSource.ps1
```diff
@@ -1,11 +1,17 @@
-$downloadPath = Resolve-Path (Join-Path $PSScriptRoot '..' 'src' 'assets' 'Azure.Core.Shared')
-$files = "ClientDiagnostics.cs", "ArrayBufferWriter.cs", "DiagnosticScope.cs"
-$baseUrl = "https://raw.githubusercontent.com/Azure/azure-sdk-for-net/master/sdk/core/Azure.Core/src/Shared/"
-
-foreach ($file in $files)
+function Invoke-DownloadSharedSource
```
pakrym: It'ss impossible to invoke this script independently now.

eng/Generate.ps1
```diff
@@ -1,64 +1,58 @@
...
+        Invoke-AutoRest $autoRestArguments
+    }
 }
 
-# Sample configuration
-$sampleDirectory = Join-Path $repoRoot 'samples'
-$projectNames = 'AppConfiguration'
-
-foreach ($projectName in $projectNames)
+if(-not $noRun)
```
pakrym: This is a very strange pattern.

eng/CodeChecks.ps1
MiYanni: That's my mistake. I had it in my commented code and removed it accidentally.

eng/DownloadSharedSource.ps1
MiYanni: Didn't know that was a requirement. Will fix.

eng/Generate.ps1
MiYanni: When the 'executable' is also a 'library', it made it difficult to source and execute. Will fix the pattern in the morning.

pakrym:
Can you please verify that diff check looks like expected? I'll take another look at the morning but in general the change looks good

pakrym:
The build passed with the exception still in. Suspicious

MiYanni:
@pakrym Yes. Cannot test the changes at home. Will test in the morning. Doing things as functions creates scopes that are now gone. So, it is likely something is falling through the cracks.

MiYanni:
See the previous run to see what a failure looks like:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=208842&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774

eng/Generate.ps1
```diff
@@ -36,19 +27,24 @@ $debugFlags = if (-not $noDebug) { '--debug', '--verbose' }
...
 {
-    Invoke-AutoRest "--reset" $repoRoot
+    'body-string',
+    'extensible-enums-swagger',
+    'url-multi-collectionFormat',
+    'url',
+    'body-complex',
+    'custom-baseUrl',
+    'custom-baseUrl-more-options',
+    'header'
```
MiYanni: Made the test list vertical instead of horizontal since we'll be adding a lot more here in the near term (avoid having a really wide line).

readme.md
```diff
@@ -1,6 +1,17 @@
...
+- [NodeJS](https://nodejs.org/en/) (10.x.x or 12.x.x)
+- `npm install` (at root)
+- `npm install -g @autorest/autorest`
+- [.NET Core SDK](https://dotnet.microsoft.com/download/dotnet-core/3.0) (3.0.100)
+- [PowerShell Core](https://github.com/PowerShell/PowerShell/releases/latest)
+
+## Build
+- `dotnet build` (at root)
+- `./eng/Generate.ps1` (at root in PowerShell Core)
+
```
MiYanni: This is just to help remember what is needed to do stuff in the repo. We can add/modify it later.

eng/DownloadSharedSource.ps1
```diff
@@ -1,11 +1,17 @@
...
 foreach ($file in $files)
 {
     $text = Invoke-WebRequest -Uri "$baseUrl/$file";
-    $text.Content.Replace("#nullable enable", "#pragma warning disable CS8600, CS8604, CS8605").Trim() | Out-File "$downloadPath/$file"
+    $text.Content.Replace('#nullable enable', '#pragma warning disable CS8600, CS8604, CS8605').Trim() | Out-File (Join-Path $downloadPath $file)
 }
 
-exit 0
\ No newline at end of file
+$ProgressPreference = $currentProgressPreference
```
pakrym: I don't think you need to restore this value anymore.

eng/Generate.ps1
```diff
@@ -1,31 +1,22 @@
...
-function Invoke-AutoRest($autoRestArguments, $repoRoot) {
-    Invoke-Block {
-        $command = "npx autorest-beta $autoRestArguments"
-        $commandText = $command.Replace($repoRoot, "`$(SolutionDir)")
-
-        Write-Host ">" $commandText
-        
-        & cmd /c "$command 2>&1"
-    }
+if ($reset)
```
pakrym: This should be on-by default so CI is always using the correct version

MiYanni: How would it not be using the right version? It downloads the version as part of CI, right?

pakrym: autorest.core is cached in user directory and --reset cleans it.

MiYanni: Fixed.

eng/DownloadSharedSource.ps1
MiYanni: Removed restoring it.

PR #328, opened 2019-12-10
Simplify CodeWriter and split renderers from code writer

pakrym:
Split concepts of things that turn code model into code (ModelWriter etc.) and ones that turn code into strings (CodeWriter).

Squash CodeWriter class structure into one level.
Made Append and Line return CodeWriter itself to allow builder pattern.


pakrym:
I think inheritance doesn't represent the relationship between these types correctly. For example:

1. We don't want to override any type from CodeWriter in inherited types.
2. Inheritance makes it impossible to replace CodeWriter implementations and  ClientWriter, ModelWriter shouldn't rely on the concrete implementation.
3. Inheritance creates a somewhat artificial limitation of one writer class == one output file. I want to be able to split writers into smaller chunks and make them reusable which is better accomplished by taking CodeWriter rather than inheriting from it.

PR #331, opened 2019-12-10
Add extensible-enums-swagger with tests

pakrym:


test/AutoRest.TestServer.Tests/Infrastructure/TestServerTestBase.cs
```diff
@@ -102,5 +102,12 @@ private async Task Test(string scenario, Func<string, HttpPipeline, Task> test,
 
             await server.DisposeAsync();
         }
+
+        private static string GetScenarioName()
+        {
+            var testName = TestContext.CurrentContext.Test.Name;
+            var indexOfUnderscore = testName.IndexOf('_');
```
pakrym: Allow test variations that hit the same endpoint.

PR #332, opened 2019-12-11
Add nullchecks for nullable required parameters

pakrym:


src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -140,6 +142,21 @@ private void WriteOperation(CodeWriter writer, ClientMethod operation, CSharpNam
...
         }
 
+        private void WriteParameterNullChecks(CodeWriter writer, ClientMethod operation)
+        {
+            foreach (ServiceClientMethodParameter parameter in operation.Parameters)
+            {
+                var cs = _typeFactory.CreateType(parameter.Type);
+                if (parameter.IsRequired && (cs.IsNullable || !cs.IsValueType))
+                {
+                    writer.Append("if (").Append(parameter.Name).Append("== null)");
```
MiYanni: Use the `writer.If` writer (for consistency). I know it'll make the genned output 'taller', but I'd rather be consistent for now. We can cleanup style later. If we want single-line statements, we can add flags to the `If` (or other) calls to simplify their output.

PR #336, opened 2019-12-11
Split modeler so we can work with fewer merge conflicts

pakrym:


MiYanni:
Since I have multiple changes in my DateTime PR, any chance we could merge that before this one?

pakrym:
Yes, yours should go first

pakrym:
This is ready to go.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -0,0 +1,255 @@
...
+
+            List<PathSegment> host = new List<PathSegment>();
+            foreach ((string text, bool isLiteral) in GetPathParts(httpRequestUri))
+            {
+                if (!isLiteral)
+                {
+
+                    if (!parameters.TryGetValue(text, out var segment))
+                    {
+                        //TODO: WORKAROUND FOR https://github.com/Azure/autorest.modelerfour/issues/58
```
MiYanni: It looks like this was fixed now. We should probably try to resolve this workaround.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -0,0 +1,255 @@
...
+                    case BinarySchema _:
+                        // skip
+                        continue;
+                    //TODO: Workaround for https://github.com/Azure/autorest.csharp/pull/275
+                    case ArraySchema arraySchema when arraySchema.ElementType is ConstantSchema constantInnerType:
+                        constantOrParameter = new ServiceClientMethodParameter(requestParameter.CSharpName(),
+                            new CollectionTypeReference(ClientModelBuilderHelpers.CreateType(constantInnerType.ValueType, false), false),
+                            CreateDefaultValueConstant(requestParameter), false);
+                        break;
+                    //TODO: Workaround for https://github.com/Azure/autorest.csharp/pull/275
```
MiYanni: This seems like the wrong link? Or something... I'm not sure how that PR relates to the code that is here now.

pakrym: You've added it :)

Issue #304, opened 2019-12-06
Encode DateTime Rfc1123 parameters properly


MiYanni:
Test: PutComplexPrimitiveDateTimeRfc1123
Description: Put complex types with datetimeRfc1123 properties

I believe date needs to be encoded in the format: `Mon, 18 May 2015 11:38:00 GMT`

MiYanni:
Done.

Issue #303, opened 2019-12-06
Encode DateTime parameters properly


MiYanni:
Test: PutComplexPrimitiveDateTime
Description: Put complex types with datetime properties

I believe date needs to be encoded in the format: `2015-05-18T11:38:00-08:00`

MiYanni:
Done.

Issue #302, opened 2019-12-06
Encode date parameters properly


MiYanni:
Test: PutComplexPrimitiveDate
Description: Put complex types with date properties

I believe date needs to be encoded in the format: `yyyy-MM-dd`

MiYanni:
Done.

PR #337, opened 2019-12-11
Datetime format fixes (encoding)

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/302
Fixes: https://github.com/Azure/autorest.csharp/issues/303
Fixes: https://github.com/Azure/autorest.csharp/issues/304

This fixes the encoding for the datetime types. It does not do decoding (will do in separate PR). Unblocked a few more tests in body-complex.


src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
```diff
@@ -58,15 +66,29 @@ internal static class Extensions
...
+        private static readonly Func<string, bool, FrameworkTypeReferenceFormat, string?> NumberSerializer =
+            (vn, nu, f) => $"writer.WriteNumberValue({vn}{(nu ? ".Value" : string.Empty)});";
+        private static Func<string, bool, FrameworkTypeReferenceFormat, string?> StringSerializer(bool includeToString = false) =>
+            (vn, nu, f) => $"writer.WriteStringValue({vn}{(includeToString ? ".ToString()" : string.Empty)});";
+
+        private static string? ToFormatSpecifier(this FrameworkTypeReferenceFormat format) => format switch
+        {
+            FrameworkTypeReferenceFormat.DateTimeRFC1123 => "R",
+            FrameworkTypeReferenceFormat.DateTimeISO8601 => "yyyy-MM-ddTHH:mm:ssZ",
+            FrameworkTypeReferenceFormat.Date => "yyyy-MM-dd",
```
MiYanni: These formats are the format of the datetimes of the testserver tests. Do we even know if these are the right formats for all swaggers?

test/AutoRest.TestServer.Tests/body-complex.cs
```diff
@@ -274,18 +270,17 @@ public void CheckComplexPrimitiveDouble()
         public Task GetComplexPrimitiveDateTimeRfc1123() => Test(async (host, pipeline) =>
         {
             var result = await PrimitiveOperations.GetDateTimeRfc1123Async(ClientDiagnostics, pipeline, host);
-            Assert.AreEqual(DateTimeOffset.Parse("Mon, 01 Jan 0001 12:00:00 GMT"), result.Value.Field);
+            Assert.AreEqual(DateTimeOffset.Parse("Mon, 01 Jan 0001 00:00:00 GMT"), result.Value.Field);
```
MiYanni: I had the actual value wrong initially because I used the description value, not the one from the actual testserver script.

src/AutoRest.CSharp.V3/ClientModels/FrameworkTypeReference.cs
```diff
@@ -9,11 +9,13 @@ internal class FrameworkTypeReference: ClientTypeReference
     {
         public Type Type { get; }
         public override bool IsNullable { get; }
+        public FrameworkTypeReferenceFormat Format { get; }
 
-        public FrameworkTypeReference(Type type, bool isNullable = false)
+        public FrameworkTypeReference(Type type, bool isNullable = false, FrameworkTypeReferenceFormat format = FrameworkTypeReferenceFormat.Default)
```
pakrym: Let's put this onto the ClientObjectProperty. I'd like to keep type and serialization information separate.

src/AutoRest.CSharp.V3/ClientModels/FrameworkTypeReferenceFormat.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace AutoRest.CSharp.V3.ClientModels
+{
+    internal enum FrameworkTypeReferenceFormat
```
pakrym: Can we share the SerializationFormat enum?

test/TestServerProjects/body-complex/Generated/Models/DateWrapper.Serialization.cs
```diff
@@ -14,12 +14,12 @@ internal void Serialize(Utf8JsonWriter writer)
             if (Field != null)
             {
                 writer.WritePropertyName("field");
-                writer.WriteStringValue(Field.ToString());
+                writer.WriteStringValue(Field.Value.ToString("yyyy-MM-dd"));
```
pakrym: Let's add an extension method for this. Same way we did with query/path/headers.

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
pakrym: Nope, we don't. We'll eventually figure out if we need more.

src/AutoRest.CSharp.V3/ClientModels/FrameworkTypeReferenceFormat.cs
MiYanni: I had mentioned this before... when it was just a header format. I'll reuse it. I wrote these changes before that change.

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
```diff
@@ -28,6 +28,14 @@ internal static class Extensions
             _ => null
         };
 
+        public static FrameworkTypeReferenceFormat ToFrameworkTypeReferenceFormat(this Schema schema) => schema switch
```
pakrym: This is very similar to GetSerializationFormat method

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
```diff
@@ -58,15 +66,29 @@ internal static class Extensions
...
-        private static readonly Func<string, bool, string?> NumberSerializer =
-            (vn, nu) => $"writer.WriteNumberValue({vn}{(nu ? ".Value" : string.Empty)});";
-        private static Func<string, bool, string?> StringSerializer(bool includeToString = false) =>
-            (vn, nu) => $"writer.WriteStringValue({vn}{(includeToString ? ".ToString()" : string.Empty)});";
+        private static readonly Func<string, bool, FrameworkTypeReferenceFormat, string?> NumberSerializer =
+            (vn, nu, f) => $"writer.WriteNumberValue({vn}{(nu ? ".Value" : string.Empty)});";
+        private static Func<string, bool, FrameworkTypeReferenceFormat, string?> StringSerializer(bool includeToString = false) =>
+            (vn, nu, f) => $"writer.WriteStringValue({vn}{(includeToString ? ".ToString()" : string.Empty)});";
+
+        private static string? ToFormatSpecifier(this FrameworkTypeReferenceFormat format) => format switch
```
pakrym: Let's unify this with the WriteSerializationFormat

MiYanni: So, the actual formats in WriteSerializationFormat are different than this. Do header/query params have a different format than the body properties?

pakrym: They are not different, they are all endup calling into https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Azure.Core.Shared/TypeFormatters.cs#L17 with formats exactly like here.

MiYanni: I was confused that there are specifiers in WriteSerializationFormat that are specifiers for the extensions, not specifiers for the ToString on DateTimeOffset. My bad.

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
MiYanni: Fixed. Used only GetSerializationFormat.

src/AutoRest.CSharp.V3/ClientModels/FrameworkTypeReferenceFormat.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/ClientModels/FrameworkTypeReference.cs
MiYanni: Done. Added RequestBody so that both ClientObjectProperty and RequestBody have the right information to do formatting during serialization generation.

test/TestServerProjects/body-complex/Generated/Models/DateWrapper.Serialization.cs
MiYanni: I added an extension method for Utf8JsonWriter called `WriteDateTimeOffsetValue`. This allows you to pass the format specifier to it. Then, that method calls the ToString in TypeFormatters, which is the same logic used for query/path/headers.

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
MiYanni: So, this is a little complicated right now. Basically, I put the format specifier code (switch statement) from WriteSerializationFormat here. Why? Because, right now, the code to do this serialization logic is in Extensions. And, Extensions doesn't use the actual CodeWriter; it just returns strings. When we move all the Json-specific serialization logic out of Extensions, we can use the same WriteSerializationFormat method. For now, I did the best I can without that refactoring.

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
```diff
@@ -58,15 +59,33 @@ internal static class Extensions
...
+            _ => null
+        };
+
+        private static readonly Func<string, bool, SerializationFormat, string?> DateTimeSerializer = (vn, nu, f) =>
+        {
+            var formatSpecifier = f.ToFormatSpecifier();
+            var valueText = $"{vn}{(nu ? ".Value" : string.Empty)}";
+            var formatText = formatSpecifier != null ? $", \"{formatSpecifier}\"" : string.Empty;
+            //TODO: Hack to call Azure.Core functionality without having the context of the namespaces specified to the file this is being written to.
+            return $"Azure.Core.Utf8JsonWriterExtensions.WriteDateTimeOffsetValue(writer, {valueText}{formatText});";
```
MiYanni: This is only temporary, because all this Json-specific serialization code needs to be moved into a CodeWriter. When it is, we'll have the context of namespaces/types, and without that right now, I don't know if the files have `using Azure.Core;`. For example, Model files generally don't use Azure.Core. The operation files use Azure.Core, which is why path/query/header calls to the extension methods even work right now.

MiYanni:
@pakrym I think I've addressed all the comments. Also, there are 3 tests failing... which is odd because those tests are ignored. Any ideas?
```
Test Run Failed.
  ! UrlStringNullAsync [< 1ms]
  ! UrlStringUnicodeAsync [< 1ms]
  ! UrlUrlDateNullAsync [< 1ms]
```

https://github.com/Azure/autorest.csharp/blob/1e868271db1c44a04657a4bfc481a10fdf373f60/test/AutoRest.TestServer.Tests/url.cs#L30-L32
https://github.com/Azure/autorest.csharp/blob/1e868271db1c44a04657a4bfc481a10fdf373f60/test/AutoRest.TestServer.Tests/url.cs#L34-L36
https://github.com/Azure/autorest.csharp/blob/1e868271db1c44a04657a4bfc481a10fdf373f60/test/AutoRest.TestServer.Tests/url.cs#L56-L58

pakrym:
> @pakrym I think I've addressed all the comments. Also, there are 3 tests failing... which is odd because those tests are ignored. Any ideas?

Those are not the failed tests. Scroll up and you'll see the one that actually fails.

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
pakrym: What about adding a UseNamespace() method to the writer? I think it would endup being a one-liner.

src/AutoRest.CSharp.V3/ClientModels/RequestBody.cs
```diff
@@ -0,0 +1,23 @@
...
+        public ConstantOrParameter Value { get; }
+        public SerializationFormat Format { get; }
+        public string Name { get; }
+        public ClientTypeReference Type { get; }
+
+        public RequestBody(ConstantOrParameter value, SerializationFormat format = SerializationFormat.Default)
+        {
+            Value = value;
+            Format = format;
+            Name = Value.IsConstant ? Value.Constant.ToValueString() : Value.Parameter.Name;
```
pakrym: `ToValueString()` is an evil method and would give a wrong result in a lot of cases. Something like `WriteConstant` should be used.

But in general we shouldn't be doing codegen on this level so I would drop name property and move this logic to the code generator itself.

src/AutoRest.CSharp.V3/ClientModels/RequestBody.cs
```diff
@@ -0,0 +1,23 @@
...
+            Type = Value.IsConstant ? Value.Constant.Type : Value.Parameter.Type;
```
pakrym: Type property is nice but let's move it to `ConstantOrParameter` I think there are a lot of callsites that would benefit from it.

src/assets/Azure.Core.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Text.Json;
+
+namespace Azure.Core
+{
+    internal static class Utf8JsonWriterExtensions
+    {
+        public static void WriteDateTimeOffsetValue(this Utf8JsonWriter writer, DateTimeOffset value, string format)
```
pakrym: Why not make this method an overload of WriteStringValue?

src/AutoRest.CSharp.V3/Pipeline/Extensions.cs
pakrym: I see you don't have the writer here. OK this is fine then

src/AutoRest.CSharp.V3/ClientModels/RequestBody.cs
MiYanni: There is only 1 place right now where RequestBody was being written. And, it did both the name and type calculation right before it was used. So, I thought moving them here made sense. But, I can move name back there. I'd love to use `WriteConstant`. However, that will have to wait until we remove all the serialization writing out of Extensions.cs. That's where the name value feeds into right now. :-/

MiYanni: Fixed.

src/AutoRest.CSharp.V3/ClientModels/RequestBody.cs
MiYanni: Done.

src/assets/Azure.Core.Shared/Utf8JsonWriterExtensions.cs
MiYanni: Great idea. Done.

Issue #307, opened 2019-12-06
Decode DateTime Rfc1123 format from response


MiYanni:
Test: GetComplexPrimitiveDateTimeRfc1123
Description: Get complex types with datetimeRfc1123 properties

The `GetDateTime()` method from System.Text.Json doesn't handle this format of date time. So, we need custom deserialization logic for it.

MiYanni:
Done.

PR #340, opened 2019-12-12
Decode DateTimeOffset formats

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/307

Follows a similar pattern to what was done for encoding DateTimeOffsets.

### DIFF
--- a/src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
+++ b/src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
@@ -47,16 +47,16 @@ namespace AutoRest.CSharp.V3.CodeGen
         private void WriteOperation(CodeWriter writer, ClientMethod operation, CSharpNamespace @namespace)
         {
             //TODO: Handle multiple responses
-            var schemaResponse = operation.Response.Type;
-            CSharpType? bodyType = schemaResponse != null ? _typeFactory.CreateType(schemaResponse) : null;
+            var responseBody = operation.Response.ResponseBody;
+            CSharpType? bodyType = responseBody != null ? _typeFactory.CreateType(responseBody.Value) : null;
             CSharpType? headerModelType = operation.Response.HeaderModel != null ? _typeFactory.CreateType(operation.Response.HeaderModel.Name) : null;
 
             CSharpType responseType = bodyType switch
             {
-                null when headerModelType == null => new CSharpType(typeof(Response)),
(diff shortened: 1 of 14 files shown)


src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -47,9 +47,9 @@ public bool WriteClient(CodeWriter writer, ServiceClient operationGroup)
         private void WriteOperation(CodeWriter writer, ClientMethod operation, CSharpNamespace @namespace)
         {
             //TODO: Handle multiple responses
-            var schemaResponse = operation.ResponseType;
-            CSharpType? responseType = schemaResponse != null ? _typeFactory.CreateType(schemaResponse) : null;
-            CSharpType returnType = schemaResponse != null && responseType != null
+            ResponseBody? responseBody = operation.ResponseBodies.FirstOrDefault();
```
pakrym: If we don't use the array in generator let's not add an array to the model yet.

pakrym:
You didn't reenable any tests here. Do we have any that cover affected scenarios?

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs

MiYanni:
`GetComplexPrimitiveDateTimeRfc1123` was the test unignored by this PR.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -47,16 +47,16 @@ public bool WriteClient(CodeWriter writer, ServiceClient operationGroup)
...
             CSharpType? headerModelType = operation.Response.HeaderModel != null ? _typeFactory.CreateType(operation.Response.HeaderModel.Name) : null;
 
             CSharpType responseType = bodyType switch
             {
-                null when headerModelType == null => new CSharpType(typeof(Response)),
-                null => new CSharpType(typeof(ResponseWithHeaders<>), headerModelType),
+                null when headerModelType != null => new CSharpType(typeof(ResponseWithHeaders<>), headerModelType),
                 { } when headerModelType == null => new CSharpType(typeof(Response<>), bodyType),
                 { } => new CSharpType(typeof(ResponseWithHeaders<>), bodyType, headerModelType),
+                _ => new CSharpType(typeof(Response)),
```
MiYanni: Resharper was like, "You ain't covering all those cases." And I was like, "Hmm, really?" So, I just put the 'all null' case as the default.

PR #341, opened 2019-12-12
Add fast generator mode

pakrym:
That reuses existing codemodel.yaml and avoids calling autorest.

### DIFF
--- a/eng/Generate.ps1
+++ b/eng/Generate.ps1
@@ -1,22 +1,26 @@
 #Requires -Version 6.0
-param($name, [switch]$noDebug, [switch]$reset)
+param($name, [switch]$noDebug, [switch]$reset, [switch]$noBuild, [switch]$fast)
 
 $ErrorActionPreference = 'Stop'
 
-function Invoke-AutoRest($autoRestArguments)
+function Invoke-AutoRest($baseOutput, $title, $autoRestArguments)
 {
-    $command = "npx autorest-beta $autoRestArguments"
+    $namespace = $title.Replace('-', '_')
+    $command = "npx autorest-beta $debugFlags $autoRestArguments --title=$title --namespace=$namespace"
+    if ($fast)
+    {
+        $codeModel = Join-Path $baseOutput $title "CodeModel.yaml"
+        $outputPath = Join-Path $baseOutput $title
+        $command = "dotnet run --project $autorestPluginProject --no-build -- --plugin=cs-modeler --title=$title --namespace=$namespace --standalone --input-file=$codeModel --output-path=$outputPath"
+    }
+    
     Write-Host "> $command"
     cmd /c "$command 2>&1"
+
     if($LastExitCode -ne 0)
     {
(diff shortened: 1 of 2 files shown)


eng/Generate.ps1
```diff
@@ -1,30 +1,40 @@
...
-param($name, [switch]$noDebug, [switch]$reset)
+param($name, [switch]$noDebug, [switch]$reset, [switch]$noBuild, [switch]$fast)
 
 $ErrorActionPreference = 'Stop'
 
-function Invoke-AutoRest($autoRestArguments)
+function Invoke-AutoRest($baseOutput, $title, $autoRestArguments)
 {
-    $command = "npx autorest-beta $autoRestArguments"
+    $command = "npx autorest-beta $debugFlags $autoRestArguments"
```
MiYanni: Use $script:debugFlags, since it isn't passed in (explicit scope access to avoid confusion).

eng/Generate.ps1
```diff
@@ -1,30 +1,40 @@
...
+    {
+        $namespace = $title.Replace('-', '_')
+        $command = "$command --title=$title --namespace=$namespace"
+    }
+
+    if ($fast)
+    {
+        $codeModel = Join-Path $baseOutput $title "CodeModel.yaml"
+        $outputPath = Join-Path $baseOutput $title
+        $command = "dotnet run --project $autorestPluginProject --no-build -- --plugin=cs-modeler --title=$title --namespace=$namespace --standalone --input-file=$codeModel --output-path=$outputPath"
```
MiYanni: Use $script:autorestPluginProject. And, there is an extra `--` in from of the `--plugin`.

src/AutoRest.CSharp.V3/JsonRpc/StandaloneAutoRestInterface.cs
```diff
@@ -59,6 +59,10 @@ public Task<string[]> ListInputs(string? artifactType = null)
 
         public async Task WriteFile(string filename, string content, string artifactType, RawSourceMap? sourceMap = null)
         {
+            if (string.IsNullOrEmpty(content))
+            {
+                return;
+            }
```
MiYanni: This guard clause should be in the other override (below) too.

eng/Generate.ps1
pakrym: It's not extra, it's separator for arguments that are for `dotnet run` and the app it's invoking.

MiYanni: That makes sense.

src/AutoRest.CSharp.V3/JsonRpc/StandaloneAutoRestInterface.cs
pakrym: We don't use that one, removed.

eng/Generate.ps1
pakrym: Done.

PR #344, opened 2019-12-13
Add singular inheritance support

pakrym:


src/AutoRest.CSharp.V3/ClientModels/ModelBuilder.cs
```diff
@@ -13,18 +13,26 @@ internal class ModelBuilder
...
             choiceSchema.Choices.Select(c => new ClientEnumValue(c.CSharpName(), ClientModelBuilderHelpers.StringConstant(c.Value))),
             true);
 
-        private static ClientModel BuildClientObject(ObjectSchema objectSchema) => new ClientObject(
-            objectSchema, objectSchema.CSharpName(),
-            objectSchema.Properties.Where(property => !(property.Schema is ConstantSchema)).Select(CreateProperty),
-            objectSchema.Properties.Where(property => property.Schema is ConstantSchema).Select(CreateConstant));
+        private static ClientModel BuildClientObject(ObjectSchema objectSchema)
+        {
+            var inheritsFrom = objectSchema.Parents?.Immediate.OfType<ObjectSchema>().SingleOrDefault();
```
MiYanni: Keep in mind that we'll have to support inheritance from schemas of types other than ObjectSchema (such as OrSchema, XorSchema, NotSchema(?), GroupSchema(?)), and DictionarySchema) later on. You can get a good idea of what may/may not apply looking at the `Schemas` class in CodeModel.cs or looking at types driving from `ComplexSchema` in there too.

test/TestServerProjects/body-complex/Generated/Models/ArrayWrapper.Serialization.cs
```diff
@@ -5,15 +5,15 @@
 
 namespace body_complex.Models.V20160229
 {
-    public partial class ArrayWrapper
+    public partial class ArrayWrapperSerializer
```
MiYanni: I think this is fine for now to have serialization be separated from the type itself (because of inheritance). However, we should figure out later on what design makes sense for serialization when dealing with derived types.

*Future thought*: I'm still thinking we have the types implement an interface for serialization. That way, we can have others create C# custom models, and have them work in our generated system, as long as those models implement that interface. It seems the most natural.

PR #352, opened 2019-12-16
Support polymorphism

pakrym:


src/AutoRest.CSharp.V3/CodeGen/SerializationWriter.cs
```diff
@@ -88,12 +112,24 @@ private void ReadProperty(CodeWriter writer, ClientTypeReference type, Serializa
...
+                WriteNullCheck(writer);
+            }
             writer.Append($"result.{name} = ");
             writer.ToDeserializeCall(type, format, _typeFactory, "property.Value", t, t);
             writer.Line(";");
         }
 
+        private static void WriteNullCheck(CodeWriter writer)
+        {
+            using (writer.If("property.Value.ValueKind == JsonValueKind.Null"))
```
MiYanni: Use `writer.Type(typeof(JsonValueKind))` instead of literal JsonValueKind. This is required as it is how the usings work (for now).

PR #353, opened 2019-12-16
Cleanup interface of JSON serialization helpers

pakrym:
Passing around delegates instead of strings allows us to use CodeWriter features inside other calls

### DIFF
--- a/src/AutoRest.CSharp.V3/CodeGen/CodeWriter.cs
+++ b/src/AutoRest.CSharp.V3/CodeGen/CodeWriter.cs
@@ -204,6 +204,12 @@ namespace AutoRest.CSharp.V3.CodeGen
             return this;
         }
 
+        public CodeWriter Append(CodeWriterDelegate writerDelegate)
+        {
+            writerDelegate(this);
+            return this;
+        }
+
         public CodeWriter Comma() => Append(", ");
 
         public CodeWriter Space() => Append(" ");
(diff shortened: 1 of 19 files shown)


src/AutoRest.CSharp.V3/CodeGen/CodeWriter.cs
```diff
@@ -248,6 +254,13 @@ public string ToFormattedCode()
 
         public override string? ToString() => _builder.ToString();
 
+        public static string Materialize(CodeWriterDelegate writer)
```
pakrym: This method is temporary. When we'll cleanup all string formatting in JSON serializer writer it can be removed.

test/TestServerProjects/body-complex/Generated/Models/ByteWrapper.Serialization.cs
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System;
```
MiYanni: Interesting. What caused this?

test/TestServerProjects/body-number/Generated/Operations/NumberOperations.cs
```diff
@@ -512,7 +512,7 @@ public static async ValueTask<Response> PutBigDecimalPositiveDecimalAsync(Client
                 request.Headers.Add("Content-Type", "application/json");
                 using var content = new Utf8JsonRequestContent();
                 var writer = content.JsonWriter;
-                writer.WriteNumberValue(99999999.99);
+                writer.WriteNumberValue(99999999.99M);
```
MiYanni: And what caused this?

pakrym: using the correct Literal implementation instead of object.ToString

test/TestServerProjects/body-complex/Generated/Models/ByteWrapper.Serialization.cs
pakrym: Not sure but I like it :)

pakrym:
CI passed https://dev.azure.com/azure-sdk/public/_build/results?buildId=213725&view=results

PR #358, opened 2019-12-17
Support additional properties

pakrym:
I'll add tests.

Move more logic into `JsonSerializationExtensions`


src/assets/Azure.Core.Shared/JsonElementExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+    internal static class JsonElementExtensions
+    {
+        public static object? GetObject(in this JsonElement element)
+        {
+            switch (element.ValueKind)
+            {
+                case JsonValueKind.String:
+                    return element.GetString();
+                case JsonValueKind.Number:
+                    return element.GetInt32();
```
MiYanni: Shouldn't we assume the largest amount of possible numeric types, and use decimal? (not int)

test/TestServerProjects/body-boolean/Generated/Models/Error.Serialization.cs
```diff
@@ -20,6 +21,7 @@ internal static void Serialize(Error model, Utf8JsonWriter writer)
                 writer.WritePropertyName("message");
                 writer.WriteStringValue(model.Message);
             }
+
```
MiYanni: It seems there is an extra carriage return added to both the Serialize and Deserialize methods.

src/AutoRest.CSharp.V3/ClientModels/ModelBuilder.cs
```diff
@@ -25,11 +25,28 @@ internal class ModelBuilder
 
         private static ClientModel BuildClientObject(ObjectSchema objectSchema)
         {
-            var inheritsFrom = objectSchema.Parents?.Immediate.OfType<ObjectSchema>().SingleOrDefault();
-            var inheritsFromTypeReference = inheritsFrom != null ? ClientModelBuilderHelpers.CreateType(inheritsFrom, false) : null;
+            ClientTypeReference? inheritsFromTypeReference = null;
+            DictionaryTypeReference? dictionaryElementTypeReference = null;
+
+            foreach (ComplexSchema complexSchema in objectSchema.Parents!.Immediate)
```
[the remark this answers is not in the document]
pakrym: Yep, https://github.com/Azure/autorest.csharp/pull/358/files#diff-5b490df2c1adda7893a3eddd9241e459R21

src/assets/Azure.Core.Shared/JsonElementExtensions.cs
pakrym: Decimal seems overkill, I did double.

PR #359, opened 2019-12-17
Standalone bug fix

MiYanni:
Simply creates the directory structure necessary for writing files when running in standalone mode.


src/AutoRest.CSharp.V3/JsonRpc/StandaloneAutoRestInterface.cs
```diff
@@ -65,6 +65,7 @@ public async Task WriteFile(string filename, string content, string artifactType
             }
             filename = Path.Combine(_basePath, filename);
             Console.WriteLine($"Writing {filename} {artifactType}");
+            (new FileInfo(filename)).Directory?.Create();
```
pakrym:
```suggestion
            Directory.CreateDirectory(Path.GetDirectoryName(filename));
```

PR #361, opened 2019-12-18
Add support to FormattableString to CodeWriter

pakrym:
Now you can use the string interpolation and get the correct result.

### DIFF
--- a/src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
+++ b/src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
@@ -80,11 +80,11 @@ namespace AutoRest.CSharp.V3.CodeGen
                 writer.Line($"using var scope = clientDiagnostics.CreateScope(\"{@namespace.FullName}.{methodName}\");");
                 //TODO: Implement attribute logic
                 //writer.Line("scope.AddAttribute(\"key\", name);");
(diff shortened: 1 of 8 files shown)


src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -176,53 +175,28 @@ private void WriteConstant(CodeWriter writer, ClientConstant constant)
...
-                    writer.Comma();
-                    writer.Literal(dateTimeValue.Millisecond);
-                    writer.Comma();
-                    writer.AppendType(typeof(TimeSpan));
-                    writer.Append(".");
-                    writer.Append(nameof(TimeSpan.Zero));
-                    writer.Append(")");
+                    var d = (DateTimeOffset)constant.Value;
+                    d = d.ToUniversalTime();
+                    writer.Append($"new {typeof(DateTimeOffset)}({d.Year:L}, {d.Month:L}, {d.Day:L} ,{d.Hour:L}, {d.Minute:L}, {d.Second:L}, {d.Millisecond:L}, {typeof(TimeSpan)}.{nameof(TimeSpan.Zero)}");
```
pakrym: ":L" format specifier renders parameter as literal.

passing typeof(bla) would also do the right thing

PR #365, opened 2019-12-19
Generate body-array/dictionary/collection

pakrym:
Tests for array and byte, not dictionary.

Support for temp variable name generation.


src/AutoRest.CSharp.V3/CodeGen/CodeWriter.cs
```diff
@@ -40,6 +50,7 @@ public CodeWriter Append(FormattableString formattableString)
             }
 
             const string literalFormatString = ":L";
+            const string declarationFormatString = ":D"; // :D :)
```
MiYanni: ;)

Issue #229, opened 2019-11-09
Process Xml serialization information


MiYanni:
![image](https://user-images.githubusercontent.com/17788297/68521740-e45e8a00-0258-11ea-94f0-af454bebcef8.png)

PR #382, opened 2019-12-27
Xml serialization

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/229


src/AutoRest.CSharp.V3/Plugins/Modeler.cs
```diff
@@ -115,5 +117,32 @@ public async Task<bool> Execute(IAutoRestInterface autoRest, CodeModel codeModel
 
             return true;
         }
+
+        private KnownMediaType[] GetMediaTypes(CodeModel codeModel)
```
MiYanni: Add a comment for https://github.com/Azure/autorest.modelerfour/issues/103, so we can make sure to change it later.

PR #388, opened 2020-01-04
Initial doc comment support

pakrym:
Not everything is great with this implementation:

1. Comments are generated at the model building time so type references can not be resolved yet.
2. Default descriptions are rough.

But it's a start and we can iterate.


pakrym:
Filed: https://github.com/Azure/autorest.csharp/issues/389 for <return/>

Issue #285, opened 2019-12-05
[Test server] body-dictionary.json


pakrym:
- [x] Code generated
- [ ] Tests for all scenarios
- [ ] API shape questions
- [ ] Fix identified issues

MiYanni:
Implemented. Additional work for the tests have their own issues. Closing.

PR #390, opened 2020-01-06
Body dictionary

MiYanni:
Adds the body-dictionary tests. Part of: https://github.com/Azure/autorest.csharp/issues/285


pakrym:
Revert  test/scripts/TestList.txt

PR #392, opened 2020-01-06
Update the test server

pakrym:


src/assets/Azure.Core.Shared/TypeFormatters.cs
```diff
@@ -15,11 +15,11 @@ internal class TypeFormatters
         public static string ToString(bool value) => value ? "true" : "false";
 
         // TODO: remove useRealIso when https://github.com/Azure/autorest.testserver/pull/111 is in
```
MiYanni: Remove comment since it is fixed.

test/AutoRest.TestServer.Tests/additionalProperties.cs
```diff
@@ -130,6 +131,7 @@ public class AdditionalPropertiesTest : TestServerTestBase
         });
 
         [Test]
+        [IgnoreOnTestServer(TestServerVersion.V2, "No match")]
```
MiYanni: Were these tests removed from V2?

pakrym: We are sending different format now that is correct but V2 match is too strict.

PR #394, opened 2020-01-07
Cleanup writers

pakrym:
Cleanup some old enum writing code


MiYanni:
You might need to update the branch from recent changes to see if there are issues.

PR #400, opened 2020-01-11
Paging support

MiYanni:
This is just the initial support. There are other issues for specific features. Keep in mind that I do have a plan to clean this up and likely name some methods differently.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "devDependencies": {
     "@autorest/autorest": "https://github.com/Azure/autorest/releases/download/autorest-3.0.6153/autorest-3.0.6153.tgz",
-    "@microsoft.azure/autorest.testserver": "^2.10.5"
+    "@microsoft.azure/autorest.testserver": "^2.10.6"
   },
   "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "devDependencies": {
     "@autorest/autorest": "https://github.com/Azure/autorest/releases/download/autorest-3.0.6153/autorest-3.0.6153.tgz",
-    "@microsoft.azure/autorest.testserver": "^2.10.1"
+    "@microsoft.azure/autorest.testserver": "^2.10.5"
   },
   "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "devDependencies": {
     "@autorest/autorest": "https://github.com/Azure/autorest/releases/download/autorest-3.0.6153/autorest-3.0.6153.tgz",
-    "@microsoft.azure/autorest.testserver": "^2.10.6"
+    "@microsoft.azure/autorest.testserver": "^2.10.7"
   },
   "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "debug": "dotnet ./artifacts/bin/AutoRest.CSharp.V3/Debug/netcoreapp3.0/AutoRest.CSharp.V3.dll --server --launch-debugger"
   },
   "devDependencies": {
-    "@autorest/autorest": "https://github.com/Azure/autorest/releases/download/autorest-3.0.6151/autorest-autorest-3.0.6151.tgz",
-    "@microsoft.azure/autorest.testserver": "^2.9.1"
+    "@autorest/autorest": "https://github.com/Azure/autorest/releases/download/autorest-3.0.6153/autorest-3.0.6153.tgz",
+    "@microsoft.azure/autorest.testserver": "^2.10.1"
   },
   "dependencies": {}
 }
--- a/samples/AppConfiguration/AppConfiguration/Generated/Operations/AllOperations.cs
+++ b/samples/AppConfiguration/AppConfiguration/Generated/Operations/AllOperations.cs
@@ -1803,12 +1803,12 @@ namespace AppConfiguration
 
             async Task<Page<Key>> FirstPageFunc(int? pageSizeHint)
             {
-                var response = await GetKeysAsync(name, after, acceptDatetime, cancellationToken);
+                var response = await GetKeysAsync(name, after, acceptDatetime, cancellationToken).ConfigureAwait(false);
                 return Page.FromValues(response.Value.Items, response.Value.NextLink, response.GetRawResponse());
             }
             async Task<Page<Key>> NextPageFunc(string nextLinkUrl, int? pageSizeHint)
             {
-                var response = await GetKeysNextPageAsync(acceptDatetime, nextLinkUrl, cancellationToken);
+                var response = await GetKeysNextPageAsync(acceptDatetime, nextLinkUrl, cancellationToken).ConfigureAwait(false);
                 return Page.FromValues(response.Value.Items, response.Value.NextLink, response.GetRawResponse());
             }
             return PageResponseEnumerator.CreateAsyncEnumerable(FirstPageFunc, NextPageFunc);
@@ -1845,12 +1845,12 @@ namespace AppConfiguration
 
             async Task<Page<KeyValue>> FirstPageFunc(int? pageSizeHint)
             {
-                var response = await GetKeyValuesAsync(key, label, after, acceptDatetime, select, cancellationToken);
+                var response = await GetKeyValuesAsync(key, label, after, acceptDatetime, select, cancellationToken).ConfigureAwait(false);
                 return Page.FromValues(response.Value.Items, response.Value.NextLink, response.GetRawResponse());
             }
             async Task<Page<KeyValue>> NextPageFunc(string nextLinkUrl, int? pageSizeHint)
             {
-                var response = await GetKeyValuesNextPageAsync(acceptDatetime, nextLinkUrl, cancellationToken);
+                var response = await GetKeyValuesNextPageAsync(acceptDatetime, nextLinkUrl, cancellationToken).ConfigureAwait(false);
                 return Page.FromValues(response.Value.Items, response.Value.NextLink, response.GetRawResponse());
             }
             return PageResponseEnumerator.CreateAsyncEnumerable(FirstPageFunc, NextPageFunc);
@@ -1888,12 +1888,12 @@ namespace AppConfiguration
 
             async Task<Page<Label>> FirstPageFunc(int? pageSizeHint)
             {
-                var response = await GetLabelsAsync(name, after, acceptDatetime, select, cancellationToken);
+                var response = await GetLabelsAsync(name, after, acceptDatetime, select, cancellationToken).ConfigureAwait(false);
                 return Page.FromValues(response.Value.Items, response.Value.NextLink, response.GetRawResponse());
             }
             async Task<Page<Label>> NextPageFunc(string nextLinkUrl, int? pageSizeHint)
             {
-                var response = await GetLabelsNextPageAsync(acceptDatetime, nextLinkUrl, cancellationToken);
+                var response = await GetLabelsNextPageAsync(acceptDatetime, nextLinkUrl, cancellationToken).ConfigureAwait(false);
                 return Page.FromValues(response.Value.Items, response.Value.NextLink, response.GetRawResponse());
             }
             return PageResponseEnumerator.CreateAsyncEnumerable(FirstPageFunc, NextPageFunc);
@@ -1931,12 +1931,12 @@ namespace AppConfiguration
 
             async Task<Page<KeyValue>> FirstPageFunc(int? pageSizeHint)
             {
-                var response = await GetRevisionsAsync(key, label, after, acceptDatetime, select, cancellationToken);
+                var response = await GetRevisionsAsync(key, label, after, acceptDatetime, select, cancellationToken).ConfigureAwait(false);
                 return Page.FromValues(response.Value.Items, response.Value.NextLink, response.GetRawResponse());
             }
             async Task<Page<KeyValue>> NextPageFunc(string nextLinkUrl, int? pageSizeHint)
             {
-                var response = await GetRevisionsNextPageAsync(acceptDatetime, nextLinkUrl, cancellationToken);
+                var response = await GetRevisionsNextPageAsync(acceptDatetime, nextLinkUrl, cancellationToken).ConfigureAwait(false);
                 return Page.FromValues(response.Value.Items, response.Value.NextLink, response.GetRawResponse());
             }
             return PageResponseEnumerator.CreateAsyncEnumerable(FirstPageFunc, NextPageFunc);
--- a/samples/AppConfiguration/AppConfiguration/Generated/Operations/AllOperations.cs
+++ b/samples/AppConfiguration/AppConfiguration/Generated/Operations/AllOperations.cs
@@ -38,7 +38,7 @@ namespace AppConfiguration
             this.clientDiagnostics = clientDiagnostics;
             this.pipeline = pipeline;
         }
-        internal HttpMessage CreateGetKeysFirstPageRequest(string? name, string? after, string? acceptDatetime)
+        internal HttpMessage CreateGetKeysRequest(string? name, string? after, string? acceptDatetime)
         {
             var message = pipeline.CreateMessage();
             var request = message.Request;
@@ -69,14 +69,14 @@ namespace AppConfiguration
         /// <param name="after"> Instructs the server to return elements that appear after the element referred to by the specified token. </param>
         /// <param name="acceptDatetime"> Requests the server to respond with the state of the resource at the specified time. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public async ValueTask<Page<Key>> GetKeysFirstPageAsync(string? name, string? after, string? acceptDatetime, CancellationToken cancellationToken = default)
+        public async ValueTask<ResponseWithHeaders<KeyListResult, GetKeysHeaders>> GetKeysAsync(string? name, string? after, string? acceptDatetime, CancellationToken cancellationToken = default)
         {
 
             using var scope = clientDiagnostics.CreateScope("AllOperations.GetKeys");
             scope.Start();
             try
             {
-                using var message = CreateGetKeysFirstPageRequest(name, after, acceptDatetime);
+                using var message = CreateGetKeysRequest(name, after, acceptDatetime);
                 await pipeline.SendAsync(message, cancellationToken).ConfigureAwait(false);
                 switch (message.Response.Status)
                 {
@@ -85,7 +85,7 @@ namespace AppConfiguration
                             using var document = await JsonDocument.ParseAsync(message.Response.ContentStream, default, cancellationToken).ConfigureAwait(false);
                             var value = KeyListResult.DeserializeKeyListResult(document.RootElement);
                             var headers = new GetKeysHeaders(message.Response);
-                            return Page.FromValues(value.Items, value.NextLink, message.Response);
+                            return ResponseWithHeaders.FromValue(value, headers, message.Response);
                         }
                     default:
                         throw await message.Response.CreateRequestFailedExceptionAsync().ConfigureAwait(false);
@@ -102,14 +102,14 @@ namespace AppConfiguration
         /// <param name="after"> Instructs the server to return elements that appear after the element referred to by the specified token. </param>
         /// <param name="acceptDatetime"> Requests the server to respond with the state of the resource at the specified time. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public Page<Key> GetKeysFirstPage(string? name, string? after, string? acceptDatetime, CancellationToken cancellationToken = default)
+        public ResponseWithHeaders<KeyListResult, GetKeysHeaders> GetKeys(string? name, string? after, string? acceptDatetime, CancellationToken cancellationToken = default)
         {
 
             using var scope = clientDiagnostics.CreateScope("AllOperations.GetKeys");
             scope.Start();
             try
             {
-                using var message = CreateGetKeysFirstPageRequest(name, after, acceptDatetime);
+                using var message = CreateGetKeysRequest(name, after, acceptDatetime);
                 pipeline.Send(message, cancellationToken);
                 switch (message.Response.Status)
                 {
@@ -118,7 +118,7 @@ namespace AppConfiguration
                             using var document = JsonDocument.Parse(message.Response.ContentStream);
                             var value = KeyListResult.DeserializeKeyListResult(document.RootElement);
                             var headers = new GetKeysHeaders(message.Response);
-                            return Page.FromValues(value.Items, value.NextLink, message.Response);
+                            return ResponseWithHeaders.FromValue(value, headers, message.Response);
                         }
                     default:
                         throw message.Response.CreateRequestFailedException();
@@ -130,19 +130,12 @@ namespace AppConfiguration
                 throw;
             }
         }
-        internal HttpMessage CreateGetKeysNextPageRequest(string? acceptDatetime, string? nextLinkUrl)
+        internal HttpMessage CreateGetKeysNextPageRequest(string? acceptDatetime, string nextLinkUrl)
         {
             var message = pipeline.CreateMessage();
             var request = message.Request;
             request.Method = RequestMethod.Get;
-            if (Uri.IsWellFormedUriString(nextLinkUrl, UriKind.Absolute))
-            {
-                request.Uri.Reset(new Uri(nextLinkUrl));
-            }
-            else
-            {
-                request.Uri.Reset(new Uri($"{host}{nextLinkUrl}"));
-            }
+            request.Uri.Reset(new Uri($"{nextLinkUrl}"));
             if (syncToken != null)
             {
                 request.Headers.Add("Sync-Token", syncToken);
@@ -157,8 +150,12 @@ namespace AppConfiguration
         /// <param name="acceptDatetime"> Requests the server to respond with the state of the resource at the specified time. </param>
         /// <param name="nextLinkUrl"> The URL to the next page of results. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public async ValueTask<Page<Key>> GetKeysNextPageAsync(string? acceptDatetime, string nextLinkUrl, CancellationToken cancellationToken = default)
+        public async ValueTask<ResponseWithHeaders<KeyListResult, GetKeysHeaders>> GetKeysNextPageAsync(string? acceptDatetime, string nextLinkUrl, CancellationToken cancellationToken = default)
         {
+            if (nextLinkUrl == null)
+            {
+                throw new ArgumentNullException(nameof(nextLinkUrl));
+            }
 
             using var scope = clientDiagnostics.CreateScope("AllOperations.GetKeys");
             scope.Start();
@@ -173,7 +170,7 @@ namespace AppConfiguration
                             using var document = await JsonDocument.ParseAsync(message.Response.ContentStream, default, cancellationToken).ConfigureAwait(false);
                             var value = KeyListResult.DeserializeKeyListResult(document.RootElement);
                             var headers = new GetKeysHeaders(message.Response);
-                            return Page.FromValues(value.Items, value.NextLink, message.Response);
+                            return ResponseWithHeaders.FromValue(value, headers, message.Response);
                         }
                     default:
                         throw await message.Response.CreateRequestFailedExceptionAsync().ConfigureAwait(false);
@@ -186,16 +183,40 @@ namespace AppConfiguration
             }
         }
         /// <summary> Gets a list of keys. </summary>
-        /// <param name="name"> A filter for the name of the returned keys. </param>
-        /// <param name="after"> Instructs the server to return elements that appear after the element referred to by the specified token. </param>
         /// <param name="acceptDatetime"> Requests the server to respond with the state of the resource at the specified time. </param>
+        /// <param name="nextLinkUrl"> The URL to the next page of results. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public AsyncPageable<Key> GetKeysAsync(string? name, string? after, string? acceptDatetime, CancellationToken cancellationToken = default)
+        public ResponseWithHeaders<KeyListResult, GetKeysHeaders> GetKeysNextPage(string? acceptDatetime, string nextLinkUrl, CancellationToken cancellationToken = default)
         {
+            if (nextLinkUrl == null)
+            {
+                throw new ArgumentNullException(nameof(nextLinkUrl));
+            }
 
-            Task<Page<Key>> FirstPageFunc(int? pageSizeHint) => GetKeysFirstPageAsync(name, after, acceptDatetime, cancellationToken).AsTask();
-            Task<Page<Key>> NextPageFunc(string continuationToken, int? pageSizeHint) => GetKeysNextPageAsync(acceptDatetime, continuationToken, cancellationToken).AsTask();
-            return PageResponseEnumerator.CreateAsyncEnumerable(FirstPageFunc, NextPageFunc);
+            using var scope = clientDiagnostics.CreateScope("AllOperations.GetKeys");
+            scope.Start();
+            try
+            {
+                using var message = CreateGetKeysNextPageRequest(acceptDatetime, nextLinkUrl);
+                pipeline.Send(message, cancellationToken);
+                switch (message.Response.Status)
+                {
+                    case 200:
+                        {
+                            using var document = JsonDocument.Parse(message.Response.ContentStream);
+                            var value = KeyListResult.DeserializeKeyListResult(document.RootElement);
+                            var headers = new GetKeysHeaders(message.Response);
+                            return ResponseWithHeaders.FromValue(value, headers, message.Response);
+                        }
+                    default:
+                        throw message.Response.CreateRequestFailedException();
+                }
+            }
+            catch (Exception e)
+            {
+                scope.Failed(e);
+                throw;
+            }
         }
         internal HttpMessage CreateCheckKeysRequest(string? name, string? after, string? acceptDatetime)
         {
@@ -281,7 +302,7 @@ namespace AppConfiguration
                 throw;
             }
         }
-        internal HttpMessage CreateGetKeyValuesFirstPageRequest(string? key, string? label, string? after, string? acceptDatetime, IEnumerable<Get6ItemsItem>? select)
+        internal HttpMessage CreateGetKeyValuesRequest(string? key, string? label, string? after, string? acceptDatetime, IEnumerable<Get6ItemsItem>? select)
         {
             var message = pipeline.CreateMessage();
             var request = message.Request;
@@ -322,14 +343,14 @@ namespace AppConfiguration
         /// <param name="acceptDatetime"> Requests the server to respond with the state of the resource at the specified time. </param>
         /// <param name="select"> Used to select what fields are present in the returned resource(s). </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public async ValueTask<Page<KeyValue>> GetKeyValuesFirstPageAsync(string? key, string? label, string? after, string? acceptDatetime, IEnumerable<Get6ItemsItem>? select, CancellationToken cancellationToken = default)
+        public async ValueTask<ResponseWithHeaders<KeyValueListResult, GetKeyValuesHeaders>> GetKeyValuesAsync(string? key, string? label, string? after, string? acceptDatetime, IEnumerable<Get6ItemsItem>? select, CancellationToken cancellationToken = default)
         {
 
             using var scope = clientDiagnostics.CreateScope("AllOperations.GetKeyValues");
             scope.Start();
             try
             {
-                using var message = CreateGetKeyValuesFirstPageRequest(key, label, after, acceptDatetime, select);
+                using var message = CreateGetKeyValuesRequest(key, label, after, acceptDatetime, select);
                 await pipeline.SendAsync(message, cancellationToken).ConfigureAwait(false);
                 switch (message.Response.Status)
                 {
@@ -338,7 +359,7 @@ namespace AppConfiguration
                             using var document = await JsonDocument.ParseAsync(message.Response.ContentStream, default, cancellationToken).ConfigureAwait(false);
                             var value = KeyValueListResult.DeserializeKeyValueListResult(document.RootElement);
                             var headers = new GetKeyValuesHeaders(message.Response);
-                            return Page.FromValues(value.Items, value.NextLink, message.Response);
+                            return ResponseWithHeaders.FromValue(value, headers, message.Response);
                         }
                     default:
                         throw await message.Response.CreateRequestFailedExceptionAsync().ConfigureAwait(false);
@@ -357,14 +378,14 @@ namespace AppConfiguration
         /// <param name="acceptDatetime"> Requests the server to respond with the state of the resource at the specified time. </param>
         /// <param name="select"> Used to select what fields are present in the returned resource(s). </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public Page<KeyValue> GetKeyValuesFirstPage(string? key, string? label, string? after, string? acceptDatetime, IEnumerable<Get6ItemsItem>? select, CancellationToken cancellationToken = default)
+        public ResponseWithHeaders<KeyValueListResult, GetKeyValuesHeaders> GetKeyValues(string? key, string? label, string? after, string? acceptDatetime, IEnumerable<Get6ItemsItem>? select, CancellationToken cancellationToken = default)
         {
 
             using var scope = clientDiagnostics.CreateScope("AllOperations.GetKeyValues");
             scope.Start();
             try
             {
-                using var message = CreateGetKeyValuesFirstPageRequest(key, label, after, acceptDatetime, select);
+                using var message = CreateGetKeyValuesRequest(key, label, after, acceptDatetime, select);
                 pipeline.Send(message, cancellationToken);
                 switch (message.Response.Status)
                 {
@@ -373,7 +394,7 @@ namespace AppConfiguration
                             using var document = JsonDocument.Parse(message.Response.ContentStream);
                             var value = KeyValueListResult.DeserializeKeyValueListResult(document.RootElement);
                             var headers = new GetKeyValuesHeaders(message.Response);
-                            return Page.FromValues(value.Items, value.NextLink, message.Response);
+                            return ResponseWithHeaders.FromValue(value, headers, message.Response);
                         }
                     default:
                         throw message.Response.CreateRequestFailedException();
@@ -385,19 +406,12 @@ namespace AppConfiguration
                 throw;
             }
         }
-        internal HttpMessage CreateGetKeyValuesNextPageRequest(string? acceptDatetime, string? nextLinkUrl)
+        internal HttpMessage CreateGetKeyValuesNextPageRequest(string? acceptDatetime, string nextLinkUrl)
         {
             var message = pipeline.CreateMessage();
             var request = message.Request;
             request.Method = RequestMethod.Get;
-            if (Uri.IsWellFormedUriString(nextLinkUrl, UriKind.Absolute))
-            {
-                request.Uri.Reset(new Uri(nextLinkUrl));
-            }
-            else
-            {
-                request.Uri.Reset(new Uri($"{host}{nextLinkUrl}"));
-            }
+            request.Uri.Reset(new Uri($"{nextLinkUrl}"));
             if (syncToken != null)
             {
                 request.Headers.Add("Sync-Token", syncToken);
@@ -412,8 +426,12 @@ namespace AppConfiguration
         /// <param name="acceptDatetime"> Requests the server to respond with the state of the resource at the specified time. </param>
         /// <param name="nextLinkUrl"> The URL to the next page of results. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public async ValueTask<Page<KeyValue>> GetKeyValuesNextPageAsync(string? acceptDatetime, string nextLinkUrl, CancellationToken cancellationToken = default)
+        public async ValueTask<ResponseWithHeaders<KeyValueListResult, GetKeyValuesHeaders>> GetKeyValuesNextPageAsync(string? acceptDatetime, string nextLinkUrl, CancellationToken cancellationToken = default)
         {
+            if (nextLinkUrl == null)
+            {
+                throw new ArgumentNullException(nameof(nextLinkUrl));
+            }
 
             using var scope = clientDiagnostics.CreateScope("AllOperations.GetKeyValues");
             scope.Start();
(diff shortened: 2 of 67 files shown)


test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
```diff
@@ -15,40 +15,35 @@ namespace AutoRest.TestServer.Tests.Infrastructure
...
             var portPhrase = "Server started at port ";
             var startup = Path.Combine(GetBaseDirectory(), "legacy", "startup", "www.js");
-
-            var processStartInfo = new ProcessStartInfo("node", startup);
-
-            // Use random port
-            processStartInfo.Environment["PORT"] = "0";
-            processStartInfo.RedirectStandardOutput = true;
-            processStartInfo.RedirectStandardError = true;
+            var port = new Random().Next(10000, 65536).ToString();
```
MiYanni: For nextLink to provide the correct port on the test server, the port must be placed into the "PORT" environment variable. This cannot be done after the process has started, so we randomize the port before starting the process. These changes should work for both V1 and V2. However, there was a V2 issue in which it wouldn't use the environment port value properly. Laurent made an [update to the test server](https://github.com/Azure/autorest.testserver/pull/120) that should allow the tests to run on V2, but this PR doesn't update the test server. So, we can try it in a subsequent PR.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -185,16 +182,41 @@ private static ServiceClientParameter[] OrderParameters(IEnumerable<ServiceClien
...
                 request,
                 OrderParameters(methodParameters),
                 clientResponse,
-                new ClientMethodDiagnostics($"{clientName}.{operationName}", Array.Empty<DiagnosticScopeAttributes>())
+                new ClientMethodDiagnostics($"{clientName}.{operationName}",Array.Empty<DiagnosticScopeAttributes>()),
+                GetClientMethodPaging(operation, schemaForPaging)
             );
         }
 
+        private static ClientMethodPaging? GetClientMethodPaging(Operation operation, ObjectSchema? schema)
```
pakrym: Hm, wonder if we need something like `SchemaPropertyReference` that would be similar to schema type reference but with a property.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -185,16 +182,41 @@ private static ServiceClientParameter[] OrderParameters(IEnumerable<ServiceClien
...
+            if (pageable == null) return null;
+
+            var nextLinkName = pageable.GetValue<string>("nextLinkName");
+            //TODO: This should actually reference an operation: https://github.com/Azure/autorest.csharp/issues/397
+            var operationName = pageable.GetValue<string>("operationName");
+
+            var itemName = pageable.GetValue<string>("itemName");
+            //TODO: Hack to figure out the property name on the model
+            var itemProperty = schema?.Properties?.FirstOrDefault(p => p.SerializedName == itemName);
+            itemName = itemProperty?.CSharpName() ?? itemName ?? "Value";
```
pakrym: Is Value an appropriate default?

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -185,16 +182,41 @@ private static ServiceClientParameter[] OrderParameters(IEnumerable<ServiceClien
...
         }
 
+        private static ClientMethodPaging? GetClientMethodPaging(Operation operation, ObjectSchema? schema)
+        {
+            var pageable = operation.Extensions.GetValue<IDictionary<object, object>>("x-ms-pageable");
+            if (pageable == null) return null;
+
+            var nextLinkName = pageable.GetValue<string>("nextLinkName");
+            //TODO: This should actually reference an operation: https://github.com/Azure/autorest.csharp/issues/397
+            var operationName = pageable.GetValue<string>("operationName");
```
pakrym: Is the operation name used anywhere?

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -203,12 +225,19 @@ private static ServiceClientParameter BuildParameter(Parameter requestParameter)
...
+            }
+
             return new ServiceClientParameter(
                 requestParameter.CSharpName(),
                 CreateDescription(requestParameter),
                 ClientModelBuilderHelpers.CreateType(requestParameter.Schema, requestParameter.IsNullable()),
                 CreateDefaultValueConstant(requestParameter) ?? defaultValue,
-                requestParameter.Required == true);
+                requestParameter.Required == true,
+                location);
```
pakrym: Why do we need the parameter location?

src/AutoRest.CSharp.V3/ClientModels/ClientMethod.cs
```diff
@@ -5,22 +5,23 @@ namespace AutoRest.CSharp.V3.ClientModels
...
         }
 
         public string Name { get; }
-
         public string? Description { get; }
         public ClientMethodRequest Request { get; }
         public ServiceClientParameter[] Parameters { get; }
         public ClientMethodResponse Response { get; }
         public ClientMethodDiagnostics Diagnostics { get; }
+        public ClientMethodPaging? Paging { get; }
```
pakrym: I thought we talked about having this on the client level?

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -96,49 +105,81 @@ private void WriteParameter(CodeWriter writer, ServiceClientParameter clientPara
...
+            var methodName = nextPage ? CreateNextPageRequestMethodName(operation) : CreateRequestMethodName(operation);
+            writer.Append($"internal {typeof(HttpMessage)} {methodName}(");
+            var parameters = nextPage
+                ? operation.Parameters.Where(IsNextPageParameter).ToArray()
+                : operation.Parameters;
+            foreach (ServiceClientParameter clientParameter in parameters)
             {
                 writer.Append($"{_typeFactory.CreateInputType(clientParameter.Type)} {clientParameter.Name},");
             }
+            if (nextPage)
```
pakrym: I'm not sure I like this logic being on the writer side. Why not create another method definition on the  builder side and avoid changing the writer?

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -96,49 +105,81 @@ private void WriteParameter(CodeWriter writer, ServiceClientParameter clientPara
...
-                //TODO: Duplicate code between query and header parameter processing logic
-                foreach (var queryParameter in operation.Request.Query)
+                foreach (var header in operation.Request.Headers)
                 {
-                    WriteQueryParameter(writer, queryParameter);
+                    WriteHeader(writer, header);
                 }
 
-                if (operation.Request.Body is ObjectRequestBody body && body.Serialization is JsonSerialization jsonSerialization)
+                if (!nextPage && operation.Request.Body is ObjectRequestBody body && body.Serialization is JsonSerialization jsonSerialization)
```
pakrym: Having `nextPage` everywhere is very confusing.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -170,51 +211,58 @@ private void WriteRequestCreation(CodeWriter writer, ClientMethod operation)
...
             {
                 null when headerModelType != null => new CSharpType(typeof(ResponseWithHeaders<>), headerModelType),
                 { } when headerModelType == null => new CSharpType(typeof(Response<>), bodyType),
                 { } => new CSharpType(typeof(ResponseWithHeaders<>), bodyType, headerModelType),
                 _ => new CSharpType(typeof(Response)),
             };
-
+            var supportsPaging = operation.Paging != null;
+            var pageType = operation.Paging?.ItemType != null ? _typeFactory.CreateType(operation.Paging.ItemType) : new CSharpType(typeof(string));
+            responseType = supportsPaging ? new CSharpType(typeof(Page<>), pageType) : responseType;
```
pakrym: Original methods keep returning original types, we can create `Page<T>` in the iterator wrapper methods.

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
```diff
@@ -0,0 +1,100 @@
...
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Azure.Core
+{
+    internal static class PageResponseEnumerator
```
pakrym: I don't see this file added to the copy sources script.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
pakrym: This makes test server startup flaky. The random port we generate might be taken

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
```diff
@@ -15,40 +15,35 @@ namespace AutoRest.TestServer.Tests.Infrastructure
...
+            {
+                RedirectStandardOutput = true,
+                RedirectStandardError = true
+            };
+            processStartInfo.Environment["PORT"] = port;
 
             _process = Process.Start(processStartInfo);
             ProcessTracker.Add(_process);
-            while (!_process.HasExited)
+            while (!(_process?.HasExited ?? false))
```
pakrym: Debug.Assert(_process != null);

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
```diff
@@ -26,29 +26,26 @@ public TestServerV2()
...
+            {
+                RedirectStandardOutput = true,
+                RedirectStandardError = true
+            };
+            processStartInfo.Environment["PORT"] = port;
 
             _process = Process.Start(processStartInfo);
             ProcessTracker.Add(_process);
-            while (!_process.HasExited)
+            while (!(_process?.HasExited ?? false))
```
pakrym: Debug.Assert(_process != null);

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
```diff
@@ -26,29 +26,26 @@ public TestServerV2()
             var wiremock = Path.Combine(nodeModules, "wiremock", "jdeploy-bundle");
             var wiremockJar = Directory.GetFiles(wiremock, "*.jar").Single();
             var root = GetBaseDirectory();
-
-            var processStartInfo = new ProcessStartInfo("java", $"-jar {wiremockJar} --root-dir {root} --port 0");
-            // Use random port
-            processStartInfo.Environment["PORT"] = "0";
-            processStartInfo.RedirectStandardOutput = true;
-            processStartInfo.RedirectStandardError = true;
+            var port = new Random().Next(10000, 65536).ToString();
```
pakrym: We are skipping all V2 tests anyway so let's not to the port change.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: Yes. I actually have tests for it. 'value' is the default when itemName is unspecified. So, our property would get cased as 'Value'.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: https://github.com/Azure/autorest.csharp/issues/397

pakrym: This would have to be a reference when the feature is implemented, we can remove the property for now.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: The location of the parameter determines which parameters are in which methods. See `IsNextPageParameter` in ClientWriter.

src/AutoRest.CSharp.V3/ClientModels/ClientMethod.cs
MiYanni: There's too much shared information at this point. I'd have to separate this out which was outside of what I was trying to accomplish here. Overall, we need to find a way to duplicate information less, like a cleaner structure. In my restructure branch, I have a better idea of how everything interacts with each other, which will allow me to clean this up significantly.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
MiYanni: I don't like it here either. But, for now, it works and passes the tests, which will allow me to refactor much easier. This PR is to get a baseline for paging working.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
MiYanni: I agree.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
MiYanni: We need to restructure ClientMethod entirely so that 1 ClientMethod instance is 1 actual output method. Right now, it is not, and I piggy backed off of that to make this paging logic simply work.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -96,49 +105,81 @@ private void WriteParameter(CodeWriter writer, ServiceClientParameter clientPara
...
-                //TODO: Add logic to escape the strings when specified, using Uri.EscapeDataString(value);
-                //TODO: Need proper logic to convert the values to strings. Right now, everything is just using default ToString().
-                //TODO: Need logic to trim duplicate slashes (/) so when combined, you don't end  up with multiple // together
-                var urlText = String.Join(String.Empty, operation.Request.HostSegments.Select(s => s.IsConstant ? s.Constant.Value : "{" + s.Parameter.Name + "}"));
-                writer.Line($"request.Uri.Reset(new {writer.Type(typeof(Uri))}($\"{urlText}\"));");
+                writer.Line($"request.Method = {typeof(RequestMethod)}.{method.ToRequestMethodName()};");
 
-                foreach (var segment in operation.Request.PathSegments)
+                var uriType = writer.Type(typeof(Uri));
+                using (nextPage ? writer.If($"{uriType}.IsWellFormedUriString(nextLinkUrl, UriKind.Absolute)") : default)
```
pakrym: We don't need this logic, next link is always absolute when operation name is not set.

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
MiYanni: You cannot download this file as I've changed it from the original source. The original class isn't partial, or else I would have downloaded the original and then added other methods to it.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
pakrym: Why isn't it?

pakrym: What restructuring is required?

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
MiYanni: The V2 tests will not be skipped once the newest test server version is used. So, this will allow that ignore to be removed as soon as we update it. No point in removing it, and it made sense to do the same change in both places at once.

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
pakrym: Do we only need partial for the static helper method? I would rather not modify shared source files too much. Or do it in an automated fashion.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
MiYanni: Will change that.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
pakrym: https://github.com/Azure/autorest.testserver/pull/121

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
MiYanni: Will change that.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
pakrym: I think building 2 instances of ClientMethod would help with this.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
pakrym: I don't see what was fixed in the linked PR to make the V2 work

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
MiYanni: I would think that partial on any shared source would make sense, right? If it is shared source, anyone using it should be able to add to it. In this case, I initially wanted to add overloads for the methods already there. Now, I've very much changed the source (classes and originally methods are different), so I couldn't use partial anyway.

pakrym: Add a find-replace to mirroring script like this https://github.com/Azure/autorest.csharp/blob/feature/v3/eng/DownloadSharedSource.ps1#L13 so it's automatic

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
MiYanni: From what I'm told, the recordings now do not use base64 encoded strings. They use regular strings. This allows the utility method that would inject the correct port to work on the recordings. Obviously, I haven't tried it but I'm trusting Laurent on it.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
MiYanni: Removed. Added https://github.com/Azure/autorest.csharp/issues/403

samples/AppConfiguration/AppConfiguration/Generated/Operations/AllOperations.cs
```diff
@@ -61,6 +54,14 @@ internal HttpMessage CreateGetKeysRequest(string? name, string? after, string? a
             {
                 request.Uri.AppendQuery("After", after, true);
             }
+            if (syncToken != null)
```
pakrym: Why is this reordered?

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
MiYanni: As we discussed offline, for now, we are keeping ClientMethod to generate the Request, Async, and Sync methods. I made the PagingMethod stuff instead to generate for just the Pageable method (one that uses both the normal method and the NextPage method). Now, both the normal and NextPage method are simply ClientMethod.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -191,10 +197,63 @@ private static ServiceClientParameter[] OrderParameters(IEnumerable<ServiceClien
                 request,
                 OrderParameters(methodParameters),
                 clientResponse,
-                new ClientMethodDiagnostics($"{clientName}.{operationName}", Array.Empty<DiagnosticScopeAttributes>())
+                new ClientMethodDiagnostics($"{clientName}.{operationName}",Array.Empty<DiagnosticScopeAttributes>())
```
pakrym: revert unrelated changes

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -191,10 +197,63 @@ private static ServiceClientParameter[] OrderParameters(IEnumerable<ServiceClien
...
+                "The URL to the next page of results.",
+                new FrameworkTypeReference(typeof(string)),
+                null,
+                true,
+                ParameterLocation.Uri);
+            var parameters = method.Parameters.Where(IsNextPageParameter).Append(nextPageUrlParameter).ToArray();
+            var request = new ClientMethodRequest(
+                method.Request.Method,
+                parameters.Where(p => p.Location == ParameterLocation.Uri).Select(p => new ConstantOrParameter(p)).ToArray(),
+                new PathSegment[0],
```
pakrym: `Array.Empty<>()`

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -203,12 +262,19 @@ private static ServiceClientParameter BuildParameter(Parameter requestParameter)
                 defaultValue = ClientModelBuilderHelpers.ParseClientConstant(constantSchema);
             }
 
+            ParameterLocation? location = null;
```
pakrym: rem

samples/AppConfiguration/AppConfiguration/Generated/Operations/AllOperations.cs
MiYanni: Because Uri appending was moved to where Uri Resetting is. Originally, it was in an if-statement since they are both related to each other. Now that I'm not using that if statement, I'm leaving it so that Uri Reset and Uri Appends are one after the other. Before, Header appending was in-between both of them.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
```diff
@@ -191,10 +197,63 @@ private static ServiceClientParameter[] OrderParameters(IEnumerable<ServiceClien
...
+        private static ClientMethod BuildNextPageMethod(ClientMethod method)
+        {
+            var nextPageUrlParameter = new ServiceClientParameter(
+                "nextLinkUrl",
+                "The URL to the next page of results.",
+                new FrameworkTypeReference(typeof(string)),
+                null,
+                true,
+                ParameterLocation.Uri);
+            var parameters = method.Parameters.Where(IsNextPageParameter).Append(nextPageUrlParameter).ToArray();
```
pakrym: Why do we need to filter original parameters here?

src/AutoRest.CSharp.V3/ClientModels/ClientMethodPaging.cs
```diff
@@ -0,0 +1,27 @@
...
+            OperationName = operationName;
+        }
+
+        public string Name { get; }
+        public ClientMethod Method { get; }
+        public ClientMethod NextPageMethod { get; }
+        public string? NextLinkName { get; }
+        public string ItemName { get; }
+        public ClientTypeReference ItemType { get; }
+        public string? OperationName { get; }
```
pakrym: You shouldn't need this considering we have NextPageMethod  reference

src/AutoRest.CSharp.V3/ClientModels/ServiceClientParameter.cs
```diff
@@ -1,23 +1,27 @@
...
             IsRequired = isRequired;
+            Location = location;
         }
 
         public ClientTypeReference Type { get; }
         public string Name { get; }
         public string? Description { get; }
         public ClientConstant? DefaultValue { get; }
         public bool IsRequired { get; }
+        public ParameterLocation? Location { get; }
```
pakrym: Remove.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -114,30 +125,32 @@ private void WriteRequestCreation(CodeWriter writer, ClientMethod operation)
...
+                writer.Line($"request.Method = {typeof(RequestMethod)}.{method.ToRequestMethodName()};");
+
+                var uriType = writer.Type(typeof(Uri));
 
                 //TODO: Add logic to escape the strings when specified, using Uri.EscapeDataString(value);
                 //TODO: Need proper logic to convert the values to strings. Right now, everything is just using default ToString().
                 //TODO: Need logic to trim duplicate slashes (/) so when combined, you don't end  up with multiple // together
                 var urlText = String.Join(String.Empty, operation.Request.HostSegments.Select(s => s.IsConstant ? s.Constant.Value : "{" + s.Parameter.Name + "}"));
-                writer.Line($"request.Uri.Reset(new {writer.Type(typeof(Uri))}($\"{urlText}\"));");
+                writer.Line($"request.Uri.Reset(new {uriType}($\"{urlText}\"));");
```
pakrym: you can use `typeof(Uri)` in format string here, `writer.Type` shouldn't be required anymore.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -114,30 +125,32 @@ private void WriteRequestCreation(CodeWriter writer, ClientMethod operation)
...
 
                 foreach (var segment in operation.Request.PathSegments)
                 {
                     WritePathSegment(writer, segment);
                 }
 
-                foreach (var header in operation.Request.Headers)
```
pakrym: Undo

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -170,51 +183,43 @@ private void WriteRequestCreation(CodeWriter writer, ClientMethod operation)
...
-                CSharpType asyncReturnType = new CSharpType(typeof(ValueTask<>), responseType);
-                writer.Append($"public async {asyncReturnType} {methodName}Async(");
-            }
-            else
-            {
-                writer.Append($"public {responseType} {methodName}(");
-            }
+            var methodName = CreateMethodName(operation.Name, async);
+            CSharpType asyncReturnType = new CSharpType(typeof(ValueTask<>), responseType);
+            writer.Append($"public {(async ? "async " : string.Empty)}{(async ? asyncReturnType : responseType)} {methodName}(");
```
pakrym: This is too hard to read.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -254,6 +261,66 @@ private void WriteOperation(CodeWriter writer, ClientMethod operation, bool asyn
...
+                var pageWrappedType = new CSharpType(typeof(Page<>), pageType);
+                var funcType = async ? new CSharpType(typeof(Task<>), pageWrappedType) : pageWrappedType;
+                var nullableInt = new CSharpType(typeof(int), true);
+
+                var continuationTokenText = pagingMethod.NextLinkName != null ? $"response.Value.{pagingMethod.NextLinkName}" : "null";
+                var asyncText = async ? "async " : string.Empty;
+                var awaitText = async ? "await " : string.Empty;
+                using (writer.Scope($"{asyncText}{funcType} FirstPageFunc({nullableInt} pageSizeHint)"))
+                {
+                    writer.Append($"var response = {awaitText}{CreateMethodName(pagingMethod.Method.Name, async)}(");
```
pakrym: You need `CofigureAwait(false)` here.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: Ah, I see. https://stackoverflow.com/a/53071419/294804

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
```diff
@@ -462,21 +529,14 @@ ref valueVariable
                         writer.Line($"var headers = new {headersModelType}(message.Response);");
                     }
 
-                    switch (responseBody)
+                    var responseCallText = responseBody switch
```
pakrym: this is a wrong change, you care preformating FormattableString into a normal string and loosing type information embedded in the FormattableString. Logic we use to add usings won't work correctly. Undo.

test/AutoRest.TestServer.Tests/paging.cs
```diff
@@ -0,0 +1,375 @@
...
+                    product = "product";
+                }
+
+                id++;
+            }
+            Assert.AreEqual(10, pageNumber);
+        }, true);
+
+        [Test]
+        [Ignore("Needs retry logic: https://github.com/Azure/autorest.csharp/issues/398")]
```
pakrym: We have retry logic in the pipeline by default. This should work.

test/AutoRest.TestServer.Tests/paging.cs
```diff
@@ -0,0 +1,375 @@
...
+using Azure;
+using Azure.Core;
+using NUnit.Framework;
+using paging;
+using CustomPagingOperations = custom_baseUrl_paging.PagingOperations;
+
+namespace AutoRest.TestServer.Tests
+{
+    [IgnoreOnTestServer(TestServerVersion.V2, "Doesn't resolve port value properly to create nextLink URLs.")]
+    public class PagingTests : TestServerTestBase
```
pakrym: Add a sync foreach test.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
pakrym: We won't be merging this kind or port generation anyway. It would make our tests very flaky.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
MiYanni: Why? Why not write the Uri information and then the Headers? It is easier to read in the generated code, and makes no difference in operation.

samples/AppConfiguration/AppConfiguration/Generated/Operations/AllOperations.cs
pakrym: Ok, keeping thing next to each other makes sense.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs
pakrym: Yep, keep it. I thought it was an artifact or unrelated change.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: I can't remove Location logic or else BuildNextPageMethod won't work. It doesn't have access to the Parameter to get Parameter.Protocol.Http to make the decision on how to filter the parameters. So, it needs to be baked into ServiceClientParameter.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: Because NextPage doesn't take Path, Body, or Query parameters. That's the only reason we can use the WriteOperation logic as is, because we filter those here.

pakrym: What about just adding a single next page parameter?

src/AutoRest.CSharp.V3/ClientModels/ClientMethodPaging.cs
MiYanni: NextPageMethod doesn't have to do with these values. They are created from the extension information from the original method. That makes sense to do because I have to determine if the extension exists ahead of time to know the operation is intended to be a pageable operation. Since I already have the pageable dictionary, we calculate those values while creating this type. It is a similar pattern to the other types of this sort.
image

pakrym: I was only talking about OperationName. I don't think it's used right now or would ever be used.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
pakrym: I'm not sure if we need to be sending the original headers with GetNext call.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
MiYanni: So, compare these 2 files side-by-side:
https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/Shared/PageResponseEnumerator.cs

https://github.com/MiYanni/autorest.csharp/blob/paging-support/src/assets/Azure.Core.Shared/PageResponseEnumerator.cs

You'll see that the public methods have the same interface. But, everything internal to how they work has changed. The classes operate with 2 funcs instead of 1. And, I added additional public methods that was used by the generator. Do you think it makes sense to name this new version a different name? Name the classes differently? Basically, this new file doesn't have anything to do with the original, other than the original 4 public methods having the same interface. But, I felt it is 'in the spirit of' the original, just enhanced to allow 2 funcs to define a paging instead of 1.

test/AutoRest.TestServer.Tests/paging.cs
MiYanni: As it stands, we don't have multiple response types implemented. So, the 500 response code won't be there for this to work, right? So, I shelved this until we have things set up. And, I don't know how retries work when doing the IEnumerable/IAsyncEnumerable stuff (haven't designed it that far yet). It is okay for this PR to not do everything related to paging. It is an initial paging implementation.

src/AutoRest.CSharp.V3/CodeGen/ClientWriter.cs

src/AutoRest.CSharp.V3/ClientModels/ClientMethodPaging.cs
MiYanni: Ah. Conceptually, it will be replaced with something like Operation, which would link to an actual operation. It doesn't hurt to be there and shows a clear path for implementing https://github.com/Azure/autorest.csharp/issues/397. Likely, a bunch of this code will change anyway as x-ms-pageable will become actual data in the code model via https://github.com/Azure/autorest.modelerfour/issues/85. So, I'm not hung up on the accuracy of the data we're using/storing right now. Putting extra information in doesn't hurt if it is intended to help us later. I'm sure we'll both want to do more cleanup as the project progresses.

test/AutoRest.TestServer.Tests/paging.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: I originally only had only nextPageUrl with it. But, I realized it needed all parameters that are not related to the Uri itself. The headers are the only extra parameters in the paging tests on the test server, which include maxresults, which was what the pageSizeHint should be. We can ask today at the autorest meeting if services ever need headers when following nextLink urls. But, it doesn't hurt to have it send them sent for now. So, as an initial implementation of paging, this is fine. If we find we can remove it because it isn't supported, we'll remove it.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/ClientModels/ClientMethodPaging.cs
pakrym: But we already have concept for this it's the `NextPageMethod` reference.

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
pakrym: I didn't even notice that. Let me review the changes.

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        public static FuncPageable<T> CreateEnumerable<T>(Func<string?, Page<T>> pageFunc) where T : notnull
+        {
+            return new FuncPageable<T>((continuationToken, pageSizeHint) => pageFunc(continuationToken));
+        }
+
+        public static FuncPageable<T> CreateEnumerable<T>(Func<string?, int?, Page<T>> pageFunc) where T : notnull
```
pakrym: Do we need all the overloads?

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
pakrym: Rename the file it's very confusing otherwise.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
pakrym: I think my biggest issue is that I don't want more code taking dependency on location property of parameter. So if the feature is required I would rather design it differently.

MiYanni: How about, when the m4 changes for paging information are implemented, we revisit what is necessary and cut back/redesign? That way, we can have a working paging mechanism for something like AppConfiguration, and make the adjustments necessary to reduce it when we have the 'final' input from m4. Sound good?

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
MiYanni: The overloads are to be consistent with the original public interfaces. Since it is Azure.Core, I figured others may need want the additional flexibility. Also, I didn't know what interfaces the generator might need. Right now, I'm not using all of them. If these were only intended for the generator, I'd likely design the interface a bit differently. For example, the delegates I created would be part of the public interface instead of using different Funcs for everything.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
MiYanni: The port was random to begin with. We were just reading it from the test server instead of sending it. Unless I'm misunderstanding the code that was already there.

pakrym: When server asks for random port from the OS its guaranteed to get a free one when you are generating a random number there are no guarantees so the server startup might fail and it happens more often than you think considering how often we restart processes.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
pakrym: I don't understand why do you want to keep features that we don't need? Do we need headers for app configuration to work?

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
MiYanni: For now, I've renamed the class itself (and fixed a bug). I'll rename the file in the branch I have with all the other file renaming. Otherwise, it'll make merging more confusing.

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
pakrym: we have a type that's used by non generated code, this is purely for generator.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
MiYanni: Any chance we could simply make that same guarantee here? I'd really like having working paging tests so, sometime in the near future, implement the rest of the features for paging, with the assurance it passes the tests. If we have to wait for the test server change, this PR will have to keep waiting. And, I have my big rename PR based on this one.

src/assets/Azure.Core.Shared/PageResponseEnumerator.cs
MiYanni: Actually, I'm not renaming this now as it will cause issues with file name conflicts. I'm simply going to put the bug fix in. I'll rename in the next PR.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
pakrym: Nope, there is an inherent race there. We can merge test server PR and update it very quickly.

pakrym: And for V2 we don't even need the change.

pakrym: Testserver PR is in.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
pakrym: Ok, since we decided to send headers let's keep copying header paramters but still remove Location from parameter definition. We can figure out which parameters need copying just by looking at `method.Request.Headers`

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
pakrym: We can figure out correct parameters by looking at original method header values.

src/AutoRest.CSharp.V3/Properties/launchSettings.json
```diff
@@ -95,6 +95,10 @@
     "body-datetime": {
       "commandName": "Project",
       "commandLineArgs": "--standalone --input-codemodel=$(SolutionDir)\\test\\TestServerProjects\\body-datetime\\CodeModel.yaml --plugin=cs-modeler --output-path=$(SolutionDir)\\test\\TestServerProjects\\body-datetime --namespace=body_datetime"
+    },
```
pakrym: You've also generated `custom-baseUrl-paging` can you update launchSettings to include it?

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
```diff
@@ -15,40 +15,36 @@ namespace AutoRest.TestServer.Tests.Infrastructure
...
             var portPhrase = "Server started at port ";
             var startup = Path.Combine(GetBaseDirectory(), "legacy", "startup", "www.js");
-
-            var processStartInfo = new ProcessStartInfo("node", startup);
-
-            // Use random port
-            processStartInfo.Environment["PORT"] = "0";
-            processStartInfo.RedirectStandardOutput = true;
-            processStartInfo.RedirectStandardError = true;
+            var port = "0";
```
pakrym: Undo the unrelated changes

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV2.cs
MiYanni: The test server auth cred was expired, so no packages were being published. For now, Laurent has used a personal credential for it. After publishing 2.10.5, we found issues with the paging tests. He fixed them and published 2.10.6. Now, I was able to run some of the tests on V2.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
MiYanni: This matches V2 and is cleaner. We should just have 1 method for this between V1 and V2, but that's outside the scope of this PR. This code has no changes in logic, just basic housekeeping.

src/AutoRest.CSharp.V3/Properties/launchSettings.json
MiYanni: Done.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
pakrym: Why do you think it is cleaner?

MiYanni: - Uses object initializer instead of setting the values after creation.
- In V2, instead of having `port` set as `"0"` twice, it uses a variable and sets it that variable once.
- In V1, it is simply to mimic V2 for consistency between them.
- Better guarantee of the process not being null using `Debug.Assert`.
- Uses a design similar to [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) to reduce unnecessary nesting.

pakrym: > Uses object initializer instead of setting the values after creation.

Which makes things inconsistent because you have to set env variable anyway.

> In V2, instead of having port set as "0" twice, it uses a variable and sets it that variable once.

In v2 environment variable is not required and in V1 local is used only once.

MiYanni: But what does any of it matter? For something like this, where it isn't 'production code', nitpicking on how it *looks* doesn't make a difference at all. We both understand what the code says. I'd value spending more time implementing the rest of the features we need than being pedantic about test server code that we'll likely clean up in the future anyway. We're both working on this codebase. We can clean up anything *at any time*. As I had mentioned before Christmas break, let's focus on getting things done and being precise later. It is a much more valuable use of our time.

src/AutoRest.CSharp.V3/ClientModels/ServiceClientParameter.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: Got it working. Fixed.

src/AutoRest.CSharp.V3/ClientModels/ClientBuilder.cs
MiYanni: Fixed.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
pakrym: That's why I say, just revert it. The change barely helps anything, is unrelated to PR at hand and makes it harder to review.

package.json
```diff
@@ -9,7 +9,7 @@
   },
   "devDependencies": {
     "@autorest/autorest": "https://github.com/Azure/autorest/releases/download/autorest-3.0.6153/autorest-3.0.6153.tgz",
-    "@microsoft.azure/autorest.testserver": "^2.10.1"
+    "@microsoft.azure/autorest.testserver": "^2.10.6"
```
pakrym: Update to 2.10.7 (https://github.com/Azure/autorest.testserver/pull/123#event-2951715090)

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
MiYanni: I'll revert, but your points are invalid. You know the history of this change, thus modifying this code is not 'unrelated' to this PR in any way, since changing this code here is what initially allowed this PR to run tests. If I told you to change something and you had cosmetic changes around that thing, I'm not going to tell you to revert the cosmetic changes. That's just unnecessary work for you.

package.json
MiYanni: Done.

PR #402, opened 2020-01-13
Update everything to the latest versions

pakrym:


eng/Generate.ps1
```diff
@@ -20,8 +20,10 @@ function Invoke-AutoRest($baseOutput, $title, $autoRestArguments)
     }
 
     Write-Host "> $command"
+    pushd $repoRoot
```
pakrym: had issues with a wrong current directory for cmd invocations

eng/Generate.ps1
```diff
@@ -5,7 +5,7 @@ $ErrorActionPreference = 'Stop'
 
 function Invoke-AutoRest($baseOutput, $title, $autoRestArguments)
 {
-    $command = "npx autorest-beta $script:debugFlags $autoRestArguments"
+    $command = "npx  @autorest/autorest $script:debugFlags $autoRestArguments"
```
pakrym: using autorest-beta sometimes calls into global installation make sure we are always calling into local one

test/AutoRest.TestServer.Tests/ExtensionsClientNameTest.cs
```diff
@@ -15,5 +16,12 @@ public void ParametersAreRenamed()
             // TODO: Add more tests here
             TypeAsserts.HasParameter(method, "renamedBodyParameter");
         }
+
+        [Test]
+        public void PropertiesAreRenamed()
+        {
+            TypeAsserts.HasProperty(typeof(OriginalSchema), "RenamedProperty");
+            TypeAsserts.HasProperty(typeof(OriginalSchema), "RenamedPropertyString");
```
MiYanni: Is this testing that we (currently) re-case the names of properties? Or is this for future renaming configurability?

test/TestServerProjects/extension-client-name/Generated/Models/OriginalSchema.cs
```diff
@@ -9,6 +9,8 @@ namespace extension_client_name.Models.V100
     public partial class OriginalSchema
     {
         /// <summary> A description about the set of tags. </summary>
-        public IDictionary<string, string>? OriginalProperty { get; set; }
+        public IDictionary<string, string>? RenamedProperty { get; set; }
+        /// <summary> A description about the set of tags. </summary>
+        public string? RenamedPropertyString { get; set; }
```
MiYanni: What in the autorest update caused this to change? Also, do we need a CodeModel update for the new autorest version?

pakrym: M4 now picks up x-ms-client-name for more things.

test/AutoRest.TestServer.Tests/ExtensionsClientNameTest.cs
pakrym: It's testing that m4 applies x-ms-client-name to properties. It didn't before.

PR #405, opened 2020-01-15
Project reorg

MiYanni:
Relies on https://github.com/Azure/autorest.csharp/pull/400

Basically, I simplified a lot of the names and structured the project to have different categories/technologies. This makes it easier to understand how different parts of the project are supposed to work with each other. To view this, I recommend getting the source and looking at the structure in Visual Studio/Rider.


src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -74,11 +77,11 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
                         case SchemaTypeReference schemaTypeReference:
                             switch (typeFactory.ResolveReference(schemaTypeReference))
                             {
-                                case ClientObject _:
+                                case ObjectType _:
```
pakrym: I don't like that we use the same suffix for type references (ISchemaType) and client models (ObjectType )

src/AutoRest.CSharp.V3/Generation/Writers/ResponseHeaderGroupWriter.cs
```diff
@@ -1,27 +1,28 @@
...
         private readonly TypeFactory _typeFactory;
 
-        public ResponseHeaderModelModelWriter(TypeFactory typeFactory)
+        public ResponseHeaderGroupWriter(TypeFactory typeFactory)
         {
             _typeFactory = typeFactory;
         }
 
-        public void WriteHeaderModel(CodeWriter writer, ResponseHeaderModel responseHeaderModel)
+        public void WriteHeaderModel(CodeWriter writer, ResponseHeaderGroup responseHeaderGroup)
```
pakrym: Why isn't ResponseHeaderGroup a ResponseHeaderGroupType?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -5,68 +5,86 @@
...
+using AutoRest.CSharp.V3.Output.Models;
+using AutoRest.CSharp.V3.Output.Models.Requests;
+using AutoRest.CSharp.V3.Output.Models.Responses;
+using AutoRest.CSharp.V3.Output.Models.Serialization;
+using AutoRest.CSharp.V3.Output.Models.Shared;
+using AutoRest.CSharp.V3.Output.Models.TypeReferences;
 using AutoRest.CSharp.V3.Utilities;
 using Azure.Core;
-using HttpHeader = AutoRest.CSharp.V3.Pipeline.Generated.HttpHeader;
+using HttpHeader = AutoRest.CSharp.V3.Input.Generated.HttpHeader;
```
pakrym: Why is this required? There should be no name conflicts.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -5,68 +5,86 @@
...
-            var allClientParameters = arg.Operations
+            var allClientParameters = operationGroup.Operations
                 .SelectMany(op => op.Request.Parameters)
                 .Where(p => p.Implementation == ImplementationLocation.Client)
                 .Distinct();
-
+            Dictionary<string, Parameter> clientParameters = new Dictionary<string, Parameter>();
             // Deduplication required because of https://github.com/Azure/autorest.modelerfour/issues/100
-            foreach (Parameter clientParameter in allClientParameters)
+            foreach (Input.Generated.Parameter clientParameter in allClientParameters)
```
pakrym: No full qualification should be required anywhere.

readme.md
```diff
@@ -19,25 +19,25 @@ version: 3.0.6184
...
 ```
 
 ```yaml $(include-csproj)
 pipeline:
-  cs-asseter:
+  csharpassets:
     input: modelerfour/identity
-  cs-asseter/emitter:
-    input: cs-asseter
+  csharpassets/emitter:
```
pakrym: Should we call it csharpproj?

src/AutoRest.CSharp.V3/Output/Models/Requests/RequestParameter.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using AutoRest.CSharp.V3.Output.Models.Shared;
+using AutoRest.CSharp.V3.Output.Models.TypeReferences;
+
+namespace AutoRest.CSharp.V3.Output.Models.Requests
+{
+    internal struct RequestParameter
```
pakrym: I think this name is less descriptive than before.

src/AutoRest.CSharp.V3/Output/Models/Types/ObjectTypeConstant.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using AutoRest.CSharp.V3.Output.Models.Shared;
+using AutoRest.CSharp.V3.Output.Models.TypeReferences;
+
+namespace AutoRest.CSharp.V3.Output.Models.Types
+{
+    internal class ObjectTypeConstant
```
pakrym: delete.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: image

src/AutoRest.CSharp.V3/Generation/Writers/ResponseHeaderGroupWriter.cs
MiYanni: Done

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: image

readme.md
MiYanni: Done

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
pakrym: Sure, rename one of those types.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
pakrym: Right, let's not have multiple types with a same name in solution.

src/AutoRest.CSharp.V3/Output/Models/Requests/RequestParameter.cs
MiYanni: Added descriptiveness

src/AutoRest.CSharp.V3/Output/Models/Types/ObjectTypeConstant.cs
MiYanni: Deleted

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: You really want me to rename the type from the CodeModel? I didn't change any CodeModel types in this PR.

MiYanni: Done

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: Done

src/AutoRest.CSharp.V3/Output/Models/Requests/RequestParameterOrConstant.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using AutoRest.CSharp.V3.Output.Models.Shared;
+using AutoRest.CSharp.V3.Output.Models.TypeReferences;
+
+namespace AutoRest.CSharp.V3.Output.Models.Requests
+{
+    internal struct RequestParameterOrConstant
```
pakrym: What's the benefit of the new name?

MiYanni: Gives context on intended usage. I generally don't like the design of this class as is. But, I haven't put in effort to see what is needed to change it. My initial thought is to have an interface that represents a parameter and have 2 classes that implement it. So, you don't particularly care if it is a constant parameter or a regular parameter.

pakrym: Can we avoid renaming this one. To me benefit seems very low and considering this type is literally a discriminated union of `Constant` type or `Parameter` type I think the current name reflects it's purpose very well.

MiYanni:
The intent with this PR is to clarify class scope. Meaning, you can see when something crosses a namespace boundary and consider if it right to use that class or make a new class or adjust the design in some way. The previous project structure was loose and I couldn't get a feel for the intent behind certain classes or workflows.

src/AutoRest.CSharp.V3/Output/Models/Requests/RequestParameterOrConstant.cs
MiYanni: Is it a union between a `parameter` and a `constant` or between a `regular parameter` and a `constant parameter`? Are both parts a parameter in nature or are they completely 2 different things?

pakrym: It represent a value that's coming either from some parameter or is just a constant.

MiYanni: In essence, `ConstantOrParameter` and `RequestParameterOrConstant` is the same thing. But, I don't know enough of the intent here. My assumption is it is more likely to be a `parameter` than a `constant`, so having the word `parameter` first shows more common usage. So, `ParameterOrConstant` would be fine. Why are you against the word `Request` prepending it when it is in the `Requests` namespace?

pakrym: I don't like renaming for the sake of renaming. This type is extremely trivial and we both know what it does. There are 7 other types in that namespace without a prefix.

pakrym:
Remove the `AutoRest.CSharp.V3.Input.Generated` namespace, move types to `AutoRest.CSharp.V3.Input`

MiYanni:
@pakrym Should I just put the types into that namespace and keep them in the `Generated` folder? Or, should I remove the folder? It had been a standard practice to put generated code into a Generated folder, so it was obvious that it was generated.

pakrym:
Don't care about the folder but namespace seems excessive.

Issue #384, opened 2020-01-02
Add x-ms-client-name support


pakrym:
- [ ] Types
- [ ] Properties
- [ ] Parameters
- [ ] Headers

pakrym:
Seems that modeler should handle all this we just need a test to make sure that things are named as expected.

PR #408, opened 2020-01-16
Add x-ms-skip-url-encoding support

pakrym:
Fixes https://github.com/Azure/autorest.csharp/issues/384


src/assets/Azure.Core.Shared/RawRequestUriBuilder.cs
```diff
@@ -0,0 +1,83 @@
...
+                        Host = value.Substring(0, separator);
+
+                        _position = value[separator] == hostSeparator ? RawWritingPosition.Rest : RawWritingPosition.Port;
+
+                        value = value.Substring(separator + 1);
+                    }
+                }
+                else if (_position == RawWritingPosition.Port)
+                {
+                    int separator = value.IndexOf(hostSeparator);
```
MiYanni: Shouldn't this be `portSeparator`?

pakrym: Nope, at the point where we are parsing port the next separator is '/'

Issue #411, opened 2020-01-22
Change path appending strategy to allow for query params in path parts


MiYanni:
`x-ms-paths` information allows for the raw path and (potentially) path parameters to have query information in them. This happens for `nextLink` in some of the paging tests on the test server.

PR #412, opened 2020-01-22
Paging operation name

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/397

This adds in supporting `operationName`. There is only 1 test that I can make pass right now with this change. The other tests need https://github.com/Azure/autorest.csharp/issues/411.

### DIFF
--- a/src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
+++ b/src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
@@ -44,21 +44,27 @@ namespace AutoRest.CSharp.V3.Output.Builders
                 }
             }
 
+            List<Method> nextPageMethods = new List<Method>();
             List<Paging> pagingMethods = new List<Paging>();
-            foreach ((string defaultOperationName, (Operation operation, Method method)) in operationMethods)
+            foreach ((string _, (Operation operation, Method method)) in operationMethods)
             {
                 IDictionary<object, object>? pageable = operation.Extensions.GetValue<IDictionary<object, object>>("x-ms-pageable");
                 if (pageable != null)
                 {
-                    string? operationName = pageable.GetValue<string>("operationName");
+                    //TODO: Assuming operationName is in this operation group: https://github.com/Azure/autorest.modelerfour/issues/85
+                    string? operationName = pageable.GetValue<string>("operationName")?.Split('_').Last();
                     (Operation NextOperation, Method NextMethod)? next =
                         operationName != null && operationMethods.TryGetValue(operationName,
                             out (Operation NextOperation, Method NextMethod) nextOperationMethod)
                             ? nextOperationMethod
                             : ((Operation NextOperation, Method NextMethod)?)null;
                     // If there is no operationName or we didn't find an existing operation, we use the original method to construct the nextPageMethod.
-                    Method nextPageMethod = BuildNextPageMethod(next?.NextMethod ?? method);
-                    operationMethods.Add(defaultOperationName, (operation, nextPageMethod));
+                    Method nextPageMethod = next?.NextMethod ?? BuildNextPageMethod(method);
+                    // Only add the method if it didn't previously exist
+                    if (next == null)
+                    {
+                        nextPageMethods.Add(nextPageMethod);
+                    }
                     //TODO: This is a hack since we don't have the model information at this point
                     ObjectSchema? schemaForPaging = ((method.Response.ResponseBody as ObjectResponseBody)?.Type as SchemaTypeReference)?.Schema as ObjectSchema;
                     Paging pagingMethod = GetClientMethodPaging(method, nextPageMethod, pageable, schemaForPaging);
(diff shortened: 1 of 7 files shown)


src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -34,31 +34,49 @@ public static Client BuildClient(OperationGroup operationGroup)
...
-                    methods.Add(method);
-                    var pageable = operation.Extensions.GetValue<IDictionary<object, object>>("x-ms-pageable");
-                    if (pageable != null)
+                    operationMethods.Add(operation.Language.Default.Name, (operation, method));
+                }
+            }
+
+            List<Method> nextPageMethods = new List<Method>();
+            List<Paging> pagingMethods = new List<Paging>();
+            foreach ((string _, (Operation operation, Method method)) in operationMethods)
```
pakrym: Lets add a private struct for `(Operation operation, Method method)`. Right now it makes code very verbose.

pakrym: operationMethods.Values?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -34,31 +34,49 @@ public static Client BuildClient(OperationGroup operationGroup)
...
+            {
+                IDictionary<object, object>? pageable = operation.Extensions.GetValue<IDictionary<object, object>>("x-ms-pageable");
+                if (pageable != null)
+                {
+                    //TODO: Assuming operationName is in this operation group: https://github.com/Azure/autorest.modelerfour/issues/85
+                    string? operationName = pageable.GetValue<string>("operationName")?.Split('_').Last();
+                    (Operation NextOperation, Method NextMethod)? next =
+                        operationName != null && operationMethods.TryGetValue(operationName,
```
pakrym: Should we throw if the `operationName` is not found?

MiYanni: I really, really want M4 to handle stuff like this. IMO, at our point in the process, we shouldn't need to validate the data we received. But, this is a weird case. Do we error for anything else related to the data right now?

pakrym: > I really, really want M4 to handle stuff like this.

Me too.

I'm just trying to figure out what would be the most confusing option for customers getting a new method generated or an error.

> Do we error for anything else related to the data right now?

We are, often not cleanly, but a parameter that's defined in path but not in parameter list would result in error.

MiYanni: Thinking about it a bit more, we should probably **not** go with a design that 'massages' issues away. I do have insight from the PowerShell generator around errors. I only had about 3-4 errors that could happen. Whenever a customer hit one, they wouldn't look into the error. They either created an issue in GitHub or messaged us directly about it. Part of the problem was where the error occurred in the code, we didn't have all the data necessary to make a detailed, actionable error message. In this situation, I can at least make a message that has enough context for them to investigate. So, I'll add that now.

MiYanni: Error added. Also created: https://github.com/Azure/autorest.testserver/issues/129

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: Done.

MiYanni: Done.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -34,34 +34,72 @@ public static Client BuildClient(OperationGroup operationGroup)
...
-                        methods.Add(nextPageMethod);
-                        //TODO: This is a hack since we don't have the model information at this point
-                        var schemaForPaging = ((method.Response.ResponseBody as ObjectResponseBody)?.Type as SchemaTypeReference)?.Schema as ObjectSchema;
-                        Paging pagingMethod = GetClientMethodPaging(method, nextPageMethod, pageable, schemaForPaging);
-                        pagingMethods.Add(pagingMethod);
+                        if (!processedMethods.TryGetValue(operationName, out OperationMethod nextOperationMethod))
+                        {
+                            string processedOperationName = processed.Operation.Language.Default.Name;
+                            throw new Exception(
+                                $"The x-ms-pageable operationName \"{operationGroup.Key}_{operationName}\" for operation {operationGroup.Key}_{processedOperationName} was not found.");
```
pakrym: Use the original value of `operationName` extension here, no need to split and then re-join.

MiYanni: Done.

PR #417, opened 2020-01-22
Null check: empty line bug

MiYanni:
Fixed inserting an empty line at the start of the operation methods if no null checks existed. (It was annoying the crap out of me)

### DIFF
--- a/src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
+++ b/src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
+++ b/src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
@@ -366,7 +367,7 @@ namespace AutoRest.CSharp.V3.Generation.Writers
                 }
             }
 
-            if (parameters.Count > 0)
+            if (parameters.Any())
             {
                 writer.Line();
             }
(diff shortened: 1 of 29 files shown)


pakrym:
Good one. I wondered if we can do something in code writer to avoid this in all cases.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -365,7 +366,11 @@ private void WriteParameterNullChecks(CodeWriter writer, IEnumerable<Parameter>
                     }
                 }
             }
-            writer.Line();
+
+            if (parameters.Count > 0)
```
pakrym: Any() and keep it enumerable?

MiYanni: Multiple enumerations of it would occur. Using IReadOnlyCollection is more accurate since we only ever provide input as an array anyway.

pakrym: Nope, `Any` knows to downcast to IList before doing the enumeration.

pakrym: https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq/src/System/Linq/AnyAll.cs#L12

MiYanni: Right, if the input is IList. But, having IEnumerable as the parameter type says it can be anything that implements IEnumerable. So it still could have multiple enumerations (that's what I meant in my first comment). At least with IReadOnlyCollection, we are guaranteeing that we are only reading the values and have access to `.Count` naturally. I believe that is the point of that interface, right? I'd rather guaranteed non-multiple enumerations of the collection (for anything, not just this method). It is less issue prone (performance) that way.

pakrym: Don't want to spend to much time on this, just know that .Any is optimized not to start enumeration when underlying structure has length.

MiYanni: Yeah, I didn't know that explicitly. I use `Any` in most cases. I'll change it to Any here, since with IReadOnlyCollection, it is guaranteed to use `.Count`.

PR #418, opened 2020-01-23
Add a CognitiveSearch sample

tg-msft:
There are a few hacks and tweaks to the swagger readme, the generator, and the supporting shared source.  It'll be easier to view by commit to catch the meaningful changes.


pakrym:
Did you manually change the code after generation?

tg-msft:
No, but I was only running `./eng/Generate.ps1 CognitiveSearch`.  Let me try running the entire script and see if that takes care of the diff check.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -134,6 +134,12 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
                             writer.Append($"({name}")
                                 .AppendNullableValue(implementationType);
 
+                            // TODO: Turn a char into a string so it can be passed to WriteStringValue
```
pakrym: Can you add an overload to https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Azure.Core.Shared/Utf8JsonWriterExtensions.cs#L12

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -347,6 +353,14 @@ public static void ToDeserializeCall(this CodeWriter writer, JsonValueSerializat
                     }
 
                     writer.AppendRaw(")");
+
+                    // TODO: Turn a GetString() into a char in a safer way
+                    if (frameworkType == typeof(char))
```
pakrym: Can you add a GetChar method to https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Azure.Core.Shared/JsonElementExtensions.cs#L65 and use it higher up

tg-msft:
Filed #419 to track fixing the `request` rename in the swagger README transform.

PR #421, opened 2020-01-23
Versions update

MiYanni:
Update autorest, autorest-core, testserver, modelerfour, and the code model to latest versions.

Created an issue for the descriptions being broken here: https://github.com/Azure/autorest.modelerfour/issues/136


src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -18,6 +18,7 @@ internal class CSharpProj : IPlugin
     <TargetFramework>netstandard2.0</TargetFramework>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Nullable>annotations</Nullable>
+    <NoWarn>CS1591</NoWarn>
```
pakrym: Don't add it here, instead include it into a shared props file https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestServerProjects/Directory.Build.props

MiYanni: Done.

PR #422, opened 2020-01-23
Use existing source code as input

pakrym:


src/AutoRest.CSharp.V3/Properties/launchSettings.json
```diff
@@ -1,108 +1,116 @@
 {
   "profiles": {
-    "body-integer": {
+    "TypeSchemaMapping": {
```
MiYanni: Resort these.

PR #430, opened 2020-01-24
Add support for generator package publish

pakrym:
As an artifact until I figure out how to publish to github releases.

### DIFF
--- a/src/AutoRest.CSharp.V3/AutoRest.CSharp.V3.csproj
+++ b/src/AutoRest.CSharp.V3/AutoRest.CSharp.V3.csproj
@@ -6,6 +6,7 @@
     <Nullable>enable</Nullable>
     <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
     <RunAnalyzersDuringLiveAnalysis>false</RunAnalyzersDuringLiveAnalysis>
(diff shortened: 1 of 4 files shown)


src/AutoRest.CSharp.V3/AutoRest.CSharp.V3.csproj
```diff
@@ -22,6 +23,23 @@
 
   <ItemGroup>
     <None Include="../../src/assets/**/*.cs" CopyToOutputDirectory="PreserveNewest" />
+    <None Include="package.json" CopyToOutputDirectory="PreserveNewest" />
+    <None Include="../../readme.md" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
+  <Target Name="NpmPackage" DependsOnTargets="Build">
+    <Exec Command="npm pack" WorkingDirectory="$(OutputPath)" ConsoleToMsBuild="true">
```
MiYanni: What's the reason to put the `npm pack` logic here (baked into csproj) instead of being in the devops pipeline?

pakrym: Easier to find the output dll and put it into the root of the package.

PR #434, opened 2020-01-27
Model flattening support (JSON only)

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/416

We don't have test server tests for model flattening for XML. So, this only does JSON. Issue for XML: https://github.com/Azure/autorest.csharp/issues/433


src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -17,9 +17,6 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
         {
             writerName ??= w => w.AppendRaw("writer");
 
-            var type = serialization.Type;
-            CSharpType implementationType = typeFactory.CreateType(type);
```
MiYanni: The type factory requires schemas created from the swagger. We make dynamic data (property + schema) for flattening. This means the factory won't be able to resolve that, so we can't use it in serialization.

The good news is that this data was redundant (in both serialization and deserialization). It is only used for nullability and if it is a value type. We know this data already, so I use the data that already exists.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -83,7 +80,7 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
 
                                 case EnumType clientEnum:
                                     writer.Append($"{writerName}.WriteStringValue({name}")
-                                        .AppendNullableValue(implementationType)
+                                        .AppendNullableValue(serialization.Type.IsNullable, isValueType: true)
```
MiYanni: Enums are value types (both `enum` and our `struct` extensible enum).

src/AutoRest.CSharp.V3/Output/Models/Serialization/Json/JsonPropertySerialization.cs
```diff
@@ -5,15 +5,17 @@ namespace AutoRest.CSharp.V3.Output.Models.Serialization.Json
 {
     internal class JsonPropertySerialization
     {
-        public JsonPropertySerialization(string name, string memberName, JsonSerialization valueSerialization)
+        public JsonPropertySerialization(string name, string memberName, JsonSerialization valueSerialization, bool? isFlattened)
```
MiYanni: Nullable bool parameter since the only time the constructor is called, the `IsBag` property on Property is nullable bool. We just feed it directly into this.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,26 +153,82 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
+            },
+            Required = true,
+            Language = new Languages { Default = new Language { Name = name } }
+        };
+
+        private static void PopulatePropertyBag(PropertyBag propertyBag, int depthIndex)
+        {
+            foreach (Property property in propertyBag.Properties.ToArray())
+            {
+                if (!(property.FlattenedNames?.Any() ?? false) || depthIndex >= property.FlattenedNames!.Count - 1)
```
pakrym: this condition is hella complicated, simplify it

test/TestServerProjects/body-complex/Generated/Models/MyBaseType.cs
```diff
@@ -16,7 +16,6 @@ public MyBaseType()
         /// <summary> The constant value Kind1. </summary>
         public string Kind { get; internal set; } = "Kind1";
         public string PropB1 { get; set; }
-        /// <summary> MISSING·SCHEMA-DESCRIPTION-OBJECTSCHEMA. </summary>
-        public MyBaseHelperType Helper { get; set; }
+        public string PropBH1 { get; set; }
```
pakrym: Didn't we ever use this in test?

test/AutoRest.TestServer.Tests/model-flattening.cs
```diff
@@ -0,0 +1,269 @@
...
+{
+    public class ModelFlatteningTests : TestServerTestBase
+    {
+        public ModelFlatteningTests(TestServerVersion version) : base(version, "model-flatten") { }
+
+        [Test]
+        public Task GetModelFlattenArray() => Test(async (host, pipeline) =>
+        {
+            var result = await new AllOperations(ClientDiagnostics, pipeline, host).GetArrayAsync();
+            Assert.AreEqual("1", result.Value.ElementAt(0).Id);
```
pakrym: Extract variables for commonly used expression like `result.Value.ElementAt(0)`

src/AutoRest.CSharp.V3/Properties/launchSettings.json
```diff
@@ -115,6 +115,10 @@
     "xml-service": {
       "commandLineArgs": "--standalone --input-codemodel=$(SolutionDir)\\test\\TestServerProjects\\xml-service\\CodeModel.yaml --plugin=csharpgen --output-folder=$(SolutionDir)\\test\\TestServerProjects\\xml-service --namespace=xml_service",
       "commandName": "Project"
+    },
+    "model-flattening": {
```
pakrym: use the generate.ps1 to update this, names should be in alphabetical order.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,26 +153,82 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
+
+        private static IEnumerable<Property> AggregateProperties(ObjectSchema schema)
+        {
+            PropertyBag propertyBag = new PropertyBag();
+            propertyBag.Properties.AddRange(schema.Properties ?? Array.Empty<Property>());
+            PopulatePropertyBag(propertyBag, 0);
+            return propertyBag.Properties;
+        }
+
+        private static Property CreatePropertyFromPropertyBag(string name, PropertyBag bag) => new Property
```
pakrym: Let's try to implement this without synthesizing new code model types.

test/TestServerProjects/body-complex/Generated/Models/MyBaseType.cs
MiYanni: It isn't used in any test server tests. It is defined in the swagger for it, but no tests set/get that value.

test/AutoRest.TestServer.Tests/model-flattening.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Properties/launchSettings.json
MiYanni: Done.

src/AutoRest.CSharp.V3/Input/CodeModelPartials.cs
```diff
@@ -30,6 +31,11 @@ public ObjectSchema()
         }
     }
 
+    internal partial class Property
```
pakrym: revert this file.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,24 +150,70 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
+        }
+
         private static JsonObjectSerialization BuildJsonObjectSerialization(ObjectSchema objectSchema, TypeReference schemaTypeReference, bool isNullable)
         {
             List<JsonPropertySerialization> serializationProperties = new List<JsonPropertySerialization>();
             foreach (var schema in EnumerateHierarchy(objectSchema))
             {
-                foreach (Property property in schema.Properties!)
+                PropertyBag propertyBag = new PropertyBag();
+                propertyBag.Properties.AddRange(schema.Properties ?? Array.Empty<Property>());
```
pakrym: I think Properties is always initialized.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,24 +150,70 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
+            public Dictionary<string, PropertyBag> Bag { get; } = new Dictionary<string, PropertyBag>();
+            public List<Property> Properties { get; } = new List<Property>();
+        }
+
+        private static void PopulatePropertyBag(PropertyBag propertyBag, int depthIndex)
+        {
+            foreach (Property property in propertyBag.Properties.ToArray())
+            {
+                int flattenNamesCount = property.FlattenedNames?.Count ?? 0;
+                if (flattenNamesCount == 0 || depthIndex >= flattenNamesCount - 1)
```
pakrym: Can't you merge these two?

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,24 +150,70 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
-                    JsonPropertySerialization propertySerialization = new JsonPropertySerialization(
-                        property.SerializedName,
-                        property.CSharpName(),
-                        BuildSerialization(property.Schema, property.IsNullable())
-                    );
+                    continue;
+                }
 
-                    serializationProperties.Add(propertySerialization);
+                string name = property!.FlattenedNames.ElementAt(depthIndex);
```
pakrym: Why do you need ! here?

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,24 +150,70 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
+                    namedBag = new PropertyBag();
+                    propertyBag.Bag.Add(name, namedBag);
                 }
+
+                namedBag.Properties.Add(property);
+                propertyBag.Properties.Remove(property);
             }
 
-            return new JsonObjectSerialization(schemaTypeReference, serializationProperties.ToArray(), CreateAdditionalProperties(objectSchema));
+            foreach ((string name, PropertyBag innerBag) in propertyBag.Bag)
```
pakrym: foreach on Bag.Values

src/AutoRest.CSharp.V3/Output/Models/TypeReferences/SchemaTypeReference.cs
```diff
@@ -7,13 +7,13 @@ namespace AutoRest.CSharp.V3.Output.Models.TypeReferences
 {
     internal class SchemaTypeReference: TypeReference
     {
-        public SchemaTypeReference(Schema schema, bool isNullable)
+        public SchemaTypeReference(Schema? schema, bool isNullable)
```
pakrym: Why is this nullable?

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
pakrym: I'd like to keep the original logic now that it's possible, we are going to have struct models soon.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,24 +150,70 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
+                    property.SerializedName,
+                    property.CSharpName(),
+                    BuildSerialization(property.Schema, property.IsNullable()),
+                    false);
+            }
+
+            foreach ((string name, PropertyBag innerBag) in propertyBag.Bag)
+            {
+                JsonPropertySerialization[] serializationProperties = GetPropertySerializationsFromBag(innerBag).ToArray();
+                JsonObjectSerialization objectSerialization = new JsonObjectSerialization(new SchemaTypeReference(null, false), serializationProperties, null);
```
pakrym: Make type property on JsonObjectSerialization nullable instead of SchemaTypeReference

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,24 +150,70 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
+                yield return new JsonPropertySerialization(name, name, objectSerialization, true);
```
pakrym: Serialized and source names are the same?

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -150,24 +150,70 @@ private static XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema o
...
+
         private static JsonObjectSerialization BuildJsonObjectSerialization(ObjectSchema objectSchema, TypeReference schemaTypeReference, bool isNullable)
         {
             List<JsonPropertySerialization> serializationProperties = new List<JsonPropertySerialization>();
             foreach (var schema in EnumerateHierarchy(objectSchema))
             {
-                foreach (Property property in schema.Properties!)
+                PropertyBag propertyBag = new PropertyBag();
+                propertyBag.Properties.AddRange(schema.Properties ?? Array.Empty<Property>());
+                PopulatePropertyBag(propertyBag, 0);
```
pakrym: Shouldn't this call be outside of the loop?

src/AutoRest.CSharp.V3/Output/Models/TypeReferences/SchemaTypeReference.cs
pakrym: Found it, let's rather make Type on object serialization nullable.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
pakrym: Should we make one of the names nullable and remove IsFlattened?

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: Fixed. Removed the first condition as it was redundant. If there are no flatten names, the `count - 1` will be -1 and the first depthIndex is 0. Therefore, 0 is greater or equal to -1, which means will continue properly.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs

src/AutoRest.CSharp.V3/Input/CodeModelPartials.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: Because we do `property.FlattenedNames?.Count ?? 0` before this, it can't guarantee that FlattenNames is not null. However, our condition itself (logically) means that we have FlattenNames to work with.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs

src/AutoRest.CSharp.V3/Output/Models/TypeReferences/SchemaTypeReference.cs
MiYanni: That was the first thing I tried. However, I went with changing Schema to nullable for a number of reasons:
1. `Type` on JsonObjectSerialization is an override of JsonSerialization's abstract Type. You cannot make an override change nullability.
2. If you did change JsonSerialization's abstract Type to nullable, it is used all over the place (switches, accessing IsNullable, etc). I felt it was outside the necessary changes for this PR.
3. In concept, this makes the most sense since I have a type reference that has no schema. I was considering making a new TypeReference type (something that just has IsNullable), but this worked well with minimal changes.
4. There are only 2 uses of the Schema property. One is my hack I put in for paging. The other is `ResolveReference` which is checking for equality. So, changing it to null affects nothing in usage for Schema.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: Explanation above.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Output/Models/TypeReferences/SchemaTypeReference.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: `GetPropertySerializationsFromBag` uses a populated property bag. So, the next line needs that to add the correct property serializations to the list.

pakrym: We can pull next line out of the loop too.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
MiYanni: Because of the changes we just made, we cannot always create a type (as it may not have one). We've worked around this by making the logic below recognize if we have a type or not.

pakrym: Are you sure? I think all places you've changed have provably not-null type.

MiYanni: Fixed. We're now using the CSharpType IsNullable instead of the TypeReference one. Also reverted the overload for AppendNullableValue.

pakrym:
Nice!

PR #435, opened 2020-01-30
Remove TypeReference concept

pakrym:
Replace them with direct CSharpType references. 

Pull more logic into model building phase.


eng/Generate.ps1
```diff
@@ -1,11 +1,25 @@
 #Requires -Version 6.0
-param($name, [switch]$noDebug, [switch]$reset, [switch]$noBuild, [switch]$fast, [switch]$updateLaunchSettings)
+param($name, [switch]$noDebug, [switch]$reset, [switch]$noBuild, [switch]$fast, [switch]$updateLaunchSettings, [switch]$removeGenerated = $true)
```
MiYanni: Maybe change `removedGenerated` to `clean` instead.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -270,28 +285,32 @@ private static Method BuildNextPageMethod(Method method)
                 method.Diagnostics);
         }
 
-        private static Paging GetClientMethodPaging(Method method, Method nextPageMethod, IDictionary<object, object> pageable, ObjectSchema? schema)
+        private Paging GetClientMethodPaging(Method method, Method nextPageMethod, IDictionary<object, object> pageable, ObjectType type)
```
MiYanni: Good job on cleanup of this method's implementation.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -270,28 +285,32 @@ private static Method BuildNextPageMethod(Method method)
...
-            itemName = itemProperty?.CSharpName() ?? itemName ?? "Value";
-            var nextLinkProperty = schema?.Properties?.FirstOrDefault(p => p.SerializedName == nextLinkName);
-            nextLinkName = nextLinkProperty?.CSharpName() ?? nextLinkName;
-            // If itemName resolves to Value, we can't use itemProperty. So, get the correct property.
-            var itemTypeProperty = schema?.Properties?.FirstOrDefault(p => p.CSharpName() == itemName);
-            var itemTypeValueSchema = (itemTypeProperty?.Schema as ArraySchema)?.ElementType;
-            var itemType = BuilderHelpers.CreateType(itemTypeValueSchema ?? new Schema(), false);
-            var name = $"{method.Name}Pageable";
-            return new Paging(method, nextPageMethod, name, nextLinkName, itemName, itemType);
+            var nextLinkName = pageable.GetValue<string>("nextLinkName") ?? "nextLink";
```
MiYanni: This is a bug. We cannot default to `nextLink` when no name is provided. There are tests specifically for that.

test/TestServerProjects/paging/Generated/Operations/PagingOperations.cs
```diff
@@ -2353,12 +2353,12 @@ public AsyncPageable<Product> GetNullNextLinkNamePagesPageableAsync(Cancellation
             async Task<Page<Product>> FirstPageFunc(int? pageSizeHint)
             {
                 var response = await GetNullNextLinkNamePagesAsync(cancellationToken).ConfigureAwait(false);
-                return Page.FromValues(response.Value.Values, null, response.GetRawResponse());
+                return Page.FromValues(response.Value.Values, response.Value.NextLink, response.GetRawResponse());
```
MiYanni: As mentioned above, it was null because there was a not a value provided. We still need to support this.

pakrym: Super strange but ok :)

PR #437, opened 2020-01-31
Http infrastructure tests

MiYanni:
So, this PR fixes a few bugs. It also added `ResponseMethodAdditional` to cover methods that aren't in the original `ResponseMethod`. I ended up generating `head`, `head-exceptions`, and `httpInfrastructure.quirks`, but didn't need them for these tests. So, I can remove those if we don't want them generating.


src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -340,16 +340,19 @@ private void WritePagingOperation(CodeWriter writer, Paging pagingMethod, bool a
...
                 WriteConstant(writer, constantOrParameter.Constant);
             }
             else
             {
-                writer.AppendRaw(constantOrParameter.Parameter.Name)
-                    .AppendNullableValue(_typeFactory.CreateType(constantOrParameter.Type));
+                writer.AppendRaw(constantOrParameter.Parameter.Name);
+                if (!ignoreNullability)
+                {
+                    writer.AppendNullableValue(_typeFactory.CreateType(constantOrParameter.Type));
```
MiYanni: This was for the thing where we had nullable parameters for the CreateRequest method. In that situation, we decided to just ignore the nullability aspect of it.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -563,6 +566,8 @@ ref valueVariable
                         case { }:
                             writer.Append($"return {typeof(Response)}.FromValue({valueVariable}, message.Response);");
                             break;
+                        case null when !statusCodes.Any():
+                            break;
```
MiYanni: This allows operations to be created that have no responses.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -362,19 +362,6 @@ PathSegment TextSegment(string text)
...
-            HttpMethod.Delete => RequestMethod.Delete,
-            HttpMethod.Get => RequestMethod.Get,
-            HttpMethod.Head => RequestMethod.Head,
-            HttpMethod.Options => (RequestMethod?)null,
-            HttpMethod.Patch => RequestMethod.Patch,
-            HttpMethod.Post => RequestMethod.Post,
-            HttpMethod.Put => RequestMethod.Put,
-            HttpMethod.Trace => null,
-            _ => null
-        };
```
MiYanni: I moved this to be in the same place were we have the 'opposite' method of it. It made more sense to keep them together so that if changes are needed, it is obvious that both need changed.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerTestBase.cs
```diff
@@ -73,7 +73,7 @@ public Task TestStatus(Func<string, HttpPipeline, Task<Response>> test, bool ign
         private Task TestStatus(string scenario, Func<string, HttpPipeline, Task<Response>> test, bool ignoreScenario = false) => Test(scenario, async (host, pipeline) =>
         {
             var response = await test(host, pipeline);
-            Assert.That(response.Status, Is.EqualTo(200).Or.EqualTo(201), "Unexpected response " + response.ReasonPhrase);
+            Assert.That(response.Status, Is.EqualTo(200).Or.EqualTo(201).Or.EqualTo(202).Or.EqualTo(204), "Unexpected response " + response.ReasonPhrase);
```
MiYanni: Allowed other accept values for tests that returned them. Pretty much anything in 200's should be good, but these tests only used 200, 201, 202, and 204.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
pakrym: https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.empty?view=netframework-4.8

src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -96,7 +96,7 @@ private KnownMediaType[] GetMediaTypes(CodeModel codeModel)
                         types.Add(bodyRequest.KnownMediaType);
                     }
 
-                    foreach (var response in operation.Responses!)
+                    foreach (var response in operation.Responses ?? Array.Empty<ServiceResponse>())
```
pakrym: I was thinking recently if we should start had writing the input model and make it nice to consume by fixing all this nullable noise.

pakrym:
> I ended up generating head, head-exceptions, and httpInfrastructure.quirks, but didn't need them for these tests. So, I can remove those if we don't want them generating.

Let's keep them not generated so the generation list is a better indication of what is covered.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -29,7 +29,7 @@ public static ObjectSerialization BuildObject(KnownMediaType mediaType, ObjectSc
             }
         }
 
-        public static ObjectSerialization Build(KnownMediaType mediaType, Schema schema, bool isNullable)
+        public static ObjectSerialization Build(KnownMediaType? mediaType, Schema schema, bool isNullable)
```
pakrym: Doesn't make sense for this to be null

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -294,9 +294,9 @@ private static Paging GetClientMethodPaging(Method method, Method nextPageMethod
             BuilderHelpers.ParseConstant(requestParameter),
             requestParameter.Required == true);
 
-        private ResponseHeaderGroupType? BuildResponseHeaderModel(Operation operation, HttpResponse httpResponse)
+        private ResponseHeaderGroupType? BuildResponseHeaderModel(Operation operation, HttpResponse? httpResponse)
         {
-            if (!httpResponse.Headers.Any())
+            if (!(httpResponse?.Headers.Any() ?? false))
```
pakrym: Super hard to read.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -216,7 +216,7 @@ public OperationMethod(Operation operation, Method method)
                 Schema schema = schemaResponse.Schema is ConstantSchema constantSchema ? constantSchema.ValueType : schemaResponse.Schema;
                 TypeReference responseType = BuilderHelpers.CreateType(schema, isNullable: false);
 
-                ObjectSerialization serialization = SerializationBuilder.Build(httpResponse.KnownMediaType, schema, isNullable: false);
+                ObjectSerialization serialization = SerializationBuilder.Build(httpResponse?.KnownMediaType, schema, isNullable: false);
```
pakrym: Can `httpResponse` be null here?

src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
MiYanni: The nullability for it is added on. Meaning, it can be turned off. However, then you never know what is expected to be nullable or not. In this case, a valid operation can have no responses.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: I'm not creating an empty enumerable here. There is a null case already (that handles the generic `return message.Response;`) and this new case is specifically for when we have no responseBody and there aren't any statusCodes. This only happens when there aren't any responses (which means every response from that operation will fall into `default` and throw).

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: If there are no responses, there is no mediaType. We built a lot of the system assuming a response is required.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: I've always followed this kind of logic: https://softwareengineering.stackexchange.com/a/314003/108631
I'm assuming you want me to write `if(!(httpResponse?.Headers.Any() == true))` instead, or separate it to a local variable. Unfortunately, C# doesn't handle tri-state booleans well syntactically. 😓

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: Yes, when you don't have any responses in the operation. These are valid cases on the test server.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: I ended up writing it as `if (httpResponse == null || !httpResponse.Headers.Any())`, which is the most straightforward to read, and allows us to no longer need the `!` at the bottom of the method.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
pakrym: If there is no response then we are we calling this method?

src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
pakrym: We can always initialize property to `Array.Empty()`

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
pakrym: `!statusCodes.Any() == statusCodes.Empty()` and .Empty is easier to read.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs

src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
MiYanni: I'll make another PR to change nullability assignment to, instead, initialize with empty collections/values.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: As discussed, `.Empty()` is a static method on `Enumerable<T>` and returns an empty enumerable of T.

pakrym:
Something doesn't build.

PR #439, opened 2020-02-04
Add test for x-ms-client-name on schemas

pakrym:
m4 recently added the feature. Test it.


pakrym:
cc @tg-msft, x-ms-clientname should be allowed on everything now.

PR #440, opened 2020-02-05
CodeModel gen cleanup

MiYanni:
- Removed nullable references on collection types and have them initialized as empty collections
- Used NameGenerators to do some of the modifications instead of string or regex
- I removed my dumb code to work around that Visual Studio extension issue (I simply removed the extension)

### DIFF
--- a/src/AutoRest.CSharp.V3/Output/Models/Types/OutputLibrary.cs
+++ b/src/AutoRest.CSharp.V3/Output/Models/Types/OutputLibrary.cs
@@ -45,16 +45,16 @@ namespace AutoRest.CSharp.V3.Output.Models.Types
 
         private Dictionary<Schema, ISchemaType> BuildModels()
         {
-            var allSchemas = (_codeModel.Schemas.Choices ?? Enumerable.Empty<ChoiceSchema>()).Cast<Schema>()
-                .Concat(_codeModel.Schemas.SealedChoices ?? Enumerable.Empty<SealedChoiceSchema>())
-                .Concat(_codeModel.Schemas.Objects ?? Enumerable.Empty<ObjectSchema>());
+            var allSchemas = _codeModel.Schemas.Choices.Cast<Schema>()
+                .Concat(_codeModel.Schemas.SealedChoices)
+                .Concat(_codeModel.Schemas.Objects);
 
             return allSchemas.ToDictionary(schema => schema, BuildModel);
         }
 
         private ISchemaType BuildModel(Schema schema) => schema switch
         {
-            SealedChoiceSchema sealedChoiceSchema => new EnumType(sealedChoiceSchema, _context),
+            SealedChoiceSchema sealedChoiceSchema => (ISchemaType)new EnumType(sealedChoiceSchema, _context),
             ChoiceSchema choiceSchema => new EnumType(choiceSchema, _context),
             ObjectSchema objectSchema => new ObjectType(objectSchema, _context),
             _ => throw new NotImplementedException()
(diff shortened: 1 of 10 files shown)


src/AutoRest.CodeModel/Program.cs
```diff
@@ -40,40 +42,13 @@ private static void Main()
                                 && !l.Contains("Newtonsoft.Json.JsonExtensionData"))
                     .Select(l => Regex.Replace(l, @"(.*\[)Newtonsoft\.Json\.JsonProperty\((.*""),?.*(\)\])",
                         "$1YamlDotNet.Serialization.YamlMember(Alias = $2$3", RegexOptions.Singleline).TrimEnd()))
-                // Workaround for anyOf SecurityScheme types
+                // Removes AdditionalProperties property from types as they are required to derive from IDictionary<string, object> for deserialization to work properly
                 .Replace($"        private System.Collections.Generic.IDictionary<string, object> _additionalProperties = new System.Collections.Generic.Dictionary<string, object>();{Environment.NewLine}{Environment.NewLine}        public System.Collections.Generic.IDictionary<string, object> AdditionalProperties{Environment.NewLine}        {{{Environment.NewLine}            get {{ return _additionalProperties; }}{Environment.NewLine}            set {{ _additionalProperties = value; }}{Environment.NewLine}        }}{Environment.NewLine}", String.Empty)
```
MiYanni: This was still needed as it adds AdditionalProperties to these types. Language does it's own implementation of this so that the deserializer doesn't fail when there are unknown keys.

src/AutoRest.CodeModel/Program.cs
```diff
@@ -40,40 +42,13 @@ private static void Main()
...
-                .Replace("class HTTPSecurityScheme", "class HTTPSecurityScheme : System.Collections.Generic.Dictionary<string, object>")
-                .Replace("class SecurityScheme", "class SecurityScheme : System.Collections.Generic.Dictionary<string, object>")
```
MiYanni: I did this in the CodeModelPartials instead of the generator.

src/AutoRest.CodeModel/Program.cs
```diff
@@ -40,40 +42,13 @@ private static void Main()
...
                 // Fixes stray blank lines from the C# generator
                 .Replace($"{Environment.NewLine}{Environment.NewLine}{Environment.NewLine}", Environment.NewLine)
                 .Replace($"{Environment.NewLine}{Environment.NewLine}    }}", $"{Environment.NewLine}    }}")
-                // Replaces the alias to correctly match the schema from the C# generation fix above
-                .Replace("\"plus\"", "\"+\"")
-                .Replace("\"minus\"", "\"-\"")
                 // Weird generation issue workaround
-                .Replace($"{Namespace}.bool.True", "true")
-                // Set properties to be single quoted
-                .Replace("Alias = \"version\"", "ScalarStyle = YamlDotNet.Core.ScalarStyle.SingleQuoted, Alias = \"version\"")
```
MiYanni: This was only needed for serialization.

src/AutoRest.CSharp.V3/Output/Models/Types/OutputLibrary.cs
```diff
@@ -45,16 +45,16 @@ private Client[] BuildClients()
...
+                .Concat(_codeModel.Schemas.SealedChoices)
+                .Concat(_codeModel.Schemas.Objects);
 
             return allSchemas.ToDictionary(schema => schema, BuildModel);
         }
 
         private ISchemaType BuildModel(Schema schema) => schema switch
         {
-            SealedChoiceSchema sealedChoiceSchema => new EnumType(sealedChoiceSchema, _context),
+            SealedChoiceSchema sealedChoiceSchema => (ISchemaType)new EnumType(sealedChoiceSchema, _context),
```
pakrym: why did this change?

MiYanni: It showed red squiggles in VS under the whole thing. It can't tell the type to use (it can't tell the common interface is `ISchemaType`). It compiled before (obviously), but I wonder why you didn't see the squiggles... Resharper, maybe?

pakrym: Not a big deal.

PR #442, opened 2020-02-07
Fix package so it can be used in the main repo

pakrym:
1. Adds the required files that were missed by my original PR
2. Splits readme.md to avoid having test related stuff inside the package
3. Updates to the latest shared source, fixes generator appropriately
4. Adds a way to specify the shared source location
5. Split shared source files in 2 folders so they are easier to manage


eng/Generate.ps1
```diff
@@ -145,6 +146,7 @@ if ($updateLaunchSettings)
         'profiles' = [ordered]@{}
     };
 
+    $sharedSourceNormalized = $sharedSource.Replace($repoRoot, "`$(SolutionDir)")
```
MiYanni: Can just use non-interpolated string: `'$(SolutionDir)'` (single quotes are literals)

package.json
```diff
@@ -2,10 +2,6 @@
   "name": "@autorest/csharp-v3",
   "private": true,
   "description": "package.json intended for in-repo use only, package.json used for publishing is located in src/AutoRest.CSharp.V3/package.json",
-  "scripts": {
-    "start": "dotnet ./artifacts/bin/AutoRest.CSharp.V3/Debug/netcoreapp3.0/AutoRest.CSharp.V3.dll --server",
-    "debug": "dotnet ./artifacts/bin/AutoRest.CSharp.V3/Debug/netcoreapp3.0/AutoRest.CSharp.V3.dll --server --launch-debugger"
-  },
```
MiYanni: If this is the in-repo package.json, why remove the scripts?

pakrym: Because it's not required. The "main" project json is now correct enough to run the server and we avoid 2 entry point.

eng/Generate.ps1
pakrym: Good point!

eng/Generate.ps1
```diff
@@ -25,7 +25,7 @@ function Invoke-AutoRest($baseOutput, $title, $autoRestArguments)
     if ($fast)
     {
         $codeModel = Join-Path $baseOutput $title "CodeModel.yaml"
-        $command = "dotnet run --project $script:autorestPluginProject --no-build -- --plugin=csharpgen --title=$title --namespace=$namespace --standalone --input-file=$codeModel --output-folder=$outputPath"
+        $command = "dotnet run --project $script:autorestPluginProject --no-build -- --plugin=csharpgen --title=$title --namespace=$namespace --standalone --input-file=$codeModel --output-folder=$outputPath --shared-source-folder=$sharedSource --save-code-model=true"
```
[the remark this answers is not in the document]
pakrym: Meh, it never changes.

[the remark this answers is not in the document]
pakrym: I add a `script:` prefix.

PR #444, opened 2020-02-11
Update to the latest m4

pakrym:


pakrym:
Blocked by https://github.com/Azure/autorest.modelerfour/issues/153#issuecomment-584810585

Issue #447, opened 2020-02-21
Support Operation Resource returning from LRO operations


MiYanni:
There are 2 options:
- Make a generic model for an operation resource (such as OperationResource) and allow those LRO operations to return `Operation<OperationResource>`.
- Have an Operation where there is no T (operation discards operation resource information). Currently, we only have `Operation<T>`.

Issue #398, opened 2020-01-11
Implement retry on 500 logic


MiYanni:
From paging tests, there is a retry mechanism that, when a status code 500 is provided, it will retry the call. This is not implemented currently.

This should be handled by a `Cookie` and `Set-Cookie` policy.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie
https://github.com/Azure/autorest.python/blob/f271e8a592299a0b03938650546fedfe581aeaba/test/azure/AcceptanceTests/conftest.py#L72-L88

PR #448, opened 2020-02-24
LRO support (JSON only)

MiYanni:
Implements ARM-based LRO for JSON responses.
- It covers any of the ARM resource scenarios from this diagram:
image
- Current issues:
  - It cannot be combined with paging: https://github.com/Azure/autorest.csharp/issues/450
  - Does not cover Operation resource implementation: https://github.com/Azure/autorest.csharp/issues/447
  - Only does JSON, not XML: https://github.com/Azure/autorest.csharp/issues/449
- This PR also correctly fixed the extra line being inserted at the start of operation methods
  - Recommend 'Ignore Whitespace changes' when reviewing


test/AutoRest.TestServer.Tests/header.cs
```diff
@@ -19,6 +19,7 @@ public class HeaderTests : TestServerTestBase
         public HeaderTests(TestServerVersion version) : base(version, "header") { }
 
         [Test]
+        [Ignore("Built pipeline already injects User-Agent, so test fails with 400 Bad Request.")]
```
pakrym: Add a parameter to TestStatus method to ask for a simple pipeline. Same for the paging one.

eng/CodeChecks.ps1
```diff
@@ -5,13 +5,13 @@ $ErrorActionPreference = 'Stop'
...
 
 Write-Host 'Downloading shared source files...'
-& (Join-Path $PSScriptRoot 'DownloadSharedSource.ps1')
+# & (Join-Path $PSScriptRoot 'DownloadSharedSource.ps1')
 
 Write-Host 'Generating test clients...'
 & (Join-Path $PSScriptRoot 'Generate.ps1')
 
 Write-Host 'Checking generated file differences...'
-git -c core.safecrlf=false diff --ignore-space-at-eol --exit-code
```
pakrym: You shouldn't need to disable this.

test/AutoRest.TestServer.Tests/header.cs
MiYanni: Done.

eng/CodeChecks.ps1
MiYanni: Done.

samples/Azure.Storage.Tables/Azure.Storage.Tables/CodeModel.yaml
```diff
@@ -1112,7 +1112,7 @@ globalParameters:
   protocol: !<!Protocols> 
     http: !<!HttpParameter> 
       in: header
-- !<!Parameter> &ref_79
```
pakrym: This one is interesting. Considering you didn't update the autorest why did this change.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -331,6 +342,144 @@ private void WritePagingOperation(CodeWriter writer, Paging pagingMethod, bool a
...
+                        );
+                        writer.Line($"return {typeof(Response)}.FromValue({valueVariable}, r);");
+                        break;
+                    default:
+                        writer.Line($"return Response.FromValue(r, r);");
+                        break;
+                }
+                writer.LineRaw("},");
+                writer.Line($"async (r, c) =>");
+                writer.LineRaw("{");
```
pakrym: Any reason you didn't use scope here?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -103,15 +107,28 @@ public Client BuildClient(OperationGroup operationGroup)
                     Paging pagingMethod = GetClientMethodPaging(processed.Method, nextPageMethod, pageable, objectType);
                     pagingMethods.Add(pagingMethod);
                 }
+                // For some reason, booleans in dictionaries are deserialized as string instead of bool.
+                bool longRunningOperation = Convert.ToBoolean(processed.Operation.Extensions.GetValue<string>("x-ms-long-running-operation") ?? "false");
+                if (longRunningOperation && pageable == null)
```
pakrym: Why do we need to check for `pageable == null`?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -311,6 +328,33 @@ private Paging GetClientMethodPaging(Method method, Method nextPageMethod, IDict
...
+            var originalResponseParameter = new Parameter(
+                "originalResponse",
+                "The original response from starting the operation.",
+                new CSharpType(typeof(AzureResponse)),
+                null,
+                true);
+            var httpMessageParameter = new Parameter(
+                "createOriginalHttpMessage",
+                "Creates the HTTP message used for the original request.",
+                new CSharpType(typeof(Func<>), new CSharpType(typeof(HttpMessage))),
```
pakrym: Does `new CSharpType(typeof(Func<HttpMessage>)` work?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -311,6 +328,33 @@ private Paging GetClientMethodPaging(Method method, Method nextPageMethod, IDict
...
+            string name = $"{method.Name}Operation";
+            return new LongRunningOperation(method, originalResponse, name, new[] { originalResponseParameter, httpMessageParameter }, finalStateVia);
+        }
+
+        private static FinalStateVia GetFinalStateVia(string? rawValue) => rawValue switch
+        {
+            "azure-async-operation" => FinalStateVia.AzureAsyncOperation,
+            "location" => FinalStateVia.Location,
+            "original-uri" => FinalStateVia.OriginalUri,
+            _ => FinalStateVia.Location
```
pakrym: Any reason we are not throwing for unknown values?

src/AutoRest.CSharp.V3/Output/Models/Requests/LongRunningOperation.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            OriginalMethod = originalMethod;
+            OriginalResponse = originalResponse;
+            Name = name;
+            CreateParameters = createParameters;
+            FinalStateVia = finalStateVia;
+        }
+
+        public string Name { get; }
+        public Method OriginalMethod { get; }
```
pakrym: Is this a method that initiates operation?

src/AutoRest.CSharp.V3/Output/Models/Requests/Method.cs
```diff
@@ -22,7 +22,7 @@ public Method(string name, string? description, Request request, Parameter[] par
         public string? Description { get; }
         public Request Request { get; }
         public Parameter[] Parameters { get; }
-        public Response Response { get; }
+        public Response Response { get; set; }
```
pakrym: Why is this mutable?

samples/Azure.Storage.Tables/Azure.Storage.Tables/CodeModel.yaml
MiYanni: 🤔

src/assets/Azure.Core.Shared/OperationHelpers.cs
```diff
@@ -0,0 +1,57 @@
...
+
+#pragma warning disable CS8600, CS8604, CS8605
+
+namespace Azure.Core
+{
+    internal static class OperationHelpers
+    {
+        public static TimeSpan DefaultPollingInterval { get; } = TimeSpan.FromSeconds(1);
+
+        public static T GetValue<T>(ref T? value) where T: class
```
pakrym: Not sure I like to have an extension method on every value. Can we get rif of it?

src/assets/Generator.Shared/FinalStateVia.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+#nullable enable
+
+namespace Azure.Core
+{
+    internal enum FinalStateVia
```
pakrym: `OperationFinalStateVia`

test/AutoRest.TestServer.Tests/body-file.cs
```diff
@@ -49,7 +49,7 @@ public class BodyFileTests : TestServerTestBase
 
             Assert.AreEqual(3000 * 1024 * 1024L, total);
             await result.Value.DisposeAsync().ConfigureAwait(false);
-        });
+        }, false, true);
```
pakrym: nit: add parameter names here

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: I originally wrote it using scope. However, you cannot append to the end of a scope. The scope always writes a newline after the `}`. Both of these sets of curly brackets needed characters after them. When I generated it with scope, it was compile-able, but was ugly when Roslyn when through and tried to prettify it.

test/AutoRest.TestServer.Tests/lro.cs
```diff
@@ -0,0 +1,1455 @@
...
+
+namespace AutoRest.TestServer.Tests
+{
+    [IgnoreOnTestServer(TestServerVersion.V2, "LRO tests are not supported yet")]
+    public class LroTest: TestServerTestBase
+    {
+        public LroTest(TestServerVersion version) : base(version, "lros") { }
+
+        [Test]
+        [Ignore("Needs to send x-ms-client-request-id: https://github.com/Azure/autorest.csharp/issues/446")]
```
pakrym: I think I asked you about this already but I forgot. Why is this skipped? The full pipeline should be sending the `x-ms-client-request-id`

test/AutoRest.TestServer.Tests/lro.cs
```diff
@@ -0,0 +1,1455 @@
...
+        {
+            var value = new SubProduct();
+            var operation = await new LrOSOperations(ClientDiagnostics, pipeline, host).StartPutSubResourceOperationAsync(value);
+            var result = await operation.WaitForCompletionAsync().ConfigureAwait(false);
+            Assert.AreEqual("100", result.Value.Id);
+            Assert.AreEqual("Succeeded", result.Value.ProvisioningState);
+        });
+
+        [Test]
+        public Tas
```
pakrym: The full pipeline should be retrying on 500 we might need to update the Azure.Core version.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: https://github.com/Azure/autorest.csharp/issues/450
You have to write some kind of logic that combines LRO and paging. It has to actually change what is being called while doing paging operations. For now, that was outside the scope of this PR, and is something we can add to the generator when it is actually needed.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: I thought we had to provide generic parameters as a parameter... At least that's how the original implementation worked. Is that no longer the case? Other places in the code use this pattern seen here.

test/AutoRest.TestServer.Tests/paging.cs
```diff
@@ -294,7 +294,7 @@ await foreach (var page in pageableAsync.AsPages())
                 }
             });
             Assert.AreEqual(2, id);
-        }, true);
+        }, true, true);
```
pakrym: nit: parameter names.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: Location is the default. It used to be azure-async-operation, but was changed which is why some tests need further investigation by Laurent. This value (final-state-via) is optional when declaring an LRO.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -331,6 +342,144 @@ private void WritePagingOperation(CodeWriter writer, Paging pagingMethod, bool a
...
+                            ref valueVariable
+                        );
+                        writer.Line($"return {typeof(Response)}.FromValue({valueVariable}, r);");
+                        break;
+                    default:
+                        writer.Line($"return Response.FromValue(r, r);");
+                        break;
+                }
+                writer.LineRaw("},");
+                writer.Line($"async (r, c) =>");
```
pakrym: Let's give these parameters full names.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -331,6 +342,144 @@ private void WritePagingOperation(CodeWriter writer, Paging pagingMethod, bool a
...
+                {
+                    case JsonSerialization jsonSerialization:
+                        writer.Append($"using var {document:D} = ");
+                        writer.Line($"await {typeof(JsonDocument)}.ParseAsync(r.ContentStream, default, c).ConfigureAwait(false);");
+                        writer.ToDeserializeCall(
+                            jsonSerialization,
+                            w => w.Append($"document.RootElement"),
+                            ref valueVariable
+                        );
+                        writer.Line($"return {typeof(Response)}.FromValue({valueVariable}, r);");
```
pakrym: Can we create `Response<T>` in the LongRunningOperation implementation? It would simplify generated code a little bit more.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -331,6 +342,144 @@ private void WritePagingOperation(CodeWriter writer, Paging pagingMethod, bool a
...
+                        writer.Line($"var {document:D} = {typeof(XDocument)}.Load(r.ContentStream, LoadOptions.PreserveWhitespace);");
+                        writer.ToDeserializeCall(
+                            xmlSerialization,
+                            w => w.Append($"document"),
+                            ref valueVariable
+                        );
+                        writer.Line($"return {typeof(Response)}.FromValue({valueVariable}, r);");
+                        break;
+                    default:
+                        writer.Line($"return Response.FromValue(r, r);");
```
pakrym: Can we avoid passing lambdas at all if there is no return type?

src/AutoRest.CSharp.V3/Output/Models/Requests/LongRunningOperation.cs
MiYanni: It is the original operation, so yes. The nomenclature in the LRO documentation uses both 'original request' and 'initial request' to describe this. In my diagram, I used 'original operation' to differentiate between that and a 'polling operation'.

src/assets/Azure.Core.Shared/OperationHelpers.cs
pakrym: Oh, it's not an extension method, my bad.

src/AutoRest.CSharp.V3/Output/Models/Requests/Method.cs
MiYanni: Required to mutate to modify the original method's response when determining LRO operations. Original methods only return Response (not Response<T>). You suggested I use this implementation as to simplify how we handle LRO when looping over the operations.

src/assets/Azure.Core.Shared/OperationHelpers.cs
MiYanni: This isn't my code. This is Azure.Core. It is a downloaded file.

src/assets/Generator.Shared/FinalStateVia.cs
MiYanni: Done.

test/AutoRest.TestServer.Tests/body-file.cs
MiYanni: Done.

src/assets/Generator.Shared/ArmOperationHelpers.cs
```diff
@@ -0,0 +1,389 @@
...
+                        return true;
+                    }
+                }
+
+                throw _clientDiagnostics.CreateRequestFailedException(response);
+            }
+        }
+
+        private static readonly string[] s_failureStates = { "failed", "canceled" };
+        private static readonly string[] s_terminalStates = s_failureStates.Append("succeeded").ToArray();
```
pakrym: Just define an array, don't need to allocate multiple things.

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: You have to send a specific value for `x-ms-client-request-id`. Is there a way to inject a value per request (that won't be overridden)?

src/assets/Generator.Shared/ArmOperationHelpers.cs
```diff
@@ -0,0 +1,389 @@
...
+
+            public override Response GetRawResponse() => _rawResponse;
+
+            public override ValueTask<Response<T>> WaitForCompletionAsync(CancellationToken cancellationToken = default) =>
+                this.DefaultWaitForCompletionAsync(OperationHelpers.DefaultPollingInterval, cancellationToken);
+
+            public override ValueTask<Response<T>> WaitForCompletionAsync(TimeSpan pollingInterval, CancellationToken cancellationToken) =>
+                this.DefaultWaitForCompletionAsync(pollingInterval, cancellationToken);
+
+            public override async ValueTask<Response> UpdateStatusAsync(CancellationToken cancellationToken = default)
```
pakrym: Let's merge two UpdateStatus* methods into one with `bool async` parameter. Don't want to have two pieces of complicated logic like this.

test/AutoRest.TestServer.Tests/paging.cs
MiYanni: Done.

src/assets/Generator.Shared/ArmOperationHelpers.cs
```diff
@@ -0,0 +1,389 @@
...
+            if ((requestMethod != RequestMethod.Put && (info.HeaderFrom == HeaderFrom.None || !info.HasLocation)) || finalStateVia == FinalStateVia.AzureAsyncOperation)
+            {
+                // Support Operation Resource: https://github.com/Azure/autorest.csharp/issues/447
+                throw clientDiagnostics.CreateRequestFailedException(originalResponse);
+            }
+
+            return new ArmOperation<T>(pipeline, clientDiagnostics, originalResponse, scopeName, info, createFinalResponse, createFinalResponseAsync);
+        }
+
+        private class ArmOperation<T> : Operation<T> where T : notnull
```
pakrym: I know it's a lot to ask but can we summerize the logic in a comment above this type. It seems to be complicated enough to warrant that.

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: Give me a bit to double check this.

src/assets/Generator.Shared/ArmOperationHelpers.cs
```diff
@@ -0,0 +1,389 @@
...
+                        throw new InvalidOperationException("The operation has not completed yet.");
+                    }
+
+                    return _value;
+                }
+            }
+            public override bool HasCompleted => _hasCompleted;
+            public override bool HasValue => _hasValue;
+
+            private HttpMessage CreateGetResponseRequest(string link)
```
pakrym: This needs a better name.

src/assets/Generator.Shared/ArmOperationHelpers.cs
```diff
@@ -0,0 +1,389 @@
...
+            {
+                HttpMessage message = _pipeline.CreateMessage();
+                Request request = message.Request;
+                request.Method = RequestMethodAdditional.Get;
+                var uri = new RawRequestUriBuilder();
+                uri.AppendRaw(link, false);
```
pakrym: `request.Uri.Reset(new Uri(link));`

src/assets/Generator.Shared/ArmOperationHelpers.cs
```diff
@@ -0,0 +1,389 @@
...
+                    return GetRawResponse();
+                }
+
+                if (_shouldPoll)
+                {
+                    _rawResponse = GetResponse(_info.PollUri, cancellationToken);
+                }
+                _shouldPoll = true;
+                _hasCompleted = IsTerminalState(out string state);
+                if (HasCompleted)
```
pakrym: Don't use property and field interchangeably in the same method.

src/assets/Generator.Shared/ArmOperationHelpers.cs
```diff
@@ -0,0 +1,389 @@
...
+
+        private enum HeaderFrom
+        {
+            None,
+            OperationLocation,
+            AzureAsyncOperation,
+            Location
+        }
+
+        private class ScenarioInfo
```
pakrym: Any reason these can't be fields on the Operation?

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
pakrym: There is a parameter for that https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs#L29

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: This is our deserialization logic that is copy-pasted from how we write deserialization already. So, this is dynamic based on what type we're trying to deserialize into. Is there a more generic process we could use everywhere? If so, it would apply here too.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
pakrym: Is LRO+Paging a real thing?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
pakrym: I'm not sure :) there is no reason to require it if all generic parameters are available as C# types.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
pakrym: Let's distinguish between no value and unknown value here and throw for unknown.

test/AutoRest.TestServer.Tests/lro.cs
pakrym: It sends a random one per request. Do we need a specific one for every request?

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: Technically, this is creating `Response<Response>`. That's because this lambda expects `Response<T>` to be returned. To avoid this, I'd have to complicate the logic to do UpdateStatus since it is what calls these lambdas. As it stands, either implementation isn't perfect because of this issue: https://github.com/Azure/autorest.csharp/issues/447
So, if we would need that logic, we can clean this up. Otherwise, the likelihood of this code actually being generated outside of the test server is almost zero.

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
pakrym: The difference is that in other cases we don't have a hand-written piece of logic in between. Usually methods return Response<T> directly so there is nowhere we can put the dynamic logic. But here there is an obvious place for it.

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: Done.

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: Since I was going to do it anyway, I'm going to create a document for this information based on my diagram information. For now, I'm putting links to the reference documents I used to get to this point.

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: Done.

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: Done.

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: They are input to create the ArmOperation. We create the ScenarioInfo before making an ArmOperation. I could expose them from the ArmOperation, but we're already in a class, so they are already in the scope where they are used.

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: In the test server, yes. In practice, no idea.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: I double checked. Currently, we don't 'pull apart' the Type passed in. We just get Name, Namespace, and IsValueType out of it. I think that is because it would involve recursion in the constructor which I avoided when I originally made the first concept of this thing. For now, this works, but I created an issue to address allowing the shorthand creation: https://github.com/Azure/autorest.csharp/issues/458

src/AutoRest.CSharp.V3/Output/Models/Requests/LongRunningOperation.cs
pakrym: Can we rename it to InitialRequest and FinalResponse?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: Done.

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: For these tests, yes. There is a value on the test server that needs to be sent to it. So, it cannot be random.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -331,6 +342,145 @@ private void WritePagingOperation(CodeWriter writer, Paging pagingMethod, bool a
...
+                            writer.Line($"return {typeof(Response)}.FromValue({valueVariable}, response);");
+                            break;
+                        case XmlElementSerialization xmlSerialization:
+                            writer.Line($"var {document:D} = {typeof(XDocument)}.Load(response.ContentStream, LoadOptions.PreserveWhitespace);");
+                            writer.ToDeserializeCall(
+                                xmlSerialization,
+                                w => w.Append($"document"),
+                                ref valueVariable
+                            );
+                            writer.Line($"return {typeof(Response)}.FromValue({valueVariable}, response);");
```
pakrym: Let's get rid of `Response<T>` creation too.

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: The original call for each of these is returning 500, which means the pipeline is not retrying it. If we have to update Azure.Core for it to work, we should do that as a separate PR, and try to enable any tests that link to https://github.com/Azure/autorest.csharp/issues/398 in the test project.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: I'm confused as the code generated here calls generated code deserializers. So, unless we make interfaces for our deserializers, there's not an easy path to have a generic deserialization mechanism. If we want that, feel free to make an issue and we can address it in a follow-up PR, for both this situation and our overall deserialization writing.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: We use both the Response itself and the Value for the UpdateStatus:
image

Do you want to return something less elegant, like a tuple of both values? Or make a separate type to just hold Response and Value? (I thought that was the point of Response<T>, though)

src/assets/Generator.Shared/ArmOperationHelpers.cs
pakrym: Sure, but why does this need to be a separate type and why cant' we extract these values in the operation ctor.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: Fixed. Now returns just the Value.

test/TestServerProjects/lro/Generated/Operations/LROsOperations.cs
```diff
@@ -0,0 +1,4220 @@
...
+        internal HttpMessage CreateDelete204SucceededRequest()
+        {
+            var message = pipeline.CreateMessage();
+            var request = message.Request;
+            request.Method = RequestMethodAdditional.Delete;
+            var uri = new RawRequestUriBuilder();
+            uri.AppendRaw(host, false);
+            uri.AppendPath("/lro/delete/204/succeeded", false);
+            request.Uri = uri;
+            return
```
pakrym: Can we return the value directly?

pakrym: Meh, it's fine.

MiYanni: I think you meant to point to a different section of code. I'm assuming to mean because it does `var value = ...;` and then `return value;`. Right now, the design for the deserialization, you construct an entire statement and provide a variable name. So, it assumes you are always doing variable assignment. Without editing that design, we always get a variable. So, in this case, we return that variable. I looked at this when I made the change, but it is really outside the scope of LROs. It is generalized deserialization mechanics.

src/assets/Generator.Shared/ArmOperationHelpers.cs
```diff
@@ -0,0 +1,389 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core.Pipeline;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// Helper methods for ARM long-running operations.
+    /// </summary>
+    public static class ArmOperationHelpers
```
pakrym: internal

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: It was done for conceptual encapsulation. But now, I've changed the code to no longer have this type. Fixed.

src/assets/Generator.Shared/ArmOperationHelpers.cs
MiYanni: Fixed.

PR #457, opened 2020-02-25
Newline method start fix

MiYanni:
- Changed it so it doesn't create new empty lines if the previous set of text ended in {.
- Added try-catch around $clean for Generate.ps1 so it doesn't halt the entire generation if it cannot remove a file.


pakrym:
Let's fix this in the CoreWriter in a more generic way (don't add new line is the previous one ends with `{`)

MiYanni:
@pakrym Changed it so it doesn't create new empty lines if the previous set of text ended in `{`.

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
```diff
@@ -269,13 +269,13 @@ public CodeWriter Literal(object? o)
         public CodeWriter Line(FormattableString formattableString)
         {
             Append(formattableString);
-            Line();
+            _builder.AppendLine();
```
MiYanni: This needs to always append a line because you are adding text (not just a blank line).

eng/Generate.ps1
```diff
@@ -30,7 +30,13 @@ function Invoke-AutoRest($baseOutput, $title, $autoRestArguments)
 
     if ($clean)
     {
-        Get-ChildItem $outputPath -Filter Generated -Directory -Recurse | Remove-Item -Force -Recurse;
+        try
```
pakrym: This has never failed for me is some file locked?

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
```diff
@@ -376,13 +376,32 @@ public void RemoveTrailingComma()
...
                 if (_builder[i] == ',')
                 {
                     _builder.Remove(i, _builder.Length - i);
                 }
+            });
+        }
 
-                break;
-            }
+        private void CreateNewLine()
```
pakrym: Why don't you have this logic in Line method itself.

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
pakrym: But your logic should be able to handle it just fine.

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
```diff
@@ -367,7 +367,7 @@ private void PopScope(CodeWriterScope expected)
             Debug.Assert(actual == expected);
         }
 
-        public void RemoveTrailingComma()
+        private void ActionOnLast(Action<int> action)
```
pakrym: Should we just have a method that finds index of last non whitespace char?

```
var lastNonSpace = FindLastNonSpace();

if (lastNonSpace != -1 && _builder[lastNonSpace] == '{') 
{
}
```

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
MiYanni: The condition for writing a new line in this situation is different than `Line()`. If any text is written during `Append(formattableString)`, it should always write a new line. The reason this is for a situation where the line we appended has the last character as `{`. Obviously, we want a new line after '{'. The normal `Line()` method is for blank line insertion (no appended text). So, what I did here now is check to see if `formattableString` has text (which allows us to know that text was written). This was significantly easier and cleaner than changing `Append(formattableString)` to 'let us know text was written'.

What I've also done is added `LineRaw()` (which always writes a new line) to follow the pattern we used for the rest of the raw methods. Then, I call `LineRaw()` here and in the `Line()` method (when the condition is met).

eng/Generate.ps1
MiYanni: Removed the changes.

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
MiYanni: Done.

PR #461, opened 2020-02-28
Add generated-code marker to generated files

pakrym:


src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
```diff
@@ -321,6 +321,8 @@ public string ToFormattedCode()
             builder.AppendLine("// Copyright (c) Microsoft Corporation. All rights reserved.");
             builder.AppendLine("// Licensed under the MIT License.");
             builder.AppendLine();
+            builder.AppendLine("// <auto-generated/>");
+            builder.AppendLine();
```
MiYanni: Good find to use this mechanism for it.

PR #465, opened 2020-03-03
Update paging to use new m4 data

MiYanni:
Replaced the paging implementation that used the x-ms-pageable extension information. Now, it uses the paging node in language.default.

### DIFF
--- a/src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
+++ b/src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
@@ -50,69 +50,64 @@ namespace AutoRest.CSharp.V3.Output.Builders
             }
 
             string clientName = operationGroup.CSharpName();
-            Dictionary<string, OperationMethod> processedMethods = new Dictionary<string, OperationMethod>(StringComparer.InvariantCultureIgnoreCase);
+            Dictionary<string, OperationMethod> operationMethods = new Dictionary<string, OperationMethod>(StringComparer.InvariantCultureIgnoreCase);
             foreach (Operation operation in operationGroup.Operations)
             {
                 Method? method = BuildMethod(operation, clientName, clientParameters);
                 if (method != null)
                 {
-                    processedMethods.Add(operation.Language.Default.Name, new OperationMethod(operation, method));
+                    operationMethods.Add(operation.Language.Default.Name, new OperationMethod(operation, method));
                 }
             }
 
             List<Method> nextPageMethods = new List<Method>();
-            List<Paging> pagingMethods = new List<Paging>();
+            List<PagingInfo> pagingMethods = new List<PagingInfo>();
             List<LongRunningOperation> longRunningOperationMethods = new List<LongRunningOperation>();
-            foreach ((string processedName, OperationMethod processed) in processedMethods)
+            foreach ((string operationName, (Operation operation, Method method)) in operationMethods)
             {
-                IDictionary<object, object>? pageable = processed.Operation.Extensions.GetValue<IDictionary<object, object>>("x-ms-pageable");
-                if (pageable != null)
+                Paging? paging = operation.Language.Default.Paging;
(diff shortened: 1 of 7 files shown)


src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -50,69 +50,64 @@ public Client BuildClient(OperationGroup operationGroup)
             }
 
             string clientName = operationGroup.CSharpName();
-            Dictionary<string, OperationMethod> processedMethods = new Dictionary<string, OperationMethod>(StringComparer.InvariantCultureIgnoreCase);
+            Dictionary<string, OperationMethod> operationMethods = new Dictionary<string, OperationMethod>(StringComparer.InvariantCultureIgnoreCase);
```
MiYanni: Originally, I attempted to remove this goofy Dictionary mechanism since the operationName (key) is not necessarily needed. However, that proved to be more difficult than expected because of the longRunningOperation code relying on updating the method, and then also knowing that it is not a paging operation (since we don't have the implementation for paging LROs). Technically, from the AutoRest meeting, all of this logic needs to get flipped around, as they said to prefer LRO over paging, which we don't do right now.

Long story short, I ended up doing the least amount of changes, but I feel this code should be redesigned or refactored at some point so it is processing things more logically (and prefers LRO over paging).

pakrym: you can still update the method without the dictionary or am I missing something ?

test/TestServerProjects/custom-baseUrl-paging/Generated/Operations/PagingOperations.cs
```diff
@@ -365,6 +365,84 @@ public Response<ProductResult> GetPagesPartialUrlNextPage(string nextLink, Cance
                 throw;
             }
         }
+        internal HttpMessage CreateGetPagesPartialUrlOperationNextNextPageRequest(string nextLink)
```
pakrym: Why do we now generate this?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: Updating the method is only part of it. I can keep the OperationMethod class and still update the method. However, converting this to a list would require us to at minimum duplicate it. Meaning, something like .ToArray when we start the loop that uses it. In the paging section, we need to access the method we created that is paired to the operation in NextLinkOperation. Now, it should be the same operation instance, so I can do an equality comparison, but that requires looping through everything to find the operation that matches (using something like .FirstOrDefault). So, I just avoided that by continuing to use the operation name as the unique key to get the method that is paired with it for paging.

test/TestServerProjects/custom-baseUrl-paging/Generated/Operations/PagingOperations.cs
MiYanni: This is due to: https://github.com/Azure/autorest.modelerfour/issues/202
When that issue is fixed, this code should be removed, and no paging code should need to be changed.

src/AutoRest.CSharp.V3/Output/Models/Requests/PagingInfo.cs
```diff
@@ -6,9 +6,9 @@
 
 namespace AutoRest.CSharp.V3.Output.Models.Requests
 {
-    internal class Paging
+    internal class PagingInfo
```
pakrym: What do you think about PagingMethod?

MiYanni: I thought about it originally, but this isn't a method. It is information that contains methods (literally method properties). However, it does produce methods. I think between this and LongRunningOperation, we would need better naming for this 'meta-method' data.

PR #478, opened 2020-03-03
Minor updates/fixes

MiYanni:
- Updated Roslyn package to latest
  - Fixed issue with not cleaning up fully-qualified names
- Removed a couple minor type writing redundancies
- Made Generate.ps1 only clean files, not folders, so my Visual Studio doesn't cause the script to fail

### DIFF
--- a/src/AutoRest.CSharp.V3/Input/Source/SourceInputModelBuilder.cs
+++ b/src/AutoRest.CSharp.V3/Input/Source/SourceInputModelBuilder.cs
@@ -18,8 +18,8 @@ namespace AutoRest.CSharp.V3.Input.Source
         private SourceInputModelBuilder(Compilation compilation)
         {
             _compilation = compilation;
-            _schemaNameAttribute = compilation.GetTypeByMetadataName(typeof(CodeGenSchemaAttribute).FullName);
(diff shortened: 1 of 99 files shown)


src/AutoRest.CSharp.V3/Input/Source/SourceInputModelBuilder.cs
```diff
@@ -18,8 +18,8 @@ public class SourceInputModelBuilder
         private SourceInputModelBuilder(Compilation compilation)
         {
             _compilation = compilation;
-            _schemaNameAttribute = compilation.GetTypeByMetadataName(typeof(CodeGenSchemaAttribute).FullName);
-            _schemaMemberAttribute = compilation.GetTypeByMetadataName(typeof(CodeGenSchemaMemberAttribute).FullName);
+            _schemaNameAttribute = compilation.GetTypeByMetadataName(typeof(CodeGenSchemaAttribute).FullName!)!;
+            _schemaMemberAttribute = compilation.GetTypeByMetadataName(typeof(CodeGenSchemaMemberAttribute).FullName!)!;
```
MiYanni: They changed this stuff to return nullables, but we don't care.

PR #480, opened 2020-03-04
Add public clients generation

pakrym:


src/AutoRest.CSharp.V3/Output/Models/Types/EnumType.cs
```diff
@@ -28,7 +28,7 @@ private EnumType(Schema schema, BuildContext context, IEnumerable<ChoiceValue> c
             Declaration = BuilderHelpers.CreateTypeAttributes(
                 schema.CSharpName(),
                 $"{context.DefaultNamespace}.Models",
-                Accessibility.Public,
+                "public",
```
MiYanni: Are we no longer using the Accessibility enum?

src/AutoRest.CSharp.V3/Output/Models/Client.cs
```diff
@@ -1,31 +1,30 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
+// Licensed under the MIT License.
```
MiYanni: I noticed some of these file headers had been changed. Why would we remove the "See License.txt" part?

src/assets/Generator.Shared/CodeGenSchemaMemberAttribute.cs
```diff
@@ -8,11 +8,8 @@ namespace Azure.Core
     [AttributeUsage(AttributeTargets.Property | AttributeTargets.Field)]
     internal class CodeGenSchemaMemberAttribute : Attribute
     {
-        public string SchemaName { get; }
```
MiYanni: Why the removal of this property, but not the parameter in the constructor?

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -300,26 +178,26 @@ private void WritePagingOperation(CodeWriter writer, PagingInfo pagingMethod, bo
...
                 var nullableInt = new CSharpType(typeof(int), true);
 
                 var continuationTokenText = pagingMethod.NextLinkName != null ? $"response.Value.{pagingMethod.NextLinkName}" : "null";
-                var asyncText = async ? "async " : string.Empty;
-                var awaitText = async ? "await " : string.Empty;
+                var asyncText = async ? "async" : string.Empty;
+                var awaitText = async ? "await" : string.Empty;
                 var configureAwaitText = async ? ".ConfigureAwait(false)" : string.Empty;
-                using (writer.Scope($"{asyncText}{funcType} FirstPageFunc({nullableInt} pageSizeHint)"))
+                using (writer.Scope($"{asyncText} {funcType} FirstPageFunc({nullableInt} pageSizeHint)"))
```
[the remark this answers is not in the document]
pakrym: Yep, Roslyn is good at removing double spaces.

src/assets/Generator.Shared/CodeGenSchemaMemberAttribute.cs
pakrym: We never read the value through the property, the way we use roslyn allows us to read the parameter value directly.

src/AutoRest.CSharp.V3/Output/Models/Client.cs
pakrym: This is the correct header and it got applied to the new files I created.

src/AutoRest.CSharp.V3/Output/Models/Types/EnumType.cs
pakrym: I don't think it was adding anything.

pakrym:
I noticed we are naming the operation group classes as Client now instead of Operations. Does that mean we intend to have every operation group as a separate client? Originally, I made operation groups as separate classes to simply keep similar operations together. They weren't initially supposed to become clients themselves.

Ha! I didn't realize that but I think I'll keep it :) We can always merge OGs if we get an ask.

Issue #475, opened 2020-03-03
Support putting models in sub-directory


annelo-msft:
Generation doesn't work if the file isn't in the top-level directory, but I'd like to have models still in a Models file to separate them from client and other types.

PR #482, opened 2020-03-04
Load existing sources recursively and process base types

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/475
Fixes: https://github.com/Azure/autorest.csharp/issues/477


pakrym:
> So, this allows you to basically split schema properties into multiple classes (though, they become base classes of the original class). Would we eventually allow the other direction (combine multiple schemas into 1 class)?

Yeah, I was thinking about allowing to use the same class for multiple schemas but this one is way easier to implement :)

PR #492, opened 2020-03-05
Speedup code-gen a bit

pakrym:
Use Roslyn APIs better to cause fewer allocations and parallelize processing. This is especially visible on  large projects.

### DIFF
--- a/src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
+++ b/src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
@@ -32,11 +32,16 @@ namespace AutoRest.CSharp.V3.AutoRest.Plugins
 
         public void AddGeneratedFile(string name, string text)
         {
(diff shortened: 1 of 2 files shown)


src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -32,11 +32,16 @@ private GeneratedCodeWorkspace(Project generatedCodeProject)
 
         public void AddGeneratedFile(string name, string text)
         {
-            _project = _project.AddDocument(GeneratedFolder + "/" + name, text, GeneratedFolders).Project;
+            var document = _project.AddDocument(GeneratedFolder + "/" + name, text, GeneratedFolders);
+            var root = document.GetSyntaxRootAsync().Result;
+            Debug.Assert(root != null);
+            root = root.WithAdditionalAnnotations(Simplifier.Annotation);
```
pakrym: Doing this beforehand instead in a loop saves a lot of resources.

pakrym:
Gives 40 sec improvement to CI build

PR #494, opened 2020-03-05
Remaining test server swaggers

MiYanni:
Update the list to accurately represent the remaining test server swaggers to do. 6 remain.


eng/Generate.ps1
```diff
@@ -52,15 +52,10 @@ $configurationPath = Join-Path $testServerDirectory 'readme.tests.md'
...
     'additionalProperties',
-    #'azure-composite-swagger',
     #'azure-parameter-grouping',
-    #'azure-report',
-    #'azure-resource',
-    #'azure-resource-x',
     #'azure-special-properties',
     'body-array',
     'body-boolean',
-    #'body-boolean.quirks',
```
pakrym: What's in all these quirks swaggers?

MiYanni: Lol.
image

pakrym: Should we delete them from the test server? Along with other swaggers without tests?

MiYanni: There might be v2 generators that use them to test things. We have no idea. 🐢

PR #495, opened 2020-03-06
Creates output directory

MiYanni:
Making directory since you need it to actually create files. AutoRest itself has no mechanism to create a directory (afaik).

### DIFF
--- a/src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
+++ b/src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using AutoRest.CSharp.V3.AutoRest.Communication;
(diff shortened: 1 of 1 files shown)


src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -27,6 +28,7 @@ internal class CSharpGen : IPlugin
     {
         public async Task<bool> Execute(IPluginCommunication autoRest, CodeModel codeModel, Configuration configuration)
         {
+            Directory.CreateDirectory(configuration.OutputFolder);
```
pakrym: I don't like going around autorest in this way.

pakrym: We are not writing files directly same should be with directories.

MiYanni: `GeneratedCodeWorkspace.Create` runs `Directory.GetFiles`. So, it really don't make a difference to use some other mechanism instead of `Directory.Create` here.

PR #497, opened 2020-03-06
Samples building

MiYanni:
Allowed samples to build on generation. Fixed build issues with samples.

### DIFF
--- a/samples/readme.samples.md
+++ b/samples/readme.samples.md
@@ -5,3 +5,19 @@ require: $(this-folder)/../readme.md
 output-folder: $(this-folder)/$(title)/$(title)
 clear-output-folder: false
 ```
+
+``` yaml $(include-csproj) != 'disable'
+pipeline:
+  csharpproj:
+    input: modelerfour/identity
+  csharpproj/emitter:
+    input: csharpproj
+    scope: output-scope
+```
+
+``` yaml $(dotnet-build) != 'disable'
+pipeline:
+  csharpgen/emitter/command:
+    input: csharpgen/emitter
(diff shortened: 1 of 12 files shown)


samples/readme.samples.md
```diff
@@ -4,4 +4,20 @@
 require: $(this-folder)/../readme.md
 output-folder: $(this-folder)/$(title)/$(title)
 clear-output-folder: false
+```
+
+``` yaml $(include-csproj) != 'disable'
```
MiYanni: For whatever reason, boolean values don't work for these conditions. However, string comparisons work! So, we use 'enable/disable' for this stuff. By default, we make a csproj and build. However, for Azure.Storage.Tables, we disable generating the csproj.

pakrym: Can we use a `true` as a string? Should we file this issue?

MiYanni: I did 'enable/disable' to avoid any confusion with booleans. And, I thought there was an issue filed before... when we ran into this like a few months back. But, I could be wrong.

samples/readme.samples.md
```diff
@@ -4,4 +4,20 @@
...
+pipeline:
+  csharpproj:
+    input: modelerfour/identity
+  csharpproj/emitter:
+    input: csharpproj
+    scope: output-scope
+```
+
+``` yaml $(dotnet-build) != 'disable'
```
pakrym: Why don't we put this logic into a common location with test projects?

MiYanni: Sure. I'll do that in a separate PR and just clean up all shared readmes (root, samples, TestProject, TestServerProjects). Give me a bit.

PR #498, opened 2020-03-06
Network interface sample

MiYanni:


pakrym:
Hm, where are all the client?

pakrym:
> Hm, where are all the client?

VS Code extension bug..

samples/Azure.Network.Management.Interface/Azure.Network.Management.Interface/ResourceNavigationLink.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using Azure.Core;
+
+namespace Azure.Network.Management.Interface.Models
+{
+    [CodeGenSchema("ResourceNavigationLink")]
+    public partial class ResourceNavigationLink : SubResource
+    {
+        [CodeGenSchemaMember("id")]
+        public string ResourceNavigationLinkId { get; internal set; }
```
pakrym: what was the conflict?

MiYanni: It is actually an issue in the swagger itself. It says allof SubResource, and SubResource has an Id. ResourceNavigationLink also has an Id. So, since we are using derived class logic for that interaction, you cannot redefine the property of the same name. So, I just changed the name of this Id.

pakrym: Is it the same serialized name? Should m4 handle this?

MiYanni: M4 can't handle it. They are 2 different properties with the same name in the 'same' schema. When you 'allof' something, you literally also have those properties. So, it is just a bad swagger or design. If ResourceNavigationLink.Id and SubResource.Id mean the same value, then ResourceNavigationLink.Id should be removed.

PR #503, opened 2020-03-07
Include csproj generation in repo only

pakrym:
The csproj generated are not suitable for the main repo and always need modifications. Having it on by default blocks updating to a new generator version.


pakrym:
I moved the dotnet build invocation to Generate.ps1 as I wasn't able to get conditions working correctly.

pakrym:
Also changed to unconditionally generated csprojes and updated the Azure.Core 1.1.0 to avoid version conflicts for search project.

eng/Generate.ps1
```diff
@@ -170,5 +169,10 @@ foreach ($key in $keys)
 {
     $definition = $swaggerDefinitions[$key];
     Invoke-AutoRest $definition.output $definition.title $definition.arguments
+    $projectPath = Join-Path $definition.output $definition.title;
+    if (!$noProjectBuild)
```
pakrym: allows to build even when running `-fast`

pakrym:
Merging to unblock generator update.

Issue #470, opened 2020-03-03
Enable changing the name of an enum


annelo-msft:
If not possible, I'll make generated enums internal and create separate enums for these with the public API name

Issue #463, opened 2020-02-28
Add support for transforming the names of enum values


annelo-msft:
I have a struct named FieldValueType, with values String, Integer and Object.  This causes compiler warnings, so I'd like to rename these values.

PR #505, opened 2020-03-08
Add support for enum mapping

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/470
Fixes: https://github.com/Azure/autorest.csharp/issues/463

cc @annelo-msft @tg-msft


test/TestProjects/TypeSchemaMapping/Customized/CustomDaysOfWeek.cs
```diff
@@ -0,0 +1,14 @@
+using Azure.Core;
+
+namespace NamespaceForEnums
+{
+    [CodeGenSchema("DaysOfWeek")]
+    internal partial struct CustomDaysOfWeek
+    {
+        [CodeGenSchemaMember("Monday")]
+        public static CustomDaysOfWeek FancyMonday { get; } = new CustomDaysOfWeek(FancyMondayValue);
```
pakrym: You can remap individual members of extensible enums.

test/TestProjects/TypeSchemaMapping/Customized/CustomFruitEnum.cs
```diff
@@ -1,9 +1,12 @@
 using Azure.Core;
 
-namespace AnotherCustomNamespace
+namespace NamespaceForEnums
 {
     [CodeGenSchema("Fruit")]
-    internal partial struct CustomFruitEnum
+    internal enum CustomFruitEnum
```
pakrym: Because `enum` types can't be partial you have to copy all the values to remap the enum.

Another cool feature is that generator would switch between generating extensible or normal enum based on what type `[CodeGenSchema]` is applied to.

PR #507, opened 2020-03-08
Filter well known headers

pakrym:


src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -25,6 +25,13 @@ namespace AutoRest.CSharp.V3.Output.Builders
 {
     internal class ClientBuilder
     {
+        private static string[] _knownResponseHeaders = new[]
+        {
+            "Date",
+            "ETag",
+            "x-ms-client-request-id",
+            "x-ms-request-id"
+        };
```
MiYanni: I'm fairly certain we need ETag for management plane. I believe it is used to update a resource in Patch operations. I'm not an expert on this, though. Date is probably fine to remove. The 2 x-ms values, I don't know how they're actually used.

pakrym: ETag is available through ReponseHeaders https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/ResponseHeaders.cs#L46 no need to have two strongly typed way to access it.

PR #508, opened 2020-03-08
Update launchsettings

pakrym:


eng/Generate.ps1
```diff
@@ -144,7 +144,7 @@ if ($updateLaunchSettings)
         $codeModel = Join-Path $outputPath 'CodeModel.yaml'
         $namespace = $definition.title.Replace('-', '_')
 
-        $settings.profiles[$key] = @{
+        $settings.profiles[$key] = [ordered]@{
```
pakrym: prevent random reordering.

PR #516, opened 2020-03-09
Multiple requests support

MiYanni:


src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -60,18 +60,26 @@ public ClientBuilder(BuildContext context)
...
-                    operationMethods.Add(operation.Language.Default.Name, new OperationMethod(operation, method));
+                    HttpRequest? httpRequest = serviceRequest.Protocol.Http as HttpRequest;
+                    if (httpRequest == null)
+                    {
+                        // Only handles HTTP requests
+                        continue;
+                    }
+
+                    RestClientMethod method = BuildMethod(operation, clientName, clientParameters, httpRequest, serviceRequest.Parameters);
+                    operationMethods.Add($"{operation.Language.Default.Name}{overloadCount++}", new OperationMethod(operation, method));
```
pakrym: Wouldn't this break other scenarios? Now everything would have a suffix so https://github.com/Azure/autorest.csharp/pull/516/files#diff-657045284e38bc0dd54ff5e452fe98eaR92 wouldn't work.

pakrym: Yep, 

```
FATAL: System.Exception: The x-ms-pageable operationName "Paging_NextFragment" for operation Paging_GetMultiplePagesFragmentNextLink was not found.
```

MiYanni: Fixed.

src/AutoRest.CSharp.V3/Properties/launchSettings.json
```diff
@@ -1,12 +1,12 @@
 {
   "profiles": {
     "additionalProperties": {
-      "commandLineArgs": "--standalone --input-codemodel=$(SolutionDir)\\test\\TestServerProjects\\additionalProperties\\CodeModel.yaml --plugin=csharpgen --output-folder=$(SolutionDir)\\test\\TestServerProjects\\additionalProperties --namespace=additionalProperties --shared-source-folder=$(SolutionDir)\\src\\assets --save-code-model=true",
-      "commandName": "Project"
+      "commandName": "Project",
```
pakrym: Argh, why did it reorder again?

MiYanni: I think you're looking at an old comparison. The only thing that is added (after my commit) was the media_types into the correct place.

pakrym:
> /azp run

What failed?

MiYanni:
@pakrym It just never started running (or the status wasn't being reported to GitHub).

Issue #523, opened 2020-03-10
Handle Binary in LRO


MiYanni:
Currently, LRO lambdas give you Response. However, Binary needs Message (the one that Response was contained in) to be able to call ExtractResponseContent() to get the Stream. So, the lambdas would need to change in a way to support this.

mayurid:
This is not something we have any asks for.

Issue #522, opened 2020-03-10
Handle multiple media types as a parameter


MiYanni:
Should be done after: https://github.com/Azure/autorest.modelerfour/issues/207

PR #524, opened 2020-03-10
Binary support

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/515

Normalized serialization cases (switches) and reuse of method serialization logic (instead of copy/paste).

Added issues:
https://github.com/Azure/autorest.csharp/issues/522
https://github.com/Azure/autorest.csharp/issues/523


src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -383,5 +383,25 @@ public static void ToDeserializeCall(this CodeWriter writer, ITypeProvider imple
             SerializationFormat.Duration_ISO8601 => "P",
             _ => null
         };
+
+        public static void WriteMethodDeserialization(this CodeWriter writer, JsonSerialization serialization, bool async,
```
pakrym: WriteResponseDeserialization.

src/AutoRest.CSharp.V3/Generation/Writers/XmlCodeWriterExtensions.cs
```diff
@@ -375,5 +376,16 @@ private static void ToDeserializeValueCall(this CodeWriter writer, XmlValueSeria
...
             }
         }
+
+        public static void WriteMethodDeserialization(this CodeWriter writer, XmlElementSerialization serialization,
+            ref string destination, string response, string document = "document")
+        {
+            writer.Line($"var {document:D} = {typeof(XDocument)}.Load({response:D}.ContentStream, LoadOptions.PreserveWhitespace);");
+            writer.ToDeserializeCall(
+                serialization,
+                w => w.Append($"{document:D}"),
```
pakrym: document seems to be declared twice here.

src/AutoRest.CSharp.V3/Output/Models/Serialization/BinarySerialization.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
+using AutoRest.CSharp.V3.Generation.Types;
+
+namespace AutoRest.CSharp.V3.Output.Models.Serialization.Xml
+{
+    internal class BinarySerialization : ObjectSerialization
```
pakrym: Not sure I like this, binary is not a way to serialize objects, it's completely different kind of request body, multipart would eventually be another one.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
MiYanni: I called it 'method' deserialization because the 2 serialization contexts are: Method and Model. Meaning, you write serialization information into the client methods, and you also write serialization information for the models. So, if we would refactor, I think that split in naming makes the most sense.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -259,63 +259,46 @@ private void WriteCreateOperationOperation(CodeWriter writer, LongRunningOperati
...
+                            {
+                                case JsonSerialization jsonSerialization:
+                                    writer.WriteMethodDeserialization(jsonSerialization, async: false, ref valueVariable, responseVariable);
+                                    break;
+                                case XmlElementSerialization xmlSerialization:
+                                    writer.WriteMethodDeserialization(xmlSerialization, ref valueVariable, responseVariable);
+                                    break;
+                                default:
+                                    throw new NotSupportedException();
+                            }
```
pakrym: Why not extract this entire switch?

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -415,42 +415,34 @@ private void WriteStatusCodeSwitch(CodeWriter writer, ResponseBody? responseBody
...
-                                );
+                                writer.WriteMethodDeserialization(xmlSerialization, ref valueVariable, responseVariable);
                                 break;
                         }
                     }
                     else if (responseBody is StreamResponseBody _)
                     {
-                        writer.Line($"var {valueVariable:D} = message.ExtractResponseContent();");
+                        writer.Line($"var {valueVariable:D} = {messageVariable:D}.ExtractResponseContent();");
+                        using (writer.Scope($"if ({valueVariable:D} == null)"))
```
pakrym: We don't check that content isn't null in other branches why do it here?

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
pakrym: Yeah but it sounds like we're deserializing a method.

MiYanni: It is like Baby Clothes. It is clothes for babies, not made out of babies. 👶 But, I'll change it to `WriteDeserializationForMethods` instead. Then, the others could be renamed to `WriteDeserializationForModels` in another PR.

src/AutoRest.CSharp.V3/Generation/Writers/XmlCodeWriterExtensions.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/Output/Models/Serialization/BinarySerialization.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: The other serializers create non-null Value variables (because of their deserialization mechanics). Here, we don't get to do that. The other deserializers will error, however, if it is null because it cannot parse the content. So, there's not a need to throw a different error in that situation.

pakrym: I'm not sure I'm a fan of how the Exception would look. This might result in an exception with  "200 success" status code in a message. Let's create an extension method for HttpPipelineMessage that checks stream for null and throws some reasonable exception.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -259,63 +259,46 @@ private void WriteCreateOperationOperation(CodeWriter writer, LongRunningOperati
...
+                            writer.Line($"return {valueVariable};");
                         }
                     }
+                    else if (lroMethod.OriginalResponse.ResponseBody is StreamResponseBody)
+                    {
+                        throw new NotSupportedException("Binary is not supported as message (not response) is required for ExtractResponseContent() call.");
```
pakrym: Add link to the issue here.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -136,31 +136,32 @@ private void WriteRequestCreation(CodeWriter writer, RestClientMethod operation)
...
+                            writer.Line($"request.Content = content;");
+                            break;
+                        case XmlElementSerialization xmlSerialization:
+                            writer.Line($"using var content = new {typeof(XmlWriterContent)}();");
+                            writer.ToSerializeCall(
+                                xmlSerialization,
+                                WriteConstantOrParameter(value, ignoreNullability: true),
+                                writerName: w => w.Append($"content.{nameof(XmlWriterContent.XmlWriter)}"));
+                            writer.Line($"request.Content = content;");
+                            break;
```
pakrym: default throw?

MiYanni: The original code didn't throw here. This switch is stylistically the same as the rest of the serialization handling. It is also equivalent to the new WriteDeserializationForMethods on CodeWriterExtensions (your suggestion to encapsulate the switch itself) which also doesn't throw.

Either we should throw whenever we can't handle something (which we aren't particularly consistent on) or we don't, and let the code write gracefully (meaning, it still writes compile-able code, but just doesn't have that section). In this case here, if we don't write this section (switch falls through), it will still write a valid method. The request simply won't have content.

pakrym: I think it might be time to tighten things down now that we have most of the features. This would allow us to easily discover corner cases.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: We already have failures that result in 200 success, as we default to this same throw at the bottom of this method (status code switch's default). While I agree with you, I think it makes more sense to look at our throws that we write as a whole and clean them up in a single PR. I love refactoring, but we need to be complete when doing it (not just looking at a specific situation at a time).

pakrym: Let's skip this check here then and add them to all places all at once.

MiYanni: Okay. Just FYI, that means it will technically be returning `Response<Stream?>` instead of `Response<Stream>`. Since we don't have nullability for those projects, they don't care, though.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: I added throws in a similar way that I found elsewhere (NotImplementedException) to any relevant serialzation processing switches.

PR #526, opened 2020-03-11
Enable more LRO scenarios

pakrym:
Fixes https://github.com/Azure/autorest.csharp/issues/525 as well.

I've reenabled only the obvious methods to unblock storage.


src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -304,27 +283,45 @@ private RestClientMethod BuildMethod(Operation operation, string clientName, IRe
...
+                    response.HttpResponse.StatusCodes.Select(ToStatusCode).ToArray(),
+                    responseHeaderModel
+                );
             }
-            else if (response is BinaryResponse)
+            else
             {
-                responseBody = new StreamResponseBody();
+                // Special case for httpInfrastructure testServer swagger, service method always fails
+                clientResponse = new Response(null, Array.Empty<int>(), null);
```
MiYanni: Is this case supposed to pass for that test? I'm confused.

pakrym: This is only to make `httpInfrastructure` weird test swagger work. Nothing else should really be hitting this.

pakrym: I would prefer to have an exception where there is no response but can't.

test/TestServerProjects/paging/Generated/Operations/PagingRestClient.cs
```diff
@@ -1048,11 +1048,7 @@ public async ValueTask<Response<ProductResult>> GetMultiplePagesLROAsync(string
                 switch (message.Response.Status)
                 {
                     case 202:
-                        {
-                            using var document = await JsonDocument.ParseAsync(message.Response.ContentStream, default, cancellationToken).ConfigureAwait(false);
-                            var value = ProductResult.DeserializeProductResult(document.RootElement);
-                            return Response.FromValue(value, message.Response);
-                        }
```
MiYanni: What made the return type change here? It is still the same status code.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: So, the generator fails for `httpInfrastructure` if that isn't there? What changed from before where this wasn't needed?

pakrym: `clientResponse` was nullable and we checked it all the time.

test/TestServerProjects/paging/Generated/Operations/PagingRestClient.cs
pakrym: This is now treated as LRO and not a paging method so we don't care about the initial response type.

PR #528, opened 2020-03-12
Updated m4 to latest

MiYanni:
Fixes: https://github.com/Azure/autorest.modeler…four/issues/196 and https://github.com/Azure/autorest.modelerfour/issues/197
Relates to: https://github.com/Azure/perks/pull/94

This fixed the reference issues in the code model schema. It also fixed the type tags in the code model not using HttpBinaryResponse. Also, the new feature where unused flattened models are removed is enabled by default. So, bye bye to those. If we want to disable it, we'd have to add `keep-unused-flattened-models: true` to the readme. However, we have no reason currently to keep them.


pakrym:
Can you remove the workaround from https://github.com/Azure/autorest.csharp/tree/feature/v3/samples/Azure.Storage.Management#swagger-tweaks-to-get-things-working to test https://github.com/Azure/autorest.modelerfour/issues/205 fix?

MiYanni:
@pakrym It fixed that issue too! Niccccceeeeeeeee.

Issue #531, opened 2020-03-12
Parse enums case-insensitively


pakrym:
It's against the spec but there are just too many violations.

PR #532, opened 2020-03-12
Ignore casing when parsing enums

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/531

### DIFF
--- a/src/AutoRest.CSharp.V3/Generation/Writers/SerializationWriter.cs
+++ b/src/AutoRest.CSharp.V3/Generation/Writers/SerializationWriter.cs
@@ -167,7 +167,7 @@ namespace AutoRest.CSharp.V3.Generation.Writers
                     {
                         foreach (EnumTypeValue value in schema.Values)
                         {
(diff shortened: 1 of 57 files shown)


src/AutoRest.CSharp.V3/Generation/Writers/SerializationWriter.cs
```diff
@@ -163,14 +163,14 @@ private void WriteSealedChoiceSerialization(CodeWriter writer, EnumType schema)
                     }
                     writer.Line();
 
-                    using (writer.Scope($"public static {declaredTypeName} To{declaredTypeName}(this string value) => value switch", end: "};"))
+                    using (writer.Scope($"public static {declaredTypeName} To{declaredTypeName}(this string value)"))
                     {
                         foreach (EnumTypeValue value in schema.Values)
                         {
-                            writer.Line($"{value.Value.Value:L} => {declaredTypeName}.{value.Declaration.Name},");
+                            writer.Line($"if (string.Equals(value, {value.Value.Value:L}, StringComparison.InvariantCultureIgnoreCase)) return {declaredTypeName}.{value.Declaration.Name};");
```
MiYanni: Should be `{typeof(string)}` and `{typeof(StringComparison)}`

PR #534, opened 2020-03-13
Updated all the things

MiYanni:


pakrym:
Let's update to latest core too

samples/Azure.Storage.Management/Azure.Storage.Management/Generated/Models/ProxyResource.Serialization.cs
```diff
@@ -1,71 +0,0 @@
...
-// Licensed under the MIT License.
-
-// <auto-generated/>
-
-#nullable disable
-
-using System.Text.Json;
-using Azure.Core;
-
-namespace Azure.Storage.Management.Models
```
pakrym: Why were these deleted?

MiYanni: The swaggers referenced, file, blob, and storage, import `common-types/resource-management/v1/types.json` for some other types that it uses. However, in this case, ProxyResource, is not used by anything in file, blob, and storage. The same for Network, as it has a shared set of resources, but doesn't use all of them in NetworkInterface.

MiYanni: Aka, it was further unreferenced schema reduction by now also looking at shared swaggers.

PR #535, opened 2020-03-13
Add support for trimming newlines and newlines between methods

pakrym:
So the generated code is a bit easier to read.


src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
```diff
@@ -2,25 +2,36 @@
...
+        private const char NewLineChar = '\n';
+        private static readonly string _newLine = "\n";
+
         private readonly List<string> _usingNamespaces = new List<string>();
-        private readonly StringBuilder _builder = new StringBuilder();
+
         private readonly Stack<CodeWriterScope> _scopes;
         private string? _currentNamespace;
 
+        private char[] _builder;
```
pakrym: Switched away from StringBuilder so I can do non-allocating Substrings using spans.

Issue #486, opened 2020-03-04
Serializer doesn't generate a using for enum.ToSerialString call


pakrym:
When serializer uses Enum ToSerialString extension method it should add an explicit using in case the enum type is in another namespace.

cc @annelo-msft

PR #536, opened 2020-03-13
Support enums in queries, urls and headers

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/325
Fixes: https://github.com/Azure/autorest.csharp/issues/486


src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -172,7 +174,8 @@ private void WriteRequestCreation(CodeWriter writer, RestClientMethod operation)
 
         private void WriteUriFragment(CodeWriter writer, PathSegment segment)
         {
-            writer.Append($"uri.AppendRaw({WriteConstantOrParameter(segment.Value)}");
+            writer.Append($"uri.AppendRaw(");
+            WriteConstantOrParameter(writer, segment.Value);
```
pakrym: Failed experiment...

Issue #419, opened 2020-01-23
Avoid collisions with the name `request`


tg-msft:
One of the Search models used the word `request` as a parameter name which caused a conflict in the generation of the `Create*Request` method that has a local named `request`.  I added a swagger README transform to rename the property for the time being.

tg-msft:
(and https://github.com/Azure/autorest.csharp/tree/feature/v3/samples/CognitiveSearch#swagger-tweaks-to-get-things-working will need to have the workaround removed)

PR #540, opened 2020-03-13
Avoid some variable conflicts

pakrym:
I'm sure there are more but at least now we have a pattern to use.

Fixes: https://github.com/Azure/autorest.csharp/issues/419


src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -145,36 +145,36 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
 
         public static void ToDeserializeCall(this CodeWriter writer, JsonSerialization serialization, CodeWriterDelegate element, ref string destination)
         {
-            destination = writer.GetTemporaryVariable(destination);
+            var destinationDeclaration = new CodeWriterDeclaration(destination);
 
             if (serialization is JsonValueSerialization valueSerialization)
             {
-                writer.Append($"var {destination:D} =")
+                writer.Append($"var {destinationDeclaration:D} =")
```
pakrym: First time a declaration is used with `:D` suffix it gets an `ActualName` assigned that is unique in this scope.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -91,49 +91,55 @@ private void WriteClientCtor(CodeWriter writer, RestClient restClient, CSharpTyp
 
         private void WriteRequestCreation(CodeWriter writer, RestClientMethod operation)
         {
+            using var methodScope = writer.AmbientScope();
```
pakrym: We need an ambient scope around methods because parameter declarations appear before main method scope starts.

MiYanni: Based on this, might make sense to go back to a code writer Method call, where it declares the parameters in the method's scope, instead of just ad-hoc appending them around parentheses like we are doing below. Not necessary for this PR, but something that would clean this and other methods up a bit.

Issue #529, opened 2020-03-12
Support model classification into input/output/both


pakrym:
And only generate appropriate serializers.

PR #541, opened 2020-03-14
Take model kind into account when generating properties and serialization

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/529


pakrym:
cc @heaths related to what you are doing

heaths:
Adding myself as a reviewer so I can view in Code. GH PRs don't handle large PRs well.

pakrym:
> I'm hoping read-only'ness will be as easy.

It won't :) The serialization code depends on type being mutable but I'm working on making serializer changes that would make it easier.

heaths:
Ideally, if we have a "property bag" of those attributes we can pass down it would be easier. I just haven't had time to go back to it since I've been wrestling to just get the models I have now out.

PR #542, opened 2020-03-16
Add initial constructor generation features

pakrym:
Next step would be to use ctors from deserialization code.

Needs https://github.com/Azure/autorest.csharp/pull/541 to generate correct accessibility for default ctor based on type of model.


src/AutoRest.CSharp.V3/Output/Models/Types/ObjectType.cs
```diff
@@ -101,7 +101,12 @@ private IEnumerable<ObjectTypeConstructor> BuildConstructors()
             }
 
             yield return new ObjectTypeConstructor(
-                BuilderHelpers.CreateMemberDeclaration(Type.Name, Type, "internal", _sourceTypeMapping?.DefaultConstructor),
+                BuilderHelpers.CreateMemberDeclaration(
+                    Type.Name,
+                    Type,
+                    // inputs have public ctor by default
+                    _objectSchema.IsInput ? "public" : "internal",
+                    _sourceTypeMapping?.DefaultConstructor),
```
MiYanni: I was wondering about them both being internal. I just assumed this initial implementation didn't matter or not. Something that would have gotten fixed up when they actually get called.

pakrym: Don't want to break existing generated clients.

PR #549, opened 2020-03-18
Parameter flattening support [JSON only]

MiYanni:
Added this issue for XML serialization changes that would be needed: https://github.com/Azure/autorest.csharp/issues/550


src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -49,18 +49,19 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
                             writer.Line($"{writerName}.WritePropertyName({property.Name:L});");
                             writer.ToSerializeCall(
                                 property.ValueSerialization,
-                                w => w.Append($"{property.MemberName}"));
+                                w => w.Append($"{property.MemberName}"),
+                                writerName);
```
MiYanni: This was a bug that we didn't notice before doing parameter flattening.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -248,7 +248,12 @@ private RestClientMethod BuildMethod(Operation operation, string clientName, IRe
                             }
                             else
                             {
-                                var serialization = _serializationBuilder.Build(httpRequestWithBody.KnownMediaType, requestParameter.Schema, requestParameter.IsNullable());
+                                VirtualParameter[] flattenedParameters = Array.Empty<VirtualParameter>();
+                                if (requestParameter.Flattened == true)
+                                {
+                                    flattenedParameters = parameters.OfType<VirtualParameter>().Where(vp => vp.OriginalParameter == requestParameter).ToArray();
```
MiYanni: This gets the VirtualParameters (those that have been flattened) based on the original requestParameter.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -86,24 +87,38 @@ private XmlValueSerialization BuildXmlValueSerialization(Schema schema, bool isN
...
+                            new JsonPropertySerialization(
+                                fp.TargetProperty.SerializedName,
+                                fp.Schema is ConstantSchema constantSchema
+                                    ? w => w.WriteConstant(BuilderHelpers.ParseConstant(constantSchema, _typeFactory))
+                                    : (CodeWriterDelegate)(w => w.Append($"{fp.CSharpName()}")),
+                                BuildSerialization(fp.TargetProperty.Schema, fp.TargetProperty.IsNullable(), Array.Empty<VirtualParameter>())))
+                            .ToArray(),
+                        null,
+                        _typeFactory.CreateImplementationType(objectSchema, isNullable)
+                    );
```
MiYanni: This thing is the main secret sauce. Basically, here, we let all ObjectSchemas get processed as JsonValueSerialization prior. Now, if we get any flattenedParameters, we recognize this should be processed as an object serialization. We convert the flattenedParameters into JsonPropertySerialization, which some extra special sauce added for constants.

src/AutoRest.CSharp.V3/Output/Models/Serialization/Json/JsonPropertySerialization.cs
```diff
@@ -1,19 +1,21 @@
...
+        public JsonPropertySerialization(string name, CodeWriterDelegate? memberName, JsonSerialization valueSerialization)
         {
             Name = name;
             MemberName = memberName;
             ValueSerialization = valueSerialization;
         }
 
         public string Name { get; }
-        public string? MemberName { get; }
+        public CodeWriterDelegate? MemberName { get; }
```
MiYanni: This change was to facilitate constants, since constant properties were written the same for serialization as normal properties. However, for flattening, we don't have a property name, so we make this delegate write the actual constant value directly; which we already had support for in both the property writer (assignment of the property) and then as a constant in the request parts.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -207,15 +207,15 @@ private RestClientMethod BuildMethod(Operation operation, string clientName, IRe
...
+                            constantOrParameter = BuilderHelpers.ParseConstant(constant, _typeFactory);
                             valueSchema = constant.ValueType;
                             break;
                         default:
                             constantOrParameter = BuildParameter(requestParameter);
                             break;
                     }
 
-                    if (!constantOrParameter.IsConstant)
+                    if (!constantOrParameter.IsConstant && requestParameter.Flattened != true)
```
MiYanni: This makes sure we don't put any of the flattened parameters into the method definition.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -86,24 +87,38 @@ private XmlValueSerialization BuildXmlValueSerialization(Schema schema, bool isN
...
-                        new JsonDynamicPropertiesSerialization(BuildSerialization(dictionarySchema.ElementType, false)),
+                        new JsonDynamicPropertiesSerialization(BuildSerialization(dictionarySchema.ElementType, false, Array.Empty<VirtualParameter>())),
                         _typeFactory.CreateImplementationType(dictionarySchema, isNullable)
                         );
+                case ObjectSchema objectSchema when flattenedParameters.Any():
+                    return new JsonObjectSerialization(
+                        _typeFactory.CreateType(objectSchema, isNullable),
+                        flattenedParameters.Select(fp =>
+                            new JsonPropertySerialization(
+                                fp.TargetProperty.SerializedName,
```
pakrym: This is very hard to understand.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -86,24 +87,38 @@ private XmlValueSerialization BuildXmlValueSerialization(Schema schema, bool isN
...
+                                fp.Schema is ConstantSchema constantSchema
+                                    ? w => w.WriteConstant(BuilderHelpers.ParseConstant(constantSchema, _typeFactory))
+                                    : (CodeWriterDelegate)(w => w.Append($"{fp.CSharpName()}")),
```
pakrym: We shouldn't be using `CSharpName` anywhere outside of model building, we should pull the name of parameters or properties.

src/AutoRest.CSharp.V3/Output/Models/Serialization/Json/JsonPropertySerialization.cs
pakrym: I'm not a fan because it kind of break the abstraction, you can't deserialize into a constant.

Can we handle constants same way we did with object properties here where we still generate a parameter with a default value until we figure out exactly what semantics are expected.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -163,33 +168,54 @@ private XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema objectSc
...
+                if (flattenedParameters.Any())
+                {
+                    VirtualParameter? virtualParameter = flattenedParameters.FirstOrDefault(fp => fp.TargetProperty == property);
+                    if (virtualParameter != null)
+                    {
+                        propertyName = virtualParameter.Schema is ConstantSchema constantSchema
+                            ? w => w.WriteConstant(BuilderHelpers.ParseConstant(constantSchema, _typeFactory))
+                            : (CodeWriterDelegate)(w => w.Append($"{virtualParameter.Language.Default.Name}"));
+                    }
+                }
```
MiYanni: Now, since we're going through the properties, I simply find the virtual parameter that matches the property, and extract the correct name from it.

src/AutoRest.CSharp.V3/Output/Models/Types/ObjectType.cs
```diff
@@ -20,7 +20,6 @@ namespace AutoRest.CSharp.V3.Output.Models.Types
     internal class ObjectType : ISchemaType
     {
         private readonly ObjectSchema _objectSchema;
-        private readonly BuildContext _context;
```
MiYanni: Unused field.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
pakrym: Why do we need to do it? don't we have them as method parameterS?

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -86,24 +89,26 @@ private XmlValueSerialization BuildXmlValueSerialization(Schema schema, bool isN
...
                 case DictionarySchema dictionarySchema:
                     return new JsonObjectSerialization(
                         _typeFactory.CreateType(dictionarySchema, isNullable),
                         Array.Empty<JsonPropertySerialization>(),
-                        new JsonDynamicPropertiesSerialization(BuildSerialization(dictionarySchema.ElementType, false)),
+                        new JsonDynamicPropertiesSerialization(BuildSerialization(dictionarySchema.ElementType, false, Array.Empty<VirtualParameter>())),
                         _typeFactory.CreateImplementationType(dictionarySchema, isNullable)
                         );
+                case ObjectSchema objectSchema when flattenedParameters.Any():
+                    return BuildJsonObjectSerialization(objectSchema, (ObjectType)_library.FindTypeForSchema(objectSchema), flattenedParameters);
```
pakrym: Why don't you pass the ObjectType that we got into the parent method here?

pakrym: Hm, shouldn't we have this logic in BuildObject?

src/AutoRest.CSharp.V3/Output/Models/Serialization/Json/JsonPropertySerialization.cs
MiYanni: The parameter flattening is only used for serialization, not deserialization, so there isn't a use case where MemberName would represent a constant in deserialization, given the way flattened parameter JsonPropertySerialization is created. 

Our overall serialization story is fractured and needs to be normalized. Right now, there is a mix between de/serialization access and de/serialization construction that is non-uniform. I still like the idea of writing 1 serializer method and 1 deserializer method per nested (model) object, even if the object isn't used. It makes our generator code much cleaner since all de/serialization is always written to 1 place in the generated output. The more different places you try to reuse the same generated output logic, the more complex the generator's code becomes.

I believe this change shows that any data that is to be written and has the potential to be swapped out should be a CodeWriterDelegate, as it allows the information to not be pre-created as a string; it has a dynamic aspect to it. It also allows it to always  work with our type system (formattable string writers) or any other system we tie into code writing.

I don't like the idea of adding a parameter for which the user could provide and break the operation. However, the more reasonable option would be to have a place to inject writing a variable to represent the constant, and use that as the MemberName. But, that requires a lot more 'spaghetti logic' to an already complex hierarchy of calls.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
MiYanni: There is no ObjectType from the parent method.

pakrym: Why not build object?

MiYanni: Because BuildSerialization is specifically for JSON (the method this code is in). BuildObject is to determine which media type builder to call. Here, we know which to call because we're in JSON.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: No. `requestParameter.Flattened` are the parameters (original) that are of the model type. The actual parameters we want to use are of type VirtualParameters. This entire request building process could be refined if we used signatureParameters, because that filtering work was already done for us. However, currently, we're creating a constant parameter, and signatureParameters won't have constants in it. Plus, that work to use signatureParameters it outside the scope of these changes.

src/AutoRest.CSharp.V3/Output/Models/Serialization/Json/JsonPropertySerialization.cs
MiYanni: Reverted this back to string, since we're generating a constant parameter instead now.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
pakrym: Why is it specific to JSON?

MiYanni: We don't have model flattening for XML (https://github.com/Azure/autorest.csharp/issues/433), so a similar method won't work until then.

This place in the code is specifically a JSON codepath. I'll update the name to be more appropriate, so it is clear. You can see usage here:
image

pakrym: Then make https://github.com/Azure/autorest.csharp/blob/8e78e6f326f361dacde8e70c3396e74601b34697/src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs#L208 public and call it directly.

pakrym: I don't like this roundabout way of getting `ObjectType` when most methods just take them as parameter.

MiYanni: I'm so confused. We are in the class that contains `BuildJsonObjectSerialization`, so making it public has no effect. The parent calls are to the method `Build`, which is public, and doesn't take `ObjectType`. `Build` itself has no association to ObjectType, because you're normally not building an object. `BuildObject` is for building objects. However, we're now utilizing the object building mechanism to build it in-line. That's the trade-off of doing this style of outputting. This section of code is for outputting values, and values normally have methods to call to do their serialization. Here, we're leveraging the object serialization in the value serialization pipeline, which is exactly what we agreed upon. Make sense?

PR #551, opened 2020-03-18
Fix bugs with coverage

pakrym:


test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
```diff
@@ -99,20 +99,20 @@ public async Task ResetAsync()
...
+            await CollectCoverage(results, "/report/optional", testName);
 
             return results.ToArray();
         }
 
-        private async Task CollectCoverage(HashSet<string> results, string url)
+        private async Task CollectCoverage(HashSet<string> results, string url, string testName)
         {
-            var coverageString = await Client.GetStringAsync($"{url}?qualifier={Environment.TickCount64}");
+            var coverageString = await Client.GetStringAsync($"{url}?qualifier={testName}");
```
pakrym: Prevent infinite spawning of files.

eng/PublishRelease.ps1
```diff
@@ -12,16 +13,31 @@ try
...
\ No newline at end of file
+}
+
+
+pushd $RepoRoot
+try
+{
+    # set the version in the root package.json so coverage-push can pick it up
+
+   npm version --no-git-tag-version $devVersion | Out-Null;
```
pakrym: coverage-push picks up the version from the root of the repo but our main package.json is not there, still update the version as a workaround.

src/AutoRest.CSharp.V3/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0",
+  "version": "3.0.0",
```
pakrym: rev the version so new coverage appears at the top of the list.

PR #576, opened 2020-03-26
LRO tests fix/unignore

MiYanni:
Allowed tests that work now to run. Fixed tests that made incorrect assumptions.


test/AutoRest.TestServer.Tests/lro.cs
```diff
@@ -104,7 +104,7 @@ public class LroTest: TestServerTestBase
         });
 
         [Test]
-        [Ignore("Handle multiple responses: https://github.com/Azure/autorest.csharp/issues/413")]
+        [Ignore("Handle empty response body: https://github.com/Azure/autorest.csharp/issues/300")]
```
MiYanni: These paths are defined to return `Product`. However, these routes on the test server don't return a body.

pakrym: Is it supposed to be an invalid response? Should we fix a test server? How do other languages handle this test?

test/AutoRest.TestServer.Tests/lro.cs
```diff
@@ -736,27 +718,27 @@ public class LroTest: TestServerTestBase
...
         public Task LROPostAsyncNoRetrySucceeded() => Test(async (host, pipeline) =>
         {
             var value = new Product();
             var operation = await new LROsClient(ClientDiagnostics, pipeline, host).StartPostAsyncNoRetrySucceededAsync(value);
             var result = await operation.WaitForCompletionAsync().ConfigureAwait(false);
             Assert.AreEqual("100", result.Value.Id);
             Assert.AreEqual("foo", result.Value.Name);
-            Assert.AreEqual("Succeeded", result.Value.ProvisioningState);
+            // Final GET does not return provisioning state
+            Assert.AreEqual(null, result.Value.ProvisioningState);
```
pakrym: We should fix the test: https://github.com/Azure/autorest.testserver/blob/master/legacy/routes/lros.js#L782

I think it needs `res.end(outStr)` after the line I've linked.

test/AutoRest.TestServer.Tests/lro.cs
```diff
@@ -480,27 +472,19 @@ public class LroTest: TestServerTestBase
...
-        [Ignore("Handle multiple responses: https://github.com/Azure/autorest.csharp/issues/413")]
         public Task LROErrorPut201NoProvisioningStatePayload() => Test(async (host, pipeline) =>
         {
             var value = new Product();
             var operation = await new LrosaDsClient(ClientDiagnostics, pipeline, host).StartPutError201NoProvisioningStatePayloadAsync(value);
-            var result = await operation.WaitForCompletionAsync().ConfigureAwait(false);
-            Assert.AreEqual("100", result.Value.Id);
-            Assert.AreEqual("foo", result.Value.Name);
-            Assert.AreEqual("Succeeded", result.Value.ProvisioningState);
+            Assert.ThrowsAsync<RequestFailedException>(async () => await operation.WaitForCompletionAsync().ConfigureAwait(false));
```
pakrym: Not sure if this one should throw. Can you double check if it does in python?

MiYanni: `LROError` are error situations. Here, we go down a failure path, which throws.

pakrym: Then we need to fix LROErrorDeleteNoLocation_Sync

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: Technically, it isn't invalid to return no ProvisioningState after the final GET. But yes, I saw that line in routes, which is a copy-pasted line from other similar routes. Yeah, it should be fixed.

pakrym: Let's fix it before merging this PR.

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: It isn't supposed to be invalid. But, there are other tests that return empty bodies... which we haven't handled yet. I don't know what the recommendation is.

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: Created: https://github.com/Azure/autorest.testserver/pull/154

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: So, `LROErrorDeleteNoLocation` simply does:
```
res.status(204).end();
```
Which means it is a DELETE operation that returns 204 (which is valid). It succeeds because of:
```csharp
// If provisioningState was not found, it defaults to Succeeded.
if (_headerFrom == HeaderFrom.None)
{
    return true;
}
```
This is supposed to be how you handle a terminal polling state in this situation. So, I don't know what is an 'error' about this test. I'll check python to see what they are doing.

test/AutoRest.TestServer.Tests/lro.cs
pakrym: I verified that empty response is invalid when swagger defines a model, so Assert.Throws should be fine

MiYanni: Fixed.

test/AutoRest.TestServer.Tests/lro.cs
MiYanni: Fixed!

PR #578, opened 2020-03-26
Add a null check for content parameter

pakrym:
Reenables one test.


pakrym:
I'll merge this to unblock other RP checcks.

PR #581, opened 2020-03-26
Add parameter grouping support

pakrym:


src/AutoRest.CSharp.V3/Input/CodeModelPartials.cs
```diff
@@ -95,6 +95,11 @@ public ObjectSchema()
         public bool IsOutput => Usage.Contains(SchemaContext.Output);
     }
 
+    // redefined manually to inherit from ObjectSchema
+    internal partial class GroupSchema : ObjectSchema
```
MiYanni: Would it make sense to make an issue for M4 to update it to do this?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -215,10 +215,24 @@ private RestClientMethod BuildMethod(Operation operation, string clientName, IRe
...
+                        else
+                        {
+                            constantOrReference = parameter;
+                        }
                     }
 
-                    if (parameter != null && requestParameter.Flattened != true)
+                    if (parameter != null &&
+                        requestParameter.Flattened != true &&
+                        requestParameter.GroupedBy == null)
```
MiYanni: This if-statement is the only thing stopping the extraction of creating `Request` as a separate method (for encapsulation and maintenance purposes). I tried this before, but it is a bit tangled. Not for this PR, but it would be nice to have `BuildMethod` just call separate methods for each of the pieces of information it needs.

src/AutoRest.CSharp.V3/Input/CodeModelPartials.cs
pakrym: Not sure, it's not like what they are doing is wrong, it's just more convenient for us to parse it this way.

PR #585, opened 2020-03-27
Support multiple response types

pakrym:


src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -307,67 +307,81 @@ private RestClientMethod BuildMethod(Operation operation, string clientName, IRe
...
+                clientResponse.ToArray(),
+                responseHeaderModel,
                 new Diagnostic($"{clientName}.{operationName}", Array.Empty<DiagnosticAttribute>())
             );
         }
 
+        // Merges operations without response types types together
+        private CSharpType? ReduceResponses(List<Response> responses)
+        {
+            foreach (var typeGroup in responses.GroupBy(r=> r.ResponseBody))
```
MiYanni: Does this not need to be a separate list if you're adding/removing items from it?

pakrym: `responses.GroupBy` consumes the entire list before starting to return results so another loop is not required.

PR #587, opened 2020-03-28
Allow overriding and suppressing any generated members

pakrym:
Now you can override any generated member by defining a member with the same name in a partial class.

If you want to remove some generated member without providing own implementation `CodeGenSuppressAttribute` can be used.

For example: `[CodeGenSuppress("CustomizedModel(CustomFruitEnum, CustomDaysOfWeek)")]` to suppress method/constructor or `[CodeGenSuppress("PropertyName)")]` for properties/fields.

Sometimes removing members would leave empty types or funny formatting, I'll deal with it later.

cc @heaths @tg-msft


src/AutoRest.CSharp.V3/Generation/Writers/ModelWriter.cs
```diff
@@ -262,7 +251,7 @@ private void WriteChoiceSchema(CodeWriter writer, EnumType schema)
                 var implementType = new CSharpType(typeof(IEquatable<>), cs);
                 using (writer.Scope($"{schema.Declaration.Accessibility} readonly partial struct {name}: {implementType}"))
                 {
-                    writer.Line($"private readonly string? _value;");
+                    writer.Line($"private readonly {typeof(string)} _value;");
```
MiYanni: 👍

test/TestProjects/TypeSchemaMapping/Customized/CustomizedModel.cs
```diff
@@ -1,16 +1,73 @@
+using System.Text.Json;
 using Azure.Core;
+using NamespaceForEnums;
 
 namespace CustomNamespace
 {
     [CodeGenSchema("Model")]
+    [CodeGenSuppress("CustomizedModel(CustomFruitEnum, CustomDaysOfWeek)")]
```
MiYanni: Would it make sense to take multiple parameters (representing the constructor parameters) instead of a single string representation? I'm a bit confused how this format of string is matched with a constructor in the MemberRemoverRewriter code.

pakrym: Yeah, let me try that.

The ToDisplayString call on this line is what's generating this string notation.
https://github.com/Azure/autorest.csharp/pull/587/files#diff-b9c3140b06337ba81fab9579487c05b4R99

heaths: Hmm, and now that I think about it, are parameter names possible as well? What if types are the same for parameters, but names are different (e.g. we want to reorder a constructor's parameters that all happen to be the same - or mostly the same - types)?

pakrym: If the parameter types are the same it would be considered the same overload and handled by overriding logic.

Issue #226, opened 2019-11-08
Debugging comments (includes swagger information)


MiYanni:
- Include debug information as comments in a debug build related to swagger names/info
- Start and end braces would also include `// class XYZ start` as a debug option.

pakrym:
We haven't found a need for this in a while. Will reopen if anyone asks.

PR #589, opened 2020-03-29
Map types by original name instead of schema name/serialized name

pakrym:
It's way easier to know what to put into an attribute when the original generated name is used as a key.

Also read property types from customized classes. **NOTE** this only really works for changing between primitive types right now but we'll be able to build up based on this functionality.

It's a breaking change in many ways but I'll fix all the usages when updating the generator.

cc @heaths @tg-msft


test/TestProjects/TypeSchemaMapping/MainClient.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace CustomNamespace
 {
-    [CodeGenClient("")]
+    [CodeGenClient("ServiceClient")]
```
pakrym: This case got way better.

test/TestProjects/TypeSchemaMapping/Customized/SecondModel.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Azure.Core;
+
+namespace TypeSchemaMapping.Models
+{
+    // This model would map without the attribute because the name and namespace is the same.
+    internal partial class SecondModel
```
pakrym: Example of how mapping just works when the full type name matches.

samples/CognitiveServices.TextAnalytics/CognitiveServices.TextAnalytics/InnerError.cs
```diff
@@ -5,10 +5,10 @@
 
 namespace CognitiveServices.TextAnalytics.Models
 {
-    [CodeGenSchema("InnerError")]
+    [CodeGenModel("InnerError")]
     public partial class InnerError
     {
-        [CodeGenSchemaMember("innerError")]
+        [CodeGenMember("innerError")]
```
pakrym: Still case insensitive.

test/TestProjects/TypeSchemaMapping/Customized/SecondModel.cs
tg-msft: How do I work around this when I want to have my own `SecondModel` class that's pretty different than the generated `SecondModelInternal`?  Does putting `[CodeGenModel("SecondModel")] internal class SecondModelInternal { ... }` tell the generator not to merge into my own `public class SecondModel { ... }`?

tg-msft:
I don't know the best way to handle this.  Using the generated type name makes getting started a lot easier.  Using the generated type name makes it a lot harder to map from the swagger.  But you already kind of have to be able to do that anyway if you're not customizing the model...

If it's easy, maybe we could add a debug flag to dump out a simple mapping from swagger names to generated names to help folks with particularly nasty swagger files that use a lot of transforms.

It's also fine to wait until there's a demonstrated problem.

pakrym:
Adding debug information is tracked by https://github.com/Azure/autorest.csharp/issues/226

test/TestProjects/TypeSchemaMapping/Customized/SecondModel.cs
pakrym: > Does putting [CodeGenModel("SecondModel")] internal class SecondModelInternal { ... } tell the generator not to merge into my own public class SecondModel { ... }?

Yes, remapping takes precedence over trying to map to existing types.

src/AutoRest.CSharp.V3/Input/Source/ModelTypeMapping.cs
```diff
@@ -2,31 +2,61 @@
...
         public SourceMemberMapping[] PropertyMappings { get; }
 
-        public ModelTypeMapping(string schemaName, INamedTypeSymbol existingType, SourceMemberMapping[] propertyMappings)
+        public ModelTypeMapping(INamedTypeSymbol memberAttribute, string originalName, INamedTypeSymbol existingType): base(originalName, existingType)
         {
-            SchemaName = schemaName;
-            ExistingType = existingType;
-            PropertyMappings = propertyMappings;
+            List<SourceMemberMapping> memberMappings = new List<SourceMemberMapping>();
+            foreach (var member in GetMembers(existingType))
```
heaths: Nit: not sure how close you're following convention in azure-sdk-for-net, but you have flipped when it's okay and not okay to use `var` here (okay for `new`; not okay otherwise).

src/AutoRest.CSharp.V3/Input/Source/SourceInputModel.cs
```diff
@@ -3,31 +3,120 @@
...
+        private readonly INamedTypeSymbol _schemaMemberNameAttribute;
+
+        public SourceInputModel(Compilation compilation)
+        {
+            _compilation = compilation;
+            _schemaNameAttribute = compilation.GetTypeByMetadataName(typeof(CodeGenModelAttribute).FullName!)!;
+            _schemaMemberNameAttribute = compilation.GetTypeByMetadataName(typeof(CodeGenMemberAttribute).FullName!)!;
+            _clientAttribute = compilation.GetTypeByMetadataName(typeof(CodeGenClientAttribute).FullName!)!;
+
+            var assembly = _compilation.Assembly;
```
heaths: Nit: other instances of improper `var` (to note: I'm a fan of `var` - just towing the line 😉).

src/AutoRest.CSharp.V3/Input/Source/SourceInputModel.cs
```diff
@@ -3,31 +3,120 @@
...
+
+                        if (TryGetName(type, _clientAttribute, out var operationName))
+                        {
+                            definedClients.Add(BuildClient(operationName, namedTypeSymbol));
+                        }
+                    }
+                }
+            }
+
+            DefinedClientTypes = definedClients.ToArray();
```
heaths: Any reason to allocate a new array and copy members? I know the properties are defined as arrays below, but why not just keep them as lists? Less time and allocation. You created the lists, so it's not like there's a shared ownership problem where copying the source list is probably a good idea.

src/AutoRest.CSharp.V3/Input/Source/SourceInputModel.cs
```diff
@@ -3,31 +3,120 @@
...
+
+            return mapping;
         }
 
-        public ClientTypeMapping? FindForOperationGroup(string name)
+        internal static bool TryGetName(ISymbol symbol, INamedTypeSymbol attributeType, [NotNullWhen(true)] out string? name)
         {
-            return DefinedClientTypes.SingleOrDefault(s => string.Compare(s.OperationGroupName, name, StringComparison.InvariantCultureIgnoreCase) == 0);
+            name = null;
+#pragma warning disable RS1024
```
heaths: Nit: could you keep the warning text as a comment?

test/TestProjects/TypeSchemaMapping/Generated/Models/CustomizedModel.cs
```diff
@@ -13,10 +13,10 @@ namespace CustomNamespace
     internal partial class CustomizedModel
     {
         /// <summary> Initializes a new instance of CustomizedModel. </summary>
-        /// <param name="customizedStringProperty"> A description about the set of tags. </param>
+        /// <param name="customizedStringProperty"> . </param>
         /// <param name="customizedFancyField"> Fruit. </param>
         /// <param name="daysOfWeek"> Day of week. </param>
-        internal CustomizedModel(string customizedStringProperty, CustomFruitEnum customizedFancyField, CustomDaysOfWeek daysOfWeek)
+        internal CustomizedModel(int? customizedStringProperty, CustomFruitEnum customizedFancyField, CustomDaysOfWeek daysOfWeek)
```
heaths: Perhaps change the name of the parameter? Very confusing here and above.

src/AutoRest.CSharp.V3/Input/Source/SourceInputModel.cs
pakrym: This is an inefficient pattern that unfortunately widely used through the codebase. I'll clean it up with a single pass in the future.

heaths: It certainly makes sense where we need to "seal" or copy and array, though. I used it intentionally from an incoming `IEnumerable<T>` a few times in KV, but because I didn't own the enumeration and didn't want to assume anything.

Issue #590, opened 2020-03-30
Support keywords in parameter name.


pakrym:
Management network has a property named `As` that results in constructor parameter `as` that fails the build.

PR #593, opened 2020-03-30
Refrest C# keyword list

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/590


src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -98,7 +98,7 @@ private void WriteClientMethod(CodeWriter writer, ClientMethod clientMethod, boo
                 writer.Append($"RestClient.{CreateMethodName(clientMethod.RestClientMethod.Name, async)}(");
                 foreach (var parameter in clientMethod.RestClientMethod.Parameters)
                 {
-                    writer.Append($"{parameter.Name}, ");
+                    writer.Append($"{parameter.Name:I}, ");
```
pakrym: `:I` == `identifier`, so CodeWriter can make sure to format it correctly

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
```diff
@@ -70,6 +70,7 @@ public CodeWriter Append(FormattableString formattableString)
 
             const string literalFormatString = ":L";
             const string declarationFormatString = ":D"; // :D :)
+            const string identifierFormatString = ":I";
```
MiYanni: So, if I think a little deeper about this... right now (in this PR), we're manually telling names (parameters) to add the `@` if they are a keyword via this `:I` formatter. However, we seem to be arriving at a similar state to how we handle types. Meaning, names of things should all flow through a single channel and get handled appropriately.

So, this situation (of a name matching a keyword) would always get checked, as long as name writing is always properly handled in the FormattableString. I can understand `:L` being a separate symbol, but maybe we just have a name class (similar to `Type` class) that handles names appropriately, instead of doing custom formatters for names. This also resolves the issue of writing other names (not just parameters) that might conflict with keywords (such as variables or fields). I know we determine variable names by hand, but it would be nice to have some consistency for name handling across the board.

pakrym: Technically we do have a class like that - `CodeWriterDeclaration`, it even has an advantage of resolving name conflicts. But it's slightly annoying to use in some cases (referencing parameters being one of them). I agree that it would be a good direction to go but I wonder if we can do without making the code too annoying to write.

pakrym:
Merging this as is to unblock partners.

PR #603, opened 2020-03-31
Fix member remover

pakrym:
Run it after all the types were generated because strange things happen if some types are not found.

Fixes: https://github.com/Azure/autorest.csharp/issues/601

### DIFF
--- a/src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
+++ b/src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
@@ -35,11 +35,6 @@ namespace AutoRest.CSharp.V3.AutoRest.Plugins
             var root = document.GetSyntaxRootAsync().Result;
             Debug.Assert(root != null);
 
-            var compilation = document.Project.GetCompilationAsync().GetAwaiter().GetResult();
-            Debug.Assert(compilation != null);
-
-            var rewriter = new MemberRemoverRewriter(compilation.GetSemanticModel(root.SyntaxTree));
-            root = rewriter.Visit(root);
             root = root.WithAdditionalAnnotations(Simplifier.Annotation);
             document = document.WithSyntaxRoot(root);
             _project = document.Project;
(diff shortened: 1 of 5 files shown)


src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -70,6 +65,16 @@ public async IAsyncEnumerable<(string Name, string Text)> GetGeneratedFilesAsync
 
         private static async Task<Document> ProcessDocument(Document document)
         {
+            var compilation = document.Project.GetCompilationAsync().GetAwaiter().GetResult();
```
heaths, line 68: Now that you're in an async method, might as well use `await` for cleaner code.

test/TestProjects/TypeSchemaMapping/Generated/Models/CustomizedModel.cs
```diff
@@ -12,16 +12,5 @@ namespace CustomNamespace
     /// <summary> The Model. </summary>
     internal partial class CustomizedModel
     {
-
-        /// <summary> Initializes a new instance of CustomizedModel. </summary>
-        /// <param name="propertyRenamedAndTypeChanged"> . </param>
-        /// <param name="customizedFancyField"> Fruit. </param>
-        /// <param name="daysOfWeek"> Day of week. </param>
-        internal CustomizedModel(int? propertyRenamedAndTypeChanged, CustomFruitEnum customizedFancyField, CustomDaysOfWeek daysOfWeek)
```
heaths, line 20: I wasn't expecting this to be remove.d Isn't this the serialization ctor you need to deserialize?

pakrym: leftovers of testing

src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
pakrym: Good point.

Issue #366, opened 2019-12-19
Add support for null items in collection deserialization


pakrym:

MiYanni:
Should include dictionary items and nested collection items in dictionaries.

Issue #299, opened 2019-12-06
Implement logic when server does not return a response


MiYanni:
Test: GetComplexBasicNotProvided 
Description: Get a basic complex type while the server doesn't provide a response payload

I'm assuming this means that, when a response isn't provided, it is supposed to still create an empty object... which doesn't sound correct. I believe you'd expect an error, since the definition says it should provide a response:
![image](https://user-images.githubusercontent.com/17788297/70293438-472e3d00-1795-11ea-8a54-fa6fd747e93a.png)

PR #605, opened 2020-04-01
Null array/dictionary item handling

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/366
Resolves: https://github.com/Azure/autorest.csharp/issues/299
Resolves: https://github.com/Azure/autorest.csharp/issues/289

Also fixes tests that return empty response bodies (since we don't support them).


samples/AppConfiguration/AppConfiguration/Generated/Models/Error.Serialization.cs
```diff
@@ -27,7 +27,14 @@ internal static Error DeserializeError(JsonElement element)
                     {
                         continue;
                     }
-                    type = property.Value.GetString();
+                    if (property.Value.ValueKind == JsonValueKind.Null)
```
MiYanni, line 30: This condition is now checked twice, which causes no issues during code execution. However, it would be nice to not have it write it again in this situation. The problem is we have no knowledge that it was handled customly when we go to write the new condition.

If we want to solve this, the best thing I can think of is allowing the new condition to have a flag to change it to use `continue;` in this situation, and no longer write the custom condition above. I haven't looked into how easy that is yet.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -308,45 +309,73 @@ private static void CollectProperties(Dictionary<ObjectTypeProperty, CodeWriterD
...
+                                w => w.Append($"{dictionaryItemVariable}.Value"),
+                                (w, returnValue) => writer.Append($"{dictionaryVariable}.Add({dictionaryItemVariable}.Name, {returnValue});"));
+                        }
 
+                        valueCallback(writer, w => w.Append(dictionaryVariable));
+                    });
                     return;
                 case JsonValueSerialization valueSerialization:
-                    valueCallback(writer, w => w.DeserializeValue(valueSerialization, element));
+                    isNullable ??= valueSerialization.Type.IsNullable || !valueSerialization.Type.IsValueType;
```
MiYanni, line 377: This allows all types that are nullable (meaning non-value types) to have the logic for handling JSON null values.

src/AutoRest.CSharp.V3/Output/Models/Serialization/Json/JsonDictionarySerialization.cs
```diff
@@ -7,13 +7,13 @@ namespace AutoRest.CSharp.V3.Output.Models.Serialization.Json
...
     {
-        public JsonDictionarySerialization(CSharpType? type, JsonSerialization valueSerialization)
+        public JsonDictionarySerialization(CSharpType type, JsonSerialization valueSerialization)
         {
             Type = type;
             ValueSerialization = valueSerialization;
         }
 
-        public CSharpType? Type { get; }
+        public CSharpType Type { get; }
```
MiYanni, line 16: This was a fix because it was no longer being set to null in any situation. I'm assuming this was from a past code cleanup.

test/AutoRest.TestServer.Tests/body-array.cs
```diff
@@ -100,9 +98,11 @@ public class BodyArray : TestServerTestBase
         });
 
         [Test]
-        public Task GetArrayByteWithNull() => Test((host, pipeline) =>
+        public Task GetArrayByteWithNull() => Test(async (host, pipeline) =>
         {
-            Assert.ThrowsAsync(Is.InstanceOf<Exception>(), async () => await new ArrayClient(ClientDiagnostics, pipeline, host).GetByteInvalidNullAsync());
+            var result = await new ArrayClient(ClientDiagnostics, pipeline, host).GetByteInvalidNullAsync();
```
pakrym, line 103: Test says invalid but we don't throw.

test/AutoRest.TestServer.Tests/body-dictionary.cs
```diff
@@ -498,11 +448,10 @@ public class BodyDictionary : TestServerTestBase
...
 
         [Test]
-        [Ignore("Null responses: https://github.com/Azure/autorest.csharp/issues/289")]
-        public Task GetDictionaryNull() => Test(async (host, pipeline) =>
+        public Task GetDictionaryNull() => Test((host, pipeline) =>
         {
-            var result = await new DictionaryClient(ClientDiagnostics, pipeline, host).GetNullAsync();
-            Assert.IsNull(result.Value);
+            // Empty response body
+            Assert.ThrowsAsync(Is.InstanceOf<JsonException>(), async () => await new DictionaryClient(ClientDiagnostics, pipeline, host).GetNullAsync());
```
pakrym, line 454: Is null returned here or an empty response?

test/AutoRest.TestServer.Tests/body-array.cs
MiYanni: It is valid for us since `byte[]` is able to be null. Don't trust the test server names as gospel.

test/AutoRest.TestServer.Tests/body-dictionary.cs
MiYanni: Empty response. No test server tests return `null`. All the test server test names that either say `NotProvided` or `Null` return an empty body. They always look like this:
```js
res.status(200).end();
```

samples/AppConfiguration/AppConfiguration/Generated/Models/Error.Serialization.cs
MiYanni: Fixed it. Instead of sending the nullability state to the property serialization writing, I always set it `false` so the null check isn't written.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -308,45 +309,73 @@ private static void CollectProperties(Dictionary<ObjectTypeProperty, CodeWriterD
             }
         }
 
-        private static void DeserializeIntoVariable(this CodeWriter writer, JsonSerialization serialization, Action<CodeWriter, CodeWriterDelegate> valueCallback, CodeWriterDelegate element)
+        private static void DeserializeIntoVariable(this CodeWriter writer, JsonSerialization serialization, Action<CodeWriter, CodeWriterDelegate> valueCallback,
+            CodeWriterDelegate element, bool? writeNullHandling = null)
         {
+            void WriteNullableScope(Action writeContent)
```
pakrym, line 315: It seems that every block in this method calls WriteNullableScope might be cleaner to add another methd like DeserializeIntoVaribleWithNullScope that would write the check and call into DeserializeIntoVarible.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -308,45 +309,73 @@ private static void CollectProperties(Dictionary<ObjectTypeProperty, CodeWriterD
...
             switch (serialization)
             {
                 case JsonArraySerialization array:
-                    var arrayVariable = new CodeWriterDeclaration("array");
-                    writer.Line($"{array.Type} {arrayVariable:D} = new {array.Type}();");
-
-                    var collectionItemVariable = new CodeWriterDeclaration("item");
-                    writer.Line($"foreach (var {collectionItemVariable:D} in {element}.EnumerateArray())");
-                    using (writer.Scope())
+                    writeNullHandling ??= true;
```
pakrym, line 337: Is it correct to inherit this value for inner serializations?

MiYanni: By 'inherit', you mean 'use the value provided by the parameter', then yes. This is how it works for property serializers, where they provide `false` to this method, and it acknowledges it appropriately. Otherwise, if a value is not provided (null), these act like non-value types, thus `true` to writing the null handler.

MiYanni: I don't believe so as inner serializations will have their own calls to this method that would have their own value passed in for this. I originally thought I had to do this, but our `property bag` recursive logic seems to handle this correctly. Meaning, inner properties will be written flat in the main property loop, so there isn't a 'depth' to keeping track of this value.

pakrym: What about the List<List<int>> case?

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
MiYanni: Done. However, personally, I don't think it is cleaner because it increases cyclomatic complexity since you have to check the serialization type twice now.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -309,73 +309,73 @@ private static void CollectProperties(Dictionary<ObjectTypeProperty, CodeWriterD
...
-        private static void DeserializeIntoVariable(this CodeWriter writer, JsonSerialization serialization, Action<CodeWriter, CodeWriterDelegate> valueCallback,
-            CodeWriterDelegate element, bool? writeNullHandling = null)
+        private static void DeserializeIntoVariableWithNullHandling(this CodeWriter writer, JsonSerialization serialization, Action<CodeWriter, CodeWriterDelegate> valueCallback,
+            CodeWriterDelegate element, bool? writeNullHandling)
         {
-            void WriteNullableScope(Action writeContent)
+            if (serialization is JsonArraySerialization || serialization is JsonDictionarySerialization)
             {
-                if (writeNullHandling == true)
+                writeNullHandling ??= true;
```
pakrym, line 317: Doesn't type based logic work here too? JsonArraySerialization.Type would be a reference type.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -227,10 +227,11 @@ void WriteNullCheck()
                             if (property.Property != null)
                             {
                                 // Reading a property value
-                                writer.DeserializeIntoVariable(
+                                writer.DeserializeIntoVariableWithNullHandling(
```
pakrym, line 230: You don't need to call DeserializeIntoVariableWithNullHandling here if you are passing false anyway.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
MiYanni: So, the inner `List<int>` isn't a property of the outer type, so the `continue;` null check isn't written to begin with. Since it is a non-value type, it should write the new null check. Similar example here in `List<Fish>` from Sawshark.Serializartion.cs in body-complex.

image

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
MiYanni: 👀 Right!

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -308,6 +309,35 @@ private static void CollectProperties(Dictionary<ObjectTypeProperty, CodeWriterD
             }
         }
 
+        private static void DeserializeIntoVariableWithNullHandling(this CodeWriter writer, JsonSerialization serialization, Action<CodeWriter, CodeWriterDelegate> valueCallback,
+            CodeWriterDelegate element, bool? writeNullHandling)
```
pakrym, line 313: I don't think we need this parameter. Outside property branch no one passes a value here and the property branch can call directly into DeserializeValue

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
MiYanni: Unfortunately, you still need a mechanism to do the type check since you need the actual type to get the `Type` property. `JsonSerialization` base type has no `Type` property.

MiYanni: Done.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
MiYanni: Done.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
MiYanni: Done.

Issue #608, opened 2020-04-01
Generated code is incorrect when multiple clients have the same operations with different set of response headers


pakrym:
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/batch/data-plane/Microsoft.Batch/stable/2020-03-01.11.0/BatchService.json
```

PR #610, opened 2020-04-01
Prefix header types with a client name to avoid conflicts

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/608


pakrym:
To avoid type name conflicts, having the same type name across namespaces is still not great.

pakrym:
Headers are not public types so I don't think their names matter too much and I don't expect devs to ever rename them.

PR #611, opened 2020-04-02
Resolve namespace-identifier conflicts

pakrym:
Fixes https://github.com/Azure/autorest.csharp/issues/609


src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
```diff
@@ -190,6 +190,7 @@ private void AppendType(CSharpType type)
             {
                 UseNamespace(type.Namespace);
 
+                AppendRaw("global::");
```
MiYanni, line 193: Nice. We ended up doing the same thing in the low-level C# generator for Azure PowerShell.

PR #612, opened 2020-04-02
Add support for text bodies

pakrym:
And a signalr sample


src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -166,6 +166,14 @@ private void WriteRequestCreation(CodeWriter writer, RestClientMethod operation)
                             writer.Line($");");
                         }
                         break;
+                    case TextRequestBody binaryBody:
```
MiYanni, line 169: Variable name should be `textBody`.

Issue #599, opened 2020-03-31
'CorsRules': member names cannot be the same as their enclosing type


erich-wang:
**Pavel provided the workaround**: The new generator allows you to rename types/properties change namespaces and accessibility by mapping generated types to a customization types. CodeGenSchema CodeGenSchemaMember attributes in the example above are used for that. Having a property with CodeGenSchemaMember on it instructs generator to use this name instead of the default it picks and resolves the build error

    
**[Erich]**: with previous c# generator, the generated source code could be built successfully without any customization, we prefer to follow same pattern instead of introducing extra customization (smile).

**[Pavel]**: Do you have ideas about how do we make sure someone fixes the name after SDK was generated?

**[Erich]**: 
    IMO, the name customization is nice to have instead of must to have. If it is must to have, I suggest to improve C# generator to avoid the customization. I'm not sure whether python/java generators have same issue

allenjzhang:
@Pavel, while the full customization should be there to give override options, we should still generate out of box that works without requiring manual touch. I suggest we use **Value** suffix to the same name property.  So CorsRules will have CorsRulesValue { get; set}. It is less than Ideal name, but should be more acceptable that Value { get; set;}.

PR #615, opened 2020-04-02
Resolve property-type name conflicts

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/599


src/AutoRest.CSharp.V3/Output/Models/Types/ObjectType.cs
```diff
@@ -333,6 +333,12 @@ private IEnumerable<ObjectTypeProperty> BuildProperties()
             foreach (Property property in _objectSchema.Properties!)
             {
                 var name = property.CSharpName();
+                // Resolve property-type name conflicts
+                // TODO: we might eventually need a more generic version of this logic
+                if (name == Type.Name)
+                {
+                    name += "Value";
```
MiYanni, line 340: This works for 99% of the cases, but keep in mind that if, somehow, there was a `[TypeName]` property and a `[TypeName]Value` property already existing, this would cause a different kind of conflict. I've seen weird stuff like that situation from Network before. So, don't be surprised if it happens.

pakrym:
> This works for 99% of the cases, but keep in mind that if, somehow, there was a [TypeName] property and a [TypeName]Value property already existing, this would cause a different kind of conflict. I've seen weird stuff like that situation from Network before. So, don't be surprised if it happens.

🤦‍♂

PR #627, opened 2020-04-03
Add smoke test

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/628


pakrym:
> Similar to doing our test server tests, why not have a list (that we can comment/uncomment) of smoke tests, so the ones that are failing we can handle in future PRs? Then, we can get the set of working smoke tests in right now, and work on the failing ones over time.

https://github.com/Azure/autorest.csharp/pull/627/files#diff-ec83444674a20e3adc31341388a9ab6b is that list

eng/Generation.psm1
```diff
@@ -0,0 +1,63 @@
+$repoRoot = Resolve-Path (Join-Path $PSScriptRoot '..')
+$autorestBinary = Join-Path $repoRoot 'node_modules' '.bin' 'autorest-beta'
+$AutorestPluginProject = Resolve-Path (Join-Path $repoRoot 'src' 'AutoRest.CSharp.V3')
+
+function Invoke($command)
+{
+    Write-Host "> $command"
+    pushd $repoRoot
+    if ($IsLinux)
```
MiYanni, line 9: If we're adding Linux support, we have to make sure we use `Join-Path`/`Resolve-Path` in all places when accessing/writing files, as those cmdlets will use Environment.PathSeparator. Additionally, Linux file/folder names are case sensitive, so we have to make sure that we have those path elements cased exactly what they are on disk.

eng/Generate.ps1
```diff
@@ -179,14 +176,8 @@ if (![string]::IsNullOrWhiteSpace($name))
...
-    $definition = $swaggerDefinitions[$key];
-    Invoke-AutoRest $definition.output $definition.title $definition.arguments
-    $projectPath = Join-Path $definition.output $definition.title;
-    if (!$noProjectBuild)
-    {
-        Invoke "dotnet build $projectPath --verbosity quiet /nologo"
-    }
-}
+$keys | %{ $swaggerDefinitions[$_] } | ForEach-Object -Parallel {
+    Import-Module "$using:PSScriptRoot\Generation.psm1" -DisableNameChecking;
```
MiYanni, line 180: Cannot use `\` for paths for Linux support. Need to use Join-Path.

eng/Generate.ps1
```diff
@@ -1,53 +1,19 @@
...
-    {
-        $codeModel = Join-Path $baseOutput $title "CodeModel.yaml"
-        $command = "dotnet run --project $script:autorestPluginProject --no-build -- --plugin=csharpgen --title=$title --namespace=$namespace --standalone --input-file=$codeModel --output-folder=$outputPath --shared-source-folder=$script:sharedSource --save-code-model=true"
-    }
-
-    if ($clean)
-    {
-        Get-ChildItem $outputPath -Filter Generated -Directory -Recurse | Get-ChildItem -File -Recurse | Remove-Item -Force
-    }
+Import-Module "$PSScriptRoot\Generation.psm1" -DisableNameChecking -Force;
```
MiYanni, line 4: Cannot use `\` for paths for Linux support. Need to use Join-Path.

eng/pipelines/build.yml
```diff
@@ -47,3 +47,26 @@ jobs:
...
+        inputs:
+          version: "$(DotNetCoreSDKVersion)"
+      - task: NodeTool@0
+        displayName: "Install Node 13.x"
+        inputs:
+          versionSpec: '13.x'
+      - script: |
+          npm ci
+        displayName: "Install packages"
+      - pwsh: .\eng\Generate.ps1 -Exclude TestServer,TestProjects,Samples -reset -noDebug
```
MiYanni, line 67: I'd recommend using forward slashes for all file paths in pipeline definitions, as forward slash works on both Windows and Linux.

pakrym: It seems powershell normalized it just fine.

eng/Generation.psm1
pakrym: It's not required, the pipeline runs just fine on linux and PS is normalizing the paths, I switched back to windows because of https://github.com/Azure/autorest/issues/3474

PR #635, opened 2020-04-05
Resolve more name conflicts

pakrym:
ToString,
Equals,
pipeline,
clientDiagnostics


src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -130,8 +130,8 @@ private void WriteClientMethod(CodeWriter writer, ClientMethod clientMethod, boo
 
         private void WriteClientFields(CodeWriter writer, Client client)
         {
-            writer.Line($"private readonly {typeof(ClientDiagnostics)} clientDiagnostics;");
-            writer.Line($"private readonly {typeof(HttpPipeline)} pipeline;");
+            writer.Line($"private readonly {typeof(ClientDiagnostics)} _clientDiagnostics;");
```
MiYanni, line 133: Maybe it makes sense to have const strings for these client field names? Then, if we ever need to change them or anything, we only need to update in 1 place.

pakrym: If I have to rename them one more time I'm making consts :)

PR #636, opened 2020-04-06
Support multiple inheritance

pakrym:
The way it works is by picking one type to inherit from and inlining all other types.

If one of the base types has discriminator it's preferred, otherwise the first type in the list is selected.

Base type can be overriden by defining in in the customized model.


src/AutoRest.CSharp.V3/Output/Models/Types/ObjectType.cs
```diff
@@ -365,19 +378,53 @@ private IEnumerable<ObjectTypeProperty> BuildProperties()
...
+
+            foreach (var objectSchema in objectSchemas)
+            {
+                // Take first schema or the one with discriminator
+                if (selectedSchema == null || objectSchema.Discriminator != null)
                 {
-                    CSharpType type = _typeFactory.CreateType(parentObjectSchema, false);
-                    Debug.Assert(!type.IsFrameworkType);
-                    return type;
+                    selectedSchema = objectSchema;
```
MiYanni, line 417: Really interesting way to write this. However, I don't think it works as it would always select the last `objectSchema` with a discriminator. You'd likely want the first one.

src/AutoRest.CSharp.V3/readme.md
```diff
@@ -4,7 +4,7 @@
 ## Configuration
 ```yaml
 use-extension:
-  "@autorest/modelerfour": "4.12.276"
+  "@autorest/modelerfour": "http://tinyurl.com/tkk28fx"
```
MiYanni, line 7: What.... is that? Should use an aka.ms link and have access shared with the team, so it can be managed/updated.

pakrym: Temporary, trying out a fix for a workaround.https://github.com/Azure/autorest.modelerfour/pull/259

pakrym:
> How does serialization work in this PR?

Same way as usual, our serialization code always read all the properties and doesn't really care about the inheritance as long as there is an appropriate ctor to initialize the object.

> I'd be interested in trying this on a management plane service that has multiple inheritance. I'll have to look for one as I don't remember one right this second.

I had one I was trying it on but the fix in https://github.com/Azure/autorest.modelerfour/pull/259 treats it like an error now :)

src/AutoRest.CSharp.V3/Output/Models/Types/ObjectType.cs
pakrym: You are right. Also this loop is kinda strange

MiYanni: Much better. 👍

src/AutoRest.CSharp.V3/readme.md
MiYanni: Just make sure to not merge with a temporary version. Good for testing, but wait until it is a real package to merge into our repo.

PR #641, opened 2020-04-07
Couple naming fixes

pakrym:
Fixes situation where we reuse the name violating C# scoping rules:
This is invalid:

```
{
   var a;
}

var a;
```
While this is valid:
```
{
   var a;
}

{
   var a;
}
```

We handle both now. Also fixes signing (to enable testing of CodeWriter) and removes `CaseExtensions` dependency (while improving how we handle names like "odata.type")


test/TestServerProjects/xml-service/Generated/Models/Blobs.Serialization.cs
```diff
@@ -24,9 +24,9 @@ internal static Blobs DeserializeBlobs(XElement element)
             }
             blobPrefix = array;
             var array0 = new List<Blob>();
-            foreach (var e0 in element.Elements("Blob"))
+            foreach (var e in element.Elements("Blob"))
```
pakrym, line 27: This got better!

test/AutoRest.TestServer.Tests/CodeWriterTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            using (codeWriter.Scope())
+            {
+                codeWriter.Line($"{cwd1:D}");
+            }
+
+            codeWriter.Line($"{cwd2:D}");
+
+            Assert.AreEqual(
+                @"{
+a
```
pakrym, line 69: This is the situation being fixed.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -199,8 +197,7 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
                     writer.Line($"{objAdditionalProperties.Type} {dictionaryVariable:D} = new {objAdditionalProperties.Type}();");
                 }
 
-                writer.Line($"foreach (var {itemVariable:D} in {element}.EnumerateObject())");
-                using (writer.Scope())
+                using (writer.Scope($"foreach (var {itemVariable:D} in {element}.EnumerateObject())"))
```
pakrym, line 200: This change makes so `itemVariable` is defined inside foreach scope and results in better names.

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
```diff
@@ -392,14 +414,16 @@ public override string ToString()
...
                     .ToArray();
-
-            builder.AppendLine("// Copyright (c) Microsoft Corporation. All rights reserved.");
-            builder.AppendLine("// Licensed under the MIT License.");
-            builder.AppendLine();
-            builder.AppendLine("// <auto-generated/>");
-            builder.AppendLine();
-            builder.AppendLine("#nullable disable");
-            builder.AppendLine();
+            if (header)
```
pakrym, line 417: Helps testing.

Directory.Build.props
```diff
@@ -36,7 +36,9 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)AzureSDKClient.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
```
pakrym, line 39: Required for InternalsVisibleTo

MiYanni: That shouldn't be the case, AFAIK. InternalsVisibleTo only requires a key (second parameter) if the project is signed. All of our projects in Azure PowerShell had blocks like this in the AssemblyInfo:
https://github.com/Azure/azure-powershell/blob/6f2df0317fb7f7170a8effea7a897ef09f00259e/src/PrivateDns/PrivateDns/Properties/AssemblyInfo.cs#L49-L53

So, unless we're signing now (which is fine if we are), we should likely have a build that does signing (Release) and a build that doesn't (Debug). But, signing isn't required for InternalsVisibleTo.

src/AutoRest.CSharp.V3/Generation/Writers/XmlCodeWriterExtensions.cs
```diff
@@ -202,7 +202,8 @@ private static void ToDeserializeCall(this CodeWriter writer, XmlElementSerializ
             }
 
             var elementVariable = new CodeWriterDeclaration(serialization.Name.ToVariableName() + "Element");
-            using (writer.Scope($"if ({element}.Element({serialization.Name:L}) is {typeof(XElement)} {elementVariable:D})"))
+            writer.Line($"if ({element}.Element({serialization.Name:L}) is {typeof(XElement)} {elementVariable:D})");
+            using (writer.Scope())
```
MiYanni, line 206: Why did this remove the using for the if-statement scope? It is the opposite change from the JSonCodeWriterExtensions.

src/AutoRest.CSharp.V3/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Runtime.CompilerServices;
+
+[assembly:InternalsVisibleTo("AutoRest.TestServer.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100d15ddcb29688295338af4b7686603fe614abd555e09efba8fb88ee09e1f7b1ccaeed2e8f823fa9eef3fdd60217fc012ea67d2479751a0b8c087a4185541b851bd8b16f8d91b840e51b1cb0ba6fe647997e57429265e85ef62d565db50a69ae1647d54d7bd855e4db3d8a91510e5bcbd0edfbbecaa20a7bd9ae74593daa7b11b4")]
```
MiYanni, line 6: As mentioned, this seems really odd to sign the DLL in every build. You'd likely not need a signed DLL for the test project, as you'd want to use a debug build for the test project (for debugging purposes).

src/AutoRest.CSharp.V3/Utilities/StringExtensions.cs
```diff
@@ -17,14 +15,54 @@ internal static class StringExtensions
         public static bool IsNullOrEmpty(this string? text) => String.IsNullOrEmpty(text);
         public static bool IsNullOrWhiteSpace(this string? text) => String.IsNullOrWhiteSpace(text);
 
-        [return: NotNullIfNotNull("text")]
-        private static string? RemoveNonWordCharacters(this string? text) => !text.IsNullOrEmpty() ? Regex.Replace(text, @"\W+", String.Empty) : text;
-        [return: NotNullIfNotNull("text")]
-        private static string? PrependUnderscoreIfNumbers(this string? text) => Regex.IsMatch(text ?? String.Empty, @"^\d") ? $"_{text}" : text;
         [return: NotNullIfNotNull("name")]
-        public static string ToCleanName(this string name) => name.ToPascalCase().RemoveNonWordCharacters().PrependUnderscoreIfNumbers();
+        public static string ToCleanName(this string name, bool camelCase = true)
```
MiYanni, line 19: So, in what cases are we cleaning names? Because, we should be getting 'good' names from AutoRest, right? So, I'm trying to understand why we still have custom name cleaning. We can use the `naming:` node in our config:
https://github.com/Azure/autorest.modelerfour/tree/master/modelerfour#modelfour-options

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
MiYanni: Right now, the header is written here because we assume ToString on a CodeWriter produces a file. How about we have something that represents a file that contains a CodeWriter? Then, we have this block be part of a file's starting text, and append the CodeWriter output to that. It is better encapsulation of concepts. Then, your tests don't need to call ToString differently.

test/TestServerProjects/xml-service/Generated/Models/Blobs.Serialization.cs
MiYanni: Is there a reason it is `array0` here? Do we need to apply the same fix for arrays?

src/AutoRest.CSharp.V3/Properties/AssemblyInfo.cs
pakrym: Signing a DLL on every build is a common practice, and you need a signed test dll for InternalsVisibleTo to work.

src/AutoRest.CSharp.V3/Generation/Writers/XmlCodeWriterExtensions.cs
pakrym: Because the variable inside the `if` is defined in the parent scope while the loop variable in `foreach` is defined in inner scope.

src/AutoRest.CSharp.V3/Properties/AssemblyInfo.cs
pakrym: I'm wrong. If both are unsigned it would still work.

Directory.Build.props
pakrym: Yeah, you are right, if both are unsigned InternalsVisibleTo would still work. I don't think there is any benefit to making signing conditional though.

src/AutoRest.CSharp.V3/Utilities/StringExtensions.cs
pakrym: Response headers is the main one.

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
pakrym: Yeah, I want to invert the logic a bit and have this handled in GeneratedCodeWorkspace as part of the formatting/cleanup.

test/TestServerProjects/xml-service/Generated/Models/Blobs.Serialization.cs
pakrym: It correct, two arrays are defined in the same scope.

It's mostly the result of the way XML deserialization is done, we can improve names by prefixing them with serialized name so we get something like `blobPrefixArray` but that's a different change :)

src/AutoRest.CSharp.V3/Utilities/StringExtensions.cs
pakrym: Filed https://github.com/Azure/autorest.modelerfour/issues/262

src/AutoRest.CSharp.V3/Generation/Writers/CodeWriter.cs
MiYanni: Sounds good. Can you create an issue so it doesn't get lost?

pakrym: https://github.com/Azure/autorest.csharp/issues/642

PR #643, opened 2020-04-07
Required-optional tests

MiYanni:
Also a bug fix for array request header parameters. Uses same mechanics for handling serializationStyle as query parameters.


test/TestServerProjects/required-optional/Generated/Operations/ExplicitRestClient.cs
```diff
@@ -0,0 +1,1649 @@
...
+        internal HttpMessage CreatePostRequiredArrayHeaderRequest(IEnumerable<string> headerParameter)
+        {
+            var message = _pipeline.CreateMessage();
+            var request = message.Request;
+            request.Method = RequestMethod.Post;
+            var uri = new RawRequestUriBuilder();
+            uri.AppendRaw(host, false);
+            uri.AppendPath("/reqopt/requied/array/header", false);
+            request.Uri = uri;
+            request.Headers.AddDelimited("headerParameter", headerParameter, ",");
```
MiYanni, line 1515: Fix appears here.

test/TestServerProjects/required-optional/Generated/Operations/ExplicitRestClient.cs
```diff
@@ -0,0 +1,1649 @@
...
+            var message = _pipeline.CreateMessage();
+            var request = message.Request;
+            request.Method = RequestMethod.Post;
+            var uri = new RawRequestUriBuilder();
+            uri.AppendRaw(host, false);
+            uri.AppendPath("/reqopt/optional/array/header", false);
+            request.Uri = uri;
+            if (headerParameter != null)
+            {
+                request.Headers.AddDelimited("headerParameter", headerParameter, ",");
```
MiYanni, line 1592: And here.

src/AutoRest.CSharp.V3/Output/Models/Requests/RequestSerializationStyle.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace AutoRest.CSharp.V3.Output.Models.Requests
 {
-    internal enum QuerySerializationStyle
+    internal enum RequestSerializationStyle
```
pakrym, line 6:
```suggestion
    internal enum RequestParameterSerializationStyle
```
??

test/TestServerProjects/required-optional/Generated/Operations/ExplicitRestClient.cs
```diff
@@ -0,0 +1,1649 @@
...
+            {
+                scope.Failed(e);
+                throw;
+            }
+        }
+
+        /// <summary> Test explicitly required string. Please put null and the client library should throw before the request is sent. </summary>
+        /// <param name="bodyParameter"> The String to use. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public Response PostRequiredStringParameter(string bodyParameter, CancellationToken cancellationToken = default)
```
pakrym, line 509: We should also add tests for scenarios that don't require service call.

src/AutoRest.CSharp.V3/Output/Models/Requests/RequestSerializationStyle.cs
MiYanni: Done.

test/TestServerProjects/required-optional/Generated/Operations/ExplicitRestClient.cs
MiYanni: Added checks to the tests that look at both the method parameter's signature and the ImplicitClient's constructor signature.

pakrym: What about the throwing part in `put null and the client library should throw before the request is sent`?

MiYanni:
The time PR and the csproj PR caused an issue in the repo where `body-time.csproj` was the wrong name. It should have been `body_time.csproj`. I just fixed it so builds will succeed again.

test/TestServerProjects/required-optional/Generated/Operations/ExplicitRestClient.cs
MiYanni: Took me a minute to figure this out. That call doesn't have a test associated with it. There are no PostRequired tests on the test server. PutRequired is there (and returns 400 appropriately) but there isn't a test or route for PostRequired.

Edit: Oh, I get it. these definitions are supposed to be there, but there isn't a test because, obviously, it isn't supposed to ever reach the test server. You're just supposed to make tests and not have any test server feedback. That's bizarre. You'd think to make sure the test is covered, you'd have at least an endpoint to signify you ran the test. Now that I think about it, there are likely other endpoints in the test server that expect you to do this, which is completely non-obvious.

MiYanni: Done. `Required` tests have been added.

MiYanni:
CI passed: https://dev.azure.com/azure-sdk/public/_build/results?buildId=356287&view=results
Seems the GitHub hook for getting the status didn't work, though. Merging...

PR #644, opened 2020-04-07
Add public management ctor generation

pakrym:
Also changes how standalone mode works - instead of passing all parameters via command line the hosted run serializes configuration into a file that standalone process later loads. This way we can use README.md to set options (title, azure-arm) and standalone invocation would see them.


src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -58,6 +67,56 @@ public void WriteClient(CodeWriter writer, Client client)
             }
         }
 
+        private void WriteManagementClientCtors(CodeWriter writer, Client client, Configuration configuration)
+        {
```
pakrym, line 71: This is pretty hardcoded but should work for the first iteration, will improve as we gather more feedback.

samples/Azure.Storage.Management/readme.md
```diff
@@ -4,7 +4,8 @@
 > see https://aka.ms/autorest
 
 ``` yaml
-title: Azure.Storage.Management
+title: Storage
```
MiYanni, line 7: So... are we just using folder names to separate things between management and data plane?

MiYanni, line 7: So... are we just using folder names to separate things between management and data plane?

src/AutoRest.CSharp.V3/AutoRest/Communication/IPluginCommunication.cs
```diff
@@ -3,15 +3,15 @@
...
+using AutoRest.CSharp.V3.AutoRest.Plugins;
 
 namespace AutoRest.CSharp.V3.AutoRest.Communication
 {
     internal interface IPluginCommunication
     {
         string PluginName { get; }
-        Task<string> ReadFile(string filename);
-        Task<T> GetValue<T>(string key);
-        Task<string[]> ListInputs(string? artifactType = null);
```
MiYanni, line 14: So, these were AutoRest communication concepts (meaning, the JsonRPC API for AutoRest). So, changing them to non-AutoRest related methods would make us not have plugin communication (multiple plugins) in the future if we wanted/needed it.

The interface here was only so other plugin communcation methods (such as our Host communication) could communicate in the same way any other AutoRest plugin would.

src/AutoRest.CSharp.V3/AutoRest/Communication/IPluginCommunication.cs
```diff
@@ -3,15 +3,15 @@
...
+        Configuration Configuration { get; }
+        Task<string> GetCodeModel();
```
MiYanni, line 14: This is very specific to the plugin pipeline, and not related to plugin communication.

src/AutoRest.CSharp.V3/AutoRest/Plugins/ClientOptionsWriter.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Linq;
+using AutoRest.CSharp.V3.Generation.Writers;
+using AutoRest.CSharp.V3.Output.Models.Types;
+using Azure.Core;
+
+namespace AutoRest.CSharp.V3.AutoRest.Plugins
+{
+    internal class ClientOptionsWriter
+    {
+        public static void WriteClientOptions(CodeWriter writer, BuildContext context)
```
MiYanni, line 14: What is this for?

src/AutoRest.CSharp.V3/AutoRest/Communication/IPluginCommunication.cs
pakrym: Yeah, you are right, I'll redo this change without messing up this abstraciton.

src/AutoRest.CSharp.V3/AutoRest/Plugins/ClientOptionsWriter.cs
pakrym: Generates ***ClientOptions classes that are required for public ctors. https://github.com/Azure/autorest.csharp/pull/644/files#diff-1facde7a0e058c6acf8cc086f0635dc8

MiYanni: Got it. All clients we generate use the same **ClientOptions, so it makes sense to generate it with the name of the library, and not the name of the client itself. For example, it isn't UsagesClientOptions, but instead StorageManagementClientOptions for any client in StorageManagement.

eng/Generate.ps1
```diff
@@ -141,11 +140,11 @@ if ($updateLaunchSettings)
         $definition = $swaggerDefinitions[$key];
         $outputPath = (Join-Path $definition.output $key).Replace($repoRoot, '$(SolutionDir)')
         $codeModel = Join-Path $outputPath 'CodeModel.yaml'
-        $namespace = $definition.title.Replace('-', '_')
+        $namespace = $definition.projectName.Replace('-', '_')
 
         $settings.profiles[$key] = [ordered]@{
             'commandName'='Project';
-            'commandLineArgs'="--standalone --input-codemodel=$codeModel --plugin=csharpgen --output-folder=$outputPath --namespace=$namespace --shared-source-folder=$sharedSourceNormalized --save-code-model=true"
+            'commandLineArgs'="--standalone $outputPath"
```
MiYanni, line 147: Looking at this change here finally made me realize why this PR seems odd. So, the command line argument for AutoRest are simply AutoRest configuration values. Meaning, `standalone`, `input-codemodel`, `plugin`, `output-folder`, `namespace`, `shared-source-folder`, and `save-code-model` are all values that can be put into a `readme.md`.

This PR invents another configuration file, `Configuration.json`, that serves the same purpose of the configuration in an AutoRest `readme.md`. All the values in this new `Configuration.json` are standard AutoRest values other than `SharedSourceFolder`. But, you can simply make a value `shared-source-folder` in the `readme.md` and read it in our plugin.

So, while I think it is a good idea for each sample to have it's own configuration, we should just use `readme.md` for that purpose. That's the point of AutoRest supporting configuration in this way.

src/AutoRest.CSharp.V3/AutoRest/Plugins/Configuration.cs
```diff
@@ -2,30 +2,32 @@
...
         {
-            OutputFolder = new Uri(GetRequiredOption(autoRest, "output-folder")).LocalPath;
-            SharedSourceFolder = new Uri(GetRequiredOption(autoRest, "shared-source-folder")).LocalPath;
-            Namespace = autoRest.GetValue<string?>("namespace").GetAwaiter().GetResult() ?? "Sample";
-            Title = autoRest.GetValue<string?>("title").GetAwaiter().GetResult() ?? "Sample";
-            SaveCodeModel = autoRest.GetValue<bool?>("save-code-model").GetAwaiter().GetResult() ?? false;
+            OutputFolder = outputFolder;
+            Namespace = ns;
+            var namespaceParts = ns.Split('.');
+            LibraryName = name ?? namespaceParts.Last();
```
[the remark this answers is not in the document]
pakrym: Chatted offline, the namespace would have to be picked manually at the time of the project creation and is the better source of a good name.

src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -28,9 +28,10 @@ internal class CSharpProj : IPlugin
 </Project>
 
 ";
-        public async Task<bool> Execute(IPluginCommunication autoRest, CodeModel codeModel, Configuration configuration)
+        public async Task<bool> Execute(IPluginCommunication autoRest)
         {
-                await autoRest.WriteFile($"{configuration.Title}.csproj", _csProjContent, "source-file-csharp");
+            var ns = await autoRest.GetValue<string>("namespace");
+            await autoRest.WriteFile($"{ns}.csproj", _csProjContent, "source-file-csharp");
```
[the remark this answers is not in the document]
pakrym: Chatted offline. For track 2 projects the root namespace, project name and package name are all need to be the same so namespace becomes a pretty good indication of what the project name should be.

PR #645, opened 2020-04-08
Remove enum workaround

pakrym:


pakrym:
Merging to unblock other tasks.

PR #647, opened 2020-04-09
Add time support

pakrym:


src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -485,6 +485,7 @@ public static void DeserializeImplementation(this CodeWriter writer, ITypeProvid
             SerializationFormat.DateTime_Unix => "U",
             SerializationFormat.Bytes_Base64Url => "U",
             SerializationFormat.Duration_ISO8601 => "P",
+            SerializationFormat.Time_ISO8601 => "T",
```
heaths, line 488: According to https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times, this format - whether passed to `DateTimeOffset.ToString` or `TimeSpan.ToString` (I haven't gotten that far yet) is invalid. That said, according to https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.1 date-time, date, and time all use ISO8601. Have you looked at response samples (not JSON examples; not always trustworthy) to see what the service actually sends back? Based on https://en.wikipedia.org/wiki/ISO_8601#Times, these are valid time zone identifiers (which do come up in full ISO8601), but that's only for `DateTimeOffset` (and `DateTime`, which we don't use). So you couldn't use a `TimeSpan`, which actually represents relative time anyway. I think you should still probably use a `DateTimeOffset` above.

src/AutoRest.CSharp.V3/Generation/Types/TypeFactory.cs
```diff
@@ -99,6 +99,7 @@ private static bool IsList(CSharpType type)
             AllSchemaTypes.Duration => typeof(TimeSpan),
             AllSchemaTypes.OdataQuery => typeof(string),
             AllSchemaTypes.String => typeof(string),
+            AllSchemaTypes.Time => typeof(TimeSpan),
```
heaths, line 102: This is absolute time, and should be a `DateTimeOffset`. See more thorough explanation below.

pakrym: It's not absolute, it's relative to the beginning of some day.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
pakrym: json-schema and openapi/swagger is not the same thing and the later is what we support.

But even than as far as I understand `"type":"time"` is azure specific extension that's not defined by OAPI.

> Have you looked at response samples (not JSON examples; not always trustworthy) to see what the service actually sends back?

This test was added to the test server based on the form recognizer service response.

> So you couldn't use a TimeSpan

I picked TimeSpan because it's what DateTime.TimeOfDay returns which coresponds to what we try to express with `type:time`

MiYanni: To add my 2 cents, I feel the purpose of TimeSpan is to be a duration. Why? Because to convert TimeSpan into a human-readable 'time of day', you have to do extra gymnastics to get there: https://stackoverflow.com/questions/13044603/convert-time-span-value-to-format-hhmm-am-pm-using-c-sharp

If you read through different answers here, you'd see that TimeSpan isn't really meant to be 'time of day'. It doesn't have a way to say AM versus PM or have different time formats naturally (aka, has no clock format, 12 or 24 hour). From my understanding of the use case for FormRecognizer, it is literally the time written on a receipt. I also think DateTime.TimeOfDay is misleading, as the description of it says: "A time interval that represents the fraction of the day that has elapsed since midnight." This isn't what we (humans) think of when we say 'time'. For time, we think 'clock time', not 'duration since midnight'. It is a weird distinction, but it exists, even in code.

Since the use case of `"type":"time"` is a clock representation of time, users of our SDKs would want a clock time. So, for FormRecognizer, if the service returns the equivalent of '3:54 pm', we'd want a type that represents that clearly. And, I don't think the connotation of that value is represented correctly in TimeSpan. If you read through some of the comments on the answers here, you'll see other users with this differentiation: https://stackoverflow.com/questions/2037283/how-do-i-represent-a-time-only-value-in-net

Even though the implementation of time in DateTime uses TimeSpan, I don't think the actual representation of time is a TimeSpan (as confusing at that sounds). Honestly, if you took the date aspects out of DateTime, you'd have the time aspects I'm referring to. Half of the methods and properties in DateTime represent time. The problem everyone has with it is that there is also the date aspects, which you don't want in this situation. If I think about it conceptually, I feel Date actually derives of Time, as it adds another concept to it, like how Fish derives of Animal, and adds Swimming and Breathing Under Water to it, for example.

Honestly, we could just make a type called Time, that contains a private DateTime, and only exposes the time-based methods/properties from DateTime. That's basically want we want here. It is just non-standard, but we have plenty of non-standard types we've created in Azure.Core already to suit our purposes of representing concepts from services. But, if we don't want to do that, I'd prefer DateTimeOffset, since you'd still have the time-based method/properties.

pakrym: I double checked with @KrzysztofCwalina, he agrees that `TimeSpan` is a reasonable type to use in this case.

pakrym: I just don't see how DateTimeOffset is better, the intended use case has no date information and no time zone information. What does using DateTimeOffset give us except for larger object and possible customer confusion?

pakrym: it also looks strange in debugger/when called ToString upon:

```
Console.WriteLine(new DateTimeOffset(1, 1, 1, 13,2,3, default).ToString());
1/1/0001 1:02:03 PM +00:00
Console.WriteLine(new TimeSpan(13,2,3).ToString());
13:02:03

```

heaths: Swagger uses a subset of JSON schema which currently does not define time, true: https://swagger.io/specification/#dataTypes

But since other types align to JSON schema (which is why I pointed to it), it's reasonable to assume if `time` was added it would follow suit, which does support time zone offsets. Sure, you could use `TimeSpan`, but 1) math changes if the property is modifiable, so it's not simply reset at the beginning of some arbitrary day, and 2) `TimeSpan` doesn't support parsing zone offsets, so you'll have to yourself.

The debugger output is easily customized with `DebuggerDisplayAttribute` if that's your concern. My concern is making sure the type format is properly supported if deserializing or round tripping.

pakrym: > math changes if the property is modifiable, so it's not simply reset at the beginning of some arbitrary day

Not sure I understand this part.

I also double checked with Johan

> My expectation is that it supports partial-time.
> But you have a good point in that supporting the json schema draft would be of value. In practical terms I’ve never seen a full-time be returned by a service.

We would always be able to support full time using the customized properties and redefining type to be DateTimeOffset.

PR #651, opened 2020-04-11
Add issue template

heaths:
Added an issue template to recommend workaround and instance labels, as well as default to adding the v3 label.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue.md
+++ b/.github/ISSUE_TEMPLATE/issue.md
@@ -1,0 +1,25 @@
+---
+name: Issue
+about: Bug or feature request
+labels: v3
+---
+**Describe the issue or request**
+Please describe the issue or feature request.
+
+**Describe your ideas for solutions**
+Please describe how you think a change or changes might work. Consider
+how these changes might work for multiple services.
+
+**Add labels**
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/issue.md
```diff
@@ -0,0 +1,25 @@
+---
+name: Issue
+about: Bug or feature request
+labels: v3
```
heaths, line 4: Seems almost all the bugs are v3, so I thought this might help. We can remove it, though.

heaths:
@pakrym something about your recent change seems to be causing the build break:

```
MSBUILD : error MSB1050: Specify which project or solution file to use because the folder "D:\a\1\s\test\TestServerProjects\body-time" contains more than one project or solution file.
```

But when I check the directory in the source repo, I see only a single .csproj (and .yml file, which shouldn't count). I rebased on feature/v3 but that did no good. Any idea? Does a project get generated or anything? I looked through the pipeline yaml as well and nothing stood out as a problem.

pakrym:
Yeah, two PR clashed. https://github.com/Azure/autorest.csharp/pull/643 would fix it.

Issue #521, opened 2020-03-10
Collection properties should be read-only


heaths:
Based on [.NET guidelines for collection properties](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections#collection-properties-and-return-values), collection properties should be read-only. The collection may be mutable or not, but the setter should be internal at most.

> I think it's a bit more complicated than that. Some lists are nullable so we have to leave the properties settable but this is something this PR didn't try to address.
>
> _Originally posted by @pakrym in https://github.com/Azure/azure-sdk-for-net/pull/10442_

In these cases, I recommend doing what Key Vault did (I hand-coded much of the serialization): if an empty array is not expected, treat null or empty as null when serializing (though, that really only matters during deserialization, since collection properties can't actually be null).

If there are cases where a collection property should serialize to an empty array, we may need additional markup in swagger or using a class template to denote that.

pakrym:
> In these cases, I recommend doing what Key Vault did (I hand-coded much of the serialization): if an empty array is not expected, treat null or empty as null when serializing (though, that really only matters during deserialization, since collection properties can't actually be null).

There is no way to know `if an empty array is not expected` from a swagger where collection is marked as not required.

heaths:
It would have to be up to the client library to define the behavior, which is what I was suggesting above. .NET guidelines are pretty clear, and the potential for null collection property values are what it's trying to avoid. It would be unfortunate for the customer to have to start checking for null or empty in some libraries for Azure SDK, while not for other libraries or .NET (at least for newer APIs).

heaths:
To note, we should also consider making this lazy using `LazyInitializer.EnsureInitialized`, though maybe that doesn't make sense everywhere. I guess purely based on generated code, it's difficult to tell (at best).

heaths:
We could provide a hint to the serializer to treat empty and null the same, perhaps a boolean property we could set in `CodeGenSchemaMemberAttribute`.

heaths:
This is blocking release. I can't work around this because, even though I can override the fields and make them read-only (no setter), the serialization code still only checks for null and sends an empty array, which Search does not like for fields that should be null/non-existent.

heaths:
While I think most services probably wouldn't want an empty JSON array, I think is safe to assume empty as a default; however, both Key Vault and Search want undefined / unserialized. Since Search uses generated code, I'd like to get a chance for this because we have a lot of models and properties in violation of the .NET guidelines.

[...]

PR #653, opened 2020-04-13
Add EmptyAsUndefined and InitializeWith parameters to CodeGenMember

pakrym:
`EmptyAsUndefined` tells the generator not to serialize collection if it's empty.
`InitializeWith` tells the generator to always initialize the property in default ctor and during deserialization even if it's not present in JSON.

Allows to implement https://github.com/Azure/autorest.csharp/issues/521


test/TestProjects/SerializationCustomization/Generated/Models/EmptyAsUndefinedTestModel.Serialization.cs
```diff
@@ -0,0 +1,32 @@
...
+using Azure.Core;
+
+namespace SerializationCustomization.Models
+{
+    public partial class EmptyAsUndefinedTestModel : IUtf8JsonSerializable
+    {
+        void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
+        {
+            writer.WriteStartObject();
+            if (EmptyAsUndefinedList != null && EmptyAsUndefinedList.Any())
```
heaths, line 19: `Any()` in .NET Framework is slow and allocates a new enumerator. In .NET Core, it checks for `ICollection<T>` or `ICollection` first to use its `Count` property. As long as you know this will be `IList<T>` or `IReadOnlyList<T>`, just use it's `Count` property for faster/less allocations code.

test/TestProjects/SerializationCustomization/Generated/Models/AlwaysInitializeTestModel.cs
```diff
@@ -0,0 +1,65 @@
...
+            AlwaysInitializeList = alwaysInitializeList ?? new List<Item>();
+            RequiredList = requiredList;
+            RequiredAlwaysInitializeList = requiredAlwaysInitializeList ?? new List<Item>();
+            DefaultList = defaultList;
+            RequiredAlwaysInitializeObject = requiredAlwaysInitializeObject ?? new Item();
+            AlwaysInitializeObject = alwaysInitializeObject ?? new Item();
+            DefaultObject = defaultObject;
+        }
+        public IList<Item> RequiredList { get; }
+        public IList<Item> DefaultList { get; set; }
```
heaths, line 62: This still violates .NET Guidelines. Collection properties should never be settable.

pakrym: Guidelines don't cover all cases that generator are used for so they would need to change to accomodate.

test/TestProjects/SerializationCustomization/Generated/Models/EmptyAsUndefinedTestModel.Serialization.cs
pakrym: Don't want to add a complicated type mapping logic yet. I'll do it if it becomes a problem.

samples/CognitiveSearch/CognitiveSearch/Generated/Models/Analyzer.cs
```diff
@@ -30,7 +30,7 @@ public Analyzer(string name)
         /// <param name="name"> The name of the analyzer. It must only contain letters, digits, spaces, dashes or underscores, can only start and end with alphanumeric characters, and is limited to 128 characters. </param>
         internal Analyzer(string odataType, string name)
         {
-            OdataType = odataType ?? null;
```
heaths, line 33: Should either of these actually be validating? These are both required in Search. Could be separate issue if so.

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -59,8 +60,33 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
...
+                            }
+
+                            if (emptyAsUndefined)
+                            {
+                                if (hasNullableType)
+                                {
+                                    writer.Append($" && ");
+                                }
+                                writer.UseNamespace(typeof(Enumerable).Namespace!);
+                                writer.Append($"{propertyName}.Any()");
```
heaths, line 83: I recommend `Count` if you know it's `IList<T>` or `IReadOnlyList<T>`. `Any()` in .NET Framework (as opposed to .NET Core) is slower and allocates an enumerator (creates enumerator, calls MoveNext() once).

src/assets/Generator.Shared/CodeGenMemberAttribute.cs
```diff
@@ -1,14 +1,31 @@
...
+        /// </summary>
+        public bool EmptyAsUndefined { get; set; }
+
+        /// <summary>
+        /// For collection and model properties. Whether the property would always be initialized on creation/deserialization.
+        /// Requires a parameterless constructor for implementation type.
+        /// </summary>
+        public Type? InitializeWith { get; set; }
+
+        public CodeGenMemberAttribute()
```
heaths, line 26: Should `OriginalName` be read-write then, or just expect them to use the right one?

test/AutoRest.TestServer.Tests/Infrastructure/JsonAsserts.cs
```diff
@@ -0,0 +1,24 @@
...
+using Azure.Core;
+using NUnit.Framework;
+
+namespace AutoRest.TestServer.Tests.Infrastructure
+{
+    internal static class JsonAsserts
+    {
+        public static void AssertSerialization(string expected, IUtf8JsonSerializable serializable)
+        {
+            var memoryStream = new MemoryStream();
```
heaths, line 13: IIRC, it doesn't actually do much for a `MemoryStream`, but should use the `using` keyword here as well (could use new one without new block).

samples/CognitiveSearch/CognitiveSearch/Generated/Models/Analyzer.cs
pakrym: This is a serialization ctor, we are not validating service responses.

src/assets/Generator.Shared/CodeGenMemberAttribute.cs
pakrym: We expect the right one to be set using ctor parameter if the value is required.

test/TestProjects/SerializationCustomization/Generated/Models/EmptyAsUndefinedTestModel.Serialization.cs
heaths: How about this, then (I did this for the search engine): write our own extension method like `IsEmpty()` (I typically do `IsEmptyOrNull`, but whichever) that does what .NET Core does, which is basically:

```csharp
if (obj is ICollection<T> c) return c.Count > 0;
else if (object is ICollection c) return c.Count > 0;
else { var e = obj.GetEnumerator(); return !e.MoveNext(); }
```

Still, maybe not worth it (especially when you consider Framework code is NGEN'd, which makes a different if called a lot (I tested)). Just wanted to call it out as an option that works in both cases.

test/TestProjects/SerializationCustomization/Generated/Models/AlwaysInitializeTestModel.cs
heaths: The guidelines are there for consistent behavior. If collection properties can sometimes be null, that's even worse. As I understand, making them settable originated because of the null-or-empty problem. I think we should focus on solving that (even if it requires "hints" like this PR) and have a consistent public API. @KrzysztofCwalina, thoughts? If we really want to entertain "sometimes-null" collection properties or collection properties that can be set - completely breaking any stored (even temporarily in a call path) references to them - this needs to be a larger discussion; though, perhaps not blocking this PR. At least this is a great start to making it possible to opt into the guidelines.

pakrym: What I don't know how to solve yet is there are 200 management plane libraries and no resources to add hints vs 2 data plane with plenty resources. This is the reason I'm reluctant to flip the switch right away and want to gather more data.

I agree we need a larger discussion, hope API review of storage management might be that.

test/TestProjects/SerializationCustomization/Generated/Models/EmptyAsUndefinedTestModel.Serialization.cs
pakrym: It would definitely work and would definitely have an effect if called a lot, just don't know if we are there yet.

test/TestProjects/SerializationCustomization/AlwaysInitializeTestClass.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Azure.Core;
+
+namespace SerializationCustomization.Models
+{
+    public partial class AlwaysInitializeTestModel
+    {
+        [CodeGenMember(InitializeWith = typeof(List<Item>))]
```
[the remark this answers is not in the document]
pakrym: Good point, I'll drop `InitializeWith` all together.

test/TestProjects/SerializationCustomization/AlwaysInitializeTestClass.cs
```diff
@@ -0,0 +1,23 @@
...
+        public IList<Item> AlwaysInitializeList { get; }
+
+        [CodeGenMember(InitializeWith = typeof(List<Item>))]
+        public IList<Item> RequiredAlwaysInitializeList { get; }
+
+        [CodeGenMember(InitializeWith = typeof(Item))]
```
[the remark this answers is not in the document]
pakrym: It doesn't, it's in the doc comment for `InitializeWith`

Issue #652, opened 2020-04-13
Customized member trimming logic removes required overloads


pakrym:
Generating https://github.com/Azure/azure-rest-api-specs/blob/master/specification/cognitiveservices/data-plane/FormRecognizer/preview/v2.0/FormRecognizer.json trims CreateAnalyzeLayoutAsync overloads

PR #655, opened 2020-04-13
Fix CreateOperation method conflicts

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/652


src/AutoRest.CSharp.V3/AutoRest/Plugins/MemberRemoverRewriter.cs
```diff
@@ -103,7 +105,7 @@ private bool ShouldRemove(ISymbol? symbol)
                     foreach (var member in members)
                     {
                         if (!member.Equals(symbol) &&
-                            member.DeclaringSyntaxReferences.Any())
+                            IsDeclaredInNonGeneratedCode(member))
```
pakrym, line 108: Don't remove members from generated files to avoid hiding the generator problems.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -300,7 +300,11 @@ private void WriteCreateOperationOperation(CodeWriter writer, LongRunningOperati
                 writer.WriteXmlDocumentationParameter(parameter.Name, parameter.Description);
             }
 
-            writer.Append($"internal {responseType} {CreateCreateOperationName(lroMethod.Name)}(");
+            // Create operation methods can clash for LROs that take different set of parameters
+            // because the input parameters are not part of Create*** signature
+            // It's an internal method so generated name is fine
+            var declaration = new CodeWriterDeclaration(CreateCreateOperationName(lroMethod.Name));
```
pakrym, line 306: The actual fix.

test/TestProjects/NameConflicts/Generated/Operations/ServiceClient.cs
```diff
@@ -55,5 +58,85 @@ public virtual Response<Struct> Operation(string request, string message, string
...
+            }
+
+            var originalResponse = RestClient.AnalyzeBody(stringBody, cancellationToken);
+            return CreateAnalyzeBody(originalResponse, () => RestClient.CreateAnalyzeBodyRequest(stringBody));
+        }
+
+        /// <summary> Analyze body, that could be different media types. </summary>
+        /// <param name="originalResponse"> The original response from starting the operation. </param>
+        /// <param name="createOriginalHttpMessage"> Creates the HTTP message used for the original request. </param>
+        internal Operation<Response> CreateAnalyzeBody0(Response originalResponse, Func<HttpMessage> createOriginalHttpMessage)
```
pakrym, line 110: The result.

src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -128,5 +128,7 @@ public async Task<CSharpCompilation> GetCompilationAsync()
             Debug.Assert(compilation != null);
             return compilation;
         }
+
+        public static bool IsGeneratedDocument(Document document) => document.Folders.Contains(GeneratedFolder);
```
MiYanni, line 132: Is it searching the file system each time to run this? Should we cache the result?

test/TestProjects/NameConflicts/Generated/Operations/ServiceClient.cs
MiYanni: I'm a bit confused. The code and parameters for this are identical to `CreateAnalyzeBody`. These Create methods (for the same operation) would always be the same?? The difference in `StartAnalyzeBody` is based on Content-Type, which only applies to body parameters. Since body parameters don't apply here, when would the `Create` methods differ?

It sounds like, based on the issue, that we'd have the right `StartAnalyzeBody` overloads, but not duplicate `CreateAnalyzeBody` methods.

src/AutoRest.CSharp.V3/AutoRest/Plugins/GeneratedCodeWorkspace.cs
pakrym: Folders is an in-memory property of Roslyn `Document` type. Not IO here.

test/TestProjects/NameConflicts/Generated/Operations/ServiceClient.cs
pakrym: Good point, deduplication might be better.

pakrym:
I've addressed comments, merging to unblock library dev work.

Issue #623, opened 2020-04-03
Unable to find object property with serialized name  in schema


pakrym:
apimanagement-control-plane
azsadmin-resource-manager-azurebridge

```
 System.InvalidOperationException: Unable to find object property with serialized name  in schema RegionListResult
   at AutoRest.CSharp.V3.Output.Models.Types.ObjectType.GetPropertyBySerializedName(String serializedName, Boolean includeParents) in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Models\Types\ObjectType.cs:line 238
   at AutoRest.CSharp.V3.Output.Builders.ClientBuilder.GetPagingInfo(RestClientMethod method, RestClientMethod nextPageMethod, Paging paging, ObjectType type) in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Builders\ClientBuilder.cs:line 468
   at AutoRest.CSharp.V3.Output.Builders.ClientBuilder.BuildClient(OperationGroup operationGroup) in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Builders\ClientBuilder.cs:line 121
   at System.Linq.Enumerable.SelectListIterator`2.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at AutoRest.CSharp.V3.Output.Models.Types.OutputLibrary.EnsureClients() in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Models\Types\OutputLibrary.cs:line 57
   at AutoRest.CSharp.V3.Output.Models.Types.OutputLibrary.get_RestClients() in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Models\Types\OutputLibrary.cs:line 32
   at AutoRest.CSharp.V3.AutoRest.Plugins.CSharpGen.Execute(IPluginCommunication autoRest, CodeModel codeModel, Configuration configuration) in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\AutoRest\Plugins\CSharpGen.cs:line 56
[...]

PR #656, opened 2020-04-13
Allow empty nextLink

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/623


MiYanni:
FYI, it isn't advised from a swagger review standpoint to have an empty nextLink. Usually, it is either provided or omitted; not empty.

pakrym:
In all fairness M4 should be handling this by synthetizing GetNextMethod...

PR #659, opened 2020-04-15
Ignore duplicated properties

pakrym:


src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -212,19 +227,6 @@ private static void PopulatePropertyBag(PropertyBag propertyBag, int depthIndex)
             }
         }
 
-        private static IEnumerable<ObjectSchema> EnumerateHierarchy(ObjectSchema schema)
-        {
-            yield return schema;
-            // WORKAROUND: https://github.com/Azure/autorest.modelerfour/issues/257
```
MiYanni, line 218: Are we testing out the fix for this in this PR?

src/AutoRest.CSharp.V3/Output/Models/Types/ObjectType.cs
```diff
@@ -358,10 +358,23 @@ private ObjectTypeDiscriminatorImplementation[] CreateDiscriminatorImplementatio
...
+                .Skip(1)
+                .SelectMany(type => type._objectSchema.Properties)
+                .Select(p => p.Language.Default.Name)
+                .ToHashSet();
+
             foreach (var objectSchema in GetCombinedSchemas())
             {
                 foreach (Property property in objectSchema.Properties!)
                 {
+                    if (existingProperties.Contains(property.Language.Default.Name))
```
MiYanni, line 372: Ah. So, we're considering it a duplicate if it has the same name. Got it.

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
pakrym: Nope, that workaround was done in 2 places, I removed one of them.

PR #663, opened 2020-04-15
Split smoke tests

pakrym:
Reenable more smoke tests and Azure.Network.Management.Interface


eng/Generate.ps1
```diff
@@ -91,7 +91,7 @@ $projectNames =
     'CognitiveSearch',
     'Azure.AI.FormRecognizer',
     'Azure.Storage.Tables',
-    'Azure.Storage.Management'
+    'Azure.Storage.Management',
```
pakrym, line 94: :(

MiYanni: image

Issue #625, opened 2020-04-03
System.InvalidOperationException: Sequence contains more than one element


pakrym:
azsadmin-resource-manager-fabric

```
FATAL: System.InvalidOperationException: Sequence contains more than one element
   at System.Linq.ThrowHelper.ThrowMoreThanOneElementException()
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source)
   at AutoRest.CSharp.V3.Output.Builders.ClientBuilder.BuildClient(OperationGroup operationGroup) in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Builders\ClientBuilder.cs:line 110
   at System.Linq.Enumerable.SelectListIterator`2.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at AutoRest.CSharp.V3.Output.Models.Types.OutputLibrary.EnsureClients() in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Models\Types\OutputLibrary.cs:line 57
   at AutoRest.CSharp.V3.Output.Models.Types.OutputLibrary.get_RestClients() in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Models\Types\OutputLibrary.cs:line 32
   at AutoRest.CSharp.V3.AutoRest.Plugins.CSharpGen.Execute(IPluginCommunication autoRest, CodeModel codeModel, Configuration configuration) in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\AutoRest\Plugins\CSharpGen.cs:line 56
```

PR #664, opened 2020-04-15
Handle multiple paged responses

pakrym:
Just pick the one with body.

Fixes https://github.com/Azure/autorest.csharp/issues/625

Filed: https://github.com/Azure/autorest.testserver/issues/169

### DIFF
--- a/src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
+++ b/src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
@@ -107,7 +107,7 @@ namespace AutoRest.CSharp.V3.Output.Builders
                         nextPageMethods.Add(nextPageMethod);
                     }
 
-                    if (!(method.Responses.Single().ResponseBody is ObjectResponseBody objectResponseBody))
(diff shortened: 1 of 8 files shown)


src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -107,7 +107,7 @@ public ClientBuilder(BuildContext context)
                         nextPageMethods.Add(nextPageMethod);
                     }
 
-                    if (!(method.Responses.Single().ResponseBody is ObjectResponseBody objectResponseBody))
+                    if (!(method.Responses.Single(r => r.ResponseBody != null).ResponseBody is ObjectResponseBody objectResponseBody))
```
MiYanni, line 110: Is there a reason this is Single instead of First? Is there not a situation in which 2 different responses return different response bodies?

Also, line 118 has `and` in the message instead of `an`.

pakrym: > Is there not a situation in which 2 different responses return different response bodies?

Never see it.

Issue #621, opened 2020-04-03
Assertion failed in ObjectType.BuildConstructors()+MoveNext()


pakrym:
> https://github.com/Azure/azure-rest-api-specs/blob/49fc16354df7211f8392c56884a3437138317d1f/specification/datashare/resource-manager/readme.md

```
  at AutoRest.CSharp.V3.Output.Models.Types.ObjectType.BuildConstructors()+MoveNext() in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Models\Types\ObjectType.cs:line 172
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at AutoRest.CSharp.V3.Output.Models.Types.ObjectType.get_Constructors() in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Output\Models\Types\ObjectType.cs:line 91
   at AutoRest.CSharp.V3.Generation.Writers.ModelWriter.WriteConstructor(CodeWriter writer, ObjectType schema) in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Generation\Writers\ModelWriter.cs:line 166
   at AutoRest.CSharp.V3.Generation.Writers.ModelWriter.WriteObjectSchema(CodeWriter writer, ObjectType schema) in D:\github\azure\autorest.csharp\src\AutoRest.CSharp.V3\Generation\Writers\ModelWriter.cs:line 76
```

same for 
> https://github.com/Azure/azure-rest-api-specs/blob/49fc16354df7211f8392c56884a3437138317d1f/specification/consumption/resource-manager/readme.md

PR #667, opened 2020-04-16
Handle property that redefines a composed base type property

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/621

Handles the following situation:

```
A { PropA }
B { PropB }
C: A, B // B get's inlined into C
D: C { PropB }
```


src/AutoRest.CSharp.V3/Output/Models/Types/ObjectType.cs
```diff
@@ -361,8 +361,8 @@ private IEnumerable<ObjectTypeProperty> BuildProperties()
             // WORKAROUND: https://github.com/Azure/autorest.modelerfour/issues/261
             var existingProperties = EnumerateHierarchy()
                 .Skip(1)
-                .SelectMany(type => type._objectSchema.Properties)
-                .Select(p => p.Language.Default.Name)
+                .SelectMany(type => type.Properties)
```
pakrym, line 364: Use ObjectType.Properties here because they were already processed and combined to handle multiple inheritance while schemas were not.

PR #669, opened 2020-04-16
Add protected mocking ctors

pakrym:
And rename Storage.Management to it's correct name.


MiYanni:
So, in Track 2, is everything that is management plan in `Azure.Management`?

pakrym:
> So, in Track 2, is everything that is management plane in Azure.Management?

That's that current plan of action that might still change.

MiYanni:
That seems a little confusing... take `Azure.AI.ConitiveServices`. So, a management plane namespace would be `Azure.Managment.AI.CognitiveServices`? It seems like it would be better as `Azure.AI.CognitiveServices.Management`. For Storage, if you want to both create a storage container and then put a blob into it, you need to access 2 different namespaces... :-/

pakrym:
Feel free to start a discussion on this topic in the channel. It's not being decided here.

PR #670, opened 2020-04-17
New tests, object-type and multiple-inheritance

MiYanni:


eng/Generate.ps1
```diff
@@ -44,14 +44,16 @@ $testNames =
...
     'non-string-enum',
+    'object-type',
     'paging',
     'required-optional',
     'url',
+    'url-multi-collectionFormat',
     'validation',
     'xml-service',
-    'xms-error-responses',
-    'url-multi-collectionFormat';
```
MiYanni, line 54: Just fixed the ordering here since I forgot to do that in the xms-error-response PR.

PR #677, opened 2020-04-21
Add customizations doc

pakrym:


readme.md
```diff
@@ -12,6 +11,601 @@
...
+
+## Customizing the generated code
+
+<!-- TOC -->
+
+- [Make a model internal](#make-a-model-internal)
+- [Rename a model class](#rename-a-model-class)
+- [Make model property internal](#make-model-property-internal)
+- [Rename a model property](#rename-a-model-property)
+- [Change a model property type](#change-a-model-property-type)
```
heaths, line 25: Might also mention you can change properties to fields, like I did for etags.

pakrym:
> What about instead using the wiki so we can add concise pages faster than going through a PR process

Want to put this into the most prominent place to start with, I've always had trouble finding things in github project wikis.

Issue #403, opened 2020-01-14
Implement partial nextLink support for paging


MiYanni:
```csharp
var uriType = writer.Type(typeof(Uri));
using (nextPage ? writer.If($"{uriType}.IsWellFormedUriString(nextLinkUrl, UriKind.Absolute)") : default)
{
    if (nextPage)
    {
        writer.Line($"request.Uri.Reset(new {uriType}(nextLinkUrl));");
    }
}
using (nextPage ? writer.Else() : default)
{
    //TODO: Add logic to escape the strings when specified, using Uri.EscapeDataString(value);
    //TODO: Need proper logic to convert the values to strings. Right now, everything is just using default ToString().
    //TODO: Need logic to trim duplicate slashes (/) so when combined, you don't end  up with multiple // together
    var urlText = String.Join(String.Empty, operation.Request.HostSegments.Select(s => s.IsConstant ? s.Constant.Value : "{" + s.Parameter.Name + "}"));
    var nextLinkText = nextPage ? "{nextLinkUrl}" : string.Empty;
    writer.Line($"request.Uri.Reset(new {uriType}($\"{urlText}{nextLinkText}\"));");

    if (!nextPage)
    {
        foreach (var segment in operation.Request.PathSegments)
        {
            WritePathSegment(writer, segment);
        }

        //TODO: Duplicate code between query and header parameter processing logic
        foreach (var queryParameter in operation.Request.Query)
        {
            WriteQueryParameter(writer, queryParameter);
        }
    }
}
```

PR #682, opened 2020-04-22
Paging: Partial & query nextLink fixes

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/403
Fixes: https://github.com/Azure/autorest.csharp/issues/411

This allows partial next links by checking if the nextLink value is absolute (via if it starts with http). If it is absolute, it does the same thing it did before, `AppendRaw`. If it is relative (not absolute), it writes the host segments (`AppendRaw`) and writes the nextLink as a path parameter (`AppendPath`). This strategy is for non-natural (artificial) nextLink parameters (those we create when crafting a new operation).

For natural (non-artificial) nextLink parameters (which is when nextLink is part of a linked operation via `operationName`), we need a different strategy as they will always be defined as path parameters naturally. These natural nextLink path parameters can apparently contain **both** path and query elements. For this situation, we now call `AppendRawNextLink` instead of `AppendPath`. This allows the path and query elements from the nextLink to be appended appropriately (via `AppendPath` and `AppendQuery` respectively).


src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -92,13 +92,13 @@ private void WriteClientCtor(CodeWriter writer, RestClient restClient, CSharpTyp
 
         private string CreateRequestMethodName(string name) => $"Create{name}Request";
 
-        private void WriteRequestCreation(CodeWriter writer, RestClientMethod operation)
+        private void WriteRequestCreation(CodeWriter writer, RestClientMethod clientMethod)
```
MiYanni, line 95: Fixed parameter name for clarity.

src/AutoRest.CSharp.V3/Output/Models/Shared/Parameter.cs
```diff
@@ -8,19 +8,22 @@ namespace AutoRest.CSharp.V3.Output.Models.Shared
...
+            IsArtificial = isArtificial;
         }
 
         public CSharpType Type { get; }
         public string Name { get; }
         public string? Description { get; }
         public Constant? DefaultValue { get; }
         public bool IsRequired { get; }
+        // Indicates the parameter was created by the generator, not defined in the swagger
+        public bool IsArtificial { get; }
```
MiYanni, line 27: Required to differentiate a nextLink parameter we create (artificial) versus a nextLink parameter that comes from swagger (natural).

test/AutoRest.TestServer.Tests/paging.cs
```diff
@@ -309,10 +309,10 @@ await foreach (var page in pageableAsync.AsPages())
             Assert.AreEqual(id, resultPage.Values.First().Properties.Id);
             Assert.AreEqual(product, resultPage.Values.First().Properties.Name);
             Assert.AreEqual("*&*#&$", resultPage.ContinuationToken);
-            Assert.ThrowsAsync<UriFormatException>(async () => await new PagingClient(ClientDiagnostics, pipeline, host).RestClient.GetMultiplePagesFailureNextPageAsync(resultPage.ContinuationToken));
+            Assert.ThrowsAsync<RequestFailedException>(async () => await new PagingClient(ClientDiagnostics, pipeline, host).RestClient.GetMultiplePagesFailureNextPageAsync(resultPage.ContinuationToken));
```
MiYanni, line 312: Since we properly concatenate host + nextLink now, this fails with RequestFailedException (as it should).

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -429,16 +424,16 @@ private static RestClientMethod BuildNextPageMethod(RestClientMethod method)
...
+                isArtificial: true);
+            List<Parameter> parameters = new List<Parameter> { nextPageUrlParameter };
             parameters.AddRange(method.Parameters.Where(p => p.Name != nextPageUrlParameter.Name));
 
             var request = new Request(
                 method.Request.HttpMethod,
-                new[] { new PathSegment(nextPageUrlParameter, false, SerializationFormat.Default),  },
-                Array.Empty<PathSegment>(),
+                method.Request.HostSegments,
+                new[] { new PathSegment(nextPageUrlParameter, false, SerializationFormat.Default) },
```
MiYanni, line 436: We need host segments now, and the artificial nextLink parameter becomes a path parameter.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -113,39 +113,58 @@ private void WriteRequestCreation(CodeWriter writer, RestClientMethod operation)
...
                 writer.Line($"var {uri:D} = new RawRequestUriBuilder();");
-                foreach (var segment in operation.Request.HostSegments)
+                Parameter? nextLinkParameter = parameters.FirstOrDefault(p => p.Name == "nextLink");
+                bool hasArtificialNextLink = nextLinkParameter?.IsArtificial == true;
+                PathSegment? nextLinkSegment = clientMethod.Request.PathSegments.FirstOrDefault(s => !s.Value.IsConstant && s.Value.Reference.Name == "nextLink");
+                // Artificial nextLink needs additional logic for relative versus absolute links
+                if (hasArtificialNextLink)
                 {
-                    WriteUriFragment(writer, uri, segment);
+                    using (writer.Scope($"if ({nextLinkParameter!.Name:D}.StartsWith({typeof(Uri)}.UriSchemeHttp, {typeof(StringComparison)}.InvariantCultureIgnoreCase))"))
```
pakrym, line 126: Can we put this logic into AppendRawNextLink?

src/AutoRest.CSharp.V3/Output/Models/Shared/Parameter.cs
pakrym: What about `IsRawNextLink`?

src/assets/Generator.Shared/RawRequestUriBuilder.cs
```diff
@@ -87,5 +90,56 @@ private enum RawWritingPosition
...
+
+        public void AppendRawNextLink(string nextLink, bool escape)
+        {
+            (string Name, string Value) GetQueryParts(string queryUnparsed)
+            {
+                int separatorIndex = nextLink.IndexOf(QueryValueSeparator);
+                return (queryUnparsed.Substring(0, separatorIndex), queryUnparsed.Substring(separatorIndex + 1));
+            }
+
+            RawNextLinkWritingPosition position = RawNextLinkWritingPosition.Path;
```
pakrym, line 102: This logic should go into AppendRaw

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -113,39 +113,58 @@ private void WriteRequestCreation(CodeWriter writer, RestClientMethod operation)
...
 
-                //TODO: Add logic to escape the strings when specified, using Uri.EscapeDataString(value);
-                //TODO: Need proper logic to convert the values to strings. Right now, everything is just using default ToString().
-                //TODO: Need logic to trim duplicate slashes (/) so when combined, you don't end  up with multiple // together
-
                 writer.Line($"var {uri:D} = new RawRequestUriBuilder();");
-                foreach (var segment in operation.Request.HostSegments)
+                Parameter? nextLinkParameter = parameters.FirstOrDefault(p => p.Name == "nextLink");
+                bool hasArtificialNextLink = nextLinkParameter?.IsArtificial == true;
+                PathSegment? nextLinkSegment = clientMethod.Request.PathSegments.FirstOrDefault(s => !s.Value.IsConstant && s.Value.Reference.Name == "nextLink");
```
pakrym, line 122: Let's mark the segment in the building code and not rely on the name here.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
pakrym: Let's merge host and paths segments together separation is not useful anymore.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: Unfortunately, we cannot. If you look at what was changed in the generated output, if this if-statement is false, we write host + non-parsed nextLink (as a parameter). If this fix statement is true, we only write nextLink as the entire url (via AppendRaw). Essentially, this is wrapping the dynamic amount of host segments.

pakrym: We can reset the URI as part of `AppendRawNextLink`

src/AutoRest.CSharp.V3/Output/Models/Shared/Parameter.cs
MiYanni: Since this is in `Parameter`, I didn't want it to be to the specific context of `nextLink`/paging. So, it could be used for any non-natural parameters. We have other ones, like `cancellationToken`, we could create in this way, which would help for parameter writing/processing consistency. For nextLink, I only care if it came from swagger or not.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: So, have Segment specify if it is host or path?

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: Hmm. I'll experiment with this. It is kinda bad that we'd write the host segments and then just remove them. But, it does simplify the `CreateRequest` writer.

src/assets/Generator.Shared/RawRequestUriBuilder.cs
MiYanni: Okay. That'll take me some time. If I'm going to modify it, I'd like to have some tests around it.

pakrym: There is a test class for it https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.Shared.Tests/RawRequestUriBuilderTest.cs

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
pakrym: Do we even care about the distiction?

src/AutoRest.CSharp.V3/Output/Models/Shared/Parameter.cs
pakrym: Oh, I though this is on the PathSegment, let's move it to be on the path segment.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: Well, host segments use WriteUriFragment and path segments use WritePathSegment, and those use AppendRaw and AppendPath respectively.

MiYanni: If I make the bigger changes to AppendRaw to also do path and query params too, then it might not need to use 2 different methods? I'd have to test it.

pakrym: We should use non-Raw methods where possible. Might need to add some bool to the segment

src/AutoRest.CSharp.V3/Output/Models/Shared/Parameter.cs
MiYanni: Fixed. PathSegment handles both host and path segments. Because of that, it now has `IsRaw`.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: Done. Combined collection of host and path segments together. They are only different if the segment has `IsRaw` as true.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: Done. Now, `AppendRawNextLink` is used for artificial nextLink values and contains this if-statement logic.

src/assets/Generator.Shared/RawRequestUriBuilder.cs
MiYanni: Done. This logic is now in AppendRaw. Added several test cases and fixed an issue when using AppendRaw to do paths after a port was specified.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: Done. We only use `IsRaw` in PathSegment now for this determination.

test/TestServerProjects/custom-baseUrl-paging/Generated/Operations/PagingRestClient.cs
```diff
@@ -245,7 +245,7 @@ internal HttpMessage CreateGetPagesPartialUrlOperationNextRequest(string account
             uri.AppendRaw(accountName, false);
             uri.AppendRaw(host, false);
             uri.AppendPath("/paging/customurl/", false);
-            uri.AppendPath(nextLink, false);
+            uri.AppendRaw(nextLink, false);
```
pakrym, line 248: I think we should keep this one as is.

test/AutoRest.Shared.Tests/RawRequestUriBuilderTest.cs
```diff
@@ -12,7 +12,30 @@ public class RawRequestUriBuilderTest
...
+        [TestCase("http://localhost|/more?one=1|&two=2|&three=3", null, null, "http://localhost/more?one=1&two=2&three=3")]
+        [TestCase("http://localhost/more?one=1&two=2|&three=3", null, null, "http://localhost/more?one=1&two=2&three=3")]
+        [TestCase("http://localhost/more|?one=1&two=2|&three=3", null, null, "http://localhost/more?one=1&two=2&three=3")]
+        [TestCase("http://localhost/more|?one=1|&two=2&three=3", null, null, "http://localhost/more?one=1&two=2&three=3")]
+        [TestCase("http://localhost|/more?one=1&two=2&three=3", null, null, "http://localhost/more?one=1&two=2&three=3")]
+        [TestCase("http://|localhost:12345|/more|?one=1|&two=2|&three=3", null, null, "http://localhost:12345/more?one=1&two=2&three=3")]
+        [TestCase("http://|localhost|:12345|/more|?one=1|&two=2|&three=3", null, null, "http://localhost:12345/more?one=1&two=2&three=3")]
+        [TestCase("http://localhost|/more?one=&two=&three=", null, null, "http://localhost/more?one=&two=&three=")]
+        [TestCase("http://localhost|/more?one&two&three", null, null, "http://localhost/more?one=&two=&three=")]
```
pakrym, line 38: Do we support `[TestCase("http://localhost|/more?one&&&&&&&three", null, null, "http://localhost/more?one=&two=&three=")]` ?

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -265,17 +265,12 @@ private RestClientMethod BuildMethod(Operation operation, string clientName, IRe
                             bodyParameter = (requestParameter, constantOrReference);
                             break;
                         case ParameterLocation.Uri:
-                            if (defaultName == "$host")
```
pakrym, line 268: :+1:

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -113,39 +113,45 @@ private void WriteRequestCreation(CodeWriter writer, RestClientMethod operation)
...
+                    {
+                        if (segment.IsRaw)
+                        {
+                            // Artificial nextLink needs additional logic for relative versus absolute links
+                            WritePathSegment(writer, uri, segment, "AppendRawNextLink");
+                        }
+                        else
+                        {
+                            // Natural nextLink parameters need to use a different method to parse path and query elements
+                            WritePathSegment(writer, uri, segment, "AppendRaw");
```
pakrym, line 132: No need to use AppendRaw here

test/TestServerProjects/custom-baseUrl-paging/Generated/Operations/PagingRestClient.cs
pakrym: I take it back.

src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
MiYanni: This was a workaround we no longer need.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: Discussed offline. Needed to parsing values with both path and query in the same nextLink.

test/TestServerProjects/custom-baseUrl-paging/Generated/Operations/PagingRestClient.cs
MiYanni: Discussed offline. Needed to parsing values with both path and query in the same nextLink.

test/AutoRest.Shared.Tests/RawRequestUriBuilderTest.cs
MiYanni: I've added a test for it now. It works as expected. 😁

PR #684, opened 2020-04-23
Re-enable client request id tests

pakrym:
Fixes https://github.com/Azure/autorest.csharp/issues/446

### DIFF
--- a/test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj
+++ b/test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.1.0" />
+    <PackageReference Include="Azure.Core" Version="1.3.0-dev.20200422.1" />
     <PackageReference Include="Azure.Identity" Version="1.0.0" />
     <PackageReference Include="nunit" Version="3.12.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.15.1" />
--- a/test/AutoRest.TestServer.Tests/Infrastructure/CustomClientRequestIdPolicy.cs
+++ b/test/AutoRest.TestServer.Tests/Infrastructure/CustomClientRequestIdPolicy.cs
@@ -1,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Core;
+using Azure.Core.Pipeline;
+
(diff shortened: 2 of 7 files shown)


test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.1.0" />
+    <PackageReference Include="Azure.Core" Version="1.3.0-dev.20200422.1" />
```
pakrym, line 11: Requires latest core

test/AutoRest.TestServer.Tests/Infrastructure/CustomClientRequestIdPolicy.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+        }
+
+        public override void OnSendingRequest(HttpMessage message)
+        {
+            base.OnSendingRequest(message);
+
+            if (ClientRequestIdScope.CurrentClientRequestId != null)
+            {
+                message.Request.ClientRequestId = ClientRequestIdScope.CurrentClientRequestId;
```
MiYanni, line 21: So, why couldn't we do this before? This was what I was thinking originally. Basically, inject a 'replace value' line like this in the pipeline.

MiYanni, line 21: So, why couldn't we do this before? This was what I was thinking originally. Basically, inject a 'replace value' line like this in the pipeline.

test/AutoRest.TestServer.Tests/lro.cs
```diff
@@ -20,45 +20,47 @@ public class LroTest: TestServerTestBase
         public LroTest(TestServerVersion version) : base(version, "lros") { }
 
         [Test]
-        [Ignore("Needs to send x-ms-client-request-id: https://github.com/Azure/autorest.csharp/issues/446")]
+        [Ignore("https://github.com/Azure/autorest.csharp/issues/683")]
```
MiYanni, line 23: What's the problem with this one?

test/AutoRest.TestServer.Tests/Infrastructure/CustomClientRequestIdPolicy.cs
pakrym: We could, with all tests but one that takes it as a parameter

test/AutoRest.TestServer.Tests/lro.cs
pakrym: Described in the issue

Issue #686, opened 2020-04-23
Incorrect implementation of xxxClient


erich-wang:
There're two issues:
1. **options** parameter may be not null, and we should respect it.
2. **options.Versionis** passed into RestClient as **host** incorrectly.

Generated code:
```
        public VaultsClient(string subscriptionId, TokenCredential tokenCredential, KeyVaultManagementClientOptions options = null)
        {
            options = new KeyVaultManagementClientOptions();
            _clientDiagnostics = new ClientDiagnostics(options);
            _pipeline = ManagementPipelineBuilder.Build(tokenCredential, options);
            RestClient = new VaultsRestClient(_clientDiagnostics, _pipeline, subscriptionId, options.Version);
        }

        public VaultsRestClient(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, string subscriptionId, string host = "https://management.azure.com", string apiVersion = "2019-09-01")


```

PR #693, opened 2020-04-27
Adjust generated ARM client constructor implementations

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/686

### DIFF
--- a/src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
+++ b/src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
@@ -102,7 +102,7 @@ namespace AutoRest.CSharp.V3.Generation.Writers
 
             using (writer.Scope())
             {
-                writer.Line($"options = new {configuration.LibraryName}ManagementClientOptions();");
+                writer.Line($"options ??= new {configuration.LibraryName}ManagementClientOptions();");
                 writer.Line($"_clientDiagnostics = new {typeof(ClientDiagnostics)}(options);");
                 writer.Line($"_pipeline = {typeof(ManagementPipelineBuilder)}.Build(tokenCredential, options);");
 
@@ -117,11 +117,11 @@ namespace AutoRest.CSharp.V3.Generation.Writers
 
                     if (ManagementClientWriterHelpers.IsApiVersionParameter(parameter))
                     {
-                        writer.Append($"options.Version, ");
+                        writer.Append($"{parameter.Name}: options.Version, ");
                         continue;
                     }
 
-                    writer.Append($"{parameter.Name:I}, ");
+                    writer.Append($"{parameter.Name}: {parameter.Name:I}, ");
                 }
                 writer.RemoveTrailingComma();
                 writer.Line($");");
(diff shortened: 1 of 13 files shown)


src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -117,11 +117,11 @@ private void WriteManagementClientCtors(CodeWriter writer, Client client, Config
 
                     if (ManagementClientWriterHelpers.IsApiVersionParameter(parameter))
                     {
-                        writer.Append($"options.Version, ");
```
MiYanni, line 120: Hmmm. You stumbled upon an awesome idea for the generator. Parameter ordering or optional parameters on calls wouldn't matter if we used named parameters everywhere. This should make calling writer logic easier and more consistent. However, the generated code might look more 'noisy'. Thoughts?

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
```diff
@@ -102,7 +102,7 @@ private void WriteManagementClientCtors(CodeWriter writer, Client client, Config
 
             using (writer.Scope())
             {
-                writer.Line($"options = new {configuration.LibraryName}ManagementClientOptions();");
+                writer.Line($"options ??= new {configuration.LibraryName}ManagementClientOptions();");
                 writer.Line($"_clientDiagnostics = new {typeof(ClientDiagnostics)}(options);");
                 writer.Line($"_pipeline = {typeof(ManagementPipelineBuilder)}.Build(tokenCredential, options);");
```
MiYanni, line 107: Should we have shared variable names for things like `_pipeline` and `_clientDiagnostics` for client (class-scoped) fields? Might make sense for `options` within this method too.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
pakrym: I'm not a fan of the noisiness part. I would use it sparingly for cases where parameter order is not trivial to match. In this case I used it because I want to use the default value for `host` parameter but have to pass arguments that go both before and after it.

MiYanni: What about... for any parameters that are defaulted, we use a named parameter. Then, it would be consistent and be very resistant to unintended bugs. We could have a parameter callsite writing (such as `:P`) that would do this automatically.

pakrym: I don't see the benefit yet, this change actually fixed a bug but doing it everywhere wouldn't.

src/AutoRest.CSharp.V3/Generation/Writers/ClientWriter.cs
pakrym: Considering the entire file is a bunch of constant strings I'm not sure if extracting all the names makes things better of just harder to read.

PR #694, opened 2020-04-27
Rename REST clients along with the public ones

pakrym:
Fixes https://github.com/Azure/autorest.csharp/issues/689


src/AutoRest.CSharp.V3/Output/Builders/ClientBuilder.cs
```diff
@@ -162,6 +165,12 @@ private string GetClientName(OperationGroup operationGroup)
                 name = name.Substring(0, name.Length - operationsSuffix.Length);
             }
 
+            var clientSuffix = "Client";
```
MiYanni, line 168: This should just be a const field in the class and used both in this method and BuildClient.

Issue #702, opened 2020-04-28
[P0] Temporarily disallow to set ApiVersion for xxxRestClient


erich-wang:
For different xxxRestClient, they may have different api versions, short-term fix is to disallow to set ApiVersion for xxxRestClient

PR #712, opened 2020-04-29
Temporary disable client version generation

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/702


pakrym:
Git would remember and we can always revert it.

The reason is that we currently presume that 1 API version is applicable to all the clients in the library (we generate one options type that get's passed in to the aggregate client ctor and into the sub clients) which is not true, in the scope of the same library multiple clients can have different API versions. So we are temporary not passing the version in so the default gets used until a better solution is designed.

MiYanni:
Makes sense. Most SDKs are composed of multiple services, each with their own API versions. So, we'd need to have separate client options. Would also make sense to make client represent a service instead of an operation group at that point too.

PR #715, opened 2020-04-29
Fix search reference to a commit

pakrym:


pakrym:
Merging to unblock build.

PR #719, opened 2020-04-30
Support additional types in WriteObjectValue

christothes:
Tables rest client attempts to serialize a few additional types that were not currently handled.


MiYanni:
The hook for reporting success seems to be broken. Just look at the little dot beside the commit number. When that is green, feel free to merge.

PR #722, opened 2020-05-01
Move scope creation into main client

pakrym:
The concept belongs to the main client as scopes correspond to main client method calls.

### DIFF
--- a/src/assets/Azure.Core.Shared/ContentTypeUtilities.cs
+++ b/src/assets/Azure.Core.Shared/ContentTypeUtilities.cs
@@ -8,13 +8,7 @@ using System.Text;
 
 namespace Azure.Core.Pipeline
 {
-#if !TESTFRAMEWORK
     internal static class ContentTypeUtilities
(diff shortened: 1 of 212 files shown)


src/assets/Azure.Core.Shared/ContentTypeUtilities.cs
```diff
@@ -8,13 +8,7 @@
 
 namespace Azure.Core.Pipeline
 {
-#if !TESTFRAMEWORK
     internal static class ContentTypeUtilities
-#else
-#pragma warning disable SA1649 // File name should match first type name
-    internal static class TestFrameworkContentTypeUtilities
-#pragma warning restore SA1649 // File name should match first type name
-#endif
```
MiYanni, line 17: Is this no longer an issue?

pakrym: Nope, we did a bit of TestFramework refactoring.

Issue #721, opened 2020-05-01
Enable 'is a duplicate'/'duplicate object schemas' smoke tests


MiYanni:
There is a bug with AutoRest's duplicate handling. Once that is resolved, re-enable these smoke tests.

MiYanni:
Closing this issue in favor of using the M4's issue for tracking: https://github.com/Azure/autorest.modelerfour/issues/257

PR #723, opened 2020-05-01
Update samples and smoke tests

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/709

- UpdateSmokeTests.ps1 gets the latest commit hash from HEAD for the smoke test files
- UpdateSamples.ps1 is a little more complicated
  - Scans the readmes for URLs that have `https://github.com`
  - Will determine the repo to get the commit hashes
  - If a comment in the readme like `#branch:someBranchName` exists, it will use that as a branch for all links
    - This means, for now, you cannot have multiple json files from different branches. We don't have a scenario for this yet, so no worries for now.
    - If it doesn't have this comment, it just uses HEAD

Both scripts use `https://github.com` links. The `raw.githubusercontent.com` links won't work for these because I depend on the `/blob` portion of the URL. This could be changed if we need to support it, but it just adds complication. Plus, I'd like to have consistent URLs, and also the `https://github.com` links get to a GitHub page and not the raw file, which allows links, and comments, and all the fun GitHub stuff.

4 tests in the smoke tests were skipped for this: https://github.com/Azure/autorest.csharp/issues/721
We can re-enable them when we get the AutoRest fixes for duplicates.


samples/Azure.Storage.Tables/readme.md
```diff
@@ -6,9 +6,9 @@
 ``` yaml
 title: Azure.Storage.Tables
 require: $(this-folder)/../../readme.md
-# https://github.com/shurd/azure-rest-api-specs/tree/tablesSwagger/specification/cosmos-db/data-plane/Microsoft.TablesStorage/preview/2018-10-10
+#branch:tablesSwagger
```
MiYanni, line 9: Since the commit-based URLs do not have branch information, we needed another mechanism to have the branch name, or else I have no idea how to get the latest commit hash from that branch.

samples/UpdateSamples.ps1
```diff
@@ -0,0 +1,32 @@
...
+            $blobIndex = $lineSplit.IndexOf('blob')
+            $oldHash = $lineSplit[($blobIndex + 1)]
+
+            $org = $lineSplit[($blobIndex - 2)]
+            $repo = $lineSplit[($blobIndex - 1)]
+            $gitLink = "$gitHubUrl/$org/$repo.git"
+            $branchList = (git ls-remote $gitLink)
+            if($branch)
+            {
+                $branchList = $branchList | Where-Object { $_ -like "*$branch*" }
```
MiYanni, line 21: This isn't full-proof since a branch name could exist within multiple branches. It'll end up taking the first branch that matches the search. But, I don't feel the need to make this more complex unless there is a specific issue with it.

PR #737, opened 2020-05-06
Unify type definition logic

pakrym:
Move all common type resolution logic to a base class.

Side effect is that now all generated types can be renamed/changed accessibility including LROs and header types


test/TestProjects/TypeSchemaMapping/Generated/MainRestClient.cs
```diff
@@ -57,53 +57,29 @@ internal HttpMessage CreateOperationRequest(CustomizedModel body)
...
-                        if (document.RootElement.ValueKind == JsonValueKind.Null)
-                        {
-                            value = null;
-                        }
-                        else
-                        {
-                            value = CustomizedModel.DeserializeCustomizedModel(document.RootElement);
-                        }
-                        return Response.FromValue(value, message.Response);
-                    }
```
MiYanni, line 79: The response type changed. Is it because it is an LRO now?

pakrym: Yes, I turned it into an LRO to exercise in a test.

Issue #734, opened 2020-05-06
Handle multipart/form-data


MiYanni:
Awaiting `multipart/form-data` implementation: https://github.com/Azure/autorest.modelerfour/issues/28

Related: https://github.com/Azure/autorest.csharp/issues/306
Related: https://github.com/Azure/autorest.csharp/issues/200
Related: https://github.com/Azure/autorest.csharp/issues/157

pakrym:
Dup https://github.com/Azure/autorest.csharp/issues/662 ?

MiYanni:
Yeah. That is the same issue. Currently, I'm marking all the smoke tests with this issue number, so I'll do the same for that one. There are a bunch that fail because of this.

Issue #733, opened 2020-05-06
"Invalid JSON pointer" smoke test failures


MiYanni:
Here is an example:
```
Error: Invalid JSON pointer: https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/cognitiveservices/data-plane/Common/Parameters.json/parameters/GlobalEndpoint
```

pakrym:
https://github.com/Azure/azure-rest-api-specs/issues/8859

PR #738, opened 2020-05-07
Smoke tests complete list

MiYanni:
Made a script that gets the list of readmes from the specs repo (local GIT clone). Then, I took that list and made it into our smoke test text document list. Lastly, I ran the smoke tests, found the readmes that fail, marked those, skipped them, and ran until I had a working set.

There are a few distinct failures, with a few random issues from the prior list still in the document, and a catch-all for some 1-off issues:
- https://github.com/Azure/autorest.csharp/issues/721
- https://github.com/Azure/autorest.csharp/issues/733
- https://github.com/Azure/autorest.csharp/issues/734
- https://github.com/Azure/autorest.csharp/issues/735

We can 'bubble up' some of the failures to M4 issues, and make sure they get P0 priority on the M4 repo. As those M4 issues are resolved, we can now use our repo to test those resolutions too. Additional validation is always nice to have. 😁


eng/GetSpecsList.ps1
```diff
@@ -0,0 +1,7 @@
+# Navigate to your local specs repo Git clone and run this script from there
+$repoPath = [regex]::Escape((Get-Location).Path)
+$partialPaths = Get-ChildItem -Path 'specification' -File -Filter 'readme.md' -Recurse |
+    Where-Object { $_.Name -clike 'readme.md' } |
+    ForEach-Object { $_.FullName -replace "^$repoPath", '' }
+$filePath = Join-Path $PSScriptRoot 'SpecsList.txt'
+$partialPaths | Out-File -FilePath $filePath
```
pakrym, line 7: This script doesn't seem to update the SmokeTestInputs.txt, does it have to be unified manually?

pakrym:
Let's move all the issues to appropriate repos and reference them there. They are not actionable in this repo.

eng/GetSpecsList.ps1
MiYanni: Yes. It simply gets the relative path to the readmes. Currently, it isn't worth the effort to make something more complicated since our ignore strategy with the `#` stuff is there... You'd have to, for every line you wanted to add, see if the file already exists in the SmokeTestInputs and only add new lines; not to mention if you wanted them in the correct (order) position in the file. Next time we decide to update the list, hopefully there is less stuff ignored, and we can just blow out the list entirely, and re-add the ignored files by hand. At that point, it would be worth making a single-script solution.

MiYanni:
I'm discussing with David on how to handle the issues. I'll update the PR once we figure out the plan.

Issue #717, opened 2020-04-30
[P0] ManagementPipelineBuilder should allow to specify scope for different Azure instance


erich-wang:
```
    internal static class ManagementPipelineBuilder
    {
        public static HttpPipeline Build(TokenCredential credential, ClientOptions options)
        {
            return HttpPipelineBuilder.Build(options, new BearerTokenAuthenticationPolicy(credential, "https://management.azure.com//.default"));
        }
    }
```
suggest to change to:
```
    internal static class ManagementPipelineBuilder
    {
        public static HttpPipeline Build(TokenCredential credential, ClientOptions options, string scope = "https://management.azure.com//.default")
        {
            //TODO: check scope
            return HttpPipelineBuilder.Build(options, new BearerTokenAuthenticationPolicy(credential, scope));
        }
    }
```

## IMPACT
Currently implementation doesn't allow specify scope for azure environment other than azure global instance, it makes it useless for other azure environment.

Issue #688, opened 2020-04-23
[P0] xxxClient should allow user to set host in the constructor


erich-wang:
**Describe the issue or request**
Currently xxxClient doesn't expose host in the constructor and the default value is "https://management.azure.com"', however the value of host is different in different environment, we should expose host in the constructor.

Issue #704, opened 2020-04-28
[P1] InstrumentClient xxxMangementClient doesn't work for underlying GetxxxClient()


erich-wang:
This is test related.
InstrumentClient(xxxManagementClient) doesn't work for underlying xxxClient, we have to call InstrumentClient(xxxClient) for each entity.

MiYanni:
@erich-wang Can you explain the issue a bit more? I don't understand what the problem is here.

erich-wang:
Basically in track 2 we have to write many redundant code for instrumenting each entity, e.g. Vaults, PrivateEndpointConnections:

In track 1
```csharp
public partial class KeyVaultManagementClient : ServiceClient<KeyVaultManagementClient>, IKeyVaultManagementClient, IAzureClient
{
        public virtual IVaultsOperations Vaults { get; private set; }
        public virtual IPrivateEndpointConnectionsOperations PrivateEndpointConnections { get; private set; }

        ...
}

//test code
KeyVaultManagementClient  client = MockContext.GetServiceClient<KeyVaultManagementClient>();
client.Vaults.Getxxx(); //client.Vaults is mocked already since its parent client is mocked
client.PrivateEndpointConnections.Getxxx(); 
```

In track 2 
```csharp
public class KeyVaultManagementClient
{
        public VaultsClient GetVaultsClient(){}
        public PrivateEndpointConnectionsClient GetPrivateEndpointConnectionsClient(){}
}

//test code
internal KeyVaultManagementClient GetKeyVaultManagementClient(TestRecording recording = null)
        {
            recording = recording ?? Recording;

            return InstrumentClient(new KeyVaultManagementClient(this.subscriptionId,
                TestEnvironment.Credential,
                recording.InstrumentClientOptions(new KeyVaultManagementClientOptions(_serviceVersion))));
        }

//Expected 
GetKeyVaultManagementClient().GetVaultsClient().Getxxx();
GetKeyVaultManagementClient().GetPrivateEndpointConnectionsClient().Getxxx();

//Actual
        internal VaultsClient GetVaultsClient(TestRecording recording = null)
        {
            recording = recording ?? Recording;

            return InstrumentClient(new VaultsClient(this.subscriptionId,
                TestEnvironment.Credential,
                recording.InstrumentClientOptions(new KeyVaultManagementClientOptions(_serviceVersion))));
        }

        internal VaultsClient GetPrivateEndpointConnectionsClient(TestRecording recording = null)
        {
[...]

PR #747, opened 2020-05-11
Allow passing host into management clients

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/688
Fixes: https://github.com/Azure/autorest.csharp/issues/717

We would definitely need some strongly typed models for these clients soon but I want to figure out more of the design first.

Make Get**Client method virtual so we can implement https://github.com/Azure/autorest.csharp/issues/704


src/AutoRest.CSharp.V3/AutoRest/Plugins/ManagementClientWriter.cs
```diff
@@ -48,13 +47,44 @@ public static void WriteAggregateClient(CodeWriter writer, BuildContext context)
...
+                            continue;
+                        }
+
+                        writer.Append($"{parameter.Name},");
+                    }
+
+                    writer.Line($"tokenCredential, options)");
+                    using (writer.Scope())
+                    {
+                    }
```
MiYanni, line 79: Lol. That's fun.

Issue #741, opened 2020-05-11
[P0] Changing from DateTime to DateTimeOffset causes error


erich-wang:
During updating EventHub, the return value is "0001-01-01T00:00:00", DateTime in track1 could handle it correctly, however DateTimeOffset in track 2 throw exception:
The UTC representation of the date '0001-01-01T00:00:00' falls outside the year range 1-9999.

pakrym:
What call is returning `0001-01-01T00:00:00` value?

erich-wang:
The corresponding swagger json is [here](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/eventhub/resource-manager/Microsoft.EventHub/stable/2017-04-01/eventhubs.json#L88), the response content is listed below FYI:


"responseBody": "{\"id\":\"/subscriptions/c9cbd920-c00c-427c-852b-8aaf38badaeb/resourceGroups/Default-EventHub-8739/providers/Microsoft.EventHub/namespaces/sdk-Namespace-2154/eventhubs/sdk-EventHub-1866\",\"name\":\"sdk-EventHub-1866\",\"type\":\"Microsoft.EventHub/Namespaces/EventHubs\",\"location\":\"Australia East\",\"properties\":{\"messageRetentionInDays\":5,\"partitionCount\":4,\"status\":\"Active\",\"**createdAt\":\"0001-01-01T00:00:00\"**,\"updatedAt\":\"0001-01-01T00:00:00\",\"partitionIds\":[],\"captureDescription\":{\"enabled\":true,\"encoding\":\"Avro\",\"destination\":{\"name\":\"EventHubArchive.AzureBlockBlob\",\"properties\":{\"storageAccountResourceId\":\"/subscriptions/c9cbd920-c00c-427c-852b-8aaf38badaeb/resourcegroups/Default-EventHub-8739/providers/Microsoft.Storage/storageAccounts/testingsdkeventhub88\",\"blobContainer\":\"container\",\"archiveNameFormat\":\"{Namespace}/{EventHub}/{PartitionId}/{Year}/{Month}/{Day}/{Hour}/{Minute}/{Second}\"}},\"intervalInSeconds\":130,\"sizeLimitInBytes\":10485900,\"skipEmptyArchives\":true}}}"

PR #754, opened 2020-05-12
Update test server version

pakrym:
Requires: https://github.com/Azure/autorest.testserver/pull/179
Fixes: https://github.com/Azure/autorest.csharp/issues/741
Fixes: https://github.com/Azure/autorest.csharp/issues/753


src/AutoRest.CSharp.V3/Generation/Writers/LongRunningOperationWriter.cs
```diff
@@ -34,7 +34,7 @@ void WriteResultFunction(bool async)
                 {
                     if (async)
                     {
-                        writer.Line($"return await new {typeof(ValueTask<Response>)}({responseVariable});");
+                        writer.Line($"return await new {typeof(ValueTask<Response>)}({responseVariable}).ConfigureAwait(false);");
```
pakrym, line 37: We really need to enable those analyzers..

PR #761, opened 2020-05-13
Add support for generator auto updating

pakrym:


src/assets/Generator.Shared/CodeGenTypeAttribute.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#nullable enable
```
pakrym, line 4: Sync shared sources with azure-sdk-for-net

eng/PublishRelease.ps1
```diff
@@ -21,6 +21,8 @@ try
    Write-Host "Publishing $file"
 
    cmd /c "npx -q publish-release --token $Token --repo autorest.csharp --owner azure --name $name --tag $devVersion --notes=prerelease-build --prerelease --editRelease false --assets $file --target_commitish $Sha 2>&1"
+
+   Write-Host "##vso[task.setvariable variable=AutorestCSharpVersion;isSecret=false]https://github.com/Azure/autorest.csharp/releases/download/$devVersion/autorest-csharp-v3-$devVersion.tgz"
```
weshaggard, line 25: Should we reduce this to just the version number?

eng/pipelines/build.yml
```diff
@@ -26,27 +39,48 @@ jobs:
...
           dotnet test AutoRest.CSharp.V3.sln
         displayName: "Test"
         env:
           DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
           DOTNET_CLI_TELEMETRY_OPTOUT: 1
           DOTNET_MULTILEVEL_LOOKUP: 0
+        workingDirectory: $(Build.SourcesDirectory)/autorest.csharp
       - ${{if and(ne(variables['Build.Reason'], 'PullRequest'), eq(variables['System.TeamProject'], 'internal'))}}:
+        - checkout: azure-sdk-for-net
         - pwsh: ./eng/PublishRelease.ps1 -WorkingDirectory ./artifacts/bin/AutoRest.CSharp.V3/Debug/netcoreapp3.0/ -PackageJson ./src/AutoRest.CSharp.V3/package.json -Token $(azuresdk-github-pat) -Artifacts $(Build.ArtifactStagingDirectory) -BuildNumber $(Build.BuildNumber) -Sha $(Build.SourceVersion) -CoverageUser $(storage-coverage-user) -CoveragePass $(storage-coverage-pass)
```
weshaggard, line 64: I'd suggest folding these line to help with readability. You can fold it by adding a backtick at the end or you can use yml folding by using `>` (https://yaml-multiline.info/)

eng/pipelines/build.yml
```diff
@@ -26,27 +39,48 @@ jobs:
...
           condition: succeeded()
           failOnStderr: false
           displayName: 'Publish to GitHub and coverage report'
+          workingDirectory: $(Build.SourcesDirectory)/autorest.csharp
```
weshaggard, line 68: You can put this in another job if you want to avoid the multi-checkout requiring an updated working directory for everything else.

eng/pipelines/build.yml
```diff
@@ -26,27 +39,48 @@ jobs:
...
+        - pwsh: ./eng/UpdateAzureSdkForNet.ps1 -Version $(AutorestCSharpVersion) -SdkRepoRoot $(Build.SourcesDirectory)/azure-sdk-for-net
+          condition: succeeded()
```
weshaggard, line 70: condition succeeded is the default so you don't have to explicitly specify it.

eng/PublishRelease.ps1
pakrym: For now I need to pass the entire path to autorest because we run from github releases. I tried to figure out how to use version only but didn't find an easy way yet. Maybe next time.

eng/pipelines/build.yml
pakrym: I think I can live with this, it's not too bad.

eng/PublishRelease.ps1
weshaggard: Can we not just pass the version and format the github release stuff around the version in the targets file?

pakrym: What does that improve? At least this way all the logic is part of this repo.

weshaggard: I guess nothing really beyond the PRTitle being a little long and scary, but it looks like you already removed the version from that so maybe nothing to see here :)

Issue #766, opened 2020-05-14
The parameter name `host` RPManagementClient.cs ctor is too generic


allenjzhang:
With name like `host` could led someone to think that should it be just the host name. It is unclear to caller whether it is a full URL and whether it should include scheme "https://" or not. Also what if some env URL that may have segments. 

I think indicating it is an Endpoint or URL seems more fitting. What about these suggestions. 
          * Uri managementEndpoint

pakrym:
renamed to `endpoint`

ArthurMa1978:
After re-design this has been fixed

PR #767, opened 2020-05-14
Use Uri for host

pakrym:
Rename the default parameter that autorest generated.

Adding to https://github.com/Azure/autorest.csharp/issues/766
Fixes https://github.com/Azure/autorest.csharp/issues/746


src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -325,6 +325,15 @@ private void WriteConstantOrParameter(CodeWriter writer, ReferenceOrConstant con
 
         private void WritePathSegment(CodeWriter writer, CodeWriterDeclaration uri, PathSegment segment, string? methodName = null)
         {
+            if (segment.Value.Type.IsFrameworkType &&
+                segment.Value.Type.FrameworkType == typeof(Uri))
+            {
+                writer.Append($"{uri}.Reset(");
+                WriteConstantOrParameter(writer, segment.Value, enumAsString: !segment.IsRaw);
+                writer.Line($");");
```
MiYanni, line 333: Does this work for every situation? What if there are multiple Uri segments?

pakrym: Multiple Uri segment wouldn't make a lot of sense because all of them would have to be well formatted Uris.

MiYanni: If we don't support relative URIs, that make sense.

Issue #739, opened 2020-05-07
Need to be able to override target namespace and accessibility


heaths:
As decided across languages, in Search we need to move all (or at most) models to a new namespace, i.e. Azure.Search.Documents.*Indexes*.Models. These models are all in a single swagger (we have two), but the old generator way of changing the target namespace (`..x-ms-code-generation-settings.namespace`) doesn't work. Having something we can define on the models via transform would be helpful, as creating custom models with the `[CodeGenModel]` attribute is not only a large effort (hence the label assigned to this issue), but untenable since new models are added with some frequency. We want them to go to the right place by default to avoid more customization work when not otherwise needed.

@pakrym also said we could take this opportunity to change the accessibility modifier as well.

This might also apply to operations that would go into Azure.Search.Documents.Indexes; however, there's only a few and easier to move. That said, the hand-written clients for Search we're still keeping in the main, i.e. Azure.Search.Documents, namespace, but we don't expose the generated operations directly.

Also, not *all* models will necessarily move. `[CodeGenModel]` support should ignore the intended namespace, i.e. if we put a custom model with a matching source name in Azure.Search.Documents.Models that should be in Azure.Search.Documents.Models.Indexes.Models otherwise, the namespace should be Azure.Search.Documents.Models.

heaths:
Alternatively, if not temporarily, Search could change the default project namespace since it would be easier to redefine the 18 models from searchindex.json than the (currently, but growing) 152 models from searchservice.json. This does add some complexity to development, though, since any new code will default to that namespace and is something we'll have to watch very carefully in review. It's not ideal. Changing something like https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp.V3/Output/Models/Types/ObjectType.cs#L45-L48 to handle this would be a more scalable solution.

Issue #451, opened 2020-02-25
Add support for making a batch of generated types internal


annelo-msft:
I would like to make most of the Form Recognizer generated types internal, with the ability to opt-out for a subset of files.  Right now, the way to make a class internal is to create a new file with a partial class and the attribute [CodeGenSchema] on it.  If I did this for most classes in my solution, it would be many files.  It would save time if I could specify in the autorest.md or elsewhere the files I wanted to make internal.

PR #768, opened 2020-05-14
Read Namespace and accessibility from swagger input

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/451
Fixes: https://github.com/Azure/autorest.csharp/issues/739


src/AutoRest.CSharp.V3/Output/Models/Client.cs
```diff
@@ -30,6 +30,7 @@ public Client(OperationGroup operationGroup, BuildContext context): base(context
             var clientName = clientPrefix + ClientSuffix;
 
             DefaultName = clientName;
+            DefaultAccessibility = operationGroup.Extensions?.Accessibility ?? "public";
```
heaths, line 33: It would be nice - though not a big deal - if we could move these as well. While we won't expose them in Search directly anyway, I can imagine some clients might expose the generated ones and may not need/want this.

pakrym: Leaving this was a mistake. Unfortunately there is no single place in swagger that represents a "client" so there is nothing to apply namespace to, I'll revert this file and keep the feature about model for now.

heaths: Ah, yes, that's a good point.

Issue #765, opened 2020-05-14
RPManagementClient's XML documentation for ctor should have <param>


allenjzhang:
There should be <param> documentation for the ctor. Particularly with the very generic name **host** being used.

`       
        /// <summary> Initializes a new instance of ComputeManagementClient. </summary>
        public ComputeManagementClient(string host, string subscriptionId, TokenCredential tokenCredential, ComputeManagementClientOptions options = null)
        {
            _options = options ?? new ComputeManagementClientOptions();
            _tokenCredential = tokenCredential;
            _host = host;
            _subscriptionId = subscriptionId;
        }

`

allenjzhang:
Let me add something more on this. The XML method doc should also list exceptions thrown by the lib.

PR #772, opened 2020-05-15
Documentation: Client Constructors

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/765

I added parameters to all of them, and the ArgumentNullException (if applicable) to the REST client and aggregate management client.


src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -65,6 +61,17 @@ private void WriteClientFields(CodeWriter writer, RestClient restClient)
         private void WriteClientCtor(CodeWriter writer, RestClient restClient, CSharpType cs)
         {
             writer.WriteXmlDocumentationSummary($"Initializes a new instance of {cs.Name}");
+            writer.WriteXmlDocumentationParameter("clientDiagnostics", "The handler for diagnostic messaging in the client.");
```
pakrym, line 64: Let's extract these strings somewhere.

src/AutoRest.CSharp.V3/Generation/Writers/ManagementClientWriter.cs
```diff
@@ -80,6 +90,17 @@ public static void WriteAggregateClient(CodeWriter writer, BuildContext context)
                     }
 
                     writer.WriteXmlDocumentationSummary($"Initializes a new instance of {title}ManagementClient");
+                    foreach (Parameter parameter in allParameters.Values)
+                    {
+                        writer.WriteXmlDocumentationParameter(parameter.Name, parameter.Description);
+                    }
+                    writer.WriteXmlDocumentationParameter("tokenCredential", "The OAuth token for making client requests.");
+                    writer.WriteXmlDocumentationParameter("options", "The options for configuring the client.");
+                    if (allParameters.Values.HasAnyNullCheck())
```
pakrym, line 99: I wonder if we should just hardcode this to always be printed

MiYanni: Since it is based on the number of parameters in a constructor, there won't ever be more than a handful of parameters. So, the check for this is very fast. I think when there would be multiple checks like this, you'd need a different strategy. But, having one for now shouldn't be an issue.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
MiYanni: Done.

Issue #773, opened 2020-05-15
Disable public client generation for data plane libraries


pakrym:
Considering diagnostic scopes are not useful.

PR #774, opened 2020-05-15
Exclude public clients by default

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/773


pakrym:
> Right now, management-plane would both need azure-arm: true and public-clients:true.

`azure-arm:` defaults public-clients to true because otherwise nothing would work.

Issue #776, opened 2020-05-18
Don't generate error types


pakrym:
We shouldn't be generating CloudError&CloudErrorBody types.

pakrym:
Requires https://github.com/Azure/autorest.modelerfour/

pakrym:
We can also do the analysis on C# generator side.

PR #780, opened 2020-05-20
Remove exception models

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/776


test/TestServerProjects/azure-parameter-grouping/Generated/Models/Error.Serialization.cs
```diff
@@ -5,39 +5,9 @@
...
 #nullable disable
 
-using System.Text.Json;
-using Azure.Core;
-
 namespace azure_parameter_grouping.Models
 {
     public partial class Error
     {
-        internal static Error DeserializeError(JsonElement element)
```
MiYanni, line 15: So, I'd have to change it so that, when exceptions are generated, we don't generate this file, right? Because, we aren't using the deserializer.

pakrym: Hm, let's generate the deserializer instead.

src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -118,7 +118,8 @@ public async Task<bool> Execute(IPluginCommunication autoRest)
                 new Uri(GetRequiredOption(autoRest, "shared-source-folder")).LocalPath,
                 autoRest.GetValue<bool?>("save-inputs").GetAwaiter().GetResult() ?? false,
                 autoRest.GetValue<bool?>("azure-arm").GetAwaiter().GetResult() ?? false,
-                autoRest.GetValue<bool?>("public-clients").GetAwaiter().GetResult() ?? false
+                autoRest.GetValue<bool?>("public-clients").GetAwaiter().GetResult() ?? false,
+                autoRest.GetValue<bool?>("include-unused").GetAwaiter().GetResult() ?? false
```
pakrym, line 122: I thought we decided not to add a flag.

test/TestServerProjects/azure-parameter-grouping/Generated/Models/Error.Serialization.cs
MiYanni: This is a bug. So, I'll fix the bug.

src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
MiYanni: Yes. After seeing what it does, I'm going to change it to do `internal` by default for any unused models (either no `usage` info or `usage: exception`). However, it uncovered 1 bug testing this (not the deserializer one, a different one), so I'm fixing that.

test/TestServerProjects/azure-parameter-grouping/Generated/Models/Error.Serialization.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
MiYanni: Fixed.

src/AutoRest.CSharp.V3/readme.md
```diff
@@ -4,7 +4,7 @@
 ## Configuration
 ```yaml
 use-extension:
-  "@autorest/modelerfour": "4.13.358"
+  "@autorest/modelerfour": "http://tinyurl.com/y6vmn7dq"
```
pakrym, line 7: Make sure to change this.

pakrym:
Add a test for the new behavior.

MiYanni:
Added tests. Merging.

PR #784, opened 2020-05-21
Support JsonElement type in serialization

pakrym:


readme.md
```diff
@@ -263,6 +264,51 @@ namespace Azure.Service.Models
...
+<details>
+
+**Generated code before (Generated/Models/Model.cs):**
+
+``` C#
+namespace Azure.Service.Models
+{
+    public partial class Model
+    {
+        public string Property { get; }
```
drwill-ms, line 280: If the property, parameter, or return value is typed as an object instead of a string, will this work as well?

pakrym: This is only for properties for now but yes, it would work with any type of original property.

PR #789, opened 2020-05-26
Use Task return types

pakrym:


pakrym:
Didn't change, the guidance is to use Task except for certain situations.

Issue #787, opened 2020-05-25
[P0] Incorrect datetime format in request


erich-wang:
**Describe the issue or request**
One change in [TypeFormatters.cs](https://github.com/Azure/azure-sdk-for-net/pull/11860/files#diff-3d9ef10043ed0ff2a49101184a3322c4) causes date format change and leads to error: **Values for request parameters are invalid: signedExpiry, signedStart.**

Before change which worked well:
![p1](https://user-images.githubusercontent.com/20198527/82830299-e8e24b80-9ee7-11ea-8628-dd8f8de412d5.png)
After change:
![p2](https://user-images.githubusercontent.com/20198527/82830716-1380d400-9ee9-11ea-90d7-636caff9a619.png)

pakrym:
These are equivalent values by specification, what service is this?

giuliohome:
Hi all.
There is still a problem, we have to use

```go
currentTime.Format("2006-01-02T03:04:05.9999999Z")
```
instead of
```go
currentTime.Format(time.RFC3339Nano)
```
because now the format shows 8 decimal digits instead of 7

PR #791, opened 2020-05-26
Revert to Z suffix for 00:00 offset for compat reasons

pakrym:
Some services don't like `+00:00` and prefer the `Z` suffix, it's allowed by the spec so add a special case.

Also fix unix time writing.

Fixes https://github.com/Azure/autorest.csharp/issues/787

### DIFF
--- a/src/assets/Generator.Shared/TypeFormatters.cs
+++ b/src/assets/Generator.Shared/TypeFormatters.cs
@@ -4,6 +4,7 @@
 #nullable enable
 
 using System;
+using System.Diagnostics;
 using System.Globalization;
 using System.Xml;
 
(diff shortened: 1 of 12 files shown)


src/assets/Generator.Shared/TypeFormatters.cs
```diff
@@ -19,6 +20,7 @@ internal class TypeFormatters
         {
             "D" => value.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture),
             "U" => value.ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture),
+            "O" when value.Offset == TimeSpan.Zero => value.ToString("yyyy-MM-ddTHH:mm:ss.fffffffZ", CultureInfo.InvariantCulture),
```
christothes, line 23: Do we need to account for casing here? Both "O" and "o" are valid.

pakrym: yep, done.

Issue #786, opened 2020-05-25
[P0] Status code 204 should be valid response when deleting VirtualMachineScaleSets


erich-wang:
**Describe the issue or request**
Based on the [swagger spec](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/compute/resource-manager/Microsoft.Compute/stable/2019-12-01/compute.json#L3673), 204 is valid response for delete operation, however the generated code doesn't treat 204 as valid, see below or [code](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/compute/Azure.Management.Compute/src/Generated/VirtualMachineScaleSetsRestClient.cs#L253)
```csharp
        public async ValueTask<Response> DeleteAsync(string resourceGroupName, string vmScaleSetName, CancellationToken cancellationToken = default)
        {
            if (resourceGroupName == null)
            {
                throw new ArgumentNullException(nameof(resourceGroupName));
            }
            if (vmScaleSetName == null)
            {
                throw new ArgumentNullException(nameof(vmScaleSetName));
            }

            using var message = CreateDeleteRequest(resourceGroupName, vmScaleSetName);
            await _pipeline.SendAsync(message, cancellationToken).ConfigureAwait(false);
            switch (message.Response.Status)
            {
                case 202:
                case 200:
                    return message.Response;
                default:
                    throw await _clientDiagnostics.CreateRequestFailedExceptionAsync(message.Response).ConfigureAwait(false);
            }
        }
```

PR #792, opened 2020-05-26
Return all Responses for LRO processing

MiYanni:
Fixes: https://github.com/Azure/autorest.csharp/issues/786

This change seems okay... but I'm 100% confident of it. Also, added Azure.Management.Compute to samples for testing/debugging purposes. I can remove it if we don't want it.


MiYanni:
Added test server issue: https://github.com/Azure/autorest.testserver/issues/183

Issue #796, opened 2020-06-02
Overriding collection properties makes enum generic type nullable


heaths:
For Azure/azure-sdk-for-net#11712 I moved collection properties to a customized class with `[CodeGenMember(EmptyAsUndefined = true, Initialize = true)]`; however, this now causes the generator to declare the internal serialization constructor as `IEnumerable<T?>` - nullable, despite these types being non-nullable:

```json
      "properties": {
        "ignoreScripts": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/CjkBigramTokenFilterScripts",
            "x-nullable": false
          },
          "description": "The scripts to ignore."
        },
      }
```

This ends up looking like:

```csharp
        internal CjkBigramTokenFilter(string oDataType, string name, IList<CjkBigramTokenFilterScripts?> ignoreScripts, bool? outputUnigrams) : base(oDataType, name)
        {
            IgnoreScripts = ignoreScripts ?? new List<CjkBigramTokenFilterScripts>();
            OutputUnigrams = outputUnigrams;
            ODataType = oDataType ?? "#Microsoft.Azure.Search.CjkBigramTokenFilter";
        }
```

`CjkBigramTokenFilterScripts` is an enum with `x-nullable` as `false`, and even instantiation of a list gets it right, but the parameter should make the type nullable - the parameter itself could be, but since it's a reference type it naturally is already.

I can work around the problem by changing my declaration, but that is not the shape of the API we want. As such, this is blocking.

heaths:
Note, the de/serialization code (in *FileName*.Serialization.cs) is also expecting it to be nullable, even though the array value is not nor was it when the generator was generating the properties:

```csharp
        void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
        {
            writer.WriteStartObject();
            if (IgnoreScripts != null && IgnoreScripts.Any())
            {
                writer.WritePropertyName("ignoreScripts");
                writer.WriteStartArray();
                foreach (var item in IgnoreScripts)
                {
                    writer.WriteStringValue(item.Value.ToSerialString());
                }
                writer.WriteEndArray();
            }

// ...

       internal static CjkBigramTokenFilter DeserializeCjkBigramTokenFilter(JsonElement element)
        {
            IList<CjkBigramTokenFilterScripts?> ignoreScripts = default;
            bool? outputUnigrams = default;
            string odataType = default;
            string name = default;
            foreach (var property in element.EnumerateObject())
            {
[...]

PR #797, opened 2020-06-02
Read nullability of value types correctly

pakrym:
Fixes https://github.com/Azure/autorest.csharp/issues/796


test/TestProjects/TypeSchemaMapping/TypeSchemaMapping.json
```diff
@@ -191,6 +192,25 @@
...
       "x-namespace": "Very.Custom.Namespace.From.Swagger",
       "x-accessibility": "internal"
+    },
+    "ModelWithArrayOfEnum": {
+      "type": "object",
+      "properties": {
+        "ArrayOfEnum": {
+          "type": "array",
+          "items": {
+            "type": "string",
```
heaths, line 202: Nit: while you don't explicitly support it now, it might be good to add `"x-nullable": false` because services *should* be doing this, which was the case for Search (they did)...unless they really do want to allow for null values.

test/TestProjects/TypeSchemaMapping/Customized/ModelWithArrayOfEnum.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System.Collections.Generic;
+using NamespaceForEnums;
+
+namespace TypeSchemaMapping.Models
+{
+    /// <summary> The ModelWithArrayOfEnum. </summary>
+    internal partial class ModelWithArrayOfEnum
+    {
+        public IReadOnlyList<EnumForModelWithArrayOfEnum> ArrayOfEnum { get; }
```
heaths, line 16: I would also declare a list (or RO list) of a nullable extensible enum or something because, in my bug, I noticed that the de/serialization code didn't always consistently check for null; e.g., during write, it just called `x.Value.ToString()`, IIRC. If not set, that would fail. In the serialization code, an extensible enum was created with `null` which would actually throw. So while, based on the sample at least, it seems my blocking problem is fixed, I don't think the whole bug is resolved.

heaths: ...or maybe you or I should open another bug on *that* behavior (bad de/serialization of actual nullable types) to move past this one. In my case, none of my value types are nullable (or, more specifically, array items).

src/AutoRest.CSharp.V3/Generation/Types/TypeFactory.cs
```diff
@@ -186,37 +186,45 @@ public bool TryCreateType(ITypeSymbol symbol, [NotNullWhen(true)] out CSharpType
...
             var fullMetadataName = GetFullMetadataName(namedTypeSymbol);
             var fullyQualifiedName = $"{fullMetadataName}, {namedTypeSymbol.ContainingAssembly.Name}";
             var existingType = Type.GetType(fullMetadataName) ?? Type.GetType(fullyQualifiedName);
+
             if (existingType != null)
             {
                 var arguments = namedTypeSymbol.TypeArguments.Select(a => CreateType(a)).ToArray();
-                type = new CSharpType(existingType, nullable, arguments);
-                return true;
+                type = new CSharpType(existingType, false, arguments);
```
heaths, line 202: What if the existing type is nullable, though? Or is that reserved for when you actually support `x-nullable`?

test/TestProjects/TypeSchemaMapping/TypeSchemaMapping.json
pakrym: `"x-nullable": false` is considered the default. We'll add coverage for that when adding support.

test/TestProjects/TypeSchemaMapping/Customized/ModelWithArrayOfEnum.cs
pakrym: I'll open a new bug.

test/TestProjects/TypeSchemaMapping/TypeSchemaMapping.json
heaths: Granted, so this should break if not modified later (in swagger) but the support is added; hence, why I was suggesting it now. But either works. 🙂

test/TestProjects/TypeSchemaMapping/Customized/ModelWithArrayOfEnum.cs
pakrym: https://github.com/Azure/autorest.csharp/issues/798

src/AutoRest.CSharp.V3/Generation/Types/TypeFactory.cs
pakrym: Yes, we assign nullability later in the code.

PR #822, opened 2020-06-16
Update to Autorest M4 v4.14.371

pakrym:
Fixes: https://github.com/Azure/autorest.modelerfour/issues/255


eng/SmokeTestInputs.txt
```diff
@@ -1,7 +1,7 @@
 https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/addons/resource-manager/readme.md
 https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/adhybridhealthservice/resource-manager/readme.md
 https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/advisor/resource-manager/readme.md
-#https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/alertsmanagement/resource-manager/readme.md:--modelerfour.lenient-model-deduplication
+https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/alertsmanagement/resource-manager/readme.md:--modelerfour.lenient-model-deduplication
```
ShivangiReja, line 4: Any reason to uncomment this? Also, why did we added `:--modelerfour.lenient-model-deduplication`, shouldn't this just be `https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/alertsmanagement/resource-manager/readme.md`?

pakrym: This file is now passing with fixes from this PR and the autorest version we used.

pakrym: `:--modelerfour.lenient-model-deduplication` allows us to add additional parameters to be passed to Autorest.

The `lenient-model-deduplication` flag allows Autorest to drop duplicated model instead of issuing an error like it usually would allowing us to process not entirely correct swaggers that we have a lot of.

ShivangiReja: Ah, got it. Thanks!

Issue #820, opened 2020-06-15
Add NuGet config to the repository


pakrym:
We've seen some flaky builds that fail to restore. Wes suggested adding a `nuget.config`:

Wes Haggard
> looking through the repo I don't see a nuget.config explicitly in the repo, I'd suggest adding one to the root of the repo. I'm not sure why this fails sometimes but it is interesting that you are hitting this failure while restoring your autorest projects and it is looking at our blob feed which as far as I can tell isn't configured in that repo. I wonder if the nuget.config is leaking from somewhere else. 

> As for the blob feeds I've not seen similar issues around them so I wouldn't send much time on that front esspecailly given we are moving away from them and moving to DevOps feeds for our repos. 
    
> I'd suggest removing 
> https://github.com/Azure/autorest.csharp/blob/feature/v3/Directory.Build.props#L7 as well. If you still need a dev package from the language try adding the devops feed to your nuget.config. Like https://github.com/Azure/azure-sdk-for-net/blob/master/NuGet.Config#L8

ShivangiReja:
Closed with https://github.com/Azure/autorest.csharp/pull/823

PR #823, opened 2020-06-17
Add NuGet config and update autorest.testserver version

ShivangiReja:
Fixes - https://github.com/Azure/autorest.csharp/issues/820


Directory.Build.props
```diff
@@ -4,7 +4,6 @@
     <RepoEngPath>$(MSBuildThisFileDirectory)eng</RepoEngPath>
     <RepoSrcPath>$(RepoRoot)src</RepoSrcPath>
     <IsTestProject Condition="$(MSBuildProjectName.EndsWith('.Tests'))">true</IsTestProject>
-    <RestoreAdditionalProjectSources>https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-net/index.json</RestoreAdditionalProjectSources>
```
pakrym, line 7: This source should be added to nuget.config

NuGet.Config
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="Local" value="tools/LocalNugetFeed" />
```
pakrym, line 5: I think we can remove this one

NuGet.Config
```diff
@@ -0,0 +1,9 @@
...
+    <add key="azure-sdk-tools" value="https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json" />
+    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
```
pakrym, line 7: We don't use these ones.

Directory.Build.props
ShivangiReja: Isn't it the same as `<add key="azure-sdk-tools" value="https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json" />`?

ShivangiReja: I accidentally substituted this with `https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json`.

ShivangiReja: Yeah, I accidentally substituted this with `https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json`.

Added this to NuGet.Config.

ShivangiReja: Yeah, my bad I accidentally substituted this with `https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json`.

Added this to NuGet.Config.

NuGet.Config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="azure-sdk-for-net" value="https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-net/index.json" />
```
weshaggard, line 5: I suggest using the Azure Devops feed like .NET, see https://github.com/Azure/azure-sdk-for-net/blob/master/NuGet.Config#L8.

weshaggard: Sorry I just realized that the .NET nuget.config doesn't have the devops feed yet either. The feed I suggest using is 
```
<add key="azure-sdk-for-net" value="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json" />
```

weshaggard: .NET repo update is still in PR https://github.com/Azure/azure-sdk-for-net/pull/11916.

ShivangiReja: Fixed here - https://github.com/Azure/autorest.csharp/pull/826

PR #826, opened 2020-06-18
Update NuGet Config to pull from DevOps feed

ShivangiReja:
Fixes - https://github.com/Azure/autorest.csharp/pull/823#discussion_r441874276

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -2,6 +2,6 @@
 <configuration>
   <packageSources>
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
-    <add key="azure-sdk-for-net" value="https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-net/index.json" />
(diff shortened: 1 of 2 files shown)


pakrym:
I think the feature we needed is in a released version of azure.core now. Can you update https://github.com/azure/autorest.csharp/blob/279e810cd54ed877632f99a893ecbbb3c00fb320/test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj#L11 to 1.2.2 and drop this feed.

NuGet.Config
```diff
@@ -2,6 +2,6 @@
 <configuration>
   <packageSources>
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
-    <add key="azure-sdk-for-net" value="https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-net/index.json" />
+    <add key="azure-sdk-for-net" value="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json" />
```
pakrym, line 5:
```suggestion
```

ShivangiReja: I don't see any difference?

PR #830, opened 2020-06-29
Always initialize the collections

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/521

I've split the PR into multiple commits:

First one is the most important to review, second is resulting generated code change and the third is some follow up changes.


test/AutoRest.Shared.Tests/ChangeTrackingListTest.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            list.CopyTo(new string[5], 0);
+
+            Assert.Throws<IndexOutOfRangeException>(() => _ = list[0]);
+
+            Assert.True(list.IsUndefined);
+        }
+
+        [Test]
+        public void CanAddElement()
```
ShivangiReja, line 40: Tests are missing for some operations like GetEnumerator(), Insert(), contains()...etc

src/AutoRest.CSharp.V3/Output/Builders/SerializationBuilder.cs
```diff
@@ -179,7 +178,7 @@ private IEnumerable<JsonPropertySerialization> GetPropertySerializationsFromBag(
                 JsonPropertySerialization[] serializationProperties = GetPropertySerializationsFromBag(innerBag, objectType).ToArray();
                 JsonObjectSerialization objectSerialization = new JsonObjectSerialization(null, serializationProperties, null);
                 yield return new JsonPropertySerialization(
-                    name, false,null, objectSerialization);
+                    name,null, objectSerialization);
```
ShivangiReja, line 181: nit: indentation is off

src/assets/Generator.Shared/ChangeTrackingExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Linq;
+
+#nullable disable
+
+namespace Azure.Core
+{
+    internal static class ChangeTrackingExtensions
+    {
+    }
+}
```
ShivangiReja, line 16: Why do we have this empty class? I don't see we are using it anywhere.

test/AutoRest.Shared.Tests/ChangeTrackingListTest.cs
pakrym: GetEnumerator is covered by foreach. You are right about the others. Adding.

PR #839, opened 2020-07-15
Consume autorest header naming feature

pakrym:
Autorest now produces a "Language" node that has both readable name and description.


samples/Azure.Storage.Tables/Azure.Storage.Tables/Generated/TableInternalQueryHeaders.cs
```diff
@@ -17,7 +17,9 @@ public TableInternalQueryHeaders(Response response)
         {
             _response = response;
         }
-        public string XMsVersion => _response.Headers.TryGetValue("x-ms-version", out string value) ? value : null;
+        /// <summary> Indicates the version of the Queue service used to execute the request. This header is returned for requests made against version 2009-09-19 and above. </summary>
+        public string Version => _response.Headers.TryGetValue("x-ms-version", out string value) ? value : null;
+        /// <summary> This header contains the continuation token value. </summary>
         public string XMsContinuationNextTableName => _response.Headers.TryGetValue("x-ms-continuation-NextTableName", out string value) ? value : null;
```
ShivangiReja, line 23: After this update `XMsVersion` got renamed to `Version`. Shouldn't `XMsContinuationNextTableName` also be renamed to `ContinuationNextTableName`?

pakrym: @daviwil, do you have an idea about this namer behavior?

daviwil: This is happening because `x-ms-version` has an `x-ms-client-name` that renames it to `Version` but `x-ms-continuation-NextTableName` does not: https://github.com/Azure/azure-rest-api-specs/blob/7f1b65db7aa74b4668d000eb27c1c0ffd8d91334/specification/cosmos-db/data-plane/Microsoft.Tables/preview/2019-02-02/table.json#L84-L107.  Might be worth adding that to the Swagger!

PR #843, opened 2020-07-18
Write ArgumentNullException doc comments

### LINKED ISSUE #602: Generated members should document exceptions
For better documentation - and consistent with .NET BCL - you should document exceptions. For example, where'd you'd throw an `ArgumentNullException`, document those params. They need to be in one line, though. Last I checked, the compiler didn't automagically merge text for duplicate exception type lines (or if it does, even easier).

heaths:
Fixes #602 using a exception message consistent with the BCL.


samples/AppConfiguration/AppConfiguration/Generated/Models/Enum4.cs
```diff
@@ -16,6 +16,7 @@ namespace AppConfiguration.Models
         private readonly string _value;
 
         /// <summary> Determines if two <see cref="Enum4"/> values are the same. </summary>
+        /// <exception cref="ArgumentNullException"> <paramref name="value"/> is null. </exception>
```
pakrym, line 19: Should we add `when` or `if`?
```
<exception cref="ArgumentNullException"> if <paramref name="value"/> is null. </exception>
```


See usages in https://github.com/dotnet/runtime/search?q=exception+cref+ArgumentNullException&unscoped_q=exception+cref+ArgumentNullException

heaths: There certainly are some abnormal cases, but if you look through a smattering of BCL docs for any exception, most elide "thrown when/if" or any prefix to the condition an exception is thrown. From some regex over the runtime repo:

```powershell
$found = dir -filter *.cs -recurse | select-string '<exception cref=\"(?<type>\w+)\">(?<first>[<>\w]+)\s+(?<second>[<>\w]+)' | select @{l='Type'; e={$_.Matches.Groups[1].Value}}, @{l='FirstWord'; e={$_.Matches.Groups[2].Value}}, @{l='SecondWord'; e={$_.Matches.Groups[3].Value}}
$found | group FirstWord -NoElement | sort Count -desc -top 10
```
Results in:
```
Count Name
----- ----
  363 <paramref
  326 The
  292 if
   89 Thrown
   41 Builder
   26 When
   21 An
   20 sourceArchive
   17 unexpected
   16 Always
```

You can also sort of pivot on the first and second words, since there's a lot of "The":
```powershell
$found | group FirstWord, SecondWord -NoElement | sort Count -desc -top 10
```
Results in:
```
Count Name
----- ----
   79 The, type
   38 The, address
   23 The, <paramref
   17 The, specified
   15 The, <see
   15 The, ZipArchive
   12 The, stream
   10 The, dictionary
    9 The, path
    8 The, caller
```

That said, we could depart from the BCL.

heaths: To break down "the" and "if":
```powershell
$found | where FirstWord -eq The | group FirstWord, SecondWord -NoElement | sort Count -desc -top 10
```
```
Count Name
----- ----
   79 The, type
   38 The, address
   23 The, <paramref
   17 The, specified
   15 The, <see
   15 The, ZipArchive
   12 The, stream
   10 The, dictionary
    9 The, path
    8 The, caller
```
```powershell
$found | where FirstWord -eq If | group FirstWord, SecondWord -NoElement | sort Count -desc -top 10
```
```
Count Name
----- ----
   76 if, the
   39 If, partUri
   30 If, parameter
   30 If, this
   14 if, <paramref
    7 If, FileAccess
    7 If, FileMode
    7 If, number
    7 If, package
    6 If, array
```
So they both tend to lead into longer, more varied sentences. I don't think adding more words necessarily makes the exception condition case clearer.

heaths: @terrajobst had mentioned that for argument exceptions, they tend to favor brevity, which this pivot shows:
```powershell
$found | where Type -like 'Argument*Exception' | group FirstWord, SecondWord -NoElement | sort Count -desc -top 10
```
```
Count Name
----- ----
  345 <paramref, name
   39 If, partUri
   30 If, parameter
   22 The, <paramref
   19 Thrown, when
   19 When, <paramref
   15 Unexpected, handle
   13 If, the
   11 The, stream
    8 A, label
```
So I think it makes sense to stick with this PR as-is.

Issue #840, opened 2020-07-15
Add ability to deserialize model types that are not used as part of any operations


JoshLove-msft:
In Event Grid, there are a bunch of model types that are output only which correspond to system event payloads (the Data property of an EventGridEvent). There are no operations defined in the swagger for these types, but we want to offer code in the SDK that can deserialize these payloads into the model types. In Track 1, this [code](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/eventgrid/Microsoft.Azure.EventGrid/src/Customization/EventGridSubscriber.cs#L173) was hand-written. One issue with doing this in Track 2, is that we can't define the JsonProperty attributes to map the Json property to the model type property, as this is a NewtonSoft.Json feature.

It would be great if the code generator were able to generate this deserialization code for us. Absent that, we will need some guidance on how to define the mapping between Json property and model type property - it seems that the Json property is always just the camelCase of the model type property, but I'm not sure if this is an assumption we want to bake into the SDK. Finally, we will need a way to be able to generate public model types for all of the models. The workaround mentioned in https://github.com/Azure/autorest.csharp/pull/837 does not work for all model types (e.g. `DeviceTwinInfoProperties`).

ellismg:
Just to confirm my understanding, it would be expected that when decoding system events, we'd always use types from `System.Text.Json` to decode them right?  We wouldn't want to use the generalized `ObjectSerializer` stuff in this case, is that right?  Or is the expectation that if the user provides a custom object serializer that is also used to deserailize the data object, event for system events?

JoshLove-msft:
> Just to confirm my understanding, it would be expected that when decoding system events, we'd always use types from `System.Text.Json` to decode them right? We wouldn't want to use the generalized `ObjectSerializer` stuff in this case, is that right? Or is the expectation that if the user provides a custom object serializer that is also used to deserailize the data object, event for system events?

That is my understanding as well - `ObjectSerializer` would not be used for system events. The system schemas are generally just primitive properties so I don't see much value to exposing this ability to customers. Our hope is that we will be able to generate the system event deserialization code which will be using System.Text.Json.

kerri-lee:
[Here](https://github.com/Azure/azure-sdk-for-net/pull/13583) our current workaround for generating deserialization code.

JoshLove-msft:
[...]

PR #847, opened 2020-07-27
Allow specifying model usage

pakrym:
Via an attribute or swagger extension

Fixes: https://github.com/Azure/autorest.csharp/issues/840


test/TestProjects/TypeSchemaMapping/TypeSchemaMapping.json
```diff
@@ -282,6 +282,26 @@
...
+    },
+    "ModelWithCustomUsage": {
+      "type": "object",
+      "properties": {
+        "ModelProperty": {
+          "type": "string",
+          "description": "."
+        }
+      },
+      "x-csharp-usage": "model,input,output",
```
JoshLove-msft, line 294: So, would we need to add this in a swagger transform for each of our generated models in Event Grid?

/cc @kerri-lee

pakrym: For you it's `output, model`

pakrym: And `x-csharp-formats : json`

PR #853, opened 2020-07-30
Handle duplicated formats

pakrym:


test/TestProjects/TypeSchemaMapping/TypeSchemaMapping.json
```diff
@@ -292,7 +292,7 @@
         }
       },
       "x-csharp-usage": "model,input,output",
-      "x-csharp-formats": "xml, json"
+      "x-csharp-formats": "xml, json, json"
```
JoshLove-msft, line 295: Is this for testing that duplicates don't show up?

pakrym: Yes

PR #857, opened 2020-08-03
Add a test for turning normal enum into extensible

pakrym:


test/TestProjects/TypeSchemaMapping/Customized/NonExtensibleEnumTurnedExtensible.cs
```diff
@@ -0,0 +1,12 @@
...
+
+using Azure.Core;
+
+namespace NamespaceForEnums
+{
+    [CodeGenModel("NonExtensibleEnum2")]
+    internal partial struct NonExtensibleEnumTurnedExtensible
+    {
+    }
+}
```
ShivangiReja, line 12: The intention of adding above is to convert normal enum into extensible enum? If yes, then I don't think you needed the above in order to generate extensible enum.  I saw that before and I believe below spec will generate extensible enum even if you don't mention the above. 

```
"NonExtensibleEnum": {
      "description": "More Letters",
      "type": "string",
      "enum": [
        "A",
        "B",
        "C"
      ]
    }
```

ShivangiReja: Please let me know if I'm missing anything here.

pakrym: Yes, this test is for ability for developers to switch between generated enum kinds without changing swaggers. It's similar to how we can rename properties, change propetry types, etc.

ShivangiReja: Yeah, but what I was trying to say that `"modelAsString": false` is missing in the spec, if we don't mention `"modelAsString"` in the spec, it by default generates extensible enum. So there is no point of below:
```
[CodeGenModel("NonExtensibleEnum2")]
    internal partial struct NonExtensibleEnumTurnedExtensible
    {
    }
```

pakrym: Yep, you are right. Fixing.

Issue #831, opened 2020-07-02
Generated extensible enum property conflicting with enum name


ShivangiReja:
[This is the guideline for extensible enums](https://azure.github.io/azure-sdk/dotnet_implementation.html#dotnet-enums), where we add property name as "enum_value"+ "Value" (ex: if the enum name is `AccountType` and value is `Standard`, then we generate struct name as `AccountType` and property name as `StandardValue`). 

In Anomaly Detector spec it has enum names as `ChangePointValue` and `AnomalyValue` and enum values as `ChangePoint` and `Anomaly`. When we try to generate Anomaly Detector API, it generates struct names as `ChangePointValue` and `AnomalyValue`.   
 The problem here is that we generate property names same as struct names i.e. `ChangePointValue` and `AnomalyValue`, so c# compiler throws below error:
`Generated\Models\ChangePointValue.cs(25,30): error CS0542: 'ChangePointValue': member names cannot be the same as their enclosing type`
`Generated\Models\AnomalyValue.cs(25,30): error CS0542: 'AnomalyValue': member names cannot be the same as their enclosing type`

ex: Generated struct: 
```csharp
public readonly partial struct ChangePointValue : IEquatable<ChangePointValue>
    {
        private const string ChangePointValue = "ChangePoint";
.
.
.
}
```
Spec for Anomaly Detector: https://github.com/Azure/azure-rest-api-specs-pr/blob/f54c64405ee2f79acc6c3f7b2f33643500be8658/specification/cognitiveservices/data-plane/AnomalyDetector.EE/preview/v1.0/metric.monitor_1.20200701_openapi.v3.json

PR #859, opened 2020-08-04
Fix: Extensible enum name conflicts with property names

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/831


test/TestProjects/NameConflicts/NameConflicts.json
```diff
@@ -305,6 +305,17 @@
             "modelAsString": true
           }
         },
+        "EnumValue": {
```
pakrym, line 308: Let's add it as a separate definition and not a parameter.

src/AutoRest.CSharp.V3/Generation/Writers/ModelWriter.cs
```diff
@@ -274,14 +274,28 @@ private void WriteChoiceSchema(CodeWriter writer, EnumType schema)
 
                     foreach (var choice in schema.Values)
                     {
-                        writer.Line($"private const {schema.BaseType} {choice.Declaration.Name}Value = {choice.Value.Value:L};");
+                        if (name == $"{choice.Declaration.Name}Value")
+                        {
+                            writer.Line($"private const {schema.BaseType} {choice.Declaration.Name}_Value = {choice.Value.Value:L};");
```
pakrym, line 279: Let's add a method instead. Something like `GetValueFieldName()` and put the logic there.

pakrym: I would also append an endex to the end (EnumValue1) I think it makes it a little bit more clear about what's happening than an underscore appearing in the middle.

test/TestProjects/NameConflicts/NameConflicts.json
```diff
@@ -305,6 +305,17 @@
...
+          "type": "string",
+          "enum": [
+            "MyEnum",
+            "SecondEnumValue"
```
pakrym, line 312: Can we also add a scenario where enum member name conflicts with other member value field.


For example: `Item, ItemValue`

src/AutoRest.CSharp.V3/Generation/Writers/ModelWriter.cs
ShivangiReja: Added!

src/AutoRest.CSharp.V3/Generation/Writers/ModelWriter.cs
```diff
@@ -272,16 +272,20 @@ private void WriteChoiceSchema(CodeWriter writer, EnumType schema)
                     }
                     writer.Line();
 
+                    int index = 1;
```
pakrym, line 275: Shouldn't this index be calculated in GetValueFieldName?

ShivangiReja: Ahh, right. But now I will get conflict with other property:

image

ShivangiReja: So, do we want to remove `ItemValue1` enum value or add underscore ex: `private const string Item_Value1 = "Item";`?

pakrym: That's the point, other property should get 2 prepended

ShivangiReja: So you're saying whatever it is generating [now](https://github.com/Azure/autorest.csharp/pull/859/files#diff-e917ec1838b6d293116e7324bd7a7fc8) is correct? If yes, then why do we want to calculate index inside `GetValueFieldName`?

ShivangiReja: Could you please tell me if we have `ItemValue` enum with  "ItemValue1", "Item", "ItemValue2" members:
```
"EnumValue": {
      "type": "string",
      "enum": [
        "ItemValue1",
        "Item",
        "ItemValue2"
      ],
      "x-ms-enum": {
        "name": "ItemValue",
        "modelAsString": true
      }
    },
```

then , what would be the private property names for "ItemValue1", "Item", "ItemValue2"?

pakrym: >So, do we want to remove ItemValue1 enum value or add underscore ex: private const string Item_Value1 = "Item";?

Sorry missed this comment.

We have multiple possible conflicts here:
1. Enum item with enum name
2. Enum item with enum backing field

Where possible we should use the defined enum item name and only workaround the backing field.

So in the following sample:

```
"EnumValue": {
      "type": "string",
      "enum": [
        "ItemValue1",
        "Item",
        "ItemValue2"
      ],
      "x-ms-enum": {
        "name": "ItemValue",
        "modelAsString": true
      }
    },
```

We should have something like this.
```
 ItemValue1 - ItemValue1Value
 Item - ItemValue3 (because 1 and 2 are taken)
 ItemValue2 - ItemValue2Value
```

ShivangiReja: Thanks for the explanation!

PR #860, opened 2020-08-05
Add null check for optional parameters

ShivangiReja:
https://github.com/Azure/azure-sdk-for-net/pull/13919#discussion_r465417155

### DIFF
--- a/test/AutoRest.TestServer.Tests/Infrastructure/TestServerSession.cs
+++ b/test/AutoRest.TestServer.Tests/Infrastructure/TestServerSession.cs
@@ -104,7 +104,7 @@ namespace AutoRest.TestServer.Tests.Infrastructure
                 {
                     foreach (var expectedStub in _expectedCoverage)
                     {
-                        if (matched.Length > 0 && !matched.Contains(expectedStub, StringComparer.InvariantCultureIgnoreCase))
+                        if (!matched.Contains(expectedStub, StringComparer.InvariantCultureIgnoreCase))
                         {
                             throw new InvalidOperationException($"Expected stub {expectedStub} was not matched, matched: {string.Join(Environment.NewLine, matched)}");
                         }
--- a/test/AutoRest.TestServer.Tests/Infrastructure/TestServerSession.cs
+++ b/test/AutoRest.TestServer.Tests/Infrastructure/TestServerSession.cs
@@ -104,7 +104,7 @@ namespace AutoRest.TestServer.Tests.Infrastructure
                 {
                     foreach (var expectedStub in _expectedCoverage)
                     {
-                        if (!matched.Contains(expectedStub, StringComparer.InvariantCultureIgnoreCase))
+                        if (matched.Length > 0 && !matched.Contains(expectedStub, StringComparer.InvariantCultureIgnoreCase))
                         {
                             throw new InvalidOperationException($"Expected stub {expectedStub} was not matched, matched: {string.Join(Environment.NewLine, matched)}");
                         }
--- a/test/AutoRest.TestServer.Tests/body-complex.cs
+++ b/test/AutoRest.TestServer.Tests/body-complex.cs
@@ -58,10 +58,10 @@ namespace AutoRest.TestServer.Tests
         });
 
         [Test]
-        public Task ThrowsIfApiVersionIsNull() => Test((host, pipeline) =>
+        public void ThrowsIfApiVersionIsNull()
         {
-            Assert.Throws<ArgumentNullException>(() => new BasicClient(ClientDiagnostics, pipeline, host, null));
-        });
+            Assert.Throws<ArgumentNullException>(() => new BasicClient(ClientDiagnostics, null, null, null));
+        }
 
         [Test]
         public Task GetComplexBasicNull() => Test(async (host, pipeline) =>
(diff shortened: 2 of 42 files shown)


pakrym:
Let's add some tests.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerSession.cs
```diff
@@ -104,7 +104,7 @@ public async ValueTask DisposeAsync(bool ignoreChecks)
                 {
                     foreach (var expectedStub in _expectedCoverage)
                     {
-                        if (!matched.Contains(expectedStub, StringComparer.InvariantCultureIgnoreCase))
+                        if (matched.Length > 0 && !matched.Contains(expectedStub, StringComparer.InvariantCultureIgnoreCase))
```
pakrym, line 107: Why do we need this?

test/AutoRest.TestServer.Tests/body-complex.cs
```diff
@@ -57,6 +57,12 @@ public class BodyComplexTest: TestServerTestBase
             Assert.ThrowsAsync(Is.InstanceOf<JsonException>(), async () => await new BasicClient(ClientDiagnostics, pipeline, host).GetNotProvidedAsync());
         });
 
+        [Test]
+        public Task ThrowsIfApiVersionIsNull() => Test((host, pipeline) =>
```
pakrym, line 61: don't use `Test(..` here.

pakrym: See CheckComplexPrimitiveInteger for an example.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerSession.cs
ShivangiReja: For [these tests](https://github.com/Azure/autorest.csharp/blob/ecf8399bb59fbf9b7df1ba34988440febf7752a4/test/AutoRest.TestServer.Tests/body-complex.cs#L63) it is going inside of this condition and giving this error:
image

and the length of `matched` is 0, that's why I added this check.

test/AutoRest.TestServer.Tests/body-complex.cs
ShivangiReja: But here we are trying to create an instance of `BasicClient` which takes host and pipeline.

ShivangiReja: But here we are trying to create an instance of `BasicClient`, which takes host and pipeline. If we remove `Test(..` then how do we get host and pipeline?

pakrym: Just use some random values, you don't need an actual test server to be running and that's what `Test((host, pipeline) =>` does

test/AutoRest.TestServer.Tests/Infrastructure/TestServerSession.cs
ShivangiReja: When I was using `Test((host, pipeline) =>` for [these tests](https://github.com/Azure/autorest.csharp/blob/ecf8399bb59fbf9b7df1ba34988440febf7752a4/test/AutoRest.TestServer.Tests/body-complex.cs#L63), it was going inside of `if` condition and giving this error:
image

because the length of `matched` was 0 for these cases, that's why I added this check but now that I've removed `Test((host, pipeline)`, this condition not needed.

test/AutoRest.TestServer.Tests/body-complex.cs
ShivangiReja: Yeah, I was using it because here we are creating an instance of `BasicClient` and it needs host and pipeline. But now I've removed `Test(..` and passed null for host and pipeline.

test/AutoRest.TestServer.Tests/validation.cs
```diff
@@ -35,5 +35,11 @@ public void ConstructorsCheckRequiredProperties()
             Assert.Throws<ArgumentNullException>(() => new Product(null, new ConstantProduct()));
             Assert.Throws<ArgumentNullException>(() => new Product(new ChildProduct(), null));
         }
+
+        [Test]
+        public void ThrowsIfApiVersionIsNull()
+        {
+            Assert.Throws<ArgumentNullException>(() => new ServiceClient(ClientDiagnostics, null, string.Empty, null, null));
```
pakrym, line 42: Which parameter is being checked for null here?

test/AutoRest.TestServer.Tests/custom-baseURL.cs
```diff
@@ -18,5 +19,11 @@ public class custom_baseURL : TestServerTestBase
         [Test]
         public Task CustomBaseUriMoreOptions() => TestStatus(async (host, pipeline) =>
             await new custom_baseUrl_more_options.PathsClient(ClientDiagnostics, pipeline, "test12", dnsSuffix: host.ToString()).GetEmptyAsync( string.Empty, string.Empty, "key1",  "v1"));
+
+        [Test]
+        public void ThrowsIfHostIsNull()
+        {
+            Assert.Throws<ArgumentNullException>(() => new custom_baseUrl.PathsClient(ClientDiagnostics, null , null));
```
pakrym, line 26: Which parameter is being checked for null here?

test/AutoRest.TestServer.Tests/validation.cs
ShivangiReja: Last parameter i.e. `apiVersion`. Replaced other params with random values to make this clear.

test/AutoRest.TestServer.Tests/custom-baseURL.cs
ShivangiReja: Last parameter i.e. `host`. Replaced other param with random value to make this clear.

PR #867, opened 2020-08-11
Always initialize Additional Properties Collection during deserialization

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/848


src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -254,7 +254,7 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
                 var objAdditionalProperties = obj.AdditionalProperties;
                 if (objAdditionalProperties != null)
                 {
-                    writer.Line($"{objAdditionalProperties.Type} {dictionaryVariable:D} = default;");
+                    writer.Line($"{objAdditionalProperties.Type} {dictionaryVariable:D} = new {objAdditionalProperties.Type}();");
```
pakrym, line 257: Nice!

PR #881, opened 2020-09-01
Update v3 code generation for HEAD API calls to return Response<bool> instead of Response

AtOMiCNebula:
This updates the v3 code generator to have the Operations/RestOperations return `Response<bool>` instead of just `Response` for HEAD API calls.  `true`/`false` is defined similarly to how I saw prior versions of autorest do it (https://github.com/Azure/autorest.common/blob/b9ca0fb9f2c5e69993054ceabf84e336b2d73f58/src/AzureExtensions.cs#L93).  I found my way here through azure/azure-sdk-for-net#14235, and code changes that come about from this can be seen here (https://github.com/AtOMiCNebula/azure-sdk-for-net/commit/fb6e34a1411299350d0c74b670a08bb3dc60c330, for viewing purposes only, I know a bot handles most of the codegen updates).  Some test fixes will be needed too once the new codegen arrives in the Resources project (https://github.com/AtOMiCNebula/azure-sdk-for-net/commit/f06b0d3b20915447276fd7f44578cd29cb0c921a), though it's currently still pinned to an older autorest.csharp, so not yet.  I'm sure you can offer suggestions on the right way to organize that. :smile:

Happy to make any changes here!  I don't grok where the line with M4 ends and this one begins, so if part of this should go into M4 or not, I'd appreciate pointers.  It'd probably be nice to have better API usability in other languages too, but I haven't looked into them at all to know if they have the same shortcoming as the C# code generator did here.

Thanks!

### DIFF
--- a/test/TestServerProjects/httpInfrastructure/Generated/HttpSuccessRestClient.cs
+++ b/test/TestServerProjects/httpInfrastructure/Generated/HttpSuccessRestClient.cs
@@ -49,14 +49,15 @@ namespace httpInfrastructure
 
         /// <summary> Return 200 status code if successful. </summary>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public async Task<Response> Head200Async(CancellationToken cancellationToken = default)
+        public async Task<Response<bool>> Head200Async(CancellationToken cancellationToken = default)
         {
             using var message = CreateHead200Request();
             await _pipeline.SendAsync(message, cancellationToken).ConfigureAwait(false);
             switch (message.Response.Status)
             {
                 case 200:
-                    return message.Response;
+                    bool value = message.Response.Status >= 200 && message.Response.Status < 300;
+                    return Response.FromValue(value, message.Response);
                 default:
                     throw await _clientDiagnostics.CreateRequestFailedExceptionAsync(message.Response).ConfigureAwait(false);
             }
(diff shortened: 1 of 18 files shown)


AtOMiCNebula:
@pakrym, any thoughts here? :smile:

pakrym:
Thank you for the contribution and sorry it took so long for me to get to it.


I think the goal is right but have a few suggestions for the implementation:
1. We should change the rest layer to remove the error handling but don't add status code conversion there.
2. Have status code to bool conversion in the public client layer instead.

We are trying to keep REST layer less opinionated and have conventions in the outer client layer.

test/TestServerProjects/httpInfrastructure/Generated/HttpSuccessRestClient.cs
```diff
@@ -668,29 +670,31 @@ internal HttpMessage CreateHead204Request()
...
         /// <summary> Return 204 status code if successful. </summary>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public async Task<Response> Head204Async(CancellationToken cancellationToken = default)
+        public async Task<Response<bool>> Head204Async(CancellationToken cancellationToken = default)
         {
             using var message = CreateHead204Request();
             await _pipeline.SendAsync(message, cancellationToken).ConfigureAwait(false);
             switch (message.Response.Status)
             {
                 case 204:
```
pakrym, line 679: Because this switch is here you would only enter the branch on 204 response code, we need to get rid of it.

ShivangiReja:
Hi @AtOMiCNebula, thank you so much for your contribution. We discussed this internally and have decided to go with [this](https://github.com/Azure/autorest/issues/3603) approach. I'm working on it and will create a PR soon :)

ShivangiReja:
Hi @AtOMiCNebula, we have merged the PR and [here](https://github.com/Azure/autorest.csharp/tree/feature/v3/test/TestProjects/HeadAsBooleanTrue) is an example of head request.

AtOMiCNebula:
@ShivangiReja, awesome!  I like your explicit approach much more than my heuristic approach.  Sorry I wasn't able to get back to my PR, but perhaps it was for the best that I didn't. :innocent:

I'll close this PR and will keep an eye on when the API changes make it into the next resources release.  Presumably that should happen when the next SDK update is picked up for the resources package, correct?

ShivangiReja:
@AtOMiCNebula we have merged the [update PR](https://github.com/Azure/azure-sdk-for-net/pull/16099) in `azure-sdk-for-net` repo. We are currently working on enabling [this](https://github.com/Azure/autorest/issues/3603) feature for `azure-arm` by default.

Issue #890, opened 2020-09-08
PageableHelpers -> AsPages(string? continuationToken) always calls the first page function whether or not the continuation token has been provided.


azabbasi:
We heard from one of our customers that the AsPages(continuationtoken) doesn't resume getting the pages and instead it starts the iteration from the first page.
So basically it doesn't provide resuming functionality. 

After looking into the `PageableHelpers.AsPages(string? continuationToken) ` implementation, we noticed that the first page is being called whether or not the continuation token has been provided.

```
public override async IAsyncEnumerable<Page<T>> AsPages(string? continuationToken = default, int? pageSizeHint = default)
{
    AsyncPageFunc<T>? pageFunc = _firstPageFunc;
    int? pageSize = pageSizeHint ?? _defaultPageSize;
    do
    {
        Page<T> pageResponse = await pageFunc(continuationToken, pageSize).ConfigureAwait(false);
        yield return pageResponse;
        continuationToken = pageResponse.ContinuationToken;
        pageFunc = _nextPageFunc;
    } while (!string.IsNullOrEmpty(continuationToken) && pageFunc != null);
}
```

Given the fact that the `CreateAsyncEnumerable` doesn't take a first page function that accepts a continuation token (and rightfully so) , this will never work in the resuming scenario.

```
public static AsyncPageable<T> CreateAsyncEnumerable<T>(Func<int?, Task<Page<T>>> firstPageFunc, Func<string?, int?, Task<Page<T>>>? nextPageFunc, int? pageSize = default) where T : notnull
{
    AsyncPageFunc<T> first = (continuationToken, pageSizeHint) => firstPageFunc(pageSizeHint);
    AsyncPageFunc<T>? next = nextPageFunc != null ? new AsyncPageFunc<T>(nextPageFunc) : null;
    return new FuncAsyncPageable<T>(first, next, pageSize);
}
```

You can contact me for more details about this issue if the provided description is not clear.

Thank you
Azad

PR #891, opened 2020-09-08
Fix paging continuation

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/890


azabbasi (approved):
LGTM
Thanks for fixing it so quickly

ShivangiReja (approved):

Issue #892, opened 2020-09-10
swagger transform to a *next* operation generates AppendRaw(nextLink.. instead of AppendRawNextLink(nextLink...


christothes:
Example transform:

```yaml
directive:
  - from: swagger-document
    where: $["x-ms-paths"]
    transform: >
      $["/{nextLink}?fooNext"] = {
        "post": {
          "tags": [
            "Foo"
          ],
          "summary": "foo",
          "operationId": "foo",
          "consumes": [
            "application/json"
          ],
          "produces": [
            "application/json"
          ],
          "parameters": [
            {
              "in": "path",
              "name": "nextLink",
              "description": "the next link",
              "required": true,
              "type": "string",
              "x-ms-skip-url-encoding": true
            },
            {
              "in": "body",
              "name": "body",
              "description": "query options",
              "required": true,
              "schema": {
                "$ref": "#/definitions/FooQueryOption"
              }
            }
          ],
          "responses": {
            "200": {
              "description": "Success",
              "schema": {
                "$ref": "#/definitions/SomeModel"
              }
            },
            "default": {
              "description": "Client error or server error (4xx or 5xx)",
              "schema": {
                "$ref": "#/definitions/ErrorCode"
              }
            }
          },
          "x-ms-pageable": {
            "nextLinkName": null
          }
        }
      }
```

Generated code:
```C#
var uri = new RawRequestUriBuilder();
uri.AppendRaw(endpoint, false);
uri.AppendRaw("/foo/v1.0", false);
uri.AppendPath("/", false);
uri.AppendRaw(nextLink, false);
```

Expected:
```C#
var uri = new RawRequestUriBuilder();
uri.AppendRaw(endpoint, false);
uri.AppendRaw("/foo/v1.0", false);
uri.AppendRawNextLink(nextLink, false);
```

PR #893, opened 2020-09-10
Always generate AppendRawNextLink for nextLink parameters

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/892

### DIFF
--- a/src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
+++ b/src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
@@ -132,16 +132,7 @@ namespace AutoRest.CSharp.V3.Generation.Writers
                 {
                     if (!segment.Value.IsConstant && segment.Value.Reference.Name == "nextLink")
                     {
-                        if (segment.IsRaw)
-                        {
-                            // Artificial nextLink needs additional logic for relative versus absolute links
-                            WritePathSegment(writer, uri, segment, "AppendRawNextLink");
-                        }
-                        else
-                        {
-                            // Natural nextLink parameters need to use a different method to parse path and query elements
-                            WritePathSegment(writer, uri, segment, "AppendRaw");
-                        }
+                        WritePathSegment(writer, uri, segment, "AppendRawNextLink");
                     }
                     else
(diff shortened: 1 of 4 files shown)


src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -132,16 +132,7 @@ private void WriteRequestCreation(CodeWriter writer, RestClientMethod clientMeth
...
-                        {
-                            // Artificial nextLink needs additional logic for relative versus absolute links
-                            WritePathSegment(writer, uri, segment, "AppendRawNextLink");
-                        }
-                        else
-                        {
-                            // Natural nextLink parameters need to use a different method to parse path and query elements
-                            WritePathSegment(writer, uri, segment, "AppendRaw");
-                        }
+                        WritePathSegment(writer, uri, segment, "AppendRawNextLink");
```
christothes, line 135: Just out of curiosity - why was this distinction between `IsRaw` ever important? Does it also make sense for `AppendRaw` to detect when the segment being appended is a superset of the target uri rather than special casing it in `AppendRawNextLink`?

pakrym: I think we tried to reduce the number of cases where we run a slightly more expensive `AppendRawNextLink`. Not sure if it was worth it.

test/TestServerProjects/paging/Generated/PagingRestClient.cs
```diff
@@ -1007,7 +1007,7 @@ internal HttpMessage CreateNextFragmentRequest(string apiVersion, string tenant,
             uri.AppendPath("/paging/multiple/fragment/", false);
             uri.AppendPath(tenant, true);
             uri.AppendPath("/", false);
```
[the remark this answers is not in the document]
pakrym: Follows swagger definition in https://github.com/Azure/autorest.testserver/blob/70c108c95aa21b934b97278c67aedc6ae22e1588/swagger/paging.json#L514

christothes (approved):

PR #894, opened 2020-09-11
Package generator into a nuget package

pakrym:


eng/pipelines/build.yml
```diff
@@ -8,6 +8,10 @@ pr:
     include:
     - feature/*
 
+variables:
```
pakrym, line 11: Easier to view in https://github.com/Azure/autorest.csharp/pull/894/files?w=1

weshaggard (approved):
Looks reasonable to me from a quick scan.

Issue #885, opened 2020-09-03
Autorest generated C# code validate optional properties even it's null, which make validation fail


blueww:
**Describe the issue or request**
Please describe the issue or feature request.

Storage team recently have new feature have [swagger change](https://github.com/Azure/azure-rest-api-specs-pr/pull/1104) as following.
The autorest generated C# code from it with always fail in validation, when any of the optional properties is null.
Since the properties are optional, we should skip validate if it’s null.

Swagger:
```json
    "DateAfterModification": {
      "properties": {
        "daysAfterModificationGreaterThan": {
          "type": "number",
          "multipleOf": 1.0,
          "minimum": 0,
          "description": "Value indicating the age in days after last modification"
        },
          "daysAfterLastAccessTimeGreaterThan": {
            "type": "number",
            "multipleOf": 1.0,
            "minimum": 0,
            "description": "Value indicating the age in days after last blob access. This property can only be used in conjuction with last access time tracking policy"
        }
      },
      "description": "Object to define the number of days after object last modification Or last access. Properties daysAfterModificationGreaterThan and daysAfterLastAccessTimeGreaterThan are mutually exclusive"
    },
```

Generated C# code:
The generated C# code to validate this item will be following (with autorest 2.0.4413, 3.0.6306)
Please note the validation will fail when any of DaysAfterModificationGreaterThan or DaysAfterLastAccessTimeGreaterThan is null (null %1 = null, null != 0). But since both properties are optional, null should be a valid value for the 2 properties.
```c#
        public virtual void Validate()
        {
            if (DaysAfterModificationGreaterThan < 0)
            {
                throw new ValidationException(ValidationRules.InclusiveMinimum, "DaysAfterModificationGreaterThan", 0);
            }
            if (DaysAfterModificationGreaterThan % 1 != 0) // ensures that the value is a multiple of 1.0
            {
                throw new ValidationException(ValidationRules.MultipleOf, "DaysAfterModificationGreaterThan", 1);
            }
            if (DaysAfterLastAccessTimeGreaterThan < 0)
            {
                throw new ValidationException(ValidationRules.InclusiveMinimum, "DaysAfterLastAccessTimeGreaterThan", 0);
            }
            if (DaysAfterLastAccessTimeGreaterThan % 1 != 0)
            {
                throw new ValidationException(ValidationRules.MultipleOf, "DaysAfterLastAccessTimeGreaterThan", 1);
            }
        }
```

**Describe your ideas for solutions**
[...]

PR #895, opened 2020-09-14
[V2 Code Generator] Fix validation method failure for optional properties

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/885


src/vanilla/ClientModelExtensions.cs
```diff
@@ -370,7 +370,7 @@ public static string CheckNull(string valueReference, string executionBlock)
 
             if (sb.ToString().Trim().Length > 0)
             {
-                if (type.IsValueType())
+                if (type.IsValueType() && sequence != null && !sequence.IsNullable)
```
pakrym, line 373: Let's create a `isNullable` local and set it `if (sequence != null && sequence.ShouldValidateChain())` branch.

ShivangiReja: Don't we want to check `sequence.IsNullable`?

pakrym: Yep, we do

ShivangiReja: ` sequence` is always null in this case, so I updated the code and passed the whole [variable ](https://github.com/Azure/autorest.csharp/pull/895/commits/3948893c945f85597fdebe3843d4183ce9360dfd#diff-99c896ba66d1a5f58410c2e2a5a02790)instead of just `valueReference` and it already has `IsNullable` method.

src/vanilla/ClientModelExtensions.cs
```diff
@@ -346,7 +347,8 @@ public static string CheckNull(string valueReference, string executionBlock)
             if (sequence != null && sequence.ShouldValidateChain())
             {
                 var elementVar = scope.GetUniqueName("element");
-                var innerValidation = sequence.ElementType.ValidateType(scope, elementVar, null);
+                variable.Name = elementVar;
```
pakrym, line 350: Why are you assigning the name here?

src/vanilla/ClientModelExtensions.cs
```diff
@@ -346,7 +347,8 @@ public static string CheckNull(string valueReference, string executionBlock)
             if (sequence != null && sequence.ShouldValidateChain())
             {
                 var elementVar = scope.GetUniqueName("element");
-                var innerValidation = sequence.ElementType.ValidateType(scope, elementVar, null);
+                variable.Name = elementVar;
+                var innerValidation = sequence.ElementType.ValidateType(scope, variable, null);
```
pakrym, line 351: Why are you passing `variable` instead of `elementVar` here?

src/vanilla/Templates/Rest/Common/ModelTemplate.cshtml
```diff
@@ -205,7 +205,7 @@ foreach (PropertyCs property in Model.InstanceProperties.Where(p => p.IsRequired
 foreach (var property in Model.InstanceProperties.Where(p => p.Constraints.Any() || !(p.ModelType is PrimaryType)))
 {
     anythingToValidate = true;
-            @:@property.ModelType.ValidateType(Model, $"this.{property.Name}", property.Constraints)
```
pakrym, line 208: This looks like a possibility to cause a name conflict

src/vanilla/ClientModelExtensions.cs
ShivangiReja: Previously we were directly passing `elementVar` to `ValidateType` method. Now that we are not passing `string` that's why I had to set `elementVar` to `variable.Name`.

src/vanilla/ClientModelExtensions.cs
ShivangiReja: Because I have changed  `string valueReference` to `IVariable variable` in ValidateType method [here](https://github.com/Azure/autorest.csharp/blob/3948893c945f85597fdebe3843d4183ce9360dfd/src/vanilla/ClientModelExtensions.cs#L322).

ShivangiReja: https://github.com/Azure/autorest.csharp/pull/895#discussion_r488332308

src/vanilla/Templates/Rest/Common/ModelTemplate.cshtml
ShivangiReja: Fixed!

src/vanilla/ClientModelExtensions.cs
```diff
@@ -346,7 +347,7 @@ public static string CheckNull(string valueReference, string executionBlock)
             if (sequence != null && sequence.ShouldValidateChain())
             {
                 var elementVar = scope.GetUniqueName("element");
-                var innerValidation = sequence.ElementType.ValidateType(scope, elementVar, null);
+                var innerValidation = sequence.ElementType.ValidateType(scope, elementVar, isNullable, null);
```
pakrym, line 350: I don't think that nullable collection means nullable element, does `sequence.ElementType` have IsNullable()?

ShivangiReja: No, `sequence.ElementType` doesn't have `IsNullable()` but only `sequence` has `sequence.IsNullable`...

pakrym: let's just pass `false` then. To keep the existing behavior.

pakrym (approved):

fearthecowboy:
Merging after customer confirmed it works.

PR #904, opened 2020-09-22
Add support for Guid when deserializing xml elements

ShivangiReja:
Fixes: https://github.com/Azure/azure-sdk-for-net-pr/issues/1337

[This](https://github.com/Azure/azure-sdk-for-net-pr/pull/1336) PR build is failing because `Guid` was not supported when deserializing xml elements, which resulted [this ](https://github.com/Azure/azure-sdk-for-net-pr/blob/742dfbad23de8476e3670a9a922ccd106ff7c1c7/sdk/customvision/Azure.AI.CustomVision/src/Generated/Models/ImagePrediction.Serialization.cs#L70) generated code.


src/AutoRest.CSharp.V3/Generation/Writers/XmlCodeWriterExtensions.cs
```diff
@@ -354,6 +354,13 @@ private static void ToDeserializeValueCall(this CodeWriter writer, XmlValueSeria
                     return;
                 }
 
+                if (frameworkType == typeof(Guid))
+                {
+                    CSharpType nonNullableType = type.WithNullable(false);
+                    writer.Append($"new {nonNullableType}({element}.Value)");
```
pakrym, line 360: You can simplify because you know the `nonNullableType` is `Guid`.

pakrym (approved):
Good job!

Issue #884, opened 2020-09-02
Modify visibility of unexposed enums from public to internal


ShivangiReja:
Example: https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/anomalydetector/Azure.AI.AnomalyDetector/api/Azure.AI.AnomalyDetector.netstandard2.0.cs#L30

PR #914, opened 2020-10-03
Modify visibility of unexposed enums from public to internal

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/884


src/AutoRest.CSharp.V3/Input/SchemaUsageProvider.cs
```diff
@@ -35,24 +35,34 @@ public SchemaUsageProvider(CodeModel codeModel)
                         Apply(operationResponse.ResponseSchema, SchemaTypeUsage.Error | SchemaTypeUsage.Output);
                     }
 
+                    foreach (var parameter in operation.Parameters)
```
pakrym, line 38: Do we use these?

pakrym (approved):
Very nice

src/AutoRest.CSharp.V3/Input/SchemaUsageProvider.cs
ShivangiReja: Discussed offline: we also use it [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp.V3/Output/Models/RestClient.cs#L40).

Issue #922, opened 2020-10-12
Re-enable x-nullable validation for string properties


pakrym:
We've disabled it because there were too many failures during update.

PR #923, opened 2020-10-12
Relax the nullability check to allow strings

pakrym:
I've made it too strict. We won't be able to ever get the update in.

I filed https://github.com/Azure/autorest.csharp/issues/922 to re-enable but not sure if we ever are going to do it.

### DIFF
--- a/src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
+++ b/src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
@@ -277,7 +277,10 @@ namespace AutoRest.CSharp.V3.Generation.Writers
                                     writer.Append($"continue;");
                                 }
                             }
-                            else if (!property.IsRequired)
+                            else if (!property.IsRequired &&
+                                     (property.Property?.ValueType.IsFrameworkType != true ||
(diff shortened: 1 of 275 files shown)


src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -277,7 +277,10 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
                                     writer.Append($"continue;");
                                 }
                             }
-                            else if (!property.IsRequired)
+                            else if (!property.IsRequired &&
```
pakrym, line 280: This is the change. So you don't have to scroll :)

ShivangiReja (approved):

ShivangiReja:
Can we write a test to this?

src/AutoRest.CSharp.V3/Generation/Writers/JsonCodeWriterExtensions.cs
ShivangiReja: Thank you! :)

pakrym:
I've reenabled one of the existing tests that I made conditional on the configuration previously.

PR #926, opened 2020-10-16
Generate Response<bool> return type for head requests when head-as-boolean flag is true

ShivangiReja:
As per [this](https://github.com/Azure/autorest/issues/3603) approach, generating return type as `Response<bool>` for head requests when `head-as-boolean` flag is set to true. For data-plane libraries the default is false and for management libraries the default is true as described [here ](https://github.com/Azure/autorest/issues/3602).

Fixes: https://github.com/Azure/autorest.csharp/issues/912

TODO: Cover all the test scenarios for head requests.


src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -517,7 +517,17 @@ private void WriteStatusCodeSwitch(CodeWriter writer, CodeWriterDeclaration mess
                 writer.Line($"var headers = new {headersModelType}({responseVariable});");
             }
 
-            using (writer.Scope($"switch ({responseVariable}.Status)"))
+            string responseStatus;
```
pakrym, line 520: We should handle this in the case statement instead.


something like

``` C#
  switch(status)
        {
            case 200:
            case int s when s > 200 && s < 300:
                Console.WriteLine("yay");
                break;
        }
```

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -549,6 +566,12 @@ private void WriteStatusCodeSwitch(CodeWriter writer, CodeWriterDeclaration mess
                             writer.Line($"var {valueVariable:D} = {message}.ExtractResponseContent();");
                             value = new Reference(valueVariable.ActualName, responseBody.Type);
                         }
+                        else if (responseBody is ConstantResponseBody _ && returnType != null)
```
pakrym, line 569: Don't discard the ConstantResponseBody variable here and avoid the cast on the next line.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -549,6 +566,12 @@ private void WriteStatusCodeSwitch(CodeWriter writer, CodeWriterDeclaration mess
...
+                        {
+                            var body = (ConstantResponseBody)responseBody;
+                            writer.Line($"{returnType} {valueVariable:D} = {body.Value.Value:L};");
```
pakrym, line 572: `writer.WriteReferenceOrConstant`

src/AutoRest.CSharp.V3/Output/Models/Responses/Response.cs
```diff
@@ -5,13 +5,46 @@ namespace AutoRest.CSharp.V3.Output.Models.Responses
...
             ResponseBody = responseBody;
             StatusCodes = statusCodes;
         }
 
         public ResponseBody? ResponseBody { get; }
-        public int[] StatusCodes { get; }
+        public StatusCodes[] StatusCodes { get; }
+    }
+
+    internal enum Family
```
pakrym, line 18: Us this used?

pakrym: I think we can just keep the numbers.

src/AutoRest.CSharp.V3/Output/Models/RestClient.cs
```diff
@@ -334,14 +335,30 @@ private RestClientMethod BuildMethod(Operation operation, HttpRequest httpReques
 
             foreach (var response in operation.Responses)
             {
+                List<StatusCodes> statusCodes = new List<StatusCodes>();
+                Array.ForEach(response.HttpResponse.IntStatusCodes, statusCode => statusCodes.Add(new StatusCodes(statusCode, null)));
```
pakrym, line 339: use a foreach statement.

src/AutoRest.CSharp.V3/Output/Models/Responses/Response.cs
```diff
@@ -5,13 +5,46 @@ namespace AutoRest.CSharp.V3.Output.Models.Responses
...
+        CLIENT_ERROR,
+
+        // 5xx HTTP status codes.
+        SERVER_ERROR,
+
+        // Other, unrecognized HTTP status codes.
+        OTHER
+    }
+
+    internal struct StatusCodes
```
pakrym, line 39: One type per file

pakrym:
Nice!

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
```diff
@@ -542,7 +532,16 @@ private void WriteStatusCodeSwitch(CodeWriter writer, CodeWriterDeclaration mess
                         }
                         else
                         {
-                           writer.Line($"case {statusCode.Family}:");
+                            if (statusCode.Family == 2)
+                            {
+                                writer.Line($"case 200:");
```
pakrym, line 537: You don't need this line for the family case. The example had it both ways to show that we can mix the individual codes and ranges.

` writer.Line($"case int s when s >= {statusCode.Family*100:L} && s < {statusCode.Family * 100 + 99:L}:");

test/AutoRest.TestServer.Tests/HeadAsBooleanTrue.cs
```diff
@@ -0,0 +1,39 @@
...
+        [Test]
+        public Task HttpSuccess204Head() => Test(async (host, pipeline) =>
+        {
+            var response = await new HttpSuccessClient(ClientDiagnostics, pipeline, host).Head204Async();
+            Assert.AreEqual(204, response.GetRawResponse().Status);
+            Assert.AreEqual(true, response.Value);
+        });
+
+        [Test]
+        public Task HttpSuccess404Head() => Test(async (host, pipeline) =>
```
pakrym, line 32: Do we have an endpoint that returns 3xx?

test/AutoRest.TestServer.Tests/HeadAsBooleanTrue.cs
```diff
@@ -0,0 +1,39 @@
...
+using NUnit.Framework;
+
+namespace AutoRest.TestServer.Tests
+{
+    public class HeadAsBooleanTrueTests : TestServerTestBase
+    {
+        public HeadAsBooleanTrueTests(TestServerVersion version) : base(version) { }
+
+        [Test]
+        public Task HttpSuccess200Head() => Test(async (host, pipeline) =>
```
pakrym, line 16: Let's add a test validates the API surface as well (check method return type).

test/AutoRest.TestServer.Tests/HeadAsBooleanTrue.cs
ShivangiReja: We don't have 3xx returns [here](https://github.com/Azure/autorest.testserver/blob/master/swagger/head.json).

ShivangiReja: But we have routes which return 3xx. Do you want me to update the spec to add use case for 3xx?

pakrym: It's ok.

src/AutoRest.CSharp.V3/Generation/Writers/RestClientWriter.cs
ShivangiReja: Ahh, yess!

pakrym (approved):

PR #935, opened 2020-10-21
Support taking JSON as a stream

pakrym:
In some cases, we don't want to parse JSON and would like to roundtrip it as a stream.

```
        "schema": {
              "type": "string",
              "format": "binary"
            }
```

Allows us to express this in swagger, this PR adds support to C# generator.


src/AutoRest.CSharp.V3/Output/Models/RestClient.cs
```diff
@@ -285,35 +273,48 @@ private RestClientMethod BuildMethod(Operation operation, HttpRequest httpReques
...
-                    var serialization = _serializationBuilder.Build(
-                        httpRequestWithBody.KnownMediaType,
-                        bodyRequestParameter.Schema,
-                        bodyParameterValue.Type);
-
-                    // This method has a flattened body
-                    if (bodyRequestParameter.Flattened == true)
+                    if (httpRequestWithBody.KnownMediaType == KnownMediaType.Binary ||
+                        // WORKAROUND: https://github.com/Azure/autorest.modelerfour/issues/360
+                        bodyRequestParameter.Schema is BinarySchema)
```
pakrym, line 278: This line is the only line of the new code here.

ShivangiReja (approved):

Issue #782, opened 2020-05-20
Ignore (do not generate) operation parameters for x-ms-client-request-id


heaths:
Based on https://github.com/Azure/azure-sdk-for-net/issues/11273, if swagger contains an `x-ms-client-request-id` parameter (or maybe any parameter marked as `"x-ms-client-request-id": true` in its definition), do not generate a parameter for it since we'll just have to pass null for it anyway and time is wasted checking for it, if not making the assembly a little larger.

pakrym:
And tracestate/traceparent

PR #937, opened 2020-10-23
Filter standard headers

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/782


pakrym:
These are added by the pipeline automatically.

ShivangiReja (approved):

PR #945, opened 2020-11-13
Enable runtime roll forward 

pakrym:
To allow running when .net 3.0 is not installed but later version is.


pakrym:
Merging to unblock migration process.

PR #954, opened 2020-12-02
Add optional cref hints to inheritdoc comments

christothes:
# Summary
The purpose of this change is to avoid inheritdoc ambiguous inheritance warnings by allowing the generator to add `cref=` hints to inheritdoc comments.

### DIFF
--- a/samples/CognitiveSearch/CognitiveSearch/Generated/Models/FacetResult.cs
+++ b/samples/CognitiveSearch/CognitiveSearch/Generated/Models/FacetResult.cs
@@ -44,7 +44,7 @@ namespace CognitiveSearch.Models
         public IEnumerable<string> Keys => AdditionalProperties.Keys;
         /// <inheritdoc />
         public IEnumerable<object> Values => AdditionalProperties.Values;
-        /// <inheritdoc cref="ICollection{T}.Count"/>
+        /// <inheritdoc cref="IReadOnlyCollection{T}.Count"/>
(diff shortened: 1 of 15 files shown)


samples/CognitiveSearch/CognitiveSearch/Generated/Models/FacetResult.cs
```diff
@@ -44,7 +44,7 @@ internal FacetResult(long? count, IReadOnlyDictionary<string, object> additional
         public IEnumerable<string> Keys => AdditionalProperties.Keys;
         /// <inheritdoc />
         public IEnumerable<object> Values => AdditionalProperties.Values;
-        /// <inheritdoc />
+        /// <inheritdoc cref="ICollection{T}.Count"/>
```
christothes, line 47: looks like this needs to be distinguished between readonly and not. Will add the change.

src/AutoRest.CSharp.V3/Generation/Writers/ModelWriter.cs
```diff
@@ -35,6 +35,7 @@ public void WriteModel(CodeWriter writer, TypeProvider model)
 
         private void WriteObjectSchema(CodeWriter writer, ObjectType schema)
         {
+            const string collectionOfT = "Collection{T}.";
```
pakrym, line 38: Can we include `I` in the name?

christothes: I did - I just forgot to push it :)

pakrym (approved):

Issue #851, opened 2020-07-29
Consider renaming your release tags to be prefixed with `v`


ckittel:
**Describe the issue or request**
When using the `--use` parameter like `autorest --input-file:myfile.json --output-folder:./out --namespace:demo --use:@autorest/csharp@3.0.0-dev.20200727.1` it doesn't work for the csharp project like it does for `autorest.python` and similar.  Autorest claims it cannot find this release.

**Describe your ideas for solutions**
This is because for this feature to work, your release tags in GitHub have to be prepended with a `v`.  Once that is done, then this built-in `autorest` feature starts to work.  So for example, if `3.0.0-dev.20200727.1` would have been tagged as `v3.0.0-dev.20200727.1` the command listed prior would have worked.

cc: @vrdmr

PR #957, opened 2020-12-16
Publish @autorest/csharp-v3 on npm

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/851


eng/PublishRelease.ps1
```diff
@@ -0,0 +1,45 @@
...
+    npm version --no-git-tag-version $devVersion | Out-Null;
+   
+    $file = npm pack -q;
+    $name = "Autorest for c# v$devVersion"
+
+    Write-Host "Publishing $file on GitHub!"
+    
+    cmd /c ""npx -q publish-release --token $GitHubToken --repo autorest.csharp --owner azure --name $name --tag v$devVersion --notes=prerelease-build --prerelease --editRelease false --assets $file --target_commitish $Sha 2>&1""
+
+    $filePath = Join-Path $WorkingDirectory '.npmrc'
```
pakrym, line 24: Should we use the recommedned pattern of checking in the npmrc and setting the token as an env var: https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#set-the-token-as-an-environment-variable-on-the-cicd-server

eng/PublishRelease.ps1
```diff
@@ -0,0 +1,45 @@
...
+try {
+    $currentVersion = node -p -e "require('./package.json').version";
+    $devVersion = "$currentVersion-dev.$BuildNumber"
+
+    Write-Host "Setting version to $devVersion"
+
+    npm version --no-git-tag-version $devVersion | Out-Null;
+   
+    $file = npm pack -q;
+    $name = "Autorest for c# v$devVersion"
```
pakrym, line 18:
```suggestion
    $name = "AutoRest C# v$devVersion"
```

pakrym (approved):

pakrym:
I've looked at the npm package and it seems to contain too much stuff in it. There are two copies of everything.

eng/PublishRelease.ps1
pakrym: I don't feel strongly.

ShivangiReja:
>I've looked at the npm package and it seems to contain too much stuff in it. There are two copies of everything.

Removed `published` folder with all the duplicate files in it.

Here is the latest NPM release: https://www.npmjs.com/package/@autorest/csharp-v3/v/3.0.0-beta.20201216.11
Here is the latest GitHub release: https://github.com/Azure/autorest.csharp/releases/tag/v3.0.0-beta.20201216.11

pakrym (approved):
👍 👍 👍

PR #962, opened 2020-12-22
docs

iscai-msft:
@jianghaolu is doing the dev section firsts, will update the PR with this section later.

I couldn't figure out how to add a samples section like [python](https://github.com/Azure/autorest.python/tree/9839bf3392d85b4661c0b1249634a6b9f1f9706b/docs/samples) has, will try to do so in the new year.

Finally, left faq and the migrate section mostly blank, (here's the [python faq](https://github.com/Azure/autorest.python/blob/9839bf3392d85b4661c0b1249634a6b9f1f9706b/docs/faq.md), and the [python migrate](https://github.com/Azure/autorest.python/tree/9839bf3392d85b4661c0b1249634a6b9f1f9706b/docs/migrate)), lmk what your breaking changes / new features are from the previous c# generator, will do this in the new year too. 

also a lot these links are probably broken (all of the ones referring to the main docs)


docs/client/initializing.md
```diff
@@ -0,0 +1,73 @@
+# <img align="center" src="../images/logo.png">  Initializing Your C# Client
+
+The first step to using your generated client in code is to import and initialize your client. Our SDKs are modelled such
+that the client is the main point of access to the generated code.
+
+## Initializing and Authenticating Your Client
+
+You can use your client if you import the namespace you specified when generating (under flag `--namespace`). For the sake of this example,
+let's say the namespace is `Azure.Pets` and your client is `PetsClient`. You could then access the client if you import the namespace.
+The default client also comes with these three parameters:
```
pakrym, line 10: We don't, currently generate a public ctor that configures the pipeline. The one that takes clientDiagnostics/pipeline is protected so the idea is that you have to define a constructor based on what authentication is to be used.

pakrym: For example https://github.com/Azure/azure-sdk-for-net/blob/3d8e6a42bd78716e9001ec54c5073429f8c9efe3/sdk/synapse/Azure.Analytics.Synapse.Monitoring/src/Customization/MonitoringClient.cs#L10

pakrym: We do it for ARM because we know that token credential-based auth is supported there

docs/client/models.md
```diff
@@ -0,0 +1,43 @@
...
+        static void Main(string[] args)
+        {
+
+            var dog = new Dog
+            {
+                ...
+            };
+        }
+    }
+}
```
pakrym, line 24: C# now supports root level statements, which might be appropriate for docs.

```suggestion
using AzurePets.Models;

var dog = new Dog
{
    ...
};
```

docs/client/models.md
```diff
@@ -0,0 +1,43 @@
...
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+
+            string myDogType= DogTypes.Dalmation;
+        }
+    }
+}
```
pakrym, line 42:
```suggestion
using AzurePets.Models;

DogTypes myDogType = DogTypes.Dalmation;
```

docs/client/operations.md
```diff
@@ -0,0 +1,262 @@
...
+is initially called.
+
+For our example, we will use the paging operation generated from [this][polling_paging_swagger] swagger. Let's say we generated this swagger with namespace `Azure.Paging`.
+
+### Sync Paging Operations
+
+By default, our sync paging operations return a [`Pageable`][pageable] object. The initial call to the function returns
+the pager, but doesn't make any network calls. Instead, calls are made when users start iterating, with each network call returning a page of data.
+
+```python
```
pakrym, line 174: Did you plan to remove this block?

docs/client/operations.md
```diff
@@ -0,0 +1,262 @@
...
+        {
+
+            internal static ClientDiagnostics clientDiagnostics = new ClientDiagnostics();
+            var pipeline = new HttpPipeline();
+            string endpoint = "http://localhost:3000";
+            var client = new PollingPagingExampleClient(clientDiagnostics, pipeline, endpoint);
+
+            var pageable = client.BasicPaging();
+
+            foreach (var page in pageable.AsPages())
```
pakrym, line 202: It might be better to show iterating items instead of pages.
```
foreach (var item in pageable) {}
```

docs/client/operations.md
```diff
@@ -0,0 +1,262 @@
...
+            var pipeline = new HttpPipeline();
+            string endpoint = "http://localhost:3000";
+            var client = new PollingPagingExampleClient(clientDiagnostics, pipeline, endpoint);
+
+            var pageableAsync = client.BasicPagingAsync();
+
+            await foreach (var page in pageableAsync.AsPages())
+            {
+                var values = page.Values;
+            }
```
pakrym, line 242:
```suggestion
            await foreach (var item in pageableAsync)
            {
            }
```

docs/generate/generate_without_sdk_repo.md
```diff
@@ -0,0 +1,42 @@
+# <img align="center" src="../images/logo.png">  Generating C# Clients Outside of the azure-sdk-for-net Repo
```
pakrym, line 1: Is it worth to duplicate this with the main readme?

docs/migrate/readme.md
```diff
@@ -0,0 +1,15 @@
+# <img align="center" src="../images/logo.png">  Migrating to Latest AutoRest
+
+See the [main docs][main_docs] for changes in versioning and flags, this section focuses on how the generated code differs.
+
+## Breaking Changes
+
+TODO
```
pakrym, line 7: I wonder if it's worth discussing all the differences because there are a LOT

iscai-msft: ah ok. I still think it's worth highliighting any main ones you want to call it. But it def depends on how many people used the previous generator (i actually don't know for .net). Here's [python's](https://github.com/Azure/autorest.python/tree/autorestv3/docs/migrate) if you're interested

docs/generate/generate_without_sdk_repo.md
iscai-msft: Yeah that's a fair point, I thiink I got carried away. Will jsut make this a link to the main readme

docs/client/operations.md
iscai-msft: yep, thakns!

pakrym:
I made a few changes. Some docs refer to features that are not available yet (i.e. `autorest --csharp` using latest) so we might want to hold off merging until those are ready. We would be able to generate sample swagger when command line support is done as well.

PR #981, opened 2021-01-15
Move customization code and csproj outside of generated folder

ShivangiReja:
I'm able to generate code using `autorest --use:@autorest/csharp@v3.0.0-beta.20201217.3 --input-file:Inputfile --public-clients:true` command but when I set to `--clear-output-folder:true`, it removes all the customizations because currently we put all the customizations inside the generated folder. 

Moving customization code and csproj outside of generated folder to fix this issue.


pakrym:
` await autoRest.Warning("output-folder path is required for customization");`

Let's change this to say something about being an absolute path instead.

```
            var projectDirectory = Path.Combine(configuration.OutputFolder, "../");
```

Let's add a configuration property ProjectRelativeDirectory and always set it to `..`. Use the value in 
CSharpGen and CSharpProj

```

        public static string GeneratedFolder = "generated";
```

This might cause issues with azure-sdk-for-net where everything is already generated to the uppercase directory.

```
          var generatedCodeDirectory = Path.Combine(projectDirectory, "generated");

            foreach (string sourceFile in Directory.GetFiles(projectDirectory, "*.cs", SearchOption.AllDirectories))
            {
                // Ignore existing generated code
                if (sourceFile.StartsWith(generatedCodeDirectory))
                {
```

this condition would fail in azure-sdk-for-net on Linux leading to broken customizations.

pakrym:
```
    <Exec Command="npx autorest@$(AutoRestVersion) --skip-upgrade-check --version=$(AutoRestCoreVersion) $(AutoRestInput) $(AutoRestAdditionalParameters) --use=$(MSBuildThisFileDirectory)../tools/netcoreapp3.0/any/ --output-folder=$(MSBuildProjectDirectory)/generated --clear-output-folder:true --namespace=$(RootNamespace) --shared-source-folders=&quot;$(AzureCoreSharedCodeDirectory);$(AutoRestSharedCodeDirectory)&quot;" />
```

We might want to keep the uppercase generated directory so it's consistent in the repo.

Wonder if we can change the default Generated value to be upper case.

`--clear-output-folder:true` should we use `--clear-output-folder=true` instead?

.gitattributes
```diff
@@ -1,6 +1,6 @@
 * text=auto
 
-**/Generated/**/*.cs linguist-generated=true
-**/Generated/**/*.csproj linguist-generated=true
+**/Generated/*.cs linguist-generated=true
+**/*.csproj linguist-generated=true
```
pakrym, line 4: This definition is too generic it would include csproject that are not generated.

We might want to have multiple definitions that are prefixed with directories we know contain generated code like samples, tests/TestServerProjects etc.

src/AutoRest.CSharp/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -42,7 +42,7 @@ public static void Initialize()
 
         public void AddGeneratedFile(string name, string text)
         {
-            var document = _project.AddDocument(GeneratedFolder + "/" + name, text, GeneratedFolders);
+            var document = _project.AddDocument("." + "/" + name, text, GeneratedFolders);
```
pakrym, line 45: ?

src/AutoRest.CSharp/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -91,17 +91,24 @@ private async Task<Document> ProcessDocument(Document document)
...
         }
 
-        public static async Task<GeneratedCodeWorkspace> Create(string projectDirectory, string[] sharedSourceFolders)
+        public static async Task<GeneratedCodeWorkspace> Create(string projectDirectory, string outputDirectory, string[] sharedSourceFolders)
         {
             var projectTask = Interlocked.Exchange(ref _cachedProject, null);
             var generatedCodeProject = projectTask != null ? await projectTask : CreateGeneratedCodeProject();
 
-            var generatedCodeDirectory = Path.Combine(projectDirectory, "Generated");
+            if (Path.IsPathRooted(projectDirectory))
```
pakrym, line 99: We might want to skip loading sources if directories are not rooted.

Issue #990, opened 2021-01-20
Generator is sending request headers that are long in scientific notation


seanmcc-msft:
### Here is the swagger:

https://github.com/Azure/azure-rest-api-specs/blob/storage-dataplane-preview/specification/storage/data-plane/Microsoft.FileStorage/preview/2020-04-08/file.json#L2639

### Here is the old generated code:

https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Files.Shares/src/Generated/FileRestClient.cs#L4972

### Here is Fiddler with the old generator:

PUT https://seanmcccanary3.file.core.windows.net/test-share-e6e81468-4acc-4000-85a7-28a4ad4a3838/test-directory-bf707c9f-6b0e-488d-856e-d22ff0a0c356/test-file-28918f61-c3d6-4825-ba2e-2c24699cf3df HTTP/1.1
Host: seanmcccanary3.file.core.windows.net
x-ms-version: 2020-04-08
**x-ms-content-length: 4398046511104**
x-ms-type: file
x-ms-file-attributes: None
x-ms-file-creation-time: Now
x-ms-file-last-write-time: Now
x-ms-file-permission: Inherit
x-ms-client-request-id: ec3f254d-d7bd-45de-ac44-0ec47fcca620
x-ms-return-client-request-id: true
User-Agent: azsdk-net-Storage.Files.Shares/12.7.0-alpha.20210120.1 (.NET 5.0.2; Microsoft Windows 10.0.19042)
x-ms-date: Wed, 20 Jan 2021 20:16:25 GMT
Authorization: SharedKey seanmcccanary3:j8qXabObAGMJQ2lptBAHAB5CzCFs9i/nIr2XyWcwzdc=
traceparent: 00-d1d272f6595c3448a081076b2c141f06-ac8acfc2bddcf840-00
Content-Length: 0

HTTP/1.1 201 Created
Content-Length: 0
Last-Modified: Wed, 20 Jan 2021 20:16:26 GMT
ETag: "0x8D8BD8043CAA3D9"
Server: Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: 1aef2513-301a-0062-3069-ef18e5000000
x-ms-client-request-id: ec3f254d-d7bd-45de-ac44-0ec47fcca620
x-ms-version: 2020-04-08
x-ms-file-change-time: 2021-01-20T20:16:26.3742425Z
[...]

PR #996, opened 2021-01-21
Send large longs in a correct notation

pakrym:
Fixes: https://github.com/Azure/autorest.csharp/issues/990


chamons (approved):
Not sure if the -MinValue version is worth testing.

Since you are using a typeformatter from the BCL it should just work.

❤️  on test.

PR #1008, opened 2021-02-04
Add autorest package reference in generated csproj for outside of autorest.csharp repo

ShivangiReja:


src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -22,15 +24,53 @@ internal class CSharpProj : IPlugin
...
+";
+         private string _csProjPackageReference = @"
+  <PropertyGroup>
+    <LangVersion>8.0</LangVersion>
+    <IncludeGeneratorSharedCode>true</IncludeGeneratorSharedCode>
+	<RestoreAdditionalProjectSources>https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json</RestoreAdditionalProjectSources>
+  </PropertyGroup>
 
+  <ItemGroup>
+	<PackageReference Include=""Microsoft.Azure.AutoRest.CSharp"" Version=""{0}"" />
```
pakrym, line 42: nit spaces

src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -22,15 +24,53 @@ internal class CSharpProj : IPlugin
...
+            AssemblyInformationalVersionAttribute? versionAttribute = clientAssembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>();
+            if (versionAttribute == null)
+            {
+                throw new InvalidOperationException($"{nameof(AssemblyInformationalVersionAttribute)} is required on client SDK assembly '{clientAssembly.FullName}'");
+            }
+
+            string version = versionAttribute.InformationalVersion;
+
+            string assemblyName = clientAssembly.GetName().Name!;
+            if (assemblyName.StartsWith(PackagePrefix, StringComparison.Ordinal))
```
pakrym, line 60: Do we use assemblyName ?

src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -44,6 +84,17 @@ public async Task<bool> Execute(IPluginCommunication autoRest)
 
             var context = new BuildContext(codeModel, configuration, null);
 
+            if (configuration.SkipCSProjPackageReference)
+            {
+                _csProjContent = string.Format(_csProjContent, "");
```
pakrym, line 89: let's store this into a local instead of the field.

pakrym (approved):

src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
ShivangiReja: What is the advantage of making `_csProjContent` and `_csProjPackageReference` local variables? If we move these strings inside the method then it will create new string for each `Execute` call.

ShivangiReja: What is the advantage of making `_csProjContent` and `_csProjPackageReference` local variables? If we move these strings inside the method then it will create new string for each method call.

src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
ShivangiReja: Removed!

src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
ShivangiReja: Fixed it. 

It was weird. In VS it was looking good but I see no spaces in VS code same as this.

src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
pakrym: Imagine `public async Task<bool> Execute(IPluginCommunication autoRest)` was called multiple times with different values of `autoRest`. Every call after the first one won't work correctly because the method's side effects leak into the containing class.


Storing the content into the variable would make it so you can invoke this method as many times as you want.

This might not matter much in this case but controlling side effects is a good habit in type design in general.

ShivangiReja: Yeah! I agree with you, actually I wanted to create a new string inside the method, instead of using the same `_csProjContent` and assigning values in that. Do you think this would resolve your concern If I create a new string, instead of moving the whole the properties inside?

pakrym: Yeah, that works.

I mean something like this:

```C#
  var version = GetVersion();
  var csProjPackageReference = string.Format(_csProjPackageReference, version);
  var csProjContent = string.Format(_csProjContent, csProjPackageReference);
```

ShivangiReja: Yeah! that's what I meant.

pakrym (approved):

PR #1011, opened 2021-02-09
Mnash 5154 skeleton operations

m-nash:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/5154


src/AutoRest.CSharp.V3/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -71,12 +72,26 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(Task<CodeModel> codeModel
                 project.AddGeneratedFile($"{responseHeaderModel.Type.Name}.cs", headerModelCodeWriter.ToString());
             }
 
-            foreach (var client in context.Library.Clients)
+            if (!context.Configuration.AzureArm)
```
chamons, line 75: We have an:

```
if (!Foo)
{
}
if (foo)
{
}
```

Any reason we couldn't combine them into an if/else?

src/AutoRest.CSharp.V3/Output/Models/ResourceOperation.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private const string ClientSuffixValue = "Client";
+        private const string OperationsSuffixValue = "Operations";
+        private OperationGroup _operationGroup;
+
+        public ResourceOperation(OperationGroup operationGroup, BuildContext context)
+            : base(context)
+        {
+            _operationGroup = operationGroup;
+            var clientPrefix = GetClientPrefix(operationGroup.Language.Default.Name);
```
chamons, line 24:
```suggestion
            string clientPrefix = GetClientPrefix(operationGroup.Language.Default.Name);
```

No var when right hand doesn't show type.

chamons (approved):
Seems reasonable enough. I'll let @pakrym comment further if he's around.

pakrym:
Can you give some context about the goal of the change?

pakrym:
You would need to run `.\eng\Generate.ps1` to re-generate all affected sources.

chamons:
I believe this is one of a number of refactors that he's doing in a branch to introduce the new style management APIs.

image

I detailed PR message for each would be helpful though.

src/AutoRest.CSharp/Output/Models/Types/TypeProvider.cs
```diff
@@ -10,7 +10,7 @@ namespace AutoRest.CSharp.Output.Models.Types
 {
     internal abstract class TypeProvider
     {
-        private readonly BuildContext _context;
+        protected readonly BuildContext _context;
```
pakrym, line 13: nit: convert to a property.

src/AutoRest.CSharp/Properties/launchSettings.json
```diff
@@ -24,17 +32,13 @@
...
     },
-    "Azure.Storage.Tables": {
-      "commandName": "Project",
-      "commandLineArgs": "--standalone $(SolutionDir)\\samples\\Azure.Storage.Tables\\Generated"
-    },
-    "azure-parameter-grouping": {
+    "Azure.ResourceManager.Compute": {
       "commandName": "Project",
-      "commandLineArgs": "--standalone $(SolutionDir)\\test\\TestServerProjects\\azure-parameter-grouping\\Generated"
+      "commandLineArgs": "--standalone $(SolutionDir)\\test\\TestProjects\\Azure.ResourceManager.Compute\\Generated"
```
pakrym, line 37: We usually don't put actual service swaggers under `TestProjects`. `TestProjects` are hand-crafted swagger files to illustrate particular features and write tests against them. I recommend samples directory.

pakrym (approved):

PR #1019, opened 2021-02-10
Add PrivateAssets="All" to the package reference

ShivangiReja:


src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -39,7 +39,7 @@ internal class CSharpProj : IPlugin
   </PropertyGroup>
 
   <ItemGroup>
-	  <PackageReference Include=""Microsoft.Azure.AutoRest.CSharp"" Version=""{0}"" />
+	  <PackageReference Include=""Microsoft.Azure.AutoRest.CSharp"" Version=""{0}"" PrivateAssets=""All"" />
```
pakrym, line 42: tabs are still here :)

pakrym (approved):

src/AutoRest.CSharp/AutoRest/Plugins/CSharpProj.cs
ShivangiReja: I don't see it. Where?

pakrym: image

That arrow is a tab.

pakrym (approved):

PR #1022, opened 2021-02-11
Add a note about the default version switch

ShivangiReja:


readme.md
```diff
@@ -1,5 +1,9 @@
 # C# code generator for AutoRest V3
 
+We have updated the default generator to the new version `--v3`, which uses `@autorest/csharp` package and it will have few side effects:
```
pakrym, line 3: I don't think the formatting works 
image

readme.md
```diff
@@ -1,5 +1,9 @@
...
+    * It will generate code based on the [.NET SDK guidelines](https://azure.github.io/azure-sdk/dotnet_introduction.html), which will be totally different than the code generated by `--v2` version.
```
pakrym, line 4:
```
 `--v2` 
```
looks like a command line parameter that can be passed but it's not. Let's use V2 and V3

readme.md
ShivangiReja: Yeah! just fixed it.

pakrym (approved):

PR #1031, opened 2021-02-15
Update code for provider name

bquantump:


chamons (commented):
Could you provide a description of the work being done here, I'm out of the loop on it.

Also, would it be possible to add a test?

src/AutoRest.CSharp/Input/CodeModelPartials.cs
```diff
@@ -85,7 +85,10 @@ public string[] Formats
 
         public string? HeaderCollectionPrefix => TryGetValue("x-ms-header-collection-prefix", out object? value) ? value?.ToString() : null;
     }
-
+    internal partial class OperationGroup
+    {
+        public string? ProviderName { get; set; }
```
pakrym, line 90: Nobody seems to be reading this property.

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
             EnsureHeaderModels().TryGetValue(operation, out var model);
             return model;
         }
+
+        private void DecorateOperationGroup()
+        {
+            foreach (var operations in _codeModel.OperationGroups)
+            {
+                operations.ProviderName = _context.Configuration.OperationGroupMapping.ContainsKey(operations.Key) ? _context.Configuration.OperationGroupMapping[operations.Key] : ConstructOperationProviderName(operations);
```
pakrym, line 185: Typically we don't modify the CodeModel during the generation. If some logic needs to be applied to the output model it needs to go into the corresponding output model class (https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/Client.cs , https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/RestClient.cs , https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/Types/ObjectType.cs , etc.)

PR #1032, opened 2021-02-16
Update code to support adding and parsing provider name

bquantump:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=5393


bquantump:
Pavel Krymets FTE 1 hour ago Member
Typically we don't modify the CodeModel during the generation. If some logic needs to be applied to the output model it needs to go into the corresponding output model class (https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/Client.cs , https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/RestClient.cs , https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/Types/ObjectType.cs , etc.)

bquantump:
https://github.com/Azure/autorest.csharp/pull/1031#pullrequestreview-591319587

bquantump:
> Pavel Krymets FTE 1 hour ago Member
> Typically we don't modify the CodeModel during the generation. If some logic needs to be applied to the output model it needs to go into the corresponding output model class (https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/Client.cs , https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/RestClient.cs , https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Models/Types/ObjectType.cs , etc.)

@pakrym This property will be needed by more than one output type for us (a new client and model output we will make). Is there any other good place to put this code given that?

pakrym:
Can you please add a description of what this feature would ultimately achieve? It would help figure out a good place for this to go. The other option might be to integrate this feature into some existing code to see it work.

bquantump:
> Can you please add a description of what this feature would ultimately achieve? It would help figure out a good place for this to go. The other option might be to integrate this feature into some existing code to see it work.

This feature is to support basically having a property somewhere (I put it in the code model for this PR) that has the full provider name for a give resource (1:1 with an operation group). For example, subnet would be Microsoft.Network/virtualNetworks/subnets. We can either get this from the swagger or the OperationsGroupMapping list in the configuration file also added in this PR.

This property will need to be used to generate multiple new output types. (our new model and client types). These types are not implemented in code yet and will be implemented in later PRs.

pakrym:
How do we know this code works?

https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Builders/BuilderHelpers.cs might be a reasonable place for this.

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+                    }
+                    else if (adding)
+                    {
+                        providerName += segment.TrimEnd('/');
+                    }
+                }
+            }
+            return providerName.TrimEnd('/');
+        }
+        private static List<string> GetPathSegments(string httpRequestUri)
```
pakrym, line 213: Can https://github.com/Azure/autorest.csharp/blob/a853571978087e1ecd64c5d14c37887da4aee92e/src/AutoRest.CSharp/Utilities/StringExtensions.cs#L67 be used here?

bquantump:
> How do we know this code works?
> 
> https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Output/Builders/BuilderHelpers.cs might be a reasonable place for this.

These are all static and we would not want to run this logic to parse the URI to get the provider name more than once, since the logic is not simple. That is why we want to really store somewhere that many output types can use.

src/AutoRest.CSharp/AutoRest/Plugins/Configuration.cs
```diff
@@ -48,7 +51,8 @@ public static Configuration GetConfiguration(IPluginCommunication autoRest)
                 autoRest.GetValue<bool?>("public-clients").GetAwaiter().GetResult() ?? false,
                 autoRest.GetValue<bool?>("model-namespace").GetAwaiter().GetResult() ?? true,
                 autoRest.GetValue<bool?>("head-as-boolean").GetAwaiter().GetResult() ?? false,
-                autoRest.GetValue<bool?>("skip-csproj-packagereference").GetAwaiter().GetResult() ?? false
+                autoRest.GetValue<bool?>("skip-csproj-packagereference").GetAwaiter().GetResult() ?? false,
+                autoRest.GetValue<JsonElement?>("OperationGroupMapping").GetAwaiter().GetResult()
```
pakrym, line 55: nit: conform to the autorest naming rules. `operation-group-mapping`

src/AutoRest.CSharp/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -90,7 +90,8 @@ private static Configuration LoadConfiguration(string basePath, string json)
                 root.GetProperty(nameof(Configuration.PublicClients)).GetBoolean(),
                 root.GetProperty(nameof(Configuration.ModelNamespace)).GetBoolean(),
                 root.GetProperty(nameof(Configuration.HeadAsBoolean)).GetBoolean(),
-                root.GetProperty(nameof(Configuration.SkipCSProjPackageReference)).GetBoolean()
+                root.GetProperty(nameof(Configuration.SkipCSProjPackageReference)).GetBoolean(),
+                root.GetProperty(nameof(Configuration.OperationGroupMapping))
```
m-nash, line 94: Can we add this to one of the readme examples / tests so that we see how it is expected to be applied

src/AutoRest.CSharp/Input/CodeModelPartials.cs
```diff
@@ -85,7 +85,10 @@ public string[] Formats
 
         public string? HeaderCollectionPrefix => TryGetValue("x-ms-header-collection-prefix", out object? value) ? value?.ToString() : null;
     }
-
+    internal partial class OperationGroup
+    {
+        public string? ProviderName { get; set; }
+    }
```
m-nash, line 91: add newline here

pakrym:
How do we know this code works?

>These are all static and we would not want to run this logic to parse the URI to get the provider name more than once, since the logic is not simple

I doubt it would noticeably affect the generation time. We can cache it if perf becomes a problem and we would have a better feel of what other moving pieces are.

src/AutoRest.CSharp/Input/CodeModelPartials.cs
```diff
@@ -85,7 +85,10 @@ public string[] Formats
 
         public string? HeaderCollectionPrefix => TryGetValue("x-ms-header-collection-prefix", out object? value) ? value?.ToString() : null;
     }
-
+    internal partial class OperationGroup
+    {
+        public string? ProviderName { get; set; }
```
m-nash, line 90: should this be ResourceType?  to keep parallels with terminology in arm?

src/AutoRest.CSharp/Input/CodeModelPartials.cs
```diff
@@ -168,6 +171,7 @@ internal partial class Paging
         public Operation? NextLinkOperation { get; set; }
     }
 
+
```
m-nash, line 174: extra newline not needed

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
             EnsureHeaderModels().TryGetValue(operation, out var model);
             return model;
         }
+
+        private void DecorateOperationGroup()
+        {
+            foreach (var operations in _codeModel.OperationGroups)
+            {
+                operations.ProviderName = _context.Configuration.OperationGroupMapping.ContainsKey(operations.Key) ? _context.Configuration.OperationGroupMapping[operations.Key] : ConstructOperationProviderName(operations);
```
m-nash, line 185: Use TryGetValue here instead

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+            }
+        }
+
+        private string ConstructOperationProviderName(OperationGroup operations)
+        {
+
+            string? providerName = "";
```
m-nash, line 192: make this a string instead of string?

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+                        adding = true;
+                        providerName = segment.Substring("providers".Length + 2).TrimEnd('/');
+                    }
+                    else if (adding)
+                    {
+                        providerName += segment.TrimEnd('/');
+                    }
+                }
+            }
+            return providerName.TrimEnd('/');
```
m-nash, line 211: shouldn't this be where we Log an error? since at this point we will not have the resourcetype for this operations group?

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+
+            string? providerName = "";
+            var request = GetBestMethod(operations);
+            bool adding = false;
+            if (request != null)
+            {
+                foreach (var segment in GetPathSegments(request.Path))
+                {
+                    var asSplit = segment.Split('/', StringSplitOptions.RemoveEmptyEntries);
+                    if (asSplit?.Length > 1 && asSplit.First().Equals("providers"))
```
m-nash, line 200: make "providers" a readonly constant

m-nash, line 200: gonna guess that asSplit[0] is more efficent than asSplit.First() since we know we got back an array this is equally as safe

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+                    {
+                        adding = true;
+                        providerName = segment.Substring("providers".Length + 2).TrimEnd('/');
```
m-nash, line 203: can we simply return here?

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
m-nash, line 213: missing new line here

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+                else
+                {
+                    canidate += ch;
+                }
+            }
+            if (canidate != "" && canidate != "/")
+            {
+                seg.Add(canidate);
+            }
+            return seg;
```
m-nash, line 241: doesn't appear this is getting path segments, it appears its getting all reference portions of the path

Can we update the name?

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+        }
+
+        private HttpRequest? GetBestMethod(OperationGroup operations)
```
m-nash, line 244: rename param to operationGroup

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+                {
+                    if (serviceRequest.Protocol.Http is HttpRequest httpRequest)
+                    {
+                        if (httpRequest.Method == HttpMethod.Put)
+                        {
+                            return httpRequest;
+                        }
+                        else if (httpRequest.Method == HttpMethod.Delete)
+                        {
+                            canidate = httpRequest;
```
m-nash, line 259: why do we not just return here?

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +177,95 @@ public RestClient FindRestClient(OperationGroup operationGroup)
...
+                        }
+                        else if (httpRequest.Method == HttpMethod.Patch)
+                        {
+                            canidate ??= httpRequest;
```
m-nash, line 263: why do we not just return here?

m-nash (changes requested):

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
bquantump: What if there is a PUT? 
 PUT->Delete->Patch is the order we want

src/AutoRest.CSharp/AutoRest/Communication/StandaloneGeneratorRunner.cs
ShivangiReja, line 94: You're reading `OperationGroupMapping` property but not writing [here](https://github.com/Azure/autorest.csharp/blob/3566a512d973f526805c99aa5bdb48e744c1caf5/src/AutoRest.CSharp/AutoRest/Communication/StandaloneGeneratorRunner.cs#L59). I think If you don't write it, it will throw an error when you will do F5 for any project.

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
bquantump: Yes, but it looks like the CI attempts to test other RPs, like search for example, and this will throw an exception for all the cases that we have not custom defined yet. That is, not defined in the readme.md for that RP for those resources we can't parse (for example resources that only have gets)

src/AutoRest.CSharp/AutoRest/Communication/StandaloneGeneratorRunner.cs
bquantump: Updated

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
bquantump: No we this is just the point to start appending to

For example: "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/virtualNetworks/{virtualNetworkName}/subnets/{subnetName}/PrepareNetworkPolicies"

we start adding to the last full provider
Microsoft.Network/virtualNetworks/

and end up adding 
subnets/

and then

PrepareNetworkPolicies/

for 

Microsoft.Network/virtualNetworks/subnets/PrepareNetworkPolicies/

strip the last '/'

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
bquantump: msg will be
           
            if (request == null || resourceName== "")
            {
                throw new ArgumentException("Could not set ResourceType for operations group " + operationsGroup.Key +
                "\nPlease try setting this value for this operations in the readme.md for this swagger in the operation-group-mapping section");
            }

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -20,11 +23,16 @@ internal class OutputLibrary
         private Dictionary<OperationGroup, RestClient>? _restClients;
         private Dictionary<Operation, LongRunningOperation>? _operations;
         private Dictionary<Operation, ResponseHeaderGroupType>? _headerModels;
+        private readonly string _proivders = "providers";
```
m-nash, line 26: should be const and const should start with Upper case

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -170,5 +178,97 @@ public RestClient FindRestClient(OperationGroup operationGroup)
             EnsureHeaderModels().TryGetValue(operation, out var model);
             return model;
         }
+
+        private void DecorateOperationGroup()
+        {
+            foreach (var operationsGroup in _codeModel.OperationGroups)
+            {
+                string? resourceType = "";
```
m-nash, line 186: why string?

Also for out params you don't need to initialize
```suggestion
                string resourceType;
```

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
m-nash: yes, but I believe the if order covers that, you are returning on first match and matching in the desired order

you can have 3 of these in the right order

if(httpRequest.Method == HttpMethod.Put)
    return httpRequest;

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
m-nash: same as above

test/TestProjects/OperationGroupMappings/Configuration.json
```diff
@@ -0,0 +1,5 @@
+{"OutputFolder":".","Namespace":"OperationGroupMappings","LibraryName":null,"SharedSourceFolders":["..\\..\\..\\src\\assets\\Generator.Shared","..\\..\\..\\src\\assets\\Azure.Core.Shared"],"AzureArm":true,"PublicClients":true,"ModelNamespace":true,"HeadAsBoolean":false,"SkipCSProjPackageReference":true,
```
m-nash, line 1: for other examples this is inside the generated folder, any reason its here for this one?

test/TestProjects/OperationGroupMappings/CodeModel.yaml
```diff
@@ -0,0 +1,948 @@
+!<!CodeModel>
```
m-nash, line 1: for other examples this is inside the generated folder, any reason its here for this one?

test/TestProjects/OperationGroupMappings/OperationGroupMappings.json
```diff
@@ -0,0 +1,325 @@
...
+    },
+    "ApiVersionParameter": {
+      "name": "api-version",
+      "in": "query",
+      "required": true,
+      "type": "string",
+      "description": "Client Api Version."
+    }
+  }
+}
```
m-nash, line 325: missing newline

test/TestProjects/OperationGroupMappings/OperationGroupMappings.json
```diff
@@ -0,0 +1,325 @@
+{
```
m-nash, line 1: for the other testprojects this file is in the generated folder.

m-nash (changes requested):

src/AutoRest.CSharp/AutoRest/Plugins/Configuration.cs
```diff
@@ -21,9 +23,9 @@ public Configuration(string outputFolder, string? ns, string? name, string[] sha
             ModelNamespace = modelNamespace;
             HeadAsBoolean = headAsBoolean;
             SkipCSProjPackageReference = skipCSProjPackageReference;
+            OperationGroupMapping = operationGroupMapping.ToString() == "" ? new Dictionary<string, string>() : JsonSerializer.Deserialize<Dictionary<string, string>>(operationGroupMapping.ToString());
```
pakrym, line 26: `operationGroupMapping.ToString() == "" ` is this a null check?

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
bquantump: yeah, but the operations can be in any order, 

so could get {delete, patch, put} in

(var operation in operationsGroup.Operations)


and we would return delete if we returned on the first entry in the operations list

hence, need check the rest to see if there is a put.

samples/Azure.Management.Storage/readme.md
```diff
@@ -15,4 +15,10 @@ payload-flattening-threshold: 2
 
 modelerfour:
   lenient-model-deduplication: true
+
+operation-group-mapping:
```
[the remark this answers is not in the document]
bquantump: so better naming you mean? Like operation-to-resourcetype? Something along those lines?

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
bquantump: because

image

test/TestProjects/OperationGroupMappings/Configuration.json
bquantump: image

Modeled it after inheritance since the standalone needs the codemodel and configuration to be there

image

test/TestProjects/OperationGroupMappings/OperationGroupMappings.json
bquantump: Looks like they have it not in the generated folders in test/testProjects

image

src/AutoRest.CSharp/AutoRest/Plugins/Configuration.cs
bquantump: No, just checking if string is empty so that we don't try to sterilize an empty string. This is the case if default is set.

bquantump:
> How do we know this code works?
> 
> > These are all static and we would not want to run this logic to parse the URI to get the provider name more than once, since the logic is not simple
> 
> I doubt it would noticeably affect the generation time. We can cache it if perf becomes a problem and we would have a better feel of what other moving pieces are.

added sample operationsOverrideMappings and update storage in testservers

pakrym:
>added sample operationsOverrideMappings and update storage in testservers

I don't see the `ResourceType` used anywhere in the generated code.

m-nash (changes requested):

bquantump:
@pakrym 

Why is this https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/imds/data-plane/readme.md

have azure-arm: true

when it is data-pane?

pakrym:
>Why is this Azure/azure-rest-api-specs@d9cf7c7/specification/imds/data-plane/readme.md
>have azure-arm: true
>when it is data-pane?

No idea.

m-nash (approved):

PR #1038, opened 2021-02-18
[Draft] generation empty mgmt resource

allenjzhang:


samples/Azure.ResourceManager.Sample/Generated/Models/AvailabilitySets.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
m-nash, line 1: I don't think these belong in the Model folder, they should be in the same level as the [Resource]Operations classes

m-nash (changes requested):

Issue #877, opened 2020-08-21
Generate public constructors for service client


ShivangiReja:
* Generate public constructors for service client which will take only the parameters required to connect to the service.
* Provide constructor overloads that allow specifying additional options such as credentials, a custom HTTP pipeline, or advanced client options.

PR #1046, opened 2021-02-22
Generate public client constructors and options

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/877


src/AutoRest.CSharp/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -75,10 +75,19 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(Task<CodeModel> codeModel
 
             foreach (var client in context.Library.Clients)
             {
-                var codeWriter = new CodeWriter();
-                clientWriter.WriteClient(codeWriter, client, context.Configuration);
+                var credentialTypes = context.Configuration.CredentialTypes;
+                if (credentialTypes.Contains("TokenCredential", StringComparer.OrdinalIgnoreCase) ||
```
pakrym, line 79: Let's not condition ClientOptions on the credentials. I don't think it's an obvious relationship. Using PublicClients might be a bit better.

src/AutoRest.CSharp/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -75,10 +75,19 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(Task<CodeModel> codeModel
...
+                    credentialTypes.Contains("AzureKeyCredential", StringComparer.OrdinalIgnoreCase))
+                {
+                    var codeWriter = new CodeWriter();
+                    ClientWriter.WriteClientOptions(codeWriter, client, context);
+                    project.AddGeneratedFile($"{client.Type.Name}Options.cs", codeWriter.ToString());
```
pakrym, line 84: Would this result in a clientoptions type per client? We only need 1 for the entire library.

src/AutoRest.CSharp/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -120,6 +129,13 @@ public async Task<bool> Execute(IPluginCommunication autoRest)
                 return CodeModelSerialization.DeserializeCodeModel(codeModelYaml);
             });
 
+            if (configuration.CredentialTypes.Contains("TokenCredential", StringComparer.OrdinalIgnoreCase) &&
+                configuration.CredentialScopes == null ||
+                configuration.CredentialScopes != null && configuration.CredentialScopes.Length < 1 )
+            {
+                await autoRest.Fatal("You are using TokenCredential wihtout passing in any credential-scopes.");
```
pakrym, line 136: should we return early here?

src/AutoRest.CSharp/AutoRest/Plugins/Configuration.cs
```diff
@@ -34,10 +36,13 @@ public Configuration(string outputFolder, string? ns, string? name, string[] sha
         public bool ModelNamespace { get; }
         public bool HeadAsBoolean { get; }
         public bool SkipCSProjPackageReference { get; }
+        public string[] CredentialTypes { get; }
+        public string[]? CredentialScopes { get; }
```
pakrym, line 40: Don't make the array nullable, default it to empty instead. Would avoid having to check for null at all callsites.

src/AutoRest.CSharp/AutoRest/Plugins/Configuration.cs
```diff
@@ -34,10 +36,13 @@ public Configuration(string outputFolder, string? ns, string? name, string[] sha
...
         public static string ProjectRelativeDirectory = "../";
 
         public static Configuration GetConfiguration(IPluginCommunication autoRest)
         {
+            string[] defaultCredentialType = { "None"};
```
pakrym, line 45: Is "None" different than an empty array?

src/AutoRest.CSharp/Generation/Writers/ClientWriter.cs
```diff
@@ -18,6 +22,65 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal class ClientWriter
     {
+
+        public static void WriteClientOptions(CodeWriter writer, Client client, BuildContext context)
```
pakrym, line 26: Let's have a separate writer per type we produce and move this into ClientOptionsWriter.

src/AutoRest.CSharp/Generation/Writers/ClientWriter.cs
```diff
@@ -134,12 +200,71 @@ private void WriteClientFields(CodeWriter writer, Client client)
...
+                    string scope = "";
+                    if (configuration.CredentialScopes != null)
+                    {
+                        if (configuration.CredentialScopes.Length == 1)
+                        {
+                            scope = $"\"{configuration.CredentialScopes[0]}\"";
+                        }
+                        else
+                        {
+                            string scopesArr = "{\"" + string.Join("\", \"", configuration.CredentialScopes) + "\"}";
```
pakrym, line 258: Don't concatenate strings, use the writer with the literal flag to get escaping.

src/AutoRest.CSharp/Generation/Writers/ClientWriter.cs
```diff
@@ -134,12 +200,71 @@ private void WriteClientFields(CodeWriter writer, Client client)
...
+                writer.Append($"public {client.Type.Name:D}({typeof(Uri)} {EndpointVariable}, {typeof(AzureKeyCredential)} {CredentialVariable}, {client.Type.Name}Options {OptionsVariable} = null)");
+                using (writer.Scope())
+                {
+                    writer.Line($"{typeof(Argument)}.AssertNotNull({EndpointVariable}, nameof({EndpointVariable}));");
+                    writer.Line($"{typeof(Argument)}.AssertNotNull({CredentialVariable}, nameof({CredentialVariable}));");
+                    writer.Line();
+
+                    writer.Line($"{OptionsVariable} ??= new {client.Type.Name}Options();");
+                    writer.Line($"{ClientDiagnosticsField} = new {typeof(ClientDiagnostics)}({OptionsVariable});");
+                    writer.Line($"{PipelineField} = {typeof(HttpPipelineBuilder)}.Build({OptionsVariable}, new {typeof(AzureKeyCredentialPolicy)}({CredentialVariable}, \"api-key\"));");
```
pakrym, line 227: Do we care about configuring this name?

src/AutoRest.CSharp/AutoRest/Plugins/CSharpGen.cs
ShivangiReja: Ahh, ok. I thought we are using ClientOptions only for constructors with `TokenCredential` and `AzureKeyCredential` but I got it now, user can create an instance of ClientOptions and use it anywhere as per their requirements.

test/AutoRest.TestServer.Tests/PublicClientCtorTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using AutoRest.TestServer.Tests.Infrastructure;
+using Azure;
+using Azure.Core;
+using NUnit.Framework;
+using PublicClientCtor;
+
+namespace AutoRest.TestServer.Tests
+{
+    public class PublicClientCtorTests : TestServerTestBase
+    {
```
pakrym, line 16: Let's add tests that verify that intended headers are being sent.
You can use https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/HeaderCollectionPrefixTests.cs#L16 as a sample of how to host and inprocess server.

src/AutoRest.CSharp/AutoRest/Plugins/CSharpGen.cs
ShivangiReja: Updated to generate on client options for the entire library.

src/AutoRest.CSharp/AutoRest/Plugins/Configuration.cs
ShivangiReja: Nope, changed it to empty array.

src/AutoRest.CSharp/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -74,12 +90,24 @@ private static Configuration LoadConfiguration(string basePath, string json)
             JsonDocument document = JsonDocument.Parse(json);
             var root = document.RootElement;
             var sharedSourceFolders = new List<string>();
+            var CredentialTypes = new List<string>();
```
pakrym, line 93: nit variable name casing

src/AutoRest.CSharp/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -73,10 +74,19 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(Task<CodeModel> codeModel
                 project.AddGeneratedFile($"{responseHeaderModel.Type.Name}.cs", headerModelCodeWriter.ToString());
             }
 
+            if (configuration.PublicClients && !configuration.AzureArm && context.Library.Clients.Count() > 0)
```
pakrym, line 77: Do we have libraries without any clients?

src/AutoRest.CSharp/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -120,6 +130,13 @@ public async Task<bool> Execute(IPluginCommunication autoRest)
                 return CodeModelSerialization.DeserializeCodeModel(codeModelYaml);
             });
 
+            if (configuration.CredentialTypes.Contains("TokenCredential", StringComparer.OrdinalIgnoreCase) &&
+                configuration.CredentialScopes.Length < 1 )
+            {
+               await autoRest.Fatal("You are using TokenCredential wihtout passing in any credential-scopes.");
```
pakrym, line 136: nit indentation

src/AutoRest.CSharp/Generation/Writers/ClientOptionsWriter.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        private static string ToVersionProperty(string s)
+        {
+            return "V" + s.Replace(".", "_").Replace('-', '_');
+        }
+
+        public static string GetClientOptionsPrefix(string name)
+        {
+            if (name.EndsWith(OperationsSuffixValue) && name.Length >= OperationsSuffixValue.Length)
```
pakrym, line 78: We seem to start having a lot of copies of this code.

src/AutoRest.CSharp/AutoRest/Plugins/CSharpGen.cs
ShivangiReja: Yes, https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/mediaservices/Azure.Media.Analytics.Edge/src/Generated

src/AutoRest.CSharp/Generation/Writers/ClientOptionsWriter.cs
ShivangiReja: Where should we put this common method?

src/AutoRest.CSharp/Generation/Writers/ClientWriter.cs
```diff
@@ -134,12 +141,122 @@ private void WriteClientFields(CodeWriter writer, Client client)
...
+                    writer.Line();
+
+                    writer.Line($"{OptionsVariable} ??= new {clientOptionsName}ClientOptions();");
+                    writer.Line($"{ClientDiagnosticsField} = new {typeof(ClientDiagnostics)}({OptionsVariable});");
+                    var scopesParam = new CodeWriterDeclaration("scopes");
+                    var credentialScopes = context.Configuration.CredentialScopes;
+                    string scope = "";
+                    if (credentialScopes.Length == 1)
+                    {
+                        scope = $"\"{credentialScopes[0]}\"";
```
pakrym, line 227: I think you can remove this branch

ShivangiReja: Yeah, I added this if condition, so that in case of 1 scope, it would call [this](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs#L25) method

src/AutoRest.CSharp/Output/Models/Client.cs
```diff
@@ -119,5 +121,54 @@ private IEnumerable<ClientMethod> BuildMethods()
...
+            }
+
+            return parameters;
+        }
+
+        public IReadOnlyCollection<Parameter> GetClientConstructorParameters(CSharpType credentialType)
+        {
+            List<Parameter> parameters = new List<Parameter>();
+
+            var requiredParameters = GetRequiredParameters().ToArray();
```
pakrym, line 157: You don't need ToArray()

src/AutoRest.CSharp/Generation/Writers/ClientWriter.cs
pakrym: it would still result in an array allocation so it's okay to simplify the generator code

src/AutoRest.CSharp/Generation/Writers/ClientOptionsWriter.cs
ShivangiReja: Removed this copy and now using the same method from ClientBase.

src/AutoRest.CSharp/Output/Models/Client.cs
```diff
@@ -119,5 +121,54 @@ private IEnumerable<ClientMethod> BuildMethods()
...
+
+            var credentialParam = new Parameter(
+                "credential",
+                "A credential used to authenticate to an Azure Service.",
+                credentialType,
+                null,
+                true);
+            parameters.Add(credentialParam);
+
+            var optionalParameters = GetOptionalParameters().ToArray();
```
pakrym, line 168: You don't need ToArray() here as well.

src/AutoRest.CSharp/Output/Models/Client.cs
```diff
@@ -119,5 +121,54 @@ private IEnumerable<ClientMethod> BuildMethods()
                 }
             }
         }
+
+        private IEnumerable<Parameter> GetRequiredParameters()
+        {
+            List<Parameter> parameters = new List<Parameter>();
+
+            if (RestClient.Parameters.TryGetRequiredParameters(out var requiredParameters))
```
pakrym, line 129: Can we use the `p.DefaultValue` check only?

ShivangiReja: Yeah, it would work.

pakrym: So do we still need the TryGetRequiredParameters call?

ShivangiReja: Ahh, updated `GetOptionalParameters()` method, forgot to update this one.

pakrym (approved):

Issue #947, opened 2020-11-17
Add System.Text.Json converter to models


pakrym:
That would forward calls to existing Deserialize/Serialize methods.

JoshLove-msft:
This would be interesting for Event Grid as it would allow deserializing System events using BinaryData by default.

PR #1063, opened 2021-03-01
Add System.Text.Json converter to models

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/947


src/AutoRest.CSharp/Generation/Writers/SerializationWriter.cs
```diff
@@ -104,6 +106,30 @@ private void WriteObjectSerialization(CodeWriter writer, ObjectType model)
...
+                    {
+                        WriteCustomJsonConverter(model, writer);
+                    }
+                }
+            }
+        }
+
+        private void WriteCustomJsonConverter(ObjectType model, CodeWriter writer)
+        {
+            writer.Append($"internal partial class {model.Declaration.Name}Converter : {typeof(JsonConverter)}<{model.Declaration.Name}>");
```
pakrym, line 120: `<{model.Declaration.Name}>")` use model.Type here.

test/TestProjects/ModelWithConverterUsage/Generated/Models/ModelStruct.cs
```diff
@@ -0,0 +1,33 @@
...
+
+#nullable disable
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace ModelWithConverterUsage.Models
+{
+    /// <summary> The ModelStruct. </summary>
+    [JsonConverter(typeof(ModelStructConverter))]
```
pakrym, line 14: Should we move this attribute to the other part of the file? The same where ModelStructConverter is?

src/AutoRest.CSharp/Output/Models/Types/ObjectType.cs
```diff
@@ -86,6 +86,8 @@ public ObjectType(ObjectSchema objectSchema, BuildContext context): base(context
 
         public ObjectTypeProperty[] Properties => _properties ??= BuildProperties().ToArray();
 
+        public SchemaTypeUsage SchemaTypeUsage => _usage;
```
pakrym, line 89: I wonder if we should put this value onto the `JsonSerialization` inside the `        public ObjectSerialization[] Serializations => _serializations ??= BuildSerializations();
`

pakrym: Or better follow the pattern of IncludeSerializer method https://github.com/Azure/autorest.csharp/blob/4c0f9cf31af56decb03c67fdf99eb516c8ae81ae/src/AutoRest.CSharp/Output/Models/Types/ObjectType.cs#L306-L307

pakrym (approved):

test/TestProjects/ModelWithConverterUsage/Generated/Models/ModelStruct.cs
ShivangiReja: Yeah, this looks better.

pakrym (approved):

PR #1076, opened 2021-03-10
[Mgmt Track 2] 5156 generate empty resource autorest change

allenjzhang:


allenjzhang:
We have a problem here. Since we keyed on PUT for the resource model, both compute and storage had different creation model than the actual model. That had forced me to add renaming and causing conflict. This would create problems in our unified API model. Let's discuss in .NET sync.

samples/Azure.Management.Storage/Generated/Configuration.json
```diff
@@ -1 +1,36 @@
-{"OutputFolder":".","Namespace":"Azure.Management.Storage","LibraryName":null,"SharedSourceFolders":["..\\..\\..\\artifacts\\bin\\AutoRest.CSharp\\Debug\\netcoreapp3.0\\Generator.Shared","..\\..\\..\\artifacts\\bin\\AutoRest.CSharp\\Debug\\netcoreapp3.0\\Azure.Core.Shared"],"AzureArm":true,"PublicClients":true,"ModelNamespace":true,"HeadAsBoolean":true,"SkipCSProjPackageReference":true,"CredentialTypes":[],"CredentialScopes":[],"CredentialHeaderName":"api-key","OperationGroupToResourceType":{"Operations":"Microsoft.Storage/operations","Skus":"Microsoft.Storage/skus","Usages":"Microsoft.Storage/locations/usages","PrivateLinkResources":"Microsoft.Storage/storageAccounts/privateLinkResources"},"OperationGroupToResource":{"Operations":"Operation","Skus":"Sku","Usages":"Usage","PrivateLinkResources":"PrivateLinkResource"},"ResourceRename":{"BlobServiceProperties":"BlobService","FileServiceProperties":"FileService","StorageAccountCreateParameters":"StorageAccount"}}
\ No newline at end of file
+{
```
m-nash, line 1: This will fail code checks.  This is a generated file and it won't match what the regeneration does

samples/Azure.ResourceManager.Sample/Generated/Configuration.json
```diff
@@ -1 +1,42 @@
...
\ No newline at end of file
+{
```
m-nash, line 1: same comment as above

src/AutoRest.CSharp/AutoRest.CSharp.csproj
```diff
@@ -61,15 +61,15 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <_GeneratedPropsFileLines Include="&lt;Project>" />
```
m-nash, line 64: why is this needed?

m-nash (changes requested):

src/AutoRest.CSharp/AutoRest.CSharp.csproj
allenjzhang: The changes were done by VS for proper XML escaping formatting. Let's keep it as other folks's save probably would have trigger this later anyway.

m-nash (approved):

Issue #1078, opened 2021-03-10
Do not apply "x-csharp-usage": "converter" recursively, apply only if it is explicitly defined for a model 


ShivangiReja:
[Here](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/ModelWithConverterUsage/ModelWithConverterUsage.json#L93) "x-csharp-usage": "converter" is defined for ModelClass but it is also applied to it's property [Product](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/ModelWithConverterUsage/ModelWithConverterUsage.json#L90). 

Fix this behavior to apply converter only for the class you have defined "x-csharp-usage": "converter" but not for its properties, unless it is explicitly defined for it’s properties.

PR #1079, opened 2021-03-11
Do not apply "x-csharp-usage": "converter" recursively, apply only if it is explicitly defined for a model

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/1078


test/TestProjects/ModelWithConverterUsage/Generated/Models/Product.Serialization.cs
```diff
@@ -36,18 +33,5 @@ internal static Product DeserializeProduct(JsonElement element)
             }
             return new Product(constProperty);
         }
-
```
pakrym, line 39: Add a test.

src/AutoRest.CSharp/Input/SchemaUsageProvider.cs
```diff
@@ -18,7 +18,9 @@ public SchemaUsageProvider(CodeModel codeModel)
 
                 if (usage != null)
                 {
-                    Apply(objectSchema, (SchemaTypeUsage) Enum.Parse(typeof(SchemaTypeUsage), usage, true));
+                    var schemaTypeUsage = (SchemaTypeUsage)Enum.Parse(typeof(SchemaTypeUsage), usage, true);
+                    var recurse = schemaTypeUsage.HasFlag(SchemaTypeUsage.Converter) ? false : true;
```
pakrym, line 22: This is not correct, all other usages need to be applied recursively, only the SchemaTypeUsage.Converter shouldn't

pakrym: I think you need to change the `Apply` method and clear the `SchemaTypeUsage.Converter` flag after it was applied to the initial schema.

ShivangiReja: Current changes also work fine without changing `Apply` method. Let me know if I'm missing something.

src/AutoRest.CSharp/Input/SchemaUsageProvider.cs
```diff
@@ -18,7 +18,18 @@ public SchemaUsageProvider(CodeModel codeModel)
...
-                    Apply(objectSchema, (SchemaTypeUsage) Enum.Parse(typeof(SchemaTypeUsage), usage, true));
+                    var schemaTypeUsage = (SchemaTypeUsage)Enum.Parse(typeof(SchemaTypeUsage), usage, true);
+
+                    if (schemaTypeUsage.HasFlag(SchemaTypeUsage.Converter))
+                    {
+                        Apply(objectSchema, SchemaTypeUsage.Converter, false);
+                        schemaTypeUsage &= ~SchemaTypeUsage.Converter;
+                    }
+
+                    if (!schemaTypeUsage.Equals(default(SchemaTypeUsage)))
```
pakrym, line 29: WHy do we need this check?

pakrym (approved):

src/AutoRest.CSharp/Input/SchemaUsageProvider.cs
ShivangiReja: I added this check if user only passes "x-csharp-usage": "converter", then this check will prevent us calling `Apply` method again. Although there will be no effect even if we remove this check because `Apply` method will return from [here](https://github.com/Azure/autorest.csharp/blob/194731d242ba2cfa55886aa8ed70493f41583d32/src/AutoRest.CSharp/Input/SchemaUsageProvider.cs#L105).

pakrym: Yeah, let's simplify logic here, the `Apply` should already handle this.

PR #1093, opened 2021-03-23
Parent Decoration 

bquantump:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=5504


samples/Azure.Management.Storage/readme.md
```diff
@@ -30,5 +30,9 @@ operation-group-to-resource:
 resource-rename:
    BlobServiceProperties: BlobService
    FileServiceProperties: FileService
+resource-to-parent:
```
m-nash, line 33: the name implies the key is the name of the resource however FileShares is actually FileShare as we see in the resource rename

is this really intended to be an operation-group-to-parent?

src/AutoRest.CSharp/Input/CodeModelPartials.cs
```diff
@@ -96,6 +96,8 @@ internal partial class OperationGroup
         public string ResourceType { get; set; }
         public string Resource { get; set; }
         public bool IsTenantResource { get; set; }
+        public bool IsExtensionResource { get; set; }
+        public string? Parent { get; set; }
```
m-nash, line 100: doesn't everything need a parent?

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -331,7 +344,17 @@ private void DecorateOperationGroup()
                 {
                     operationsGroup.Resource = nameOverride;
                 }
+
             }
+            //now that we have resolved all operations groups to resource types above, can try to solve for the parent
+            foreach (var operationsGroup in _codeModel.OperationGroups)
+            {
+                if (operationsGroup.Parent != null && !ResourceTypes.Contains(operationsGroup.Parent))
```
m-nash, line 352: not sure this is ever possible since line 338 always adds if it wasn't null

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
```diff
@@ -322,7 +329,13 @@ private void DecorateOperationGroup()
                 MapHttpMethodToOperation(operationsGroup);
                 string? resourceType;
                 operationsGroup.ResourceType = _context.Configuration.OperationGroupToResourceType.TryGetValue(operationsGroup.Key, out resourceType) ? resourceType : ResourceTypeBuilder.ConstructOperationResourseType(operationsGroup);
+                ResourceTypes.Add(operationsGroup.ResourceType);
                 operationsGroup.IsTenantResource = TenantDetection.IsTenantOnly(operationsGroup);
+                operationsGroup.IsExtensionResource = ExtensionDetection.IsExtension(operationsGroup);
+                string? parent;
+                operationsGroup.Parent = _context.Configuration.ResourceToParent.TryGetValue(operationsGroup.Key, out parent) ? parent : ((operationsGroup.IsTenantResource || operationsGroup.IsExtensionResource) ? null : ParentDetection.GetParent(operationsGroup));
```
m-nash, line 336: shouldn't we throw if its null?

src/AutoRest.CSharp/Output/Models/Types/ParentDetection.cs
```diff
@@ -0,0 +1,94 @@
...
+            if (operations.TryGetValue(HttpMethod.Get, out requests)) // optimized which get to return here
+            {
+                return (HttpRequest?)requests[0].Protocol?.Http;
+            }
+            return null;
+        }
+
+        private static ProviderSegment? GetFullProvider(List<ProviderSegment> providerSegments)
+        {
+            for (int i = providerSegments.Count - 1; i > -1; i--)
```
m-nash, line 58: can we set this as a property of the method so we don't need to loop each time?

src/AutoRest.CSharp/Output/Models/Types/ParentDetection.cs
```diff
@@ -0,0 +1,94 @@
...
+                throw new ArgumentException($"Could not set parent for operations group {operationGroup.Key}. Please add to readme.md");
+            }
+            return canidateParent;
+        }
+
+        public static HttpRequest? GetBestMethod(Dictionary<HttpMethod, List<ServiceRequest>> operations)
+        {
+            List<ServiceRequest>? requests;
+
+            if (operations.TryGetValue(HttpMethod.Put, out requests))
```
m-nash, line 37: Add a new TestProject with a swagger that has an operationgroup that hits each of these cases and assert the predetermined parent is chosen

src/AutoRest.CSharp/Output/Models/Types/ParentDetection.cs
```diff
@@ -0,0 +1,94 @@
...
+            return null;
+        }
+
+        private static string ParseMethodForParent(ProviderSegment fullProvider, string path, string resourceType)
+        {
+            //case 1:
+            // Microsoft.Network/virtualNetworks/ == lastFullProvider
+            // resourceType = Microsoft.Network/virtualNetworks
+            //
+            if (fullProvider.TokenValue.Trim('/').Equals(resourceType))
```
m-nash, line 74: unit tests

m-nash (changes requested):

src/AutoRest.CSharp/Output/Models/Types/OutputLibrary.cs
bquantump: it is check "parent" null not "Parent"

So if it was overridden, add it. If it was not, verify it was actually in the swagger

src/AutoRest.CSharp/Output/Models/Types/ParentDetection.cs
bquantump: https://github.com/AME-Redmond/codegen-test-code/blob/main/extensions-types/Program.cs

PR #1118, opened 2021-04-01
Handle long query parameters

pakrym:
Large long values were falling into the float overload and having scientific format.


chamons (approved):

ShivangiReja (approved):

kasobol-msft (approved):

mikeharder (changes requested):
I'm not sure if this is the correct fix.  Please don't merge until we have done more testing and investigation.

mikeharder (approved):
After more investigation I approve this fix.  All `double` values were falling into the `float` overload.  But only `double` values `>= 10,000,000` were being formatted with scientific notation.

PR #1119, opened 2021-04-01
Put in place a fork folder structure

m-nash:
- Common will contain all of the files that are intended to be used by any of the other 3 clients
- Dataplane, Mgmt, or Lowlevel will only contain the files they overload or add
- No files in Common should reference a concrete type like MgmtOutputLibrary
- This should make it easy for us going forward to identify scope of PRs coming in.
- Also removed bulk of the duplicate code that mgmt plane initially had

ADO item for tracking https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/5725


pakrym:
Can you please rename Dataplane to DataPlane and Lowlevel to LowLevel to be consistent with type names.

ShivangiReja:
Love this new structure :)

chamons (approved):
You mixed in a few non-movement changes in the PR, but I didn't see anything that wrong.

ShivangiReja (approved):

PR #1123, opened 2021-04-06
Adjusting mgmt file namespace and few file relocation

allenjzhang:
1. [Mgmt] Update the Mgmt file namespaces
2. [All] Lift out all output library into autorest folder with XXTarget instead of mixed in with output models
3. [Mgmt] Moved inheritencechoser into mgmt
4. [Mgmt] Renamed ArmResource.cs and ArmResourceWriter.cs to be ResourceXXX.cs. Writer is also moved into Generation folder.


src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
```diff
@@ -1,15 +1,15 @@
...
-using System.Collections.Generic;
 using System.Text;
 using AutoRest.CSharp.Input;
+using AutoRest.CSharp.Mgmt.AutoRest;
 using AutoRest.CSharp.Output.Builders;
 using AutoRest.CSharp.Output.Models.Types;
-using AutoRest.CSharp.Utilities;
 
-namespace AutoRest.CSharp.Output.Models
+#nullable enable
```
pakrym, line 10: You shouldn't need this. Nullable is enabled project-wide https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/AutoRest.CSharp.csproj#L8

allenjzhang: Mmm, it didn't kick in my IDE. Let me check again.

m-nash (approved):
Fix conflict then this looks good

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
allenjzhang: It may be an IDE issue as ? still get red squiggles in editor but didn't affect build. I removed this from each file level.

chamons (approved):

PR #1132, opened 2021-04-08
Generate Container class for mgmt library

isra-fel:
This PR is still very rough
(Requires to upgrade ref to Azure.ResourceManager.Core after merging Azure/azure-sdk-for-net/pull/20222)

## What I did

- Generated XXXContainer class for mgmt library

## How to test

TBD


isra-fel:
Abandon this PR as we are going to redesign ArmResposne...

PR #1135, opened 2021-04-08
[LLC] Expose Pipeline, Hide CreateXYZ methods

ellismg:
For Low Level Clients, expose the HttpPipeline via a public read-only
property. Previously, this was `protected` and so a customer who
wished to send arbitrary requests would need to inherit from the
client to access this property. In addition, we make our `CreateXYZ`
methods private for now. We do believe we will want to expose these
somehow at some point, but we aren't exactly sure how (perhaps via
some nested `RequestBuilders` static class or something else).

### DIFF
--- a/src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
+++ b/src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
@@ -116,7 +116,7 @@ namespace AutoRest.CSharp.Generation.Writers
         private void WriteClientFields(CodeWriter writer, LowLevelRestClient client, BuildContext context)
         {
             writer.WriteXmlDocumentationSummary("The HTTP pipeline for sending and receiving REST requests and responses.");
-            writer.Append($"public {typeof(HttpPipeline)} {PipelineField}");
+            writer.Append($"public virtual {typeof(HttpPipeline)} {PipelineField}");
             writer.AppendRaw("{ get; }\n");
 
             if (HasKeyAuth (context))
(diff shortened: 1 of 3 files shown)


chamons (approved):

pakrym:
Why expose the pipeline while hiding Create* methods as the same time? Do we expect customers to handcraft entire requests by themselves?

ellismg:
> Why expose the pipeline while hiding Create* methods at the same time? Do we expect customers to handcraft entire requests by themselves?

There are definitely cases where folks will want to hand craft the entire request by themselves (for example to invoke operations we might not yet have Create* methods for (i.e. experimental APIs)).  This is something we want to enable at some point so we will want to bring this back, but I think we all agree that `protected Create<XYZ>` members is not the way to do it.

The reason to not also make them public now is that we found during user testing that having both the `Create<XYZ>` and `<XYZ>` methods on the same type ended up being very confusing to users, they were not sure what to end up using.

pakrym (approved):

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -116,7 +116,7 @@ private void WriteClientMethod(CodeWriter writer, RestClientMethod clientMethod,
         private void WriteClientFields(CodeWriter writer, LowLevelRestClient client, BuildContext context)
         {
             writer.WriteXmlDocumentationSummary("The HTTP pipeline for sending and receiving REST requests and responses.");
-            writer.Append($"protected {typeof(HttpPipeline)} {PipelineField}");
+            writer.Append($"public {typeof(HttpPipeline)} {PipelineField}");
```
pakrym, line 119: virtual

PR #1142, opened 2021-04-09
update to use latest version of Azure.ResourceManager.Core

m-nash:
- update project references
- update inherited types to include new generic type


samples/AppConfiguration/AppConfiguration.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.11.0" />
+    <PackageReference Include="Azure.Core" Version="1.12.0" />
```
pakrym, line 10: 1.13.0 is out

ShivangiReja:
Update here - https://github.com/Azure/autorest.csharp/blob/7697ac4a2ced42722b88863c100576e8778d36ae/test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj#L10

ShivangiReja:
Update version -
* https://github.com/Azure/autorest.csharp/blob/7697ac4a2ced42722b88863c100576e8778d36ae/test/AutoRest.TestServerLowLevel.Tests/AutoRest.TestServerLowLevel.Tests.csproj#L11
* https://github.com/Azure/autorest.csharp/blob/7697ac4a2ced42722b88863c100576e8778d36ae/test/AutoRest.TestServerLowLevel.Tests/AutoRest.TestServerLowLevel.Tests.csproj#L14

samples/AppConfiguration/AppConfiguration.csproj
m-nash: Not sure we can go to 1.13 yet since the azure.resourcemanager.core is pointed at 1.12 and then we will get a downgrade error

ShivangiReja:
@m-nash - https://github.com/Azure/autorest.csharp/pull/1142#issuecomment-817014733

ShivangiReja (approved):

PR #1149, opened 2021-04-12
remove partial class additions

m-nash:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/5749

- To maintain mutability we removed the code model partial additons for, OperationGroup, HttpRequest, and Schema
- These are now defined on demand when needed


src/AutoRest.CSharp/Mgmt/Decorator/ExtensionDetection.cs
```diff
@@ -1,24 +1,39 @@
...
+using System.Collections.Concurrent;
+using AutoRest.CSharp.AutoRest.Plugins;
 using AutoRest.CSharp.Input;
 
 namespace AutoRest.CSharp.Mgmt.Decorator
 {
     internal static class ExtensionDetection
     {
-        public static bool IsExtension(OperationGroup operationGroup)
+        private static ConcurrentDictionary<string, bool> _valueCache = new ConcurrentDictionary<string, bool>();
```
pakrym, line 12: I would recommend not making this static. We don't currently do it but it's possible that multiple swaggers would be processed in the same process. This might also make the in-proc generator running tests tricky.

pakrym: Recommend putting this on management BuildContext and passing that around.

pakrym (approved):

PR #1156, opened 2021-04-14
Skip resource generation via config

isra-fel:
## What I did

- The `operation-group-to-resource` configuration in `readme.md` now supports a "SKIP" keyword to skip an operation group when generating
- In order to do that, I introduced a series of `FindOperationGroup()` methods in `MgmtOutputLibrary`, which should be handful

## How to test

- In Azure.ResourceManager.Sample I marked `Usage` and `LogAnalytics` to be skipped.


src/AutoRest.CSharp/Mgmt/Decorator/SchemaDetection.cs
```diff
@@ -47,5 +48,23 @@ private static RequestParameter GetBodyParameter(ServiceRequest request, Operati
...
         }
+
+        /// <summary>
+        /// Indicates if the given operation group should skip generation or not.
+        /// If the operation group is marked as "SKIP" in configuration, will return false indicating the corresponding container, data... classes should not be generated.
+        /// </summary>
+        /// <param name="og">Operation group.</param>
+        /// <param name="config">Management plane configuration.</param>
+        /// <returns></returns>
+        public static bool ShouldSkipGeneration(this OperationGroup og, MgmtConfiguration config)
```
isra-fel, line 59: The original idea we had in the meeting was to add extention methods on `MgmtRestClient`, `ResourceContainer`...
However they don't expose `OperationGroup`, so I cannot get their `Resource`.
That's why I came to the idea to setup lookup methods to get OperationGroup from MgmtRestClient...

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtTarget.cs
```diff
@@ -40,31 +41,31 @@ public static void Execute(GeneratedCodeWorkspace project, CodeModel codeModel,
                 project.AddGeneratedFile($"Models/{name}.Serialization.cs", serializerCodeWriter.ToString());
             }
 
-            foreach (var client in context.Library.RestClients)
+            foreach (var client in context.Library.RestClients.Where(client => !context.Library.FindOperationGroup(client).ShouldSkipGeneration(configuration.MgmtConfiguration)))
```
pakrym, line 44: Instead of filtering during the writing, we should avoid building these models altogether.

pakrym:
I think autorest directives might be great for this purpose.

Similar to https://github.com/Azure/autorest/blob/master/docs/generate/built-in-directives.md#remove-operation but to remove the operation group.

```
directive:
  - remove-operation-group: <operationId>
```

It would allow us to reuse the same feature for DP and LLC.

m-nash:
> I think autorest directives might be great for this purpose.
> 
> Similar to https://github.com/Azure/autorest/blob/master/docs/generate/built-in-directives.md#remove-operation but to remove the operation group.
> 
> ```
> directive:
>   - remove-operation-group: <operationId>
> ```
> 
> It would allow us to reuse the same feature for DP and LLC.

We don't want to remove it, we simply want to place its methods on a different operation class

pakrym:
Even better: 
```
directive:
  - rename-operation-group:
      from: <operation-group>
      to: <new-operation-group>
```

m-nash:
> Even better:
> 
> ```
> directive:
>   - rename-operation-group:
>       from: <operation-group>
>       to: <new-operation-group>
> ```
> 
> And we'll merge operations from the operation groups with the same name.

They often will not be in the same RP and will need to be done as extension methods so combining them won't be ideal either.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -240,6 +239,50 @@ public override CSharpType FindTypeForSchema(Schema schema)
...
+            return EnsureResourceContainers().FirstOrDefault(ogToContainer => ogToContainer.Value == container).Key;
+        }
+
+        /// <summary>
+        /// Looks for an <see cref="OperationGroup" /> by <see cref="ResourceData" />.
+        /// </summary>
+        /// <param name="resourceData">The <see cref="ResourceData" /> instance which is built from the operation group.</param>
+        /// <returns>The operation group. `null` if not found.</returns>
+        public OperationGroup FindOperationGroup(ResourceData resourceData) {
+            return EnsureResourceData().FirstOrDefault(ogToData => ogToData.Value == resourceData).Key;
```
pakrym, line 275: You can make the OperationGroup property public: https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs#L30 and avoid this roundtrip.

m-nash: Agreed, doesn't look like these are even used so I will remove them for this PR

pakrym (approved):

PR #1159, opened 2021-04-14
Support to find [Resource]Data base class with IsSuperset comparison

YalinLi0312:
work item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=5384

Also fixed this one:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=5724


src/AutoRest.CSharp/Mgmt/Decorator/InheritanceChoser.cs
```diff
@@ -27,54 +27,99 @@ private static IList<System.Type> GetReferenceClassCollection()
...
+            return output;
+        }
 
-            RebuildExactModelInheritance(schemaMap);
-            RebuildExactModelInheritance(resourceSchemaMap);
+        private static IList<System.Type> ConvertGenericType(IList<System.Type> referenceClassCollection)
+        {
+            for (int i = 0; i < referenceClassCollection.Count; i++)
+            {
+                referenceClassCollection[i] = referenceClassCollection[i].MakeGenericType(referenceClassCollection[i].GetResourceIdentifierType());
```
m-nash, line 65: this should be coming from the customeAttribute property we added in Azure.ResourceManager.Core.ReferenceTypeAttribute

src/AutoRest.CSharp/Mgmt/Generation/ResourceIdentifierChooser.cs
```diff
@@ -18,7 +19,7 @@ internal static string GetResourceIdentifierType(this OperationGroup operation)
         }
 
         //This overload might not be needed after the final inheritance work from 5724 and 5384
-        internal static Type GetResourceIdentifierType(this CSharpType csharpType)
+        internal static Type GetResourceIdentifierType(this Type type)
```
m-nash, line 22: not needed with the comment from above

m-nash (changes requested):

m-nash (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -429,10 +429,10 @@ private ObjectTypeDiscriminatorImplementation[] CreateDiscriminatorImplementatio
             )).ToArray();
         }
 
-        private IEnumerable<ObjectTypeProperty> BuildProperties()
+        protected IEnumerable<ObjectTypeProperty> BuildProperties(bool includeParent = true)
         {
             // WORKAROUND: https://github.com/Azure/autorest.modelerfour/issues/261
-            var existingProperties = GetParentProperties();
+            var existingProperties = includeParent ? GetParentProperties() : new HashSet<string?>();
```
pakrym, line 435: The parameter name seems to be wrong and I'm not sure if you need to make this change. This HashSet is used to remove parent properties that are already defined in the base class.

src/AutoRest.CSharp/Mgmt/Decorator/Node.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AutoRest.CSharp.Output.Models.Types
+{
+    internal class Node
```
pakrym, line 10: This type can be nested into InheritanceChoser or replaced with` KeyVaultPair<>`

test/AutoRest.TestServer.Tests/Mgmt/TestProjects/SupersetInheritanceTests.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public class SupersetInheritanceTests
+    {
+        [TestCase(typeof(Resource<TenantResourceIdentifier>), typeof(SupersetModel1Data))]
+        [TestCase(typeof(Object), typeof(SupersetModel2Data))]
+        [TestCase(typeof(Object), typeof(SupersetModel3Data))]
+        [TestCase(typeof(TrackedResource<TenantResourceIdentifier>), typeof(SupersetModel4Data))]
+        public void ValidateInheritanceType(Type expectedBaseType, Type generatedClass)
+        {
+            Assert.AreEqual(expectedBaseType, generatedClass.BaseType);
```
pakrym, line 19: Don't we need to verify that properties were filtered out/

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
m-nash: It refers to whether or not you want to include the parent properties (or base properties) and is used by this
        public ObjectTypeProperty[] MyProperties => _myProperties ??= BuildProperties(false).ToArray();
which is a new property on MgmtObjectType used when we only want to see the properties that the current type explicitly adds

pakrym: This is not how the HashSet is used though. It's used to skip properties in the following line.

```
                    if (existingProperties.Contains(property.Language.Default.Name))
```

We don't use it to add more properties to the type.

PR #1170, opened 2021-04-15
Follow up items for PR 1159

m-nash:
- Follow up items for https://github.com/Azure/autorest.csharp/pull/1159
- Adds a new SystemObjectType which inherits from ObjectType but wraps a System.Type so that it can be used in all writers seemlessly
- Moved majority of implementation for ObjectType into SchemaObjectType and made ObjectType mostly abstract implementation to support the SystemObjectType
- Eliminated the need for a special ModelWriter and SerializationWriter in mgmt plane since we can now use the SystemObjectType


src/AutoRest.CSharp/Mgmt/Decorator/InheritanceChoser.cs
```diff
@@ -213,5 +213,16 @@ private static bool IsSuperset(MgmtObjectType childType, System.Type parentType)
             }
             return parentProperties.Count == matchCount;
         }
+
+        private class Node
```
chamons, line 217: Maybe use a record class here?

chamons:
As Pavel did the last review I'll let him give the 👍

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeConstructor.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Linq;
+using System.Reflection;
```
chamons, line 6: Unnecessary diff here?

chamons:
This is turning out to be a large change, I'd like time for Pavel and I to review it before it goes in.

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -568,6 +572,11 @@ private static void DeserializeFrameworkTypeValue(CodeWriter writer, CodeWriterD
             writer.AppendRaw(")");
         }
 
+        private static bool IsMgmtReferenceType(Type frameworkType)
```
allenjzhang, line 575: nit, as this list may change, would like to keep this code in Mgmt rather than Common. Can we call out to a static checked to there?

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeConstructor.cs
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System;
```
chamons, line 4: Strange diff to include.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -2,129 +2,88 @@
...
+
+            if (!IncludeDeserializer)
+            {
+                yield break;
+            }
+
+            // Skip serialization ctor if they are the same
+            if (!InitializationConstructor.Parameters
+                .Select(p => p.Type)
+                .SequenceEqual(SerializationConstructor!.Parameters.Select(p => p.Type)))
```
chamons, line 91: This used to be:
-                .SequenceEqual(SerializationConstructor.Parameters.Select(p => p.Type)))
+                .SequenceEqual(SerializationConstructor!.Parameters.Select(p => p.Type)))

Could SerializationConstructor be null here?

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,181 @@
...
+                }
+            }
+
+            return new List<ParameterInfo>();
+        }
+
+        internal IEnumerable<PropertyInfo> GetAllProperties()
+        {
+            var type = _type;
+            while (type is not null)
```
chamons, line 65: Same as above.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,181 @@
...
+            {
+                foreach (var property in type.GetProperties())
+                {
+                    yield return property;
+                }
+                type = type.BaseType;
+            }
+        }
+
+        protected override ObjectTypeDiscriminator? BuildDiscriminator()
```
chamons, line 75: `protected override ObjectTypeDiscriminator? BuildDiscriminator() => null` ?

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        protected override IEnumerable<ObjectTypeProperty> BuildProperties()
+        {
+            foreach (var property in _myProperties)
+            {
+                var getter = property.GetGetMethod();
+                var setter = property.GetSetMethod();
+                MemberDeclarationOptions memberDeclarationOptions = new MemberDeclarationOptions(
+                    getter is not null && getter.IsPublic ? "public" : "internal",
```
chamons, line 118: Again.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,181 @@
...
+                }
+                yield return new ObjectTypeProperty(memberDeclarationOptions, prop.Summary, prop.IsReadOnly, prop, GetTypeFromSystem(property.PropertyType));
+            }
+        }
+
+        private bool GetReadOnly(PropertyInfo property)
+        {
+            if (property.Name == "Tags")
+                return false;
+            return property.GetSetMethod() is null;
```
chamons, line 140: Again.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,181 @@
...
+                {
+                    args.Add(GetTypeFromSystem(argType));
+                }
+            }
+            return new CSharpType(type, false, args.ToArray());
+        }
+
+        private string GetPropertySummary(MethodInfo? setter)
+        {
+            return setter is not null ? " or sets" : string.Empty;
```
chamons, line 158:
```suggestion
            return setter != null ? " or sets" : string.Empty;
```

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        protected override CSharpType? CreateInheritedDictionaryType()
+        {
+            throw new NotImplementedException();
+        }
+
+        protected override CSharpType? CreateInheritedType()
+        {
+            return _type.BaseType is null ? null : CSharpType.FromSystemType(Context, _type.BaseType); //new CSharpType(_type.BaseType);
```
chamons, line 178:
```suggestion
            return _type.BaseType == null ? null : CSharpType.FromSystemType(Context, _type.BaseType); //new CSharpType(_type.BaseType);
```

chamons (changes requested):
This needs @pakrym 👀 as it touched deep type system types.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,181 @@
...
+                    getter is not null && getter.IsPublic ? "public" : "internal",
+                    property.Name,
+                    GetTypeFromSystem(property.PropertyType));
+                Property prop = new Property();
+                prop.Nullable = false;
+                prop.ReadOnly = GetReadOnly(property); //TODO read this from attribute from reference object
+                prop.SerializedName = ToCamelCase(property.Name);
+                prop.Summary = $"Gets{GetPropertySummary(setter)} {property.Name}";
+                prop.Required = true;
+                if (property.PropertyType.IsGenericType && property.PropertyType.GetGenericTypeDefinition() == typeof(IDictionary<,>))
```
allenjzhang, line 127: Can we add comment here to capture the node on the just enough type supported for Mgmt.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -37,13 +37,15 @@ public CSharpType(Type type, bool isNullable, params CSharpType[] arguments)
...
 
-        public CSharpType(TypeProvider implementation, string ns, string name, bool isValueType = false, bool isNullable = false)
+        public CSharpType(TypeProvider implementation, string ns, string name, bool isValueType = false, bool isNullable = false, CSharpType[]? arguments = default)
         {
             _implementation = implementation;
             Name = name;
             IsValueType = isValueType;
             IsNullable = isNullable;
             Namespace = ns;
+            if (arguments is not null)
```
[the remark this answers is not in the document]
m-nash: is not null is new syntax that avoids the operator overload null ref exception.  same as "is null" but they finally added a not version of it.

The property is initialized to an empty set if its never used so this maintains that default behavior

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -87,5 +89,18 @@ public override string ToString()
         {
             return new CodeWriter().Append($"{this}").ToString(false);
         }
+
+        internal static CSharpType FromSystemType(BuildContext context, Type parentType)
+        {
+            var genericTypes = parentType.GetGenericArguments().Select<Type, CSharpType>(t => new CSharpType(t));
```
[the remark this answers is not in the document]
m-nash: to use var yes, if i had an explicit variable type then maybe it could determine it

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeConstructor.cs
m-nash: I just did a remove all unused usings in a bunch of files i was touching.  Happy to undo this if you want

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
m-nash: Its null for the SystemObjectType path and not null in the SchemaObjectType path.  So in this class its safe that it won't be null.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeConstructor.cs
chamons: That's a fine reason.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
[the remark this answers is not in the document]
pakrym: I'm not a fan of `is not null` or `is null` because while I understand the theoretical benefit I'm yet to see the problem in practice.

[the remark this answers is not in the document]
pakrym: Consider `Arguments = arguments ?? Array.Empty<CSharpType>()`

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -126,7 +126,8 @@ internal static bool IsDictionary(CSharpType type)
         {
             return type.IsFrameworkType &&
                    (type.FrameworkType == typeof(IDictionary<,>) ||
-                   type.FrameworkType == typeof(IReadOnlyDictionary<,>));
+                   type.FrameworkType == typeof(IReadOnlyDictionary<,>) ||
+                   type.FrameworkType.IsGenericType && type.FrameworkType.GetGenericTypeDefinition() == typeof(IDictionary<,>));
```
pakrym, line 130: The `type.FrameworkType` should already be the generic definition.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -228,7 +228,7 @@ public static CodeWriter WriteReferenceOrConstant(this CodeWriter writer, Refere
             this CodeWriter writer,
             Action<CodeWriter, CodeWriterDelegate> valueCallback,
             ObjectType objectType,
-            ObjectTypeConstructor constructor,
+            ObjectTypeConstructor? constructor,
```
pakrym, line 231: Why do we need the ability to pass null here? There should be a ctor that we are calling.

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -159,7 +157,8 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
                         }
                         else if (frameworkType == typeof(string) ||
                                  frameworkType == typeof(char) ||
-                                 frameworkType == typeof(Guid))
+                                 frameworkType == typeof(Guid) ||
+                                 ReferenceTypes.IsMgmtReferenceType(frameworkType))
```
pakrym, line 161: I wonder if we can extend the `JsonValueSerialization` with the concept of "SerializeAs" that would specify the resulting json type...

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -3,107 +3,72 @@
...
+        public ObjectTypeConstructor? SerializationConstructor => _serializationConstructor ??= BuildSerializationConstructor();
+        public CSharpType? ImplementsDictionaryType => _implementsDictionaryType ??= CreateInheritedDictionaryType();
+        public ObjectTypeConstructor InitializationConstructor => _initializationConstructor ??= BuildInitializationConstructor();
         public ObjectSerialization[] Serializations => _serializations ??= BuildSerializations();
+        public bool IsStruct => ExistingType?.IsValueType == true;
+        public virtual string? Description { get; protected set; }
+        public ObjectTypeDiscriminator? Discriminator => _discriminator ??= BuildDiscriminator();
 
-        public ObjectTypeProperty[] Properties => _properties ??= BuildProperties().ToArray();
+        public abstract bool IncludeSerializer { get; }
```
pakrym, line 41: I think this makes sense on SchemaObject type only. As well as IncludeDeserializer and IncludeConverter. And possibly AdditionalPropertiesProperty

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -3,107 +3,72 @@
...
-        protected override string DefaultName => OjectSchema.CSharpName();
-        protected override string DefaultAccessibility { get; } = "public";
-        protected override string DefaultNamespace { get; }
-        protected override TypeKind TypeKind => IsStruct ? TypeKind.Struct : TypeKind.Class;
-
-        public string? Description { get; }
+        public ObjectTypeConstructor[] Constructors => _constructors ??= BuildConstructors().ToArray();
+        public virtual ObjectTypeProperty[] Properties => _properties ??= BuildProperties().ToArray();
 
         public virtual CSharpType? Inherits => _inheritsType ??= CreateInheritedType();
```
pakrym, line 32: both this property and CreateInheritedType are virtual. One should be enough.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -3,107 +3,72 @@
...
-
-        public bool IsStruct => ExistingType?.IsValueType == true;
-        protected override string DefaultName => OjectSchema.CSharpName();
-        protected override string DefaultAccessibility { get; } = "public";
-        protected override string DefaultNamespace { get; }
-        protected override TypeKind TypeKind => IsStruct ? TypeKind.Struct : TypeKind.Class;
-
-        public string? Description { get; }
+        public ObjectTypeConstructor[] Constructors => _constructors ??= BuildConstructors().ToArray();
+        public virtual ObjectTypeProperty[] Properties => _properties ??= BuildProperties().ToArray();
```
pakrym, line 30: Both this and BuildProperties are virtual. We should pick one.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -3,107 +3,72 @@
...
-        public ObjectTypeProperty[] Properties => _properties ??= BuildProperties().ToArray();
+        public abstract bool IncludeSerializer { get; }
+        public abstract bool IncludeDeserializer { get; }
+        public abstract bool IncludeConverter { get; }
+        public abstract ObjectTypeProperty? AdditionalPropertiesProperty { get; }
+        protected abstract ObjectTypeDiscriminator? BuildDiscriminator();
+        protected abstract ObjectSerialization[] BuildSerializations();
+        protected abstract ObjectTypeConstructor BuildInitializationConstructor();
+        protected abstract CSharpType? CreateInheritedDictionaryType();
+        protected abstract ObjectTypeConstructor? BuildSerializationConstructor();
```
pakrym, line 49: I'm not a fan of having the mixed extensibility story where some properties are virtual by themselves and other have virtual caching builders. I would pick one.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                foreach (var property in type.GetProperties())
+                {
+                    yield return property;
+                }
+                type = type.BaseType;
+            }
+        }
+
+        protected override ObjectTypeDiscriminator? BuildDiscriminator() => null;
```
pakrym, line 75: Do we need the disctiminator on the ObjectType then?

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,174 @@
...
+        public override bool IncludeConverter => false;
+
+        public override string DefaultName => GetNameWithoutGeneric(_type);
+
+        private string ToCamelCase(string name)
+        {
+            return name.Substring(0, 1).ToLower(CultureInfo.InvariantCulture) + name.Substring(1);
+        }
+
+        private IEnumerable<PropertyInfo> _myProperties => _type.GetProperties().Where(p => p.DeclaringType == _type);
```
pakrym, line 47: Inline?

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -0,0 +1,174 @@
...
+            }
+            return new CSharpType(type, false, args.ToArray());
+        }
+
+        private string GetPropertySummary(MethodInfo? setter)
+        {
+            return setter is not null ? " or sets" : string.Empty;
+        }
+
+        protected override ObjectTypeConstructor? BuildSerializationConstructor() => null;
```
pakrym, line 163: I have a suspicion that everything that's not implemented on the SystemObjectType can be moved from ObjectType to SchemaObjectType

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -27,10 +28,10 @@ protected TypeProvider(BuildContext context)
             TypeKind == TypeKind.Struct || TypeKind == TypeKind.Enum);
         public TypeDeclarationOptions Declaration => _type ??= BuildType();
 
-        protected BuildContext Context { get; private set; }
-        protected abstract string DefaultName { get; }
+        internal BuildContext Context { get; private set; }
+        public abstract string DefaultName { get; }
```
pakrym, line 32: We shouldn't use this publicly. As sometimes it's redefined.

src/AutoRest.CSharp/DataPlane/Output/DataPlaneClient.cs
```diff
@@ -29,6 +29,7 @@ public DataPlaneClient(OperationGroup operationGroup, BuildContext<DataPlaneOutp
         {
             _operationGroup = operationGroup;
             _context = context;
+            DefaultAccessibility = "public";
```
pakrym, line 32: public is the default

src/AutoRest.CSharp/DataPlane/Output/DataPlaneResponseHeaderGroupType.cs
```diff
@@ -22,6 +22,7 @@ internal class DataPlaneResponseHeaderGroupType: TypeProvider
 
         public DataPlaneResponseHeaderGroupType(OperationGroup operationGroup, Operation operation, HttpResponseHeader[] httpResponseHeaders, BuildContext<DataPlaneOutputLibrary> context) : base(context)
         {
+            DefaultAccessibility = "internal";
```
pakrym, line 25: nit: move next to the default name assignment

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -18,6 +18,7 @@ protected TypeProvider(BuildContext context)
         {
             Context = context;
             _existingType = new Lazy<INamedTypeSymbol?>(() => Context.SourceInputModel?.FindForType(DefaultNamespace, DefaultName));
+            DefaultAccessibility = "public";
```
pakrym, line 21: What's the benefit of this change?

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
m-nash: no longer needed

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
m-nash: yeah, was thinking the same thing, but given how large this change is already Id suggest we take that up in a later PR

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
m-nash: The problem is the value is used in places where its expecting an ObjectType which is why we need the property here.  Otherwise we have to cast in dozens of places.  Having it here allows us to treat SystemObjectType and SchemaObjectType identically from all the writers perspectives

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
m-nash: not sure I follow

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
m-nash: yes because its used by writers which expect an ObjectType

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
m-nash: sure, i was originally using it in multiple places

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
m-nash: this has an implementation now

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
m-nash: none, originally it was needed for something, but now its not ill revert

src/AutoRest.CSharp/DataPlane/Output/DataPlaneResponseHeaderGroupType.cs
m-nash: obsolete given comment below

src/AutoRest.CSharp/DataPlane/Output/DataPlaneClient.cs
m-nash: obsolete given comment below

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
m-nash: this one is tricky due to the infinite loop.  Its easier if I can explain by showing you tomorrow.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
m-nash: In the case of SystemObjectType it is not

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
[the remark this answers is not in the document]
chamons: https://xkcd.com/285/ on it being a best practice.

I'm unable to find that listed on MSDN or from dotnet core after a brief search.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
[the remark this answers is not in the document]
chamons: I checked out your branch and was able to remove it without any issues:

https://gist.github.com/chamons/1da9e7258f7a28f57ba13291a1400b26

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
chamons: Please file a tracking issue and i'm fine w\ punting this bit.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
chamons: I really do not like this "it can be null in one case and not null in another", it is really easy to trip in a future PR. However it is in Mgmt code, so I'm not blocking here.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
chamons: To restate, why do we need both these:

`public ObjectTypeConstructor SerializationConstructor => _serializationConstructor ??= BuildSerializationConstructor();`
And
`protected abstract ObjectTypeConstructor BuildSerializationConstructor();`

and then later have

`public abstract bool IncludeDeserializer { get; }`

Some things are properties you need to define values from, others are methods that the base class uses to define properties for.

If we made `ObjectTypeConstructor SerializationConstructor { get; }` abstract then it would be consistent is the point I believe.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
chamons: The implementation is to return empty array. I still don't see why it couldn't move to SchemaObjectType?

src/AutoRest.CSharp/DataPlane/Output/DataPlaneClient.cs
chamons: I guess each client has to define accessibility now.

chamons (changes requested):
Work is promising, but it needs some more hammering before going in.

Major
------
- ObjectType.cs double virtual methods and "an infinite loop" explaination
- is not null/is null vs != null/==null
- File tracking issue for JsonCodeWriterExtensions extensibility and mgmt types
- Wait for pavel to respond to 4 other comments

Minor
-------
- Unnecessary generic constraits on Select call
- Builders vs caching abstract methods in ObjectType.cs
- SystemObjecType BuildSerializationConstructor on base type?

Not blocking
-------------------
- Nullability concerns in MgmtObjectType.cs

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
pakrym: then we should make it so. That's the contract of CSharpType.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
pakrym: Why do they expect an ObjectType and not the SchemaType?

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
m-nash: it doesn't return an empty array, and why the ObjectType needs the definition is explained here https://github.com/Azure/autorest.csharp/pull/1170#discussion_r618908466

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
m-nash: because there was no objecttype originally, and the whole point of this change was to allow the common writers to work with both schemaobjecttype (the old objecttype) and systemobjecttype in the exact same manner

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
m-nash: that is how it was before.  I am not introducing this its a copy paste from the old objecttype

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
m-nash: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/5812

src/AutoRest.CSharp/Mgmt/Decorator/InheritanceChoser.cs
m-nash: prefer to leave this as is for now, it works and its isolated in mgmt library

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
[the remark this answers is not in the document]
m-nash: Ill revert these as it seems you are very passionate about it.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
m-nash: It is no longer null based on some changes I made in another area

pakrym (approved):

chamons (approved):

PR #1226, opened 2021-05-08
[LLC] Fix issues blocking azure-sdk-for-net update

ellismg:
- Missing a `ConfigureAwait(false)` in the low level client policy
  which invokes a policy from request options.

- Add an explicit `#nullable enable` in the policy since it can be
  used in libraries that have not turned on nullable annotations
  project wide.

- Generate suppressions for AZC0002 on low level client methods. We
  should update the linters to understand this pattern but for now
  this feels like a quick way to solve this issue without disabling
  the warning in every LLC project at the project level.


ellismg:
Validated locally that with the changes (plus adding project refs to Azure.Core.Experimental) we can build all the code in azure-sdk-for-net without errors.

PR #1275, opened 2021-05-26
Add CodeGenSuppressTypeAttribute to remove types from codegen

AlexanderSher:


chamons (approved):

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -91,6 +110,34 @@ private async Task<Document> ProcessDocument(Document document)
             return document;
         }
 
+        private bool ContainsSuppressedType(SyntaxTree syntaxTree, SemanticModel semanticModel)
```
pakrym, line 113: Why isn't this logic in MemberRemoverRewriter?

What if there are multiple types per file?

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -91,6 +110,34 @@ private async Task<Document> ProcessDocument(Document document)
...
+            if (typeDeclarationSyntax == null)
+            {
+                return false;
+            }
+
+            var typeSymbol = semanticModel.GetDeclaredSymbol(typeDeclarationSyntax);
+            while (typeSymbol != null)
+            {
+                var fullName = typeSymbol.ToDisplayString(_fullyQualifiedNameFormat);
+                if (_suppressedTypeNames.Contains(typeSymbol.Name) || _suppressedTypeNames.Contains(fullName))
```
pakrym, line 128: Should we do equals?

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -53,7 +61,13 @@ public void AddGeneratedFile(string name, string text)
 
         public async IAsyncEnumerable<(string Name, string Text)> GetGeneratedFilesAsync()
         {
-            List<Task<Document>> documents = new List<Task<Document>>();
+            var compilation = await _project.GetCompilationAsync();
+            if (compilation == null)
```
pakrym, line 65: should we debug.Assert instead?

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
AlexanderSher: Equals what to what?

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
AlexanderSher: Can we have multiple files per type? I thought that will produce a compiler error.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -120,7 +167,15 @@ public static async Task<GeneratedCodeWorkspace> Create(string projectDirectory,
                 }
             }
 
-            return new GeneratedCodeWorkspace(generatedCodeProject);
+            var compilation = (await generatedCodeProject.GetCompilationAsync())!;
```
pakrym, line 170: Building multiple compilations is expensive. Lets move this logic into GetGeneratedFilesAsync right before the values are needed.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
pakrym: My bad, though we were doing stirng.Contains.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
pakrym: Some analyzers might but I don't think we should build our logic based on this assumption.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
AlexanderSher: Ok

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
AlexanderSher: Fixed

test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj
```diff
@@ -25,6 +25,9 @@
     <Compile Include="../TestServerProjects/**/*.cs" LinkBase="TestServerProjects" />
     <Compile Include="../TestProjects/**/*.cs" LinkBase="TestProjects" />
     <None Include="../TestServerProjects/**/*.yaml" LinkBase="TestServerProjects" />
+    <Compile Remove="..\TestProjects\TypeSchemaMapping\Generated\Models\EnumToBeSkipped.cs" />
```
pakrym, line 28: Why do we need to remove these?

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
AlexanderSher: Ok, let me check if it will remove the whole file if there are no types.

test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj
AlexanderSher: We don't.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
pakrym: I think we would have to have additional logic in GeneratedCodeWorkspace to check if a file contains no types after the rewriter runs.

AlexanderSher: Fixed

src/AutoRest.CSharp/Common/AutoRest/Plugins/MemberRemoverRewriter.cs
```diff
@@ -80,7 +107,31 @@ public MemberRemoverRewriter(Project project, SemanticModel semanticModel)
...
+            while (typeSymbol != null)
+            {
+                var fullName = typeSymbol.ToDisplayString(_fullyQualifiedNameFormat);
+                if (_suppressedTypeNames.Contains(typeSymbol.Name) || _suppressedTypeNames.Contains(fullName))
+                {
+                    return true;
+                }
+
+                typeSymbol = SymbolEqualityComparer.Default.Equals(typeSymbol.BaseType?.ContainingAssembly, typeSymbol.ContainingAssembly)
+                    ? typeSymbol.BaseType
```
pakrym, line 127: Just noticed. nice!

pakrym (approved):

PR #1277, opened 2021-05-26
 Update generator to handle the Operation OperationGroup renamed to RestApi

m-nash:
# Description

- Update DownloadSharedSource to include mgmt shared code ([This ADO](https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/5880) item will adjust the branch to master after we ship preview)
- Update autorest to write out the Operation OperationGroup as RestApi to avoid naming issue with Operation class in Azure.Core
- Update ArmClientExtension to include a getter to retrieve the RestApiClient

# Checklist

# Ready to Land?


eng/DownloadSharedSource.ps1
```diff
@@ -2,13 +2,28 @@
...
 $downloadPath = Resolve-Path (Join-Path $PSScriptRoot '..' 'src' 'assets' 'Azure.Core.Shared')
 Get-ChildItem $downloadPath -Filter *.cs | Remove-Item;
 $files = 'ClientDiagnostics.cs', 'DiagnosticScope.cs', 'DiagnosticScopeFactory.cs', 'ContentTypeUtilities.cs', 'HttpMessageSanitizer.cs', 'OperationHelpers.cs', 'TaskExtensions.cs', 'Argument.cs', 'Multipart/MultipartFormDataContent.cs', 'Multipart/MultipartContent.cs', 'AzureKeyCredentialPolicy.cs', 'Argument.cs', 'AzureKeyCredentialPolicy.cs'
 $baseUrl = 'https://raw.githubusercontent.com/Azure/azure-sdk-for-net/master/sdk/core/Azure.Core/src/Shared/'
+DownloadAll $files $baseUrl $downloadPath
 
-foreach ($file in $files)
+#Download management Shared
+$files = 'ClientContext.cs'
+$testPath = Join-Path $PSScriptRoot '..' 'src' 'assets' 'Management.Shared'
```
pakrym, line 22: I think we can manually create the directory on the first run

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -32,8 +32,12 @@ internal class CSharpProj : IPlugin
 ";
 
         private string _armCsProjContent = @"
+  <PropertyGroup>
+    <IncludeManagementSharedCode>true</IncludeManagementSharedCode>
```
pakrym, line 36: Who would set this for azure-sdk-for-net?

m-nash: in azure-sdk-for-net its automatically included if you are a mgmt track 2 library [here](https://github.com/Azure/azure-sdk-for-net/blob/feature/mgmt-track2/eng/Directory.Build.Common.targets#L146)

test/TestServerProjectsLowLevel/Directory.Build.targets
```diff
@@ -0,0 +1,10 @@
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
chamons, line 1: There are a number of compiles of this file, can we create one and relative include it to reduce duplication?

chamons (commented):
General question. Should the added ItemGroups be in a Directory.Build.props not Directory.Build.targets? I thought targets always had target / execution and props and item and property groups.

src/AutoRest.CSharp/Common/Generation/Writers/ClientWriter.cs
```diff
@@ -116,5 +119,37 @@ protected void WriteDiagnosticScope(CodeWriter writer, Diagnostic diagnostic, st
                 writer.Line($"throw;");
             }
         }
+
+        protected Parameter[] GetNonPathParameters(RestClientMethod clientMethod)
```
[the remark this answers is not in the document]
m-nash: yes, but the functionality is applicable in both if anyone wants this.  If we don't do this we need to create a new base just to hold this one method since the various classes in mgmt all inherit from clientwriter.cs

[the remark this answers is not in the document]
allenjzhang: I think this is useful helper methods that does NOT look out of place here.

allenjzhang (approved):
The only comment I have is to echo Chris's if we can moving the build.target's compile into a higher level and eliminate the duplications.

m-nash:
@chamons @allenjzhang given that the duplication already existed (it was just in the Directory.Builds.props file and now its in a different file) should we open an issue against the repo to consolidate the build targets / props holistically and keep the scope of this PR as is?

test/TestServerProjectsLowLevel/Directory.Build.targets
AlexGhiondea: @m-nash and I had a chat -- we'll try to refactor this out into a common file -- if we run into issues we'll go ahead with the changes as-is and we'll follow-up when @chamons and @pakrym are back 🙂.

PR #1287, opened 2021-06-02
fix(model): property not found in azure resource flatten model

archerzz:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -554,7 +554,7 @@ public ObjectTypeProperty GetPropertyForGroupedParameter(RequestParameter groupe
             return objectProperty;
         }
 
-        private bool TryGetPropertyForSchemaProperty(Func<ObjectTypeProperty, bool> propertySelector, [NotNullWhen(true)] out ObjectTypeProperty? objectProperty, bool includeParents = false)
+        protected bool TryGetPropertyForSchemaProperty(Func<ObjectTypeProperty, bool> propertySelector, [NotNullWhen(true)] out ObjectTypeProperty? objectProperty, bool includeParents = false)
```
pakrym, line 557: Is this used in the overriding type?

chamons (commented):
I'm not 100% sure we want to slice SchemaObjectType like this, can you explain why is it necessary?

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
archerzz: Yes, in `MgmtObjectType.GetPropertyForSchemaProperty` below.

test/TestProjects/ExactMatchInheritance/Generated/Models/AzureResourceFlattenModelData.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the MIT License.
+
+// <auto-generated/>
+
+#nullable disable
+
+namespace ExactMatchInheritance
+{
+    /// <summary> A class representing the AzureResourceFlattenModel data model. </summary>
+    public partial class AzureResourceFlattenModelData
```
archerzz, line 11: This is not correct. Probably I didn't compose a correct test data.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
pakrym: I don't see `TryGetPropertyForSchemaProperty` used there.

m-nash: I think the Get* methods are called but not the TryGet, looks like we can revert this line

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -112,5 +112,15 @@ private IEnumerable<ObjectTypeProperty> BuildMyProperties()
 
             return operationGroup;
         }
+
+        public override ObjectTypeProperty GetPropertyForSchemaProperty(Property property, bool includeParents = false)
+        {
+            if (_isResourceType)
```
m-nash, line 118: I think this needs to check if Inherits is a SystemObjectType

test/AutoRest.TestServer.Tests/Mgmt/TestProjects/ExactMatchInheritanceTests.cs
```diff
@@ -28,6 +28,7 @@ public ExactMatchInheritanceTests()
         [TestCase(typeof(object), typeof(ExactMatchModel9))]
         [TestCase(typeof(object), typeof(ExactMatchModel7))]
         [TestCase(typeof(object), typeof(ExactMatchModel11))]
+        [TestCase(typeof(object), typeof(AzureResourceFlattenModelData))]
```
m-nash, line 31: need to add a flattenModel that should inherit from one of the reference types as well such as TrackedResource<T>

test/TestProjects/ExactMatchInheritance/ExactMatchInheritance.json
```diff
@@ -519,6 +598,26 @@
                     "type": "string"
                 }
             }
+        },
+        "AzureResourceFlattenModel": {
+            "type": "object",
+            "description": "This model is x-ms-azure-resource and has flatten properties",
+            "x-ms-azure-resource": true,
+            "properties": {
+                "new": {
```
m-nash, line 607: you need to end up with a set of properties that are equal to the set on the reference types.

Couple scenarios i can think of

foo => {bar}
flat1 =>{id, name, type, location, tags, ref:Foo (flatten)} -expect it to inherit from trackedresource<T> with 1 extra property bar
flatreference =>{id, name, type}
flat2 =>{location, tags, bar, ref:flatreference(flatten)} -expect it inherit from trackedresource<T> with 1 extra property bar

m-nash (changes requested):

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
archerzz: You are correct. I forgot to roll back the changes here. Previously I copied the codes and then refactored to remove them, but forgot to change the calling method visibility.

archerzz:
Follow up changes on the fix:
1. roll back change of the visibility of `TryGetPropertyForSchemaProperty` method, it should be private.
2. change to check the inheritance type of the current object type, if it's `SystemObjectType`, then we should search property by `SerializedName`
3. update test cases regarding `x-ms-azure-resource: true` and `x-ms-client-flatten: true`.
  3.1 3 models will be deduced as reference to system object type, 1 model will be deduced as normal model
  3.2 update the generated codes accordingly

test/TestProjects/ExactMatchInheritance/Generated/Models/AzureResourceFlattenModelData.cs
archerzz: Updated according to review comments. Now the correct type will be deduced. Thanks @m-nash !

archerzz:
> I'm not 100% sure we want to slice SchemaObjectType like this, can you explain why is it necessary?

@chamons The current logic in `RestClientBuilder` to determine the flatten property has a defect.
- If the current object type is deduced as `SystemObjectType`, then it will inherit some default properties. For example, in the reported error case, `location` has a default property object instance.
- Our parsing will generate another `location` property object instance.
- The logic to search property is by reference, e.g. `==` which will fail in this case, since both `location` object instances are different instances, though semantically they should be the same.
- Michael and I had a discussion, and it probably too risky to change the generator logic. A safe way is to specialized the behavior of `MgmtObjectType` only. So we override the method here, change to use different search logic, depending on the base object type.

archerzz:
OK. My change reveals another issue...

Some resource definitions have duplicated properties. For example, our build failed on [`VirtualMachineScaleSetExtension`](https://github.com/Azure/autorest.csharp/blob/6356d73c05e3301f31a890a0c2c6858eb381ad66/samples/Azure.ResourceManager.Sample/sample.json#L10179) which has duplicated `type` property defined in `VirtualMachineScaleSetExtensionProperties`. So there will be 2 `type` property instances created at the same level of `SchemaObjectType` of `VirtualMachineScaleSetExtension`.

So we have a dilemma here:

- Searching by reference can avoid duplicated entry found error, but in some cases it will miss the target property.
- Searching by name can ensure we found the target property, but it cannot handle cases when there are duplicated definitions.

## Possible Solutions

### Solution 1: Search by reference first, then fall back to search by name

This is a short term workaround, since we don't fix the root cause.

### Solution 2: Eliminate duplicated property definitions during parsing

This should be the right way to go. But after this change, we probably have to remove searching by reference and change to use search by name, since we will kill references of duplicated properties.

m-nash (approved):

PR #1324, opened 2021-06-17
[LLC] Rename `requestContent` to just `content`

ellismg:
The shorter name is nicer. We also considered `body` but like
`content` more for cases where an API accepts multiple content
types. In this cases we will create a parameter named `contentType`
and so using `content` makes more sense when paired with that
parameter.


ellismg:
@chamons @KrzysztofCwalina Let me know what you think about the new documentation text for the parameter.

chamons (approved):

KrzysztofCwalina (approved):

PR #1366, opened 2021-06-28
Fix #1197: Autorest C# parser which is use to parse the format string is not working well with {{

### LINKED ISSUE #1197: Autorest C# parser which is use to parse the format string is not working well with {{
This works fine:

>writer.Append($"public {resource.Type} Data");
writer.Append($"{{ get; private set; }}");

But when I add the above code in one line, it throws an Exception - `Input string was not in a correct format.`
>writer.Append($"public {resource.Type} Data {{ get; private set; }}");

AlexanderSher:
Also replaced `StringBuilder` with `ReadOnlySpan` to minimize string allocation.


chamons (commented):
Seems reasonable to me, like use of Span. But I didn't write the original code, so there might be something subtle going on.

Waiting for @pakrym 's 👍

pakrym (approved):

PR #1367, opened 2021-06-28
Fix #1312: Suppress model factories for customized models

### LINKED ISSUE #1312: Suppress model factories for customized models
If the property was hidden we shouldn't generate a model factory.

AlexanderSher:


chamons (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -40,6 +40,11 @@ private static bool RequiresModelFactory(SchemaObjectType model)
                 return false;
             }
 
+            if (model.Properties.Any(p => p.Declaration.Accessibility != "public" && p.SchemaProperty?.IsDiscriminator != true))
```
pakrym, line 43: Why the discriminator part?

pakrym (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
AlexanderSher: Discriminator property is always internal.

PR #1388, opened 2021-07-07
Bump Autorest Version

AlexanderSher:


pakrym:
This also needs a modeler and core version bump.

AlexanderSher:
Those versions are the same:
- autorest/core is still 3.4.5: https://www.npmjs.com/package/@autorest/core
- autorest/modelerfour is still 4.19.3: https://www.npmjs.com/package/@autorest/modelerfour

pakrym (approved):

PR #1392, opened 2021-07-09
Make the parent of a Tenant Resource to be OperationsBase

fengzhou-msft:
# Description

A container of tenant resource whose direct parent is tenant (for instance, the billing account resource in `TenantOnly` tests) should be able to get created with `TenantOperations` which is of type `OperationsBase` instead of `ResourceOperationsBase`. 

This PR makes corresponding changes to support that and relies on core changes https://github.com/Azure/azure-sdk-for-net/pull/22589

Todo: 

1. Add `TenantOperationsExtensionWriter`.
2. In `ListAsGenericResourceAsync()`, `Parent` is cast as `ResourceGroupOperations`. This does not apply for resources whose ancestor is subscription or tenant (here a subscription ancestor means subscription has a direct child resource that is not a resource group, that child resource and the children of that child resource, their ancestor is subscription. A tenant ancestor is similar. Each resource's ancestor is either a resource group, a subscription or a tenant).  

# Checklist

# Ready to Land?


samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetContainer.cs
```diff
@@ -121,7 +121,7 @@ public AvailabilitySetsCreateOrUpdateOperation StartCreateOrUpdate(string availa
                 }
 
                 var originalResponse = _restClient.CreateOrUpdate(Id.ResourceGroupName, availabilitySetName, parameters, cancellationToken: cancellationToken);
-                return new AvailabilitySetsCreateOrUpdateOperation(Parent, originalResponse);
+                return new AvailabilitySetsCreateOrUpdateOperation(this, originalResponse);
```
ArcturusZhang, line 124: Is this expected?

fengzhou-msft: Originally, I wanted to avoid using the Parent as much as possible. But it will other issues to use `this` here. I reverted the change.

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -23,7 +23,10 @@ public Resource(OperationGroup operationGroup, BuildContext context)
             if (isExtension)
             {
                 var parentOperationGroup = operationGroup.ParentOperationGroup(context);
-                parentValue = parentOperationGroup!.Key.ToSingular(false);
+                if (parentOperationGroup != null)
+                {
+                    parentValue = parentOperationGroup.Key.ToSingular(false);
+                }
```
ArcturusZhang, line 29: How about the following which just takes one line
```suggestion
                parentValue = parentOperationGroup?.Key.ToSingular(false) ?? string.Empty;
```

test/TestProjects/TenantOnly/TenantOnly.csproj
```diff
@@ -11,11 +11,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.ResourceManager.Core" Version="1.0.0-alpha.20210706.2" />
+    <!-- <PackageReference Include="Azure.ResourceManager.Core" Version="1.0.0-alpha.20210706.2" /> -->
+    <ProjectReference Include="../../../../azure-sdk-for-net/sdk/resourcemanager/Azure.ResourceManager.Core/src/Azure.ResourceManager.Core.csproj" />
```
fengzhou-msft, line 15: After https://github.com/Azure/azure-sdk-for-net/pull/22589 is merged and published, use the new version of core.

PR #1399, opened 2021-07-13
Support scope resource

fengzhou-msft:
This PR supports generating SDK code for a scope resource, which may have combinations of the following 3 paths:
1. Paths with a {scope} variable
  `/{scope}/providers/Microsoft.Foo/bar/{barName}`
2. Paths at a specific scope

- `/providers/Microsoft.Foo/bar/{barName}`
- `/providers/Microsoft.Management/managementGroups/{managementGroupId}/providers/Microsoft.Foo/bar/{barName}`
- `/subscriptions/{subscriptionId}/providers/Microsoft.Foo/bar/{barName}`
- `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Foo/bar/{barName}`
- `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/{resourceProviderNamespace}/{parentResourcePath}/{resourceType}/{resourceName}/providers/Microsoft.Foo/bar/{barName}`

3. Paths with a single path-parameter of resource id
`/{resourceId}`

# Description

# Checklist

# Ready to Land?


m-nash (changes requested):
Can we add the specific cases in a TestProject instead of including all of Azure.Resources?  It would help illustrate what this change is targeting.

I am not sure the Parent override is needed since its not used anywhere.

fengzhou-msft:
> Can we add the specific cases in a TestProject instead of including all of Azure.Resources? It would help illustrate what this change is targeting.

Yes, your comment makes total sense. I was thinking about having multiple smaller PRs and adding tests for the changes in each PR. The concern is that I may not have enough time to do that before Preview. I'm not targeting to merge this PR directly, just want to use it to generate the correct Resources code without needing too much manual changes. Some of the code are written in a hasty with the assumption that the scope resource have some fixed patterns as the ones in Resources, but as we discussed that there could be various combinations of the 3 paths in a single operation group. I have left TODO in comment and will improve the code in smaller PRs after Preview. 

> I am not sure the Parent override is needed since its not used anywhere.

I agree. Let's just simply use `OperationsBase` as the `Parent` type for now.

fengzhou-msft:
Replaced by #1429

PR #1413, opened 2021-07-16
Add missing methods in resource operations, containers and extensions

ShivangiReja:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Output/ResourceContainer.cs
```diff
@@ -36,24 +37,51 @@ public ResourceContainer(OperationGroup operationGroup, BuildContext<MgmtOutputL
...
+        private ResourceListMethod? FindListMethod()
+        {
+            var listMethods = GetListMethods(true, true);
+            foreach (var listMethod in listMethods)
+            {
+                var restClientMethod = GetRestClientMethod(listMethod);
+                var pathSegments = restClientMethod?.Request.PathSegments;
+
+                // resourceGroups scope
+                if (pathSegments.Any(p => p.Value.IsConstant && p.Value.Constant.Value?.ToString() == "/resourceGroups/"))
```
m-nash, line 61: not sure if this should be hard coded.

Type 1 was List me at my parent scope.  parent is variable it could be resourcegroup but it could be subscription

We can validate this with a test case that has a subscription based resource, and its subscription level list method should go in the container.

I wonder if we name these slightly differently
ParentListMethod
AncestorListMethod -> this will be multiple since if your parent is RG then you might have 1 sub scope and 1 tenant scope

m-nash (changes requested):

ArcturusZhang (commented):
Hi @ShivangiReja please be sure to also remove the `Ignore` attribute on this test: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/Mgmt/TestProjects/MgmtListMethods.cs

src/AutoRest.CSharp/Mgmt/Output/ResourceContainer.cs
```diff
@@ -36,24 +37,51 @@ public ResourceContainer(OperationGroup operationGroup, BuildContext<MgmtOutputL
...
-                ?? PagingMethods.FirstOrDefault(m => m.Method.Name.StartsWith("List", StringComparison.InvariantCultureIgnoreCase));
+            return method.Request.HttpMethod.Equals(RequestMethod.Put);
         }
 
-        private RestClientMethod? GetPutMethod()
+        private bool IsCreateResourceMethod(RestClientMethod method)
         {
-            return RestClient.Methods.FirstOrDefault(m => m.Request.HttpMethod.Equals(RequestMethod.Put));
+            return method.Request.HttpMethod.Equals(RequestMethod.Put) &&
+                (method.Name.Equals("CreateOrUpdate") || method.Name.Equals("Create") || method.Name.Equals("Put"));
```
ArcturusZhang, line 84: Should we also add a criteria that this method is returning a resource data? Or to be more specific, the resource data that corresponds to the current resource container

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
```diff
@@ -68,24 +72,185 @@ internal class ResourceOperation : TypeProvider
...
+                }
+            }
+
+            return null;
+        }
+
+        protected IEnumerable<ResourceListMethod> GetListMethods(bool parentExistsInPathParam, bool returnTypeIsResourceData)
+        {
+            List<ResourceListMethod> listMethods = new List<ResourceListMethod>();
+            var parentResourceType = OperationGroup.ParentResourceType(_context.Configuration.MgmtConfiguration);
```
ArcturusZhang, line 193: We are not using this in this scope, maybe we could remove this line

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
```diff
@@ -68,24 +72,185 @@ internal class ResourceOperation : TypeProvider
...
+        {
+            var parentResourceType = OperationGroup.ParentResourceType(_context.Configuration.MgmtConfiguration);
+            bool isParentExistsInPathParams = false;
+            bool isReturnTypeResourceData = false;
+
+            var parentParamName = "";
+            if (clientMethod.Operation?.Requests.FirstOrDefault().Protocol.Http is HttpRequest httpRequest)
+            {
+                var path = httpRequest.Path;
+                var parentIndex = path.IndexOf(parentResourceType);
```
ArcturusZhang, line 224: I have a question: what will this `httpRequest.Path` return? For instance, if a request url is `/subscriptions/{subsID}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/disks/{diskName}`, is this `httpRequest.Path` returning `/subscriptions/resourceGroups/providers/Microsoft.Compute/disks`?
If not, we might never find the parentResourceType in it.
Also we might want to find this in a case insensitive way - some API has casing problems (for instance `resourceGroups` -> `resourcegroups`

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
```diff
@@ -68,24 +72,185 @@ internal class ResourceOperation : TypeProvider
...
+                if (parentIndex > 0)
+                {
+                    path = path.Substring(parentIndex + parentResourceType.Length);
+                    var pathArr = path.Split('/');
+                    char[] charsToTrim = { '{', '}' };
+                    parentParamName = pathArr[1].Trim(charsToTrim);
+
+                    isParentExistsInPathParams = clientMethod.PathParameters.Any(p => p.Name == parentParamName);
```
ArcturusZhang, line 232: Based on the "we have to trim the `{` and `}`" logic here, I assume the answer to my above question should be NO - then we might never get the correct parent other than the parent is `subscriptions`, `resourceGroups` or first level resources.
The string of sub resource type (e.g. `Microsoft.Network/virtualNetworks/subnets`) will never show in the request path.

Also it would help tons if we have some utility functions to do this pattern, I feel like we also need this when we are trying to get the correct parameters to pass in the operation writer or container writer

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
```diff
@@ -68,24 +72,185 @@ internal class ResourceOperation : TypeProvider
...
+                    return listMethod.PagingMethod.Method;
+                }
+                else if (listMethod.ClientMethod != null)
+                {
+                    return listMethod.ClientMethod.RestClientMethod;
+                }
+            }
+
+            return null;
+        }
```
ArcturusZhang, line 188: This looks more natural if this method belongs to the class `ResourceListMethod` instead of here

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
```diff
@@ -100,5 +265,18 @@ protected virtual string CreateDescription(OperationGroup operationGroup, string
...
+        {
+            public PagingMethod? PagingMethod { get; }
+
+            public ClientMethod? ClientMethod { get; }
+
+            public ResourceListMethod(PagingMethod? pagingMethod = null, ClientMethod? clientMethod = null)
+            {
+                PagingMethod = pagingMethod;
+                ClientMethod = clientMethod;
+            }
```
ArcturusZhang, line 279: Since we should always only assign one of these two (and never both `null`), maybe we need to introduce two extra constructor to assign only one of them each

I hope we could have union type in C# soon
```suggestion
            public ResourceListMethod(PagingMethod pagingMethod)
            {
                PagingMethod = pagingMethod;
                ClientMethod = null;
            }
            public ResourceListMethod(ClientMethod clientMethod)
            {
                PagingMethod = null;
                ClientMethod = clientMethod;
            }
```

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
```diff
@@ -68,24 +72,185 @@ internal class ResourceOperation : TypeProvider
...
-                return $"List{ResourceName.ToPlural()}";
+                var restClientMethod = GetRestClientMethod(listMethod);
+                var pathSegments = restClientMethod?.Request.PathSegments;
+
+                // subscriptions scope
+                if (pathSegments.Any(p => p.Value.IsConstant && p.Value.Constant.Value?.ToString() == "/subscriptions/") &&
+                    !pathSegments.Any(p => p.Value.IsConstant && (p.Value.Constant.Value?.ToString() == "/resourceGroups/" || p.Value.Constant.Value?.ToString() == "/providers/Microsoft.Compute/locations/")))
+                {
+                    return listMethod;
+                }
```
ArcturusZhang, line 110: It will look more clear if we could have some extension method to do something like this

ArcturusZhang (changes requested):
Hi @ShivangiReja please be sure to remove the `Ignore` attribute on [this test](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/Mgmt/TestProjects/MgmtListMethods.cs) so that we could validate the list methods are properly generated. And please also feel free to add more test cases.
The test project `MgmtListMethods` was initially created to test list method related things - but when I created this, the logic to generate the list methods is still buggy, therefore I marked it by `Ignore`

ShivangiReja:
Hi @ArcturusZhang, thanks for the feedback.

>Hi @ShivangiReja please be sure to also remove the Ignore attribute on this test: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/Mgmt/TestProjects/MgmtListMethods.cs

 I've already removed the `ignore` and added few extra test cases. Check here https://github.com/Azure/autorest.csharp/pull/1413/files#diff-84defb6c2e6662aad5f7135bff5a75137b3a3f6e3ad14c56f8b10f4634ea2701

src/AutoRest.CSharp/Mgmt/Output/ResourceContainer.cs
ShivangiReja: I believe Create method doesn't always return resource data. Check few examples here - [DedicatedHosts](https://github.com/Azure/autorest.csharp/blob/feature/v3/samples/Azure.ResourceManager.Sample/Generated/DedicatedHostsRestOperations.cs#L84), [Images](https://github.com/Azure/autorest.csharp/blob/feature/v3/samples/Azure.ResourceManager.Sample/Generated/ImagesRestOperations.cs#L81), [VirtualMachines](https://github.com/Azure/autorest.csharp/blob/feature/v3/samples/Azure.ResourceManager.Sample/Generated/VirtualMachinesRestOperations.cs#L243)...etc

Above checks work fine for all the resources.

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
ShivangiReja: Fixed this issue, now [this](https://github.com/Azure/autorest.csharp/blob/33a18da16c450c6204a16b262f58f229ea857227/src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs#L261) checks parent in `/subscriptions/resourceGroups/providers/Microsoft.Compute/disks` path.

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
ShivangiReja: I've added four methods for each extensions(Subscription, Resource Group, Tenant, Management Group) to get the List methods for the clarity and added conditions as per the extensions.

src/AutoRest.CSharp/Mgmt/Output/ResourceOperation.cs
ShivangiReja: [Here](https://github.com/Azure/autorest.csharp/blob/33a18da16c450c6204a16b262f58f229ea857227/test/TestProjects/MgmtMultipleParentResource/mgmtMultipleParentResource.json#L1412) is the example for this and [here](https://github.com/Azure/autorest.csharp/blob/33a18da16c450c6204a16b262f58f229ea857227/test/TestProjects/MgmtMultipleParentResource/Generated/ChildBodySubParentContainer.cs#L326) is the generated code.

ArcturusZhang (approved):
Approved with some comments

src/AutoRest.CSharp/Mgmt/Generation/ResourceContainerWriter.cs
```diff
@@ -279,14 +279,28 @@ private void WriteGetVariants(RestClientMethod method)
 
         private void WriteListVariants()
         {
-            if (_resourceContainer.ListMethod != null)
+            if (_resourceContainer.ListMethods != null)
```
m-nash, line 282: this null check is probably not needed since we can rely on foreach'ing over an empty list?

src/AutoRest.CSharp/Mgmt/Generation/ResourceContainerWriter.cs
```diff
@@ -279,14 +279,28 @@ private void WriteGetVariants(RestClientMethod method)
...
+                        WriteClientMethod(_writer, listMethod.ClientMethod, _resourceContainer.GetDiagnostic(listMethod.ClientMethod.RestClientMethod), _resourceContainer.OperationGroup, _context, true);
+                        WriteClientMethod(_writer, listMethod.ClientMethod, _resourceContainer.GetDiagnostic(listMethod.ClientMethod.RestClientMethod), _resourceContainer.OperationGroup, _context, false);
+                    }
+                }
 
-            WriteListAsGenericResource(async: false);
-            WriteListAsGenericResource(async: true);
+                // WriteRemainingPagingMethods();
+                WriteListAsGenericResource(async: false);
+                WriteListAsGenericResource(async: true);
```
m-nash, line 302: I think these still exist even if the ListMethods is empty, this is a separate API that we are exposing on each resource type for convenience, but allows you to generically request resources by resourceType but doesn't rely on the current resources RestClient at all.

src/AutoRest.CSharp/Mgmt/Generation/ResourceOperationWriter.cs
```diff
@@ -250,17 +250,34 @@ private void WriteClientMethods(CodeWriter writer, ResourceOperation resourceOpe
             // 3. Listing children (might be resource or not) -> on the operations
 
             // write rest of the methods
+            var resourceContainer = context.Library.GetResourceContainer(resourceOperation.OperationGroup);
             foreach (var clientMethod in resourceOperation.Methods)
             {
+                var isMethodExistInContainer = false;
```
m-nash, line 256: This logic should be in the property returned by resourceOperations otherwise any time we want the list of all the methods unique to the operations vs container we have to perform the same logic you have here for isMethodExistInContainer

src/AutoRest.CSharp/Mgmt/Generation/ResourceOperationWriter.cs
```diff
@@ -272,16 +289,24 @@ private void WriteClientMethods(CodeWriter writer, ResourceOperation resourceOpe
...
                 }
             }
 
             // write rest of the LRO methods
             foreach (var clientMethod in resourceOperation.RestClient.Methods)
             {
+                var isMethodExistInContainer = false;
+                if (resourceContainer != null)
+                {
+                    isMethodExistInContainer = clientMethod == resourceContainer.CreateMethod ||
```
m-nash, line 303: example of duplicate i mentioned above

m-nash (changes requested):

src/AutoRest.CSharp/Mgmt/Generation/ResourceContainerWriter.cs
ShivangiReja: Since this property is nullable, this check is needed in order to access _resourceContainer.ListMethod property.

test/AutoRest.TestServer.Tests/Mgmt/TestProjects/MgmtListMethods.cs
```diff
@@ -19,11 +20,34 @@ public class MgmtListMethods : TestProjectTests
         [TestCase("TheExtensionFakeContainer", "ListFeatures", false)]
         [TestCase("TheExtensionFakeOperations", "ListUsages", true)]
         [TestCase("TheExtensionFakeOperations", "ListFeatures", true)]
-        public void ValidateListOfNonResourceMethod(string operationOrContainerName, string methodName, bool exist)
+        [TestCase("SubscriptionExtensions", "ListFakes", true)]
+        [TestCase("FakeContainer", "ListFakes", false)]
+        [TestCase("FakeOperations", "ListFakes", false)]
+        [TestCase("SubscriptionExtensions", "ListByLocations", true)]
+        [TestCase("FakeContainer", "ListByLocations", false)]
```
m-nash, line 27: shouldn't the name of the method still be List with just an overload?

test/TestProjects/MgmtListMethods/Generated/Extensions/SubscriptionExtensions.cs
```diff
@@ -5,10 +5,210 @@
...
+            }
+            );
+        }
+
+        /// <summary> Lists all of the virtual machines in the specified subscription. Use the nextLink property in the response to get the next page of virtual machines. </summary>
+        /// <param name="subscription"> The <see cref="SubscriptionOperations" /> instance the method will execute against. </param>
+        /// <param name="location"> The location for which virtual machines under the subscription are queried. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="location"/> is null. </exception>
+        public static async Task<Response<IReadOnlyList<Fake>>> ListByLocationsAsync(this SubscriptionOperations subscription, string location, CancellationToken cancellationToken = default)
```
m-nash, line 126: this should be ListFakes so its an overload to the one that doesn't take in a location

m-nash (changes requested):

test/AutoRest.TestServer.Tests/Mgmt/TestProjects/MgmtListMethods.cs
ShivangiReja: Renamed it to List<ResourceName>sByLocation, since the parameter count and types are same for ListAll and ListByLocation.

m-nash (approved):

PR #1430, opened 2021-07-27
[LLC] Refactoring to share more code with DataPlane

ellismg:
As we've continued to work on low level client, we've started to find that there a lot of logic in the existing Track 2 dataplane libraries that we can leverage.  As I started to work on implementing pageable, I realized it would make sense to take a moment and try to remove some of the complexity we had built up by running fast in the LLC project.

As part of this, we get to address some low-hanging code issues we had.


src/AutoRest.CSharp/Common/Output/Builders/ClientBuilder.cs
```diff
@@ -18,9 +18,11 @@ namespace AutoRest.CSharp.Common.Output.Builders
     internal static class ClientBuilder
     {
         private const string ClientSuffixValue = "Client";
+        private const string RestClientPrefix = "Rest";
```
AlexanderSher, line 21: Do we need this one?

src/AutoRest.CSharp/Common/Output/Models/RestClient.cs
```diff
@@ -29,22 +29,21 @@ public RestClient(OperationGroup operationGroup, BuildContext context, string? c
...
             DefaultName = ClientPrefix + RestClientSuffix;
-            Description = "";
         }
 
         protected RestClientBuilder Builder;
         internal OperationGroup OperationGroup { get; }
         protected string RestClientSuffix { get; }
-        public Parameter[] Parameters { get; }
-        public string Description { get; }
+        public virtual Parameter[] Parameters { get; }
```
AlexanderSher, line 45: I'm not sure I understand why it has to be virtual?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -163,6 +161,8 @@ private void WriteClientMethodDecleration(CodeWriter writer, LowLevelClientMetho
             {
                 writer.WriteXmlDocumentationParameter(parameter.Name, $"{parameter.Description}");
             }
+            writer.WriteXmlDocumentationParameter("options", $"The request options.");
```
AlexanderSher, line 164: Can we add `options` to the list of `parameters`?

src/AutoRest.CSharp/Common/Output/Models/RestClient.cs
ellismg: I don't think they need to be either - possibly this is just left over from some exploratory stuff I was doing.  I will remove.  Thanks for pointing this out!

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
ellismg: It could be made to work (and indeed, previously we did this).  However, the above is more in line with how things are modeled in the standard data-plane builders/writers and it's nice that `Parameters` matches just the parameters of the underlying service operation.  I'd personally prefer to keep it like this,

src/AutoRest.CSharp/Common/Output/Builders/ClientBuilder.cs
ellismg: We only use it in a single place - I'll inline it.

src/AutoRest.CSharp/Common/Output/Models/RestClient.cs
ellismg: Actually - I take it back after I tried to undo it.  The issue is that `LowLevelRestClient` (a subclass of this) needs to override the `Parameters` to remove the api-version parameter (since it instead is handled as part of the client options).  You can see that `LowLevelRestClient.cs` provides implementations for both this and `Description`,

ellismg: FYI - The changes I am currently working on for Pageable should allow me to remove this.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
AlexanderSher: Sorry, I don't understand. In several places there is an ad-hoc handling of the `options` parameter that otherwise will be called inside iteration. It is a guaranteed error prone.

src/AutoRest.CSharp/Common/Output/Models/RestClient.cs
ellismg: This did get cleaned up with my Pageable changes.  Going to leave this unresolved for now but it will get fixed up with my next PR.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
ellismg: I would like us to handle the `options` parameter here the same way we handle the `cancellationToken` in the other generators.  In these generators, we explicitly add the `cancellationToken` parameter in the writers instead of adding it to the `Parameters` property of the `RestClientMethod`.

I'm worried that we are going to end up trading one set of special cases for another - the entries in the `Parameter` array today correspond to service parameters which end up being included in the generated HTTP message, unlike `cancellationToken` or `options`.

I'm happy to try to make the change if you want, but I think if we are going to go down this path we want to follow a similar strategy for the management and data-plane writers, which is starting to feel like a lot of work.

Do you have problems with me landing what we have here and then exploring in a separate PR for all the generators?

AlexanderSher: No, I've been talking about `parameters` variable, not a `Parameters` property of `RestClientMethod`.

ellismg: > No, I've been talking about parameters variable, not a Parameters property of RestClientMethod.

Ahh!  Sorry for misunderstanding.  I can make that change!

ellismg: Pushed another change, @AlexanderSher that does what I think you were after.  Thanks for explaining what you were looking for and sorry for not picking it up sooner.  Let me know what you think.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -120,9 +118,12 @@ private void WriteClientMethodBody(CodeWriter writer, LowLevelClientMethod clien
             writer.Line();
         }
 
+        private static readonly CSharpType RequestOptionsParameterType = new CSharpType(typeof(RequestOptions), true);
+        private static readonly Parameter RequestOptionsParameter = new Parameter("options", "The request options", RequestOptionsParameterType, Constant.Default(RequestOptionsParameterType), false);
+
         private void WriteClientMethodDecleration(CodeWriter writer, LowLevelClientMethod clientMethod, bool async)
         {
-            var parameters = clientMethod.Parameters;
+            var parameters = clientMethod.Parameters.Concat(new Parameter[] { RequestOptionsParameter });
```
AlexanderSher, line 126: `.Append(RequestOptionsParameter)`

AlexanderSher (approved):

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
ellismg: Wow! Somehow I've gone this long not knowing `Append` exits!  Thanks for the pointer and saving me a bunch of characters down the road.

PR #1433, opened 2021-07-28
update base class names

m-nash:
# Description

# Checklist

# Ready to Land?

### DIFF
--- a/samples/Azure.Management.Storage/Generated/StorageAccountOperations.cs
+++ b/samples/Azure.Management.Storage/Generated/StorageAccountOperations.cs
@@ -19,7 +19,7 @@ using Azure.ResourceManager.Resources.Models;
 namespace Azure.Management.Storage
 {
     /// <summary> A class representing the operations that can be performed over a specific StorageAccount. </summary>
-    public partial class StorageAccountOperations : ResourceOperationsBase<StorageAccount>
+    public partial class StorageAccountOperations : ResourceOperations
     {
         private readonly ClientDiagnostics _clientDiagnostics;
         private StorageAccountsRestOperations _restClient { get; }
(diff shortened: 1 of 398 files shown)


samples/Azure.Management.Storage/Generated/StorageAccountOperations.cs
```diff
@@ -45,40 +45,6 @@ protected internal StorageAccountOperations(OperationsBase options, ResourceIden
...
-                var response = _restClient.GetProperties(Id.ResourceGroupName, Id.Name, null, cancellationToken);
-                return Response.FromValue(new StorageAccount(this, response.Value), response.GetRawResponse());
-            }
-            catch (Exception e)
-            {
-                scope.Failed(e);
-                throw;
-            }
-        }
-
```
ShivangiReja, line 81: If we are removing this method, then I think we should pass `StorageAccountExpand? expand` parameter as optional, like [rest client's method](https://github.com/Azure/autorest.csharp/blob/8efc2364091e61ced50cec9c9305a8e5aa063d9c/samples/Azure.Management.Storage/Generated/RestOperations/StorageAccountsRestOperations.cs#L309).

Example: 
```csharp
public async virtual Task<Response<StorageAccount>> GetAsync(StorageAccountExpand? expand = null, CancellationToken cancellationToken = default)
```

m-nash: that already exists, before we had 2 overloads, now we only have the one which is the one you mentioned on line 70

ShivangiReja: Yes, the get method overload exists, but it takes nullable parameter(StorageAccountExpand? expand) but it should take nullable and optional parameter(StorageAccountExpand? expand = null) just like rest client method.

ShivangiReja: Yes, the get method overload exists, but it takes nullable parameter(StorageAccountExpand? expand), now that we removed this get method, user has to always pass something in expand  parameter. To fix this, it should take nullable and optional parameter(StorageAccountExpand? expand = null) just like rest client method.

ShivangiReja (approved):

PR #1434, opened 2021-07-28
fix(generator): incorrect assignment of some special read-only collection properties

archerzz:
# Description
There are some special resource definitions which will make some properties read-only, regardless of their original definition in Swagger.

An example is `DdosProtectionPlan`: https://github.com/Azure/azure-rest-api-specs/blob/83c0c488dfe3f49d7c666511d5d6714ec6b4502f/specification/network/resource-manager/Microsoft.Network/stable/2021-02-01/ddosProtectionPlan.json#L376
- Its `Tags` property is writable.
- But finally `DdosProtectionPlan` will be recognized as `TrackedResource` of which the `Tags` property is read-only.

In our code generator, since `Tags` original definition is writable, the generated code will use `Tags = tags` to initialize, which will cause compilation error.

This commit will remove the restriction, so that collection properties are always initialized using `Collection.Add`, regardless of whether it's read-only or not. That will bring slight performance penalty, but:
- It's the neat and safe way to initialize a collection.
- It also avoids side effect of any changes in source collections.

AB#6068

# Checklist

# Ready to Land?


ArcturusZhang (approved):
LGTM

ShivangiReja:
@pakrym Could you please take a look?

archerzz:
After inspecting the CI error, I realized that the proposed solution will bring severe regression: we **cannot** always append items to initialize collection properties.

In JSON output, a collection property can have different form of zero value:
- `null`
- doesn't exist
- empty `[]`

`null` or `[]` relies on assignment of the collection properties. So if we always append items on initialization, then `null` or `[]` will all be transformed as `doesn't exist`.

The thorough fix need to understand how to correctly determined the `read-only` of a property eventually.

archerzz:
This PR is close to fixing the original problem. The broken test cases are for data plane, because in data plane, users are allowed to assign whole collection properties. So we can make further change to only use `append` behavior for control plane (e.g. mgmt library) codes.

Anyway, after discussion, we decide to disable payload flattening first. In the future, if there is requirement to flatten payload parameters for control plane, we can resume the work here.

For details, see ADO 6068.

PR #1456, opened 2021-08-09
Use PropertyReferenceType from Core package and do not generate them in other RP packages

fengzhou-msft:
# Description
This PR fixed some cases of Property matching for Resources SDK so that for the shared models with Core package, Resources SDK will reference them from vCore and do not generate them again.

**TODO:**
Property matching logic needs further improvement, for instances, ErrorResponse will only be matched in Resources and not in other RPs with a different namespace. The tricky part is that `ErrorResponse` has a property of `List<ErrorResponse>`. 
Eventually, we may need to compare all subproperties recursively but store the Class types we are comparing and return true for a Type that was already stored for comparison to avoid infinite loop. 

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
@@ -522,6 +522,11 @@ namespace AutoRest.CSharp.Generation.Writers
                 writer.Append($"({frameworkType}){element}.GetString()");
                 return;
             }
+            else if (frameworkType.Namespace?.StartsWith("Azure.ResourceManager") == true)
+            {
+                writer.Append($"{frameworkType}.Deserialize{frameworkType.Name}({element})");
+                return;
+            }
             else
                 writer.Append($"{element}.");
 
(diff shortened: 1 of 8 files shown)


src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -522,6 +522,11 @@ private static void DeserializeFrameworkTypeValue(CodeWriter writer, CodeWriterD
                 writer.Append($"({frameworkType}){element}.GetString()");
                 return;
             }
+            else if (frameworkType.Namespace?.StartsWith("Azure.ResourceManager") == true)
+            {
+                writer.Append($"{frameworkType}.Deserialize{frameworkType.Name}({element})");
+                return;
+            }
```
fengzhou-msft, line 529: When matching some Resources models for Core package, we will not use PropertyReferenceType (which is themselves) for them and as they can find the same type in Core, they will be treated as frameworkType. Need to add this deserialize logic for them.

Further thinking: We should probably rethink the definition of frameworktype, should we treat those classes in `Azure.Resourcemanager` as frameworktypes?

PR #1468, opened 2021-08-17
Fix property match and remove models that are replaced

fengzhou-msft:
# Description
This PR fixed the property match for PropertyReferenceType so it will not only support the existing ResourceIdentity, but also support more common matches like UserAssignedIdentity, ProviderResourceType, and the ErrorResponse case which contains a property of List<ErrorResponse>.

This PR also changed the property name to be the same as the one before the property is replaced. This would make the SDK more consistent in public interface when we add more PropertyReferenceTypes later and also avoid the issue like in some RP there is an ErrorResponse containing an Error property with ErrorDetail type and ErrorDetail actually matches with the ErrorResponse in Core, the previous code then changes the property name from Error to ErrorResponse, causing the error of `member names cannot be the same as their enclosing type`.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/AutoRest.CSharp.csproj
```diff
@@ -17,7 +17,8 @@
   <ItemGroup>
     <PackageReference Include="Azure.Core" Version="1.18.0-alpha.20210811.1" />
     <PackageReference Include="Azure.Core.Experimental" Version="0.1.0-preview.12" />
-    <PackageReference Include="Azure.ResourceManager" Version="1.0.0-alpha.20210811.3" />
+    <!-- <PackageReference Include="Azure.ResourceManager" Version="1.0.0-alpha.20210811.3" /> -->
+    <ProjectReference Include="../../../azure-sdk-for-net/sdk/resourcemanager/Azure.ResourceManager/src/Azure.ResourceManager.csproj" />
```
allenjzhang, line 21: hope this is temporary.

fengzhou-msft: Yes, once https://github.com/Azure/azure-sdk-for-net/pull/23091 is merged and the new version of Azure.ResourceManager is published, will use that new version.

test/TestProjects/MgmtPropertyChooser/Generated/Models/CloudError.cs
```diff
@@ -0,0 +1,34 @@
...
+// <auto-generated/>
+
+#nullable disable
+
+using Azure.ResourceManager.Resources.Models;
+
+namespace MgmtPropertyChooser.Models
+{
+    /// <summary> An error response for a resource management request. </summary>
+    internal partial class CloudError
```
allenjzhang, line 13: I assume this is also waiting for the PropertyReferenceType attribute on ErrorResponse?

fengzhou-msft: Yes, now I generate the Models based on local changes of Azure.ResourceManager.

src/AutoRest.CSharp/Mgmt/Decorator/ReferenceTypePropertyChooser.cs
```diff
@@ -33,48 +37,56 @@ private static IList<System.Type> GetReferenceClassCollection()
...
+            if (_valueCache.TryGetValue(typeToReplace, out var result))
+                return result;
             foreach (System.Type replacementType in GetReferenceClassCollection())
             {
-                // flatten properties
-                List<PropertyInfo> replacementTypeProperties = replacementType.GetProperties(BindingFlags.Public | BindingFlags.Instance).ToList();
-                List<PropertyInfo> flattenedReplacementTypeProperties = new List<PropertyInfo>();
-                foreach (var parentProperty in replacementTypeProperties)
+                // Quick match for models in Resources SDK with models in the same namespace of Core package through the full name of "Namespace.Name"
+                if ($"{typeToReplace.Type.Namespace}.{typeToReplace.Type.Name}" == $"{replacementType.Namespace}.{replacementType.Name}")
```
fengzhou-msft, line 47: As the ErrorResponse used by Resources SDK and the ErrorResponse in common-type v2 of the Azure.REsourceManager package will be 2 different models, will dump this quick match logic for Resources SDK.

fengzhou-msft:
This PR is replaced by #1479 to consolidate changes in #1477.

PR #1470, opened 2021-08-18
Fix #1247: Remapping a property to a field results in invalid parameter capitalization

### LINKED ISSUE #1247: Remapping a property to a field results in invalid parameter capitalization
``` C#
    [CodeGenModel("Response")]
    public partial class MetricQueryResult
    {
        [CodeGenMember("Timespan")]
        private readonly string _timespan;
```

Results in the following constructor:

```C#
internal MetricQueryResult(string Timespan, IEnumerable<Metric> metrics)
        {
```

AlexanderSher:


test/AutoRest.TestServer.Tests/TypeSchemaMappingTest.cs
```diff
@@ -100,7 +100,7 @@ public void ObjectsAreMappedToSchemas()
         [Test]
         public void NullableParamsAsNull()
         {
-            var model = new RenamedModelStruct("test", null, null);
```
pakrym, line 103: should we use an explicit parameter name here to test that we get the expected name?

pakrym: Maybe worth adding something to https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs

pakrym:
This PR seems to have some unrelated changes.

pakrym (approved):

test/TestServerProjects/lro/Generated/LROsRestClient.cs
```diff
@@ -86,60 +86,6 @@ public Response Put200Succeeded(Product product = null, CancellationToken cancel
...
-            await _pipeline.SendAsync(message, cancellationToken).ConfigureAwait(false);
-            var headers = new LROsPatch200SucceededIgnoreHeadersHeaders(message.Response);
-            switch (message.Response.Status)
-            {
-                case 200:
-                    return ResponseWithHeaders.FromValue(headers, message.Response);
-                default:
-                    throw await _clientDiagnostics.CreateRequestFailedExceptionAsync(message.Response).ConfigureAwait(false);
-            }
-        }
```
ShivangiReja, line 124: autorest.testserver version has been updated. Please ensure to run `npm install` first before you regenerate all the test projects.

PR #1508, opened 2021-09-10
Fix memory issue when running UpdateAzureSdkForNet.ps1

fengzhou-msft:
# Description
The [AutoRest C# - Publish](https://dev.azure.com/azure-sdk/internal/_build?definitionId=1375&_a=summary) pipeline is failing, caused by no enough memory when generating Network SDK. This PR fixes it by setting `max-old-space-size` to 4G.

# Checklist

# Ready to Land?

### DIFF
--- a/eng/UpdateAzureSdkForNet.ps1
+++ b/eng/UpdateAzureSdkForNet.ps1
@@ -13,5 +13,5 @@ $PackagesProps = "$SdkRepoRoot\eng\Packages.Data.props"
     "<PackageReference Update=`"Microsoft.Azure.AutoRest.CSharp`" Version=`"$Version`" PrivateAssets=`"All`" />" | `
     Set-Content $PackagesProps -NoNewline
 
-$Env:NODE_OPTIONS = "--max-old-space-size=8192"
+$Env:NODE_OPTIONS = "--max-old-space-size=4096"
 dotnet msbuild /restore /t:GenerateCode "$SdkRepoRoot\eng\service.proj"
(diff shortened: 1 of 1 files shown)


eng/UpdateAzureSdkForNet.ps1
```diff
@@ -13,4 +13,5 @@ $PackagesProps = "$SdkRepoRoot\eng\Packages.Data.props"
     "<PackageReference Update=`"Microsoft.Azure.AutoRest.CSharp`" Version=`"$Version`" PrivateAssets=`"All`" />" | `
     Set-Content $PackagesProps -NoNewline
 
+$Env:NODE_OPTIONS = "--max-old-space-size=8192"
```
ellismg, line 16: Just wondering - it looks like from the documentation for the hosted runners (https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#hardware) the machines end up having 7GB of memory.  I have confirmed locally that the generation completes without error with `max-old-space-size=4096`.  Maybe we want to use that smaller value so we don't end up having to swap in CI?

ellismg (approved):
LGTM - Happy to keep `max-old-space-size` at 8GB, but do wonder if we should move it back to 4GB for now given the hardware we end up running these pipelines on.

Thanks so much for fixing this!

weshaggard:
Is there any way we can do this in a more generic place like @AlexanderSher is doing in https://github.com/Azure/azure-sdk-for-net/pull/23961?

Also do we know why we need the extra memory? Have we done any profiling to see if we can get that reduced so everyone that runs our code generator doesn't end up needing to set this?

pakrym:
We should do it in https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/build/CodeGeneration.targets#L34

fengzhou-msft:
> Is there any way we can do this in a more generic place like @AlexanderSher is doing in [Azure/azure-sdk-for-net#23961](https://github.com/Azure/azure-sdk-for-net/pull/23961) or #1513.
> 
> Also do we know why we need the extra memory? Have we done any profiling to see if we can get that reduced so everyone that runs our code generator doesn't end up needing to set this?

Currently, the OOM is caused by the generation of Network SDK. Add owner @archerzz to take a look.

AlexanderSher:
https://github.com/Azure/autorest.csharp/pull/1513 has fixed the issue, so I'm closing this PR

PR #1514, opened 2021-09-14
[LLC] Pageable Support

ellismg:
This set of changes adds support to the code generator for LLC to emit `AsyncPageable<BinaryData>/Pageable<BinaryData>`  for operations which return paged data.  This allows you to access each element (as a BinaryData of the JSON text) of the paged collection using our traditional abstractions.

The change is broken into two parts, the first does some more refactoring in our client generation to have two clients, an internal REST client which just returns Response and a public "Low Level Data Plane" client which has the public facing methods and uses abstractions like `Operation` and `Pageable` for LROs and Pageable (similar to how the HLC client works).

As part of this work I took the existing pageable tests we had (which used models) and used them to produce a set of tests for the LLC version.


src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelOutputLibrary.cs
```diff
@@ -18,16 +18,20 @@ internal class LowLevelOutputLibrary : OutputLibrary
...
         private readonly BuildContext<LowLevelOutputLibrary> _context;
         private CachedDictionary<OperationGroup, LowLevelRestClient> _restClients;
+        private CachedDictionary<OperationGroup, LowLevelDataPlaneClient> _clients;
 
         public LowLevelOutputLibrary(CodeModel codeModel, BuildContext<LowLevelOutputLibrary> context) : base(codeModel, context)
         {
             _codeModel = codeModel;
             _context = context;
             _restClients = new CachedDictionary<OperationGroup, LowLevelRestClient>(EnsureRestClients);
+            _clients = new CachedDictionary<OperationGroup, LowLevelDataPlaneClient>(EnsureClients);
```
AlexanderSher, line 28: We need better names for these two fields.

src/assets/Generator.Shared/LowLevelBinaryDataOperation.cs
```diff
@@ -12,47 +12,14 @@
 
 namespace Azure.Core
 {
-    /// <summary>
-    /// The common operation type used by all low level client methods. The result of the operation is the `Response.Content` property of the final state.
-    /// </summary>
-    internal class LowLevelBinaryDataOperation : Operation<BinaryData>, IOperationSource<BinaryData>
+    internal class LowLevelBinaryDataOperation : FuncOperation<BinaryData>
```
AlexanderSher, line 15: What is the other type derived from `FuncOperation`?

src/assets/Generator.Shared/LowLevelPagableHelpers.cs
```diff
@@ -0,0 +1,121 @@
...
+                }
+            }
+
+            throw new InvalidOperationException("invalid JSON");
+        }
+
+        /// <summary>
+        /// Reads the items and next link from a response for a pageable operation. The values returned are BinaryDatas formed over the underlying content.
+        /// </summary>
+        public static (IEnumerable<BinaryData> Items, string? NextLink) GetItemsAndNextLinkFromJson(BinaryData content, string itemPropertyName = "value", string nextLinkPropertyName = "nextLink")
```
AlexanderSher, line 83: Does it have to be public?

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -1358,7 +862,71 @@ public virtual async Task<Response> GetDocumentsStatusAsync(Guid id, int? top =
...
+                }
+            }
+
+            async Task<Page<BinaryData>> NextPageFunc(string nextLink, int? pageSizeHint)
+            {
+                using var scope = _clientDiagnostics.CreateScope("DocumentTranslationClient.GetTranslationsStatus");
+                scope.Start();
+                try
+                {
+                    Response response = await RestClient.GetTranslationsStatusNextPageAsync(nextLink, top, skip, maxpagesize, ids, statuses, createdDateTimeUtcStart, createdDateTimeUtcEnd, orderBy, options).ConfigureAwait(false);
```
AlexanderSher, line 892: Multiple enumerations of `ids` in several async contexts. Potential deadlock.

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
```diff
@@ -0,0 +1,2332 @@
...
+{
+    /// <summary> The Paging service client. </summary>
+    public partial class PagingClient
+    {
+        /// <summary> The HTTP pipeline for sending and receiving REST requests and responses. </summary>
+        public virtual HttpPipeline Pipeline { get => _pipeline; }
+        private HttpPipeline _pipeline;
+        private const string AuthorizationHeader = "Fake-Subscription-Key";
+        private readonly AzureKeyCredential _keyCredential;
+        private Uri endpoint;
```
ShivangiReja, line 24: nit: rename endpoint -> _endpoint

src/assets/Generator.Shared/LowLevelBinaryDataOperation.cs
ellismg: Nothing else derives from it, but we code-gen instances of `FuncOperation<BinaryData>` to return in cases where the result of the LRO is pageable (https://github.com/Azure/autorest.csharp/pull/1514/files/57772ced9de4fff85a47bf465a96adb3e61d2e01#diff-dd69b9ac81b22be268f126e383ffffde84f2520679523a6691c808008c1fd3a6R213)

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelOutputLibrary.cs
ellismg: Do you have any thoughts on better names?  I might be too close to the problem here to come up with some other terms.  This mirrors the "Clients" and "RestClients" design we have in `DataPlaneOutputLibrary.cs`.

Would `_internalRestClients` and `_publicClients` be better here (and would give a clearer indication that one of these is internal machinery to the library and the other is about something we publicly expose to clients)?

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
ellismg: The Pageable abstraction does ensure only a single one of these async calls is ever running at once (it waits for `FirstPageFunc` to complete before calling `NextPageFunc` and ensures there are not multiple calls in flight.

Would you prefer that for any IEnumerable argument we force enumeration once (via calling .ToArray()) at the start of the method and just use that copy to the underlying service methods?

For example, here we could instead of something like:

```
options ??= new RequestOptions();
enumeratedIds = (ids)?.ToArray();
enumeratedStatueses = (statuses)?.ToArray();
enumeratedOrderBy = (orderBy)?.ToArray();
```

and then use the `enumerated` counterparts in the underlying calls to the RestClient methods.

src/assets/Generator.Shared/LowLevelPagableHelpers.cs
ellismg: It does not, thanks for catching this.  I've made it private.

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
ellismg: This may be a little tricky to address, I'd like your feedback on the approach before I start.

The `LowLevelDataPlaneClient` and `LowLevelRestClient` both `Parameters` properties which are the collection of service parameters which are placed on the client via `x-ms-client-location`.  We use the names exactly both in method signatures and local variables.

I can update the logic such that we prepend all service level parameters with a `_`, or we can treat `endpoint` special for low level client.  Would we want to do this for the other generators (like the existing Track 2 generator)?

If we want to make this change, do you mind me deferring the work for another PR, I think it will be slightly involved.

As a side note, I am going to make one small change here, which is to move these service parameters to the end of this giant list of class level variables so it's not intermixed with the fields that the code generator always writes.

ShivangiReja: Why do we even need the `private Uri endpoint`? We are not using it anywhere in the class.

ellismg: That's a really good point, @ShivangiReja. We shouldn't need to retain any of these parameters here.  The Rest Level Client will have them and store them, I can remove them.  Thanks for pointing this out and digging into this!

ShivangiReja: >The LowLevelDataPlaneClient and LowLevelRestClient both Parameters properties which are the collection of service parameters which are placed on the client via x-ms-client-location. We use the names exactly both in method signatures and local variables.

For RestClient I'm fine with using the same name as service parameters because that's what we do for [data plane](https://github.com/Azure/autorest.csharp/blob/57772ced9de4fff85a47bf465a96adb3e61d2e01/test/TestProjects/PublicClientCtor/Generated/PublicClientCtorRestClient.cs#L20-L25) as well.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
AlexanderSher: > Would you prefer that for any IEnumerable argument we force enumeration once (via calling .ToArray()) at the start of the method and just use that copy to the underlying service methods?

Yes!

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelOutputLibrary.cs
AlexanderSher: Yes, that would probably be better names.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -23,8 +23,8 @@ public partial class DocumentTranslationClient
         private const string AuthorizationHeader = "Ocp-Apim-Subscription-Key";
         private readonly AzureKeyCredential _keyCredential;
         private string endpoint;
-        private readonly string apiVersion;
         private readonly ClientDiagnostics _clientDiagnostics;
+        private readonly DocumentTranslationRestClient RestClient;
```
AlexanderSher, line 27: `_restClient`?

ellismg: Fixed!

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
ellismg: Fixed!

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelOutputLibrary.cs
ellismg: Adopted these names.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
ellismg: Pushed https://github.com/Azure/autorest.csharp/pull/1514/commits/1c8c4498ba04f4b6a0c285571172e131b7eac830 which does this (I used the suffix `Values` for the locals we create).

src/assets/Generator.Shared/LowLevelBinaryDataOperation.cs
ellismg: If you would rather `LowLevelBinaryDataOperation` not inherit from `FuncOperations` (and have me seal both types) I'm happy to do that.  I extracted the guts of `LowLevelBinaryDataOperation` into `FuncOperation<BinaryData>` so we had an easy way to return an operation which returns a pageable without having to go codegen a specific type per operation (which we wouldn't want to expose).

ellismg:
Thanks for taking a look, @AlexanderSher and @ShivangiReja.  I've responded to all your feedback with a series of follow on commits, please take another look when you have some time.

eng/Generate.ps1
```diff
@@ -84,19 +84,20 @@ if (!($Exclude -contains "TestServer"))
     }
 }
 
-$llcArgs = "--low-level-client=true --security=AzureKey --security-header-name=Fake-Subscription-Key"
+$llcArgs = "--low-level-client=true --security=AzureKey --security-header-name=Fake-Subscription-Key --modelerfour.group-parameters=false"
```
AlexanderSher, line 87: Looks like you don't need this change anymore with your previous PR

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationRestClient.cs
```diff
@@ -0,0 +1,1283 @@
...
+        /// <summary> Initializes a new instance of DocumentTranslationRestClient. </summary>
+        /// <param name="clientDiagnostics"> The handler for diagnostic messaging in the client. </param>
+        /// <param name="pipeline"> The HTTP pipeline for sending and receiving REST requests and responses. </param>
+        /// <param name="endpoint"> Supported Cognitive Services endpoints (protocol and hostname, for example: https://westus.api.cognitive.microsoft.com). </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="endpoint"/> is null. </exception>
+        public DocumentTranslationRestClient(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, string endpoint)
+        {
+            this.endpoint = endpoint ?? throw new ArgumentNullException(nameof(endpoint));
+            _clientDiagnostics = clientDiagnostics;
+            _pipeline = pipeline;
```
AlexanderSher, line 33: Do we need to store an `apiVersion`?

eng/Generate.ps1
ellismg: Correct!  I will rebase on top of that change and remove this.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationRestClient.cs
ellismg: I think the code we are generating here is correct.  The TA services don't use an `apiVersion` parameter, instead the version is part of the resource path, and encoded as part of the [`hostTemplate`](https://github.com/Azure/azure-rest-api-specs/blob/3196a62202976da192d6da86f44b02246ca2aa97/specification/cognitiveservices/data-plane/TranslatorText/preview/v1.0-preview.1/TranslatorBatch.json#L1545).

ellismg: We emit the [same shape of code](https://github.com/Azure/azure-sdk-for-net/blob/e602d34b34084d4fff731f237b7a84fd3acaeca2/sdk/translation/Azure.AI.Translation.Document/src/Generated/DocumentTranslationRestClient.cs#L32-L36) from the Track 2 Data Plane generator for this service in `azure-sdk-for-net`.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelDataPlaneClientWriter.cs
```diff
@@ -0,0 +1,701 @@
...
+                    writer.Line();
+                }
+
+                writer.Line($"return PageableHelpers.{enumerableFactoryMethod}(FirstPageFunc, {(clientMethod.PagingResponseInfo.NextPageMethod != null ? "NextPageFunc" : "null")});");
+            }
+
+            writer.Line();
+        }
+
+        private bool IsIEnumerableType(CSharpType type)
```
AlexanderSher, line 105: Can you move this to `TypeFactory`?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelDataPlaneClientWriter.cs
```diff
@@ -0,0 +1,701 @@
...
+        {
+            return type.IsFrameworkType && (type.FrameworkType == typeof(IEnumerable) || (type.FrameworkType.IsGenericType && type.FrameworkType.GetGenericTypeDefinition() == typeof(IEnumerable<>)));
+        }
+
+        private void WriteEnumerableArgumentEnumerationStatements(CodeWriter writer, RestClientMethod clientMethod, out Dictionary<string, CodeWriterDeclaration> enumeratedParameterMap)
+        {
+            enumeratedParameterMap = new Dictionary<string, CodeWriterDeclaration>();
```
AlexanderSher, line 112: Why `out` instead of just return?

ellismg: I will change it.  My thought here was that it would be a little strange to have a non void returning `WriteXyz` method, but I'll just use a `return` here.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelDataPlaneClientWriter.cs
ellismg: Fixed.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelDataPlaneClientWriter.cs
ellismg: Fixed.

eng/Generate.ps1
ellismg: Merged the parent branch in and undid this change.

src/assets/Generator.Shared/LowLevelBinaryDataOperation.cs
ellismg: @AlexanderSher Thanks for explaining your concern to me earlier today.  I pushed https://github.com/ellismg/autorest.csharp/pull/1514/commits/2c6b0d9f29c976f6ab0cb0d4fae8b3dc08658389 to the PR which removes the `LowLevelBinaryDataOperation` type in favor of just using `LowLevelFuncOperation<T>` everywhere.  I think this gets at the issue you raised, but please let me know if you'd like something else done.

Thanks again for all your help improving the code here and the very in depth review!

AlexanderSher (approved):

PR #1519, opened 2021-09-15
Handle double.NaN in Utf8JsonWriterExtensions

christothes:


src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -75,7 +75,14 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
                     writer.WriteNumberValue(d);
                     break;
                 case double d:
-                    writer.WriteNumberValue(d);
+                    if (double.IsNaN(d))
```
pakrym, line 78: What's the default behavior? Do we know if using "NaN" string for non is consistent across Azure?

christothes: The default behavior is to throw System.ArgumentException: .NET number values such as positive and negative infinity cannot be written as valid JSON.

I'm not sure if "NaN" is consistent across Azure, but it is for storage and [System.Text.Json](https://github.com/dotnet/runtime/issues/30255)

christothes: "NaN" seems fairly common based on [session records](https://grep.app/search?q=%22NaN%22&filter[repo][0]=Azure/azure-sdk-for-net&filter[lang][0]=JSON)

AlexanderSher (approved):

PR #1536, opened 2021-09-25
Fix build script

AlexanderSher:


test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
```diff
@@ -25,13 +25,15 @@ public TestServerV1()
         {
             var portPhrase = "Started server on port ";
             var startup = Path.Combine(GetBaseDirectory(), "dist", "cli", "cli.js");
+            var coverageDirectory = Path.Combine(GetArtifactsDirectory(), "coverage");
```
pakrym, line 28: Does the test server create a directory when it doesn't exist?

pakrym (approved):

test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj
```diff
@@ -7,6 +7,12 @@
     <NoWarn>SA1649;SA1633</NoWarn>
   </PropertyGroup>
 
+  <ItemGroup>
+    <AssemblyAttribute Include="AutoRest.TestServer.Tests.Infrastructure.BuildProperties">
```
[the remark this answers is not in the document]
AlexanderSher: I think I would add another value to the attribute, so that we can control both locations from build script.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerV1.cs
AlexanderSher: yes, it does.

PR #1538, opened 2021-09-27
[LLC] Return Response<bool> for head operations when head-as-boolean flag is true

ShivangiReja:
As per [this](https://github.com/Azure/autorest/issues/3603)(just like data plane), returning `Response<bool>` type for head operations when `head-as-boolean` flag is set to true. The default is set to false.

Fixes: https://github.com/Azure/autorest.csharp/issues/1334


src/AutoRest.CSharp/LowLevel/Generation/LowLevelRestClientWriter.cs
```diff
@@ -148,30 +150,29 @@ private void WriteOperation(CodeWriter writer, RestClientMethod operation, bool
...
-
-                if (async)
-                {
-                    writer.Line($"await {PipelineField:I}.SendAsync({messageVariable}, options.CancellationToken).ConfigureAwait(false);");
-                }
-                else
-                {
-                    writer.Line($"{PipelineField:I}.Send({messageVariable}, options.CancellationToken);");
-                }
+                writer.Line($"{GetAwait(async)}{PipelineField:I}.Send{GetAsyncSuffix(async)}({messageVariable}, options.CancellationToken){GetConfigureAwait(async)};");
```
AlexanderSher, line 153: I'd rather leave it as two separate lines.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelRestClientWriter.cs
```diff
@@ -191,19 +192,49 @@ private void WriteStatusCodeSwitch(CodeWriter writer, RestClientMethod clientMet
...
+                        writer.Line($"return {messageVariable}.{nameof(HttpMessage.Response)};");
+                    }
                 }
 
                 writer.Line($"default:");
-                if (async)
+                if (isNoThrow)
                 {
-                    writer.Line($"throw await {ClientDiagnosticsField}.CreateRequestFailedExceptionAsync({messageVariable}.{nameof(HttpMessage.Response)}).ConfigureAwait(false);");
+                    writer.Append($"return {typeof(ResponseWithError)}.FromError<{clientMethod.ReturnType}>({messageVariable}.{nameof(HttpMessage.Response)}, {GetAwait(async)}{ClientDiagnosticsField}.CreateRequestFailedException{GetAsyncSuffix(async)}({messageVariable}.{nameof(HttpMessage.Response)}){GetConfigureAwait(async)});");
```
AlexanderSher, line 211: So the real difference here is `return {typeof(ResponseWithError)}.FromError<{clientMethod.ReturnType}>({messageVariable}.{nameof(HttpMessage.Response)},` vs `throw`, is that correct?

I'd also extract this block into separate method.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelRestClientWriter.cs
```diff
@@ -191,19 +192,49 @@ private void WriteStatusCodeSwitch(CodeWriter writer, RestClientMethod clientMet
...
+                    }
+                    else
+                    {
+                        writer.Line($"return {messageVariable}.{nameof(HttpMessage.Response)};");
+                    }
                 }
 
                 writer.Line($"default:");
-                if (async)
+                if (isNoThrow)
```
annelo-msft, line 209: I'm not familiar with the LLC CodeGen code yet, but it looks strange to me that we changed this if/else statement to use a different condition.  Do we still handle the async/sync case somewhere?

src/assets/Generator.Shared/ResponseWithError.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Azure.Core
+{
+    internal static class ResponseWithError
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="ErrorResponse{T}"/> with the provided HTTP response and <see cref="RequestFailedException"/> exception.
+        /// </summary>
+        /// <param name="response">The HTTP response.</param>
+        /// <param name="exception"><see cref="RequestFailedException"/> exception.</param>
+        public static ErrorResponse<T> FromError<T>(Response response, RequestFailedException exception)
```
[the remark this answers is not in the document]
annelo-msft: @ellismg had suggested this approach in [this comment](https://github.com/Azure/autorest.csharp/issues/1334#issuecomment-924369290).  I believe the reason was to follow the existing response creation pattern, but if it only creates the error response instance, I agree it may not be needed.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelRestClientWriter.cs
AlexanderSher: Yes, below there are several methods that handle async/sync, but I agree that we need to reformat this a bit.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelDataPlaneClientWriter.cs
```diff
@@ -317,12 +319,13 @@ private void WriteClientMethodBody(CodeWriter writer, LowLevelClientMethod clien
         private void WriteClientMethodDecleration(CodeWriter writer, RestClientMethod clientMethod, LowLevelOperationSchemaInfo operationSchemas, bool async)
         {
             var parameters = clientMethod.Parameters.Concat(new Parameter[] { RequestOptionsParameter });
+            var isConstantResponseBody = clientMethod.Responses.All(response => response.ResponseBody is ConstantResponseBody body);
```
AlexanderSher, line 322: Why `clientMethod.ReturnType == typeof(bool)` doesn't work here?

src/assets/Generator.Shared/ResponseWithError.cs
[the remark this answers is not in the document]
ShivangiReja: I was just following [`ValueResponse<T>`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Internal/ValueResponse.cs) and [`FromValue<T>`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Response.cs#L121-L124) pattern from Core but since we are using it only in generator I removed this class.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelRestClientWriter.cs
ShivangiReja: Updated!

src/AutoRest.CSharp/LowLevel/Generation/LowLevelDataPlaneClientWriter.cs
ShivangiReja: Added this condition.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelDataPlaneClientWriter.cs
```diff
@@ -317,12 +319,13 @@ private void WriteClientMethodBody(CodeWriter writer, LowLevelClientMethod clien
         private void WriteClientMethodDecleration(CodeWriter writer, RestClientMethod clientMethod, LowLevelOperationSchemaInfo operationSchemas, bool async)
         {
             var parameters = clientMethod.Parameters.Concat(new Parameter[] { RequestOptionsParameter });
+            var isConstantResponseBody = clientMethod.Responses.All(response => response.ResponseBody is ConstantResponseBody body) && clientMethod.ReturnType != null && clientMethod.ReturnType.FrameworkType == typeof(bool);
```
AlexanderSher, line 322: Ok, so in what case `clientMethod.ReturnType != null && clientMethod.ReturnType.Equals(typeof(bool))` is `true`, but `clientMethod.Responses.All(response => response.ResponseBody is ConstantResponseBody body)` is `false`?

src/assets/Generator.Shared/ResponseWithError.cs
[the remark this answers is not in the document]
ShivangiReja: Also, we follow this pattern in the generator as well. Check [ResponseWithHeaders](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Generator.Shared/ResponseWithHeaders.cs)

@annelo-msft @AlexanderSher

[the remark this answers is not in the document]
AlexanderSher: yeah, these factory methods look excessive to me.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelDataPlaneClientWriter.cs
ShivangiReja: For an example: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestServerProjects/body-boolean/Generated/BoolClient.cs#L41-L71, in these cases return type will be true.

AlexanderSher: Ok, so the `isConstantResponseBody` is `true` only when `request.HttpMethod == RequestMethod.Head && _context.Configuration.HeadAsBoolean` is true. Condition above is implicit.

I'd prefer to either add a "HeadAsBoolean" read-only property to the `RestClientMethod` or check HttpMethod here directly.

ShivangiReja: Instead of adding the property, we can directly check `Configuration.HeadAsBoolean`

AlexanderSher (approved):

PR #1548, opened 2021-10-01
[LLC] Support Generating "Sub Clients"

### LINKED ISSUE #1527: [LLC] Generate SubClients
The .NET Design Guidelines have recently adopted a ["SubClient"](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-subclients) pattern.  We've been using this pattern by hand via [customizations for some of our LLC]() clients to get an [API surface](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/purview/Azure.Analytics.Purview.Account/api/Azure.Analytics.Purview.Account.netstandard2.0.cs) that follows this design.

We should teach the code generator how to do this for us.  I took a first stab at how this might be done here: https://github.com/ellismg/autorest.csharp/tree/ellismg/llc-sub-clients

ellismg:
Previously, the code generator would generate a publicly constructible
client type for each operation group. This leads to many top level
clients which can lead to user confusion, because it is not clear
which client is the "entry point" to the service. The Sub Clients
feature allows a way to configure the code generator such that it does
not generate publicly constructible clients for certain operation
groups and instead generates a factory method on a different client
which can be used to construct the client (which reuses the same
underlying HttpPipeline).

A new sample based on the Purview Accounts service is added which
shows how this can be used to change the factoring of the client
libraries (and allows us to machine generate what was previously
handled via hand authored customizations to the client library)

Fixes #1527


samples/Azure.Analytics.Purview.Account/Generated/CollectionsClient.cs
```diff
@@ -0,0 +1,771 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace Azure.Analytics.Purview.Account
+{
+    /// <summary> The Collections service client. </summary>
+    public partial class CollectionsClient
```
annelo-msft, line 20: Per the [subclient guidelines](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-subclients) subclients shouldn't have the `Client` suffix.  And this makes me think that `Collections` is going to be too general a name as well.  Have we given any thought to automating renaming in these instances?  IIRC, I saw a proposal go by that suggesting prefixing a resource name with the service name - @ellismg, do you know if we closed on this?

ellismg: @annelo-msft I don't think we made a decision.  We may just want to start with "use the operation group name as the type name (without the `Client` suffix, unless the sub client is also supposed to be publicly constructible?)" and allow someone to use the normal strategy to rename the generated type if we want to add a prefix.

I do think that the prefix should be on the type name only, not also in the name of the factory method that mints a sub client (i.e. `PurviewCollections` as the type name but still `GetCollectionsClient()` instead of `GetPurviewCollectionsClient()`)

annelo-msft: @AlexanderSher, has this been decided in your subclients PR?

PR #1549, opened 2021-10-05
Join back LLC Client and RestClient types

### LINKED ISSUE #1535: [LLC] Set "default response classifier" on message
In order for us to allow `Response.IsError` to reflect default behavior (i.e. which status codes are defined in the OpenAPI spec as success vs. error), we need to pass this information to the pipeline so it can be set on the response.

We achieve this via setting the `ResponseClassifier` on the message when the request is created.

In the LLC generator, we will need to:

1. [Generate properties on the client for each group of successful status codes in the OpenAPI spec](https://github.com/Azure/azure-sdk-for-net/pull/24083/files#diff-2eb5a04baba528d0f0008b55cd9d21a610926491c26d00012ae2ef9c6c6d1435R29)
2. [Generate classifier type definitions](https://github.com/Azure/azure-sdk-for-net/pull/24083/files#diff-2eb5a04baba528d0f0008b55cd9d21a610926491c26d00012ae2ef9c6c6d1435R162)
3. [Set the classifier on the message in CreateXxRequest](https://github.com/Azure/azure-sdk-for-net/pull/24083/files#diff-2eb5a04baba528d0f0008b55cd9d21a610926491c26d00012ae2ef9c6c6d1435R158)

In addition, we will:
4. [Remove the `RequestOptions` allocation from the default path](https://github.com/Azure/azure-sdk-for-net/pull/24083/files#diff-2eb5a04baba528d0f0008b55cd9d21a610926491c26d00012ae2ef9c6c6d1435L61)

AlexanderSher:
Also:
- Incapsulate common code into shared code to reduce amount of generated code
- Fix #1535: [LLC] Set "default response classifier" on message

While working on the #1535, I've noticed that I need to duplicate lots of code for both `*Client` and `*RestClient` types. Merging pages into azure-sdk-for-net repo has shown that separate types multiply the work users have to do:
- Fields must be duplicated in both types
- Constructors and, in come cases, methods, must be overridden in both types
- Custom paging has become even more complicated

Since the split of the `*Client` types into two separate parts have not been discussed in archboard, I've decided to merge them back. 

This has also allowed to move some of the absolutely identical generated code into helper methods, which should simplify customization for users as well.


src/AutoRest.CSharp/AutoRest.CSharp.csproj
```diff
@@ -12,11 +12,12 @@
     <SatelliteResourceLanguages>en-US</SatelliteResourceLanguages>
     <PackAsTool>true</PackAsTool>
     <RollForward>Minor</RollForward>
+    <DefineConstants>$(DefineConstants);EXPERIMENTAL</DefineConstants>
```
AlexanderSher, line 15: Not all projects reference `Azure.Core.Experimental`, hence we need some way to exclude bits that rely on types from `Azure.Core.Experimental`. 
When do we plan to move `RequestOptions` into `Azure.Core`?

src/AutoRest.CSharp/Common/Output/Models/Responses/StatusCodes.cs
```diff
@@ -1,21 +1,7 @@
...
-        public StatusCodes(int? code, int? family)
-        {
-            Code = code;
-            Family = family;
-        }
-
-        public int? Code { get; }
-        public int? Family { get; }
-    }
+    internal record StatusCodes(int? Code, int? Family);
```
AlexanderSher, line 6: by default, records are compared by stored values

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -606,11 +600,27 @@ private static string CreateDescription(OperationGroup operationGroup, string cl
...
                 BuilderHelpers.EscapeXmlDescription(requestParameter.Language.Default.Description);
+
+            return requestParameter.Schema switch
+            {
+                ChoiceSchema choiceSchema when type.IsFrameworkType => AddAllowedValues(description, choiceSchema.Choices),
+                SealedChoiceSchema sealedChoiceSchema when type.IsFrameworkType => AddAllowedValues(description, sealedChoiceSchema.Choices),
+                _ => description
+            };
+
+            static string AddAllowedValues(string description, ICollection<ChoiceValue> choices)
```
AlexanderSher, line 616: If allowed enum values are stored in the parameter description, they will be written everywhere the parameter is used.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelRestClientWriter.cs
```diff
@@ -2,239 +2,658 @@
...
+        {
+            var startMethod = clientMethod.StartMethod;
+            var pagingResponseInfo = clientMethod.PagingResponseInfo;
+            var nextPageMethod = clientMethod.PagingResponseInfo?.NextPageMethod;
+            var finalStateVia = startMethod.Operation.LongRunningFinalStateVia;
+            var scopeName = clientMethod.Diagnostic.ScopeName;
+            var isPageable = nextPageMethod != null;
+
+            var returnType = isPageable
+                ? async ? typeof(Task<Operation<AsyncPageable<BinaryData>>>) : typeof(Operation<Pageable<BinaryData>>)
```
AlexanderSher, line 333: `pageable` test project has LRO of pages, so this case can't be simply avoided.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelRestClientWriter.cs
```diff
@@ -2,239 +2,658 @@
...
                 }
-                writer.RemoveTrailingComma();
-                writer.Line($");");
+            }
+
+            writer.Line();
+        }
 
-                writer.Line($"{typeof(RequestOptions)}.{nameof(RequestOptions.Apply)}(options, {messageVariable});");
+        private static void WriteResponseClassifier(CodeWriter writer, string responseClassifierTypeName, StatusCodes[] statusCodes)
```
AlexanderSher, line 394: this method creates classifier that is used in `RequestWriterHelpers`

src/assets/Generator.Shared/HttpPipelineExtensions.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core.Pipeline;
+
+namespace Azure.Core
+{
+    internal static class HttpPipelineExtensions
+    {
+#if EXPERIMENTAL
+        public static async ValueTask<Response> ProcessMessageAsync(this HttpPipeline pipeline, HttpMessage message, ClientDiagnostics clientDiagnostics, RequestOptions? requestOptions, CancellationToken cancellationToken = default)
```
AlexanderSher, line 16: This method incapsulates some of the code that has been previously generated, as well as new `if (statusOption == ResponseStatusOption.NoThrow || !message.ResponseClassifier.IsErrorResponse(message))` piece

test/AutoRest.TestServerLowLevel.Tests/AccessibilityTests.cs
```diff
@@ -14,7 +14,7 @@ public class AccessibilityTests
         [Test]
         public void AccessibilityHonoredOnOperations()
         {
-            var client = typeof(AccessibilityClient);
+            var client = typeof(AccessibilityRestClient);
```
AlexanderSher, line 17: Since we're planning to enable incremental HLC functionality in the future, `*RestClient` suffix seems better to me.

test/TestServerProjectsLowLevel/paging/Generated/PagingRestClient.cs
```diff
@@ -1378,320 +1448,271 @@ public Response NextFragmentWithGrouping(string apiVersion, string tenant, strin
...
+        ///       }
+        ///     }
+        ///   ],
+        ///   nextLink: string
+        /// }
+        /// </code>
+        /// 
+        /// </remarks>
+#pragma warning disable AZC0002
+        public virtual async Task<Operation<AsyncPageable<BinaryData>>> GetMultiplePagesLROAsync(string clientRequestId = null, int? maxresults = null, int? timeout = null, RequestOptions options = null)
```
AlexanderSher, line 1560: A "combo" test case of LRO and Pageable

test/TestServerProjectsLowLevel/paging/Generated/PagingRestClient.cs
```diff
@@ -1892,302 +1939,220 @@ internal HttpMessage CreateGetMultiplePagesWithOffsetNextPageRequest(string next
                 request.Headers.Add("timeout", timeout.Value);
             }
             request.Headers.Add("Accept", "application/json");
+            message.ResponseClassifier = ResponseClassifier202.Instance;
```
AlexanderSher, line 1942: Custom `ResponseClassifier` is assigned to the message

test/TestServerProjectsLowLevel/body-complex/Generated/PrimitiveRestClient.cs
```diff
@@ -14,849 +14,1802 @@
...
+        ///   message: string
+        /// }
+        /// </code>
+        /// 
+        /// </remarks>
+#pragma warning disable AZC0002
+        public virtual async Task<Response> GetIntAsync(RequestOptions options = null)
+#pragma warning restore AZC0002
+        {
+            using var scope = _clientDiagnostics.CreateScope("PrimitiveRestClient.GetInt");
```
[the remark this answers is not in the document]
pakrym: This also doesn't allow client authors to add more attributes to the scope.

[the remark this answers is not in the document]
AlexanderSher: Hm... there may be no wrapping method. Should we add a property to `RequestOptions`?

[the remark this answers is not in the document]
pakrym: There are two scenarios that might be conflicted here. There might be a path to unify them but I'd like to understand it first before we shake the world.

1. We generate internal rest clients to be wrapped by data-plane T2 libraries. 
2. We generate public LLC clients

For the first case, we don't know what scopes need to be created and what attributes are attached. It's possible that multiple rest calls would share a single diagnostic scope.

For the second case, we know the scope name is correct because the method is public.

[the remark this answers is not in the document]
pakrym: >Should we add a property to RequestOptions?

That doesn't work. `ClientDiagnostics` is designed in a way where it's completely 0-cost when no one is listening. Forcing the caller to always allocate RequestOptions would change that.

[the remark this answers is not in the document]
AlexanderSher: This whole change is LLC-only scope

src/AutoRest.CSharp/Common/Output/Models/Shared/ParameterExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace AutoRest.CSharp.Output.Models.Shared
+{
+    internal static class ParameterExtensions
+    {
+        public static FormattableString GetNamesForMethodCall(this Parameter[] parameters)
```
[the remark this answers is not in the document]
pakrym: Where is this used?

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -190,13 +197,23 @@ public static void WriteRequestCreation(CodeWriter writer, RestClientMethod clie
...
+                {
+                    writer.Line($"{message}.{nameof(HttpMessage.ResponseClassifier)} = {responseClassifierType}.Instance;");
+                }
 
                 writer.Line($"return {message};");
             }
             writer.Line();
         }
 
+        private static ReferenceOrConstant FixFieldReference(IReadOnlyDictionary<string, string>? fields, ReferenceOrConstant value) =>
```
pakrym, line 214: What case does this handle?

AlexanderSher: If parameter is actually has been saved as a field, it swaps the name of parameter with the name of the field. E.g. `endpoint` -> `_endpoint`

pakrym (approved):
overall change LGTM, some small implementation questions

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -0,0 +1,660 @@
...
+            foreach (Parameter clientParameter in client.Parameters)
+            {
+                fieldNames.Add(clientParameter.Name, "_" + clientParameter.Name);
+                writer.Line($"private readonly {clientParameter.Type} _{clientParameter.Name};");
+            }
+
+            writer.Line();
+
+            writer.WriteXmlDocumentationSummary($"The HTTP pipeline for sending and receiving REST requests and responses.");
+            writer.Line($"public virtual {typeof(HttpPipeline)} {PipelineProperty} {{ get => {PipelineField}; }}");
```
pakrym, line 142: Do we need both the field and the property?

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
AlexanderSher: Let me try

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
AlexanderSher: Probably not. Good catch.

src/AutoRest.CSharp/Common/Output/Models/Shared/ParameterExtensions.cs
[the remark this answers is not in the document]
AlexanderSher: In `LowLevelClientWriter`, e.g.: https://github.com/Azure/autorest.csharp/pull/1549/files#diff-b87e7de27287cc647b959201c9791670b032dc207b3ebd68e858aea44e26d071R290
```cs
.Line($"using var {messageVariable:D} = Create{method.Name}Request({method.Parameters.GetNamesForMethodCall()});")
```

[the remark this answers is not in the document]
pakrym: Hm, I thought we have an `writer.WriteParameters` extension method similar to https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs#L118. 

It's a nit but in most existing places we do writing in a linear way where it's a bunch of Append calls.
This introduces the pattern where we pre-calculate strings values first.

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
AlexanderSher: https://github.com/Azure/autorest.csharp/issues/1552

src/assets/Generator.Shared/HttpPipelineExtensions.cs
```diff
@@ -0,0 +1,112 @@
...
+#nullable enable
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core.Pipeline;
+
+namespace Azure.Core
+{
+    internal static class HttpPipelineExtensions
```
annelo-msft, line 13: @AlexanderSher, what is the thinking behind putting `ProcessMessage` in an extension method, rather than on the Pipeline directly?  Will this logic eventually migrate into Pipeline?

PR #1560, opened 2021-10-09
First PR for Mgmt Test generation

changlong-liu:
# Description

## How to use test generation with 'dotnet build /t:GenerateTest'
The 'GenerateTest' command will be supported in azure-sdk-for-net after this PR is merged, released and referred by azure-sdk-for-net.
The Mgmt tests can be generated by below steps:
1. cd to Azure.ResourceManager.xxx\tests
2. create file autorest.tests.md, the content are:
~~~
``` yaml
require:
  - ../src/autorest.md
```
~~~
3. execute 'dotnet build /t:GenerateTest'

## How to use test generation with autorest command line
The test generation can be triggered with an '--testmodeler' option in autorest command line, like:
> autorest --use=D:\projects\codegen\autorest.csharp\artifacts\bin\AutoRest.CSharp\Debug\netcoreapp3.1 D:\projects\codegen\azure-sdk-for-net\sdk\compute\Azure.ResourceManager.Compute\tests\autorest.tests.md --testmodeler

## Features description for this PR:
1. Test generation: 
      + Add testModeler data structures: src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
      + Generate: /src/AutoRest.CSharp/MgmtTest/**/*.cs
2. Added a TestProject to show the geneated result:   test/TestProjects/MgmtKeyvault
3. Modify 'Generate.ps1' to enable test generation in pipeline.
4. Geneate *.csproj for test project: src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpProj.cs

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
```diff
@@ -279,5 +283,192 @@ public override string ToString()
...
+        public System.Collections.Generic.ICollection<TestScenario> TestScenarios;
+
+        [YamlMember(Alias = "_filePath")]
+        public string FilePath;
+
+        [YamlMember(Alias = "scope")]
+        public string? scope;
+
+        [YamlMember(Alias = "scope")]
+        public System.Collections.Generic.ICollection<string>? requiredVariables;
```
weidongxu-microsoft, line 317: Alias seems wrong.

test/AutoRest.TestServer.Tests/Mgmt/TestProjects/MgmtKeyvaultTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        [TestCase("GetExampleInstance", "PrivateEndpointConnectionContainer", "String")]
+        [TestCase("GetExampleInstanceAsync", "PrivateEndpointConnectionContainer", "String")]
+        [TestCase("CreateOrUpdateExampleInstance", "ManagedHsmContainer", "String")]
+        [TestCase("CreateOrUpdateExampleInstanceAsync", "ManagedHsmContainer", "String")]
+        [TestCase("GetExampleInstance", "ManagedHsmContainer", "String")]
+        [TestCase("GetExampleInstanceAsync", "ManagedHsmContainer", "String")]
+        [TestCase("CreateResourceGroupAsync", "String", "ArmClient")]
+        [TestCase("CreateResourceGroup", "String", "ArmClient")]
+        [TestCase("ReplaceWith")]
+        public void validateHelperMethods(string methodName, params string[] argTypes)
```
fengzhou-msft, line 29: First letter of the method name should be capitalized.

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
```diff
@@ -279,5 +283,190 @@ public override string ToString()
...
+
+    internal partial class TestScenario {
+        [YamlMember(Alias = "requiredVariablesDefault")]
+        public System.Collections.Generic.Dictionary<string, string> RequiredVariablesDefault;
+
+        [YamlMember(Alias = "steps")]
+        public System.Collections.Generic.ICollection<TestStep> Steps;
+
+        [YamlMember(Alias = "_testDef")]
+        public TestDefinitionModel? TestDef;
```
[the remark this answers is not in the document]
changlong-liu: The underline is coming from the oav library exports, we can suggest to align the naming convention in next oav release: https://github.com/Azure/oav/blob/d5632320034ddcc32714ac7b856df6fe385c61e4/lib/testScenario/testResourceTypes.ts#L451

As they are all normal data structure in test generator's point of view, I think we don't need to differentiate public/private among them?
And the classes in CodeModel.cs&CodeModelPartials.cs are mostly data structure definition, so I tend to define all properties as public and add logic code in other modules.

src/AutoRest.CSharp/MgmtTest/Generation/ResourceTestWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// </summary>
+    internal class ResourceTestWriter : MgmtBaseTestWriter
+    {
+        protected Resource _resource;
+        protected string TestNamespace => _resource.Type.Namespace + ".Tests.Mock";
+        protected override string TypeNameOfThis => _resource.Type.Name + "MockTests";
+
+        protected string TestBaseName => $"MockTestBase";
+
+        protected ResourceCollectionTestWriter? _collectionTestWriter;
```
ArcturusZhang, line 32: We have a writer inside another writer? This look weird

src/AutoRest.CSharp/MgmtTest/Generation/ResourceTestWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+                    foreach (var childResource in This.ChildResources)
+                    {
+                        _writer.Line();
+                        if (childResource.IsSingleton) {
+                            WriteMethodTestIfExist(childResource);
+                        }
+                    }
+                }
+            }
+        }
```
ArcturusZhang, line 81: This method makes more sense if it belongs to the CollectionTestWriter

src/AutoRest.CSharp/MgmtTest/Generation/ResourceTestWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        protected void WriteMethodTestIfExist(Resource resource) {
```
ArcturusZhang, line 83: We need to format the code
```suggestion
        protected void WriteMethodTestIfExist(Resource resource)
        {
```

src/AutoRest.CSharp/MgmtTest/Generation/ResourceTestWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+
+            return _resource.Type;
+        }
+
+        protected override bool IsResourceDataType(CSharpType? type, MgmtRestOperation operation)
+        {
+            return _resource.Type.Equals(type);
+        }
+
+        protected void WriteMethodTest(Resource resource, MgmtClientOperation clientOperation, bool async)
```
ArcturusZhang, line 171: Could we rename this method to `WriteTestMethod`?

src/AutoRest.CSharp/MgmtTest/Generation/ResourceTestWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            Debug.Assert(clientOperation != null);
```
ArcturusZhang, line 173: This is redundant since the `clientOperation` can never be `null` because its type is non-nullable

src/AutoRest.CSharp/MgmtTest/Generation/ResourceTestWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+            var methodName = clientOperation.Name;
+
+            BuildParameters(clientOperation, out var operationMappings, out var parameterMappings, out var methodParameters);
+            foreach ((var branch, var operation) in getSortedOperationMappings(clientOperation))
```
ArcturusZhang, line 177: Instead of using this `getSortedOperationMappings`, I recommend we override the `BuildParameters` method here, to sort the out variables correspondingly

ArcturusZhang: And for those variables that are not actually used, we can use the discard identifier to discard it explicitly.
```suggestion
            BuildParameters(clientOperation, out var operationMappings, out _, out var methodParameters);
            foreach ((var branch, var operation) in getSortedOperationMappings(clientOperation))
```

ArcturusZhang: Also please feel free to change the interface in `MgmtBaseClientWriter` if you feel some interface is not reasonable to use.

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,713 @@
...
+    internal partial class MgmtBaseTestWriter: MgmtClientBaseWriter
+    {
+        public static CodeWriter _tagsWriter = new CodeWriter();
+        public static HashSet<string>  variableNames = new HashSet<string>();
+
+        public MgmtBaseTestWriter(CodeWriter writer, MgmtTypeProvider provider, BuildContext<MgmtOutputLibrary> context) : base(writer, provider, context)
+        {
+        }
+
+        public void WriteTestDecorator()
```
ArcturusZhang, line 36: It is called `Attribute` in DotNet
```suggestion
        public void WriteTestAttributes()
```

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,713 @@
...
+        {
+            _writer.Line($"[RecordedTest]");
+
+            var testsConfiguration = Context.Configuration.MgmtConfiguration.TestModeler;
+            System.Text.Json.JsonElement ignoreReason = default;
+            if (testsConfiguration is not null && testsConfiguration.IsObject() && testsConfiguration?.TryGetProperty("ignore-reason", out ignoreReason) == true)
```
ArcturusZhang, line 42: These will be a lot easier if we refactor the `TestModeler` configuration

src/AutoRest.CSharp/MgmtTest/Generation/TestHelperWriter.cs
```diff
@@ -0,0 +1,94 @@
...
+                    WriteCreateResourceGroupMethod();
+                    WriteReplaceWith();
+                }
+            }
+        }
+
+        public void WriteCreateResourceGroupMethod()
+        {
+            _writer.UseNamespace("Azure.ResourceManager");
+            using (_writer.Scope($"public static async Task<ResourceGroup> CreateResourceGroupAsync(string resourceGroupName, ArmClient client)"))
```
ArcturusZhang, line 65: We can use the Type in the string interpolation to automatically use the corresponding namespaces, we do not need to use them explicitly
```suggestion
            using (_writer.Scope($"public static async Task<{typeof(ResourceGroup)}> CreateResourceGroupAsync(string resourceGroupName, {typeof(ArmClient)} client)"))
```

src/AutoRest.CSharp/MgmtTest/Generation/TestHelperWriter.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                _writer.Line($"var defaultSubscription = await client.GetDefaultSubscriptionAsync();");
+                using (_writer.Scope($"var rgop = await defaultSubscription.GetResourceGroups().CreateOrUpdateAsync", start: "(", end: ")", newLine: false))
```
ArcturusZhang, line 68:
```suggestion
                using (_writer.Scope($"var rgOp = await defaultSubscription.GetResourceGroups().CreateOrUpdateAsync", start: "(", end: ")", newLine: false))
```

src/AutoRest.CSharp/MgmtTest/Generation/TestHelperWriter.cs
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    _writer.Line($"resourceGroupName,");
+                    _writer.Line($"new ResourceGroupData(defaultSubscription.ToString()) {{ Tags = {{ {{ \"test\", \"env\" }} }} }}");
+                }
+                _writer.Line($";");
+                _writer.Append($"return rgop.Value;");
```
ArcturusZhang, line 74:
```suggestion
                _writer.Append($"return rgOp.Value;");
```

test/AutoRest.TestServer.Tests/Mgmt/OutputLibrary/OutputLibraryTestBase.cs
```diff
@@ -22,16 +22,26 @@ namespace AutoRest.TestServer.Tests.Mgmt.OutputLibrary
...
 
-        public OutputLibraryTestBase(string projectName)
+        public OutputLibraryTestBase(string projectName, string subFolder = null)
         {
             _projectName = projectName;
+            _subFolder = subFolder;
         }
 
-        internal static async Task<(CodeModel Model, BuildContext<MgmtOutputLibrary> Context)> Generate(string testProject)
+        internal static async Task<(CodeModel Model, BuildContext<MgmtOutputLibrary> Context)> Generate(string testProject, string subFolder=null)
```
ArcturusZhang, line 33:
```suggestion
        internal static async Task<(CodeModel Model, BuildContext<MgmtOutputLibrary> Context)> Generate(string testProject, string subFolder = null)
```

ArcturusZhang (changes requested):
Several comments about coding style, etc

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,713 @@
...
+            {
+                return dict;
+            }
+            var ret = new Dictionary<string, object?>();
+            foreach (KeyValuePair<object, object> entry in (IEnumerable< KeyValuePair<object, object>>)dict)
+            {
+                ret.Add(entry.Key.ToString()!, ConvertToStringDictionary(entry.Value));
+            }
+            return ret;
+        }
```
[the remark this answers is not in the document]
changlong-liu: It's used to serialize example value of type 'any' in example: https://github.com/Azure/autorest.csharp/blob/5669bdaacbd2c9f603b41226c3670c6e20379386/src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs#L488

It return 'object?' since the 'JsonSerializer.Serialize()' can handle this type of parameter

src/AutoRest.CSharp/MgmtTest/AutoRest/MgmtTestTarget.cs
```diff
@@ -0,0 +1,62 @@
...
+            bool hasCollectionTest = false;
+            foreach (var resourceCollection in context.Library.ResourceCollections)
+            {
+                var codeWriter = new CodeWriter();
+                var collectionTestWriter = new ResourceCollectionTestWriter(codeWriter, resourceCollection, context);
+                if (!collectionTestWriter.CanCreateParentResourceFromExample(context, resourceCollection))
+                {
+                    continue;
+                }
+                if (!MgmtBaseTestWriter.HasCreateExample(context, resourceCollection)
```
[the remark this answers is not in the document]
changlong-liu: this function is deleted after refactor

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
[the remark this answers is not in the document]
changlong-liu: Redesigned into this function: https://github.com/Azure/autorest.csharp/blob/3deaeff2c8a814fbcf9159ef117288cecdf67343/src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs#L343

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -47,7 +47,14 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(Task<CodeModel> codeModel
             }
             else if (configuration.AzureArm)
             {
-                MgmtTarget.Execute(project, codeModel, sourceInputModel, configuration);
+                if (configuration.MgmtConfiguration.TestModeler is not null)
```
ArcturusZhang, line 50: Here is the problem. If we enter the debug mode, this will always be non-null, therefore during the debug, only MgmtTestTarget will be executed.

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+        public static CodeWriter _tagsWriter = new CodeWriter();
+        public static HashSet<string>  variableNames = new HashSet<string>();
+
+        public MgmtBaseTestWriter(CodeWriter writer, MgmtTypeProvider provider, BuildContext<MgmtOutputLibrary> context) : base(writer, provider, context)
+        {
+        }
+
+        public void WriteTestDecorator()
+        {
+            _writer.Line($"[RecordedTest]");
```
ArcturusZhang, line 40: if possible, please use `_writer.Line($"[{typeof(RecordedTest)}]");` instead

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+                    Regex regex = new Regex("^{[A-Z0-9]{8}-([A-Z0-9]{4}-){3}[A-Z0-9]{12}}$");
+                    Match match = regex.Match(elements[i]);
+                    if (!match.Success)
+                    {
+                        elements[i] = "00000000-0000-0000-0000-000000000000";
+                    }
+                }
+            }
+            return String.Join("/", elements);
+        }
```
ArcturusZhang, line 96: Every time we need to output a resource identifier, we need to use the `CreateResourceIdentifier` function to do that.

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+        public static string GetTaskOrVoid(bool isAsync)
+        {
+            return isAsync ? "Task" : "void";
+        }
+
+        public static ExampleGroup? FindExampleGroup(BuildContext<MgmtOutputLibrary> context, MgmtRestOperation restOperation)
+        {
+            if (restOperation is null)
+                return null;
+            foreach (var exampleGroup in context.CodeModel.TestModel?.MockTest.ExampleGroups ?? Enumerable.Empty<ExampleGroup>())
```
ArcturusZhang, line 107: I believe if the code can run here, we have guaranteed the TestModel is not null, therefore we do not need the null short cut here

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+            {
+                return false;
+            }
+
+            foreach ((var branch, var operation) in GetSortedOperationMappings(clientOperation))
+            {
+                var exampleGroup = MgmtBaseTestWriter.FindExampleGroup(context, operation);
+                if (exampleGroup is not null && exampleGroup.Examples.Count > 0)
+                    return true;
+                break;
```
ArcturusZhang, line 150: Why we break here? Do we need to iterate all the collection?

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+            {
+                var obj = (JsonSerialization)sot.Serializations.Single(x => typeof(JsonSerialization).IsInstanceOfType(x));
+                foreach (var exampleProperty in ev.Properties ?? new DictionaryOfExamplValue())
+                {
+                    var property = FindPropertyThroughSerialization(exampleProperty.Value.FlattenedNames is not null ? exampleProperty.Value.FlattenedNames.ToList()! : new List<string> { exampleProperty.Key }, obj);
+                    if (property == targetProperty)
+                    {
+                        return exampleProperty.Value;
+                    }
+                }
```
ArcturusZhang, line 188: These two methods look quite low-level. In a writer, I believe we should not do anything low-level. The writer should only write things without calculating anything.
Do we have a better place for these two methods?

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+                            return exampleProperty.Value;
+                        }
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        private ObjectTypeConstructor FindSuitableConstructor(ObjectType sot, ExampleValue ev)
```
ArcturusZhang, line 207: And so does this.

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+                                writer.Append($", ");
+                            }
+                            consumedProperties.Add(targetProperty);
+                        }
+                    }
+                }
+            }
+        }
+
+        internal readonly struct JsonRawValue
```
ArcturusZhang, line 344: Please consider to put these Json related things into a specific class

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+                case EnumType enumType:
+                    WriteEnumTypeExampleValue(writer, enumType, exampleValue);
+                    break;
+                default:
+                    WriteFrameworkTypeExampleValue(writer, cst, exampleValue, variableName);
+                    break;
+            }
+        }
+
+        public static string useVariableName(string variableName)
```
ArcturusZhang, line 558: CSharp method name should always have its first letter in upper case.

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+                if (!variableNames.Contains(newName))
+                {
+                    variableNames.Add(newName);
+                    return newName;
+                }
+            }
+            throw new Exception($"Can't find a valid variable name start with {variableName}");
+        }
+
+        public static void clearVariableNames() {
```
ArcturusZhang, line 582: First letter should be in upper case.

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+            variableNames = new HashSet<string>();
```
ArcturusZhang, line 583:
```suggestion
variableNames.Clear();
```

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+using System.Text.Json.Serialization;
+using Azure.Core.Serialization;
+using System.Runtime.InteropServices.WindowsRuntime;
+using System.Text.RegularExpressions;
+
+namespace AutoRest.CSharp.MgmtTest.Generation
+{
+    internal partial class MgmtBaseTestWriter: MgmtClientBaseWriter
+    {
+        public static CodeWriter _tagsWriter = new CodeWriter();
```
ArcturusZhang, line 31: Why we make this variable static? This is shared between all instance of the `TestWriter`s?

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+
+        protected override void WriteResourceCollectionEntry(Resource resource)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected override ResourceType GetBranchResourceType(RequestPath branch)
+        {
+            throw new NotImplementedException();
+        }
```
ArcturusZhang, line 624: If we do not implement these methods in this class, please remove these overrides and make this class `abstract`

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+
+        public List<string> WriteOperationParameters(IEnumerable<Parameter> methodParameters, IEnumerable<Parameter> testMethodParameters, ExampleModel exampleModel)
```
ArcturusZhang, line 626: What does the testMethodParameters mean?

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -0,0 +1,773 @@
...
+                return GetResponseType(returnType, async);
+            }
+            else
+            {
+                var returnType = WrapResourceDataType(clientOperation.ReturnType, clientOperation.First());
+                return GetResponseType(returnType, async);
+            }
+        }
+
+        public string? ParseRequestPath(MgmtTypeProvider tp, string requestPath, ExampleModel exampleModel)
```
ArcturusZhang, line 686: We provide similar APIs in the `RequestPath` class, therefore we should no longer need to parse the request path by its raw string.

ArcturusZhang (commented):
Please make sure to format each source code file to keep the format consistent.

PR #1563, opened 2021-10-12
Support SubClients

AlexanderSher:
This work is based on the @ellismg prototype: https://github.com/Azure/autorest.csharp/pull/1548 It is in progress.

Two main differences are configuration and implementation of the writer:
- We usually don't use `autorest.md` to configure changes in the .net-specific part of autorest, using features of C# instead. It allows to control various issues inside C# project directly and doesn't require additional knowledge of yaml. So instead of adding multi-level `csharp-client-configuration` section, I propose [a couple of parameters](https://github.com/Azure/autorest.csharp/compare/feature/v3...AlexanderSher:SubClient?expand=1#diff-77c6d1c359873bf4caded5d712fd6b495e2e77b435fe96513802f155f052ab13R13) to the existing `CodeGenClientAttribute`:
```cs
[CodeGenClient(ParentClient = typeof(AccountsClient), ForcePublicConstructors = true)]
public partial class ResourceSetRulesClient {}
```
- One of the general problems in autorest.csharp codebase is that it leaks logic into writers code, making it complicated and producing lots of duplicates. For example, [this switch](https://github.com/Azure/autorest.csharp/pull/1548/files#diff-b87e7de27287cc647b959201c9791670b032dc207b3ebd68e858aea44e26d071R476) in slightly different variations is repeated 5 times. And we constantly have bugs that are rooted in duplicated code that wasn't fixed in all places. So I'm planning to move maximum data into the type providers in the future. In this PR, however, I want to address at least low hanging fruits and avoid creating more duplicates.


AlexGhiondea:
@AlexanderSher is this a replacement for #1548? Could the conversation be continued on that PR instead of having a brand new PR out? Trying to follow along.

AlexanderSher:
@KrzysztofCwalina, reviving @annelo-msft's question about subclient's name, do we have any decision at least for RC1? Are we removing `Client` suffix? If I remember correctly, we have discussed several options:
- Remove `Client` suffix: `CollectionsClient` -> `Collections`
- Remove `Client` suffix and add library name as prefix: `CollectionsClient` -> `PurviewAccountCollections`
- Remove `Client` suffix and addlast part of namespace as prefix: `CollectionsClient` -> `AccountCollections`
- Replace `Client` suffix with `Operations` suffix: `CollectionsClient` -> `CollectionsOperations`

ellismg:
> So instead of adding multi-level csharp-client-configuration section, I propose a couple of parameters to the existing CodeGenClientAttribute:

I think this is a very cool idea and I think moves us in the right longer term direction that we've been discussing where more and more of this stuff is driven via C# source customizations.

samples/Azure.Analytics.Purview.Account/Customizations.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License.
+
+// <auto-generated/>
+
+using Azure.Core;
+
+namespace Azure.Analytics.Purview.Account
+{
+    [CodeGenClient(ParentClient = typeof(AccountsClient))]
+    public partial class PurviewAccountCollections { }
```
AlexanderSher, line 11: This has become a bit unintuitive: user has to be aware of the different naming rules for subclients to apply customization. In this example, the top level name would be `CollectionsClient`, but customization type should be `PurviewAccountCollections`. As an alternative, we can always specify old client name, e.g.:
```cs
    [CodeGenClient("CollectionsClient", ParentClient = typeof(AccountsClient))]
    public partial class PurviewAccountCollections { }
```

That would also solve the naming problem cause user will always specify the name of subclient explicitly.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -393,6 +367,77 @@ private static void WriteLongRunningOperationMethod(CodeWriter writer, LowLevelL
...
+                    }
+                }
+
+                var subClientName = subClient.Type.Name;
+                var libraryName = context.DefaultLibraryName;
+                var methodName = subClientName.StartsWith(libraryName)
+                    ? subClientName[libraryName.Length..]
+                    : subClientName;
+
+                var methodSignature = new MethodSignature($"Get{methodName}{ClientBuilder.GetClientSuffix(context)}", $"Initializes a new instance of {subClient.Type.Name}", "public virtual", subClient.Type, null, methodParameters.ToArray());
```
AlexanderSher, line 398: I still don't like the way we guess about the name of the method.
Maybe it will work for now, and later we can implement https://github.com/Azure/autorest.csharp/issues/1600 and users will be able to change the name of the method.

test/TestServerProjectsLowLevel/body-array/Generated/ArrayClient.cs
```diff
@@ -18,13 +18,12 @@ public partial class ArrayClient
     {
         private const string AuthorizationHeader = "Fake-Subscription-Key";
         private readonly AzureKeyCredential _keyCredential;
-
         private readonly HttpPipeline _pipeline;
         private readonly ClientDiagnostics _clientDiagnostics;
         private readonly Uri _endpoint;
 
         /// <summary> The HTTP pipeline for sending and receiving REST requests and responses. </summary>
-        public virtual HttpPipeline Pipeline { get => _pipeline; }
```
annelo-msft, line 27: It's a bit challenging to review PRs that include new features as well as small code-style changes, as the code-style changes touch many files, so it's hard to know which files are relevant to the feature work, and not cosmetic updates.  It would be helpful if future feature PRs could be separated out from code-style changes.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountsClient.cs
```diff
@@ -0,0 +1,1282 @@
...
+                throw new ArgumentNullException(nameof(collectionName));
+            }
+
+            return new PurviewAccountCollections(_clientDiagnostics, _pipeline, _tokenCredential, _endpoint, collectionName, _apiVersion);
+        }
+
+        /// <summary> Initializes a new instance of PurviewAccountResourceSetRules. </summary>
+        public virtual PurviewAccountResourceSetRules GetResourceSetRulesClient()
+        {
+            return _cachedPurviewAccountResourceSetRules ??= new PurviewAccountResourceSetRules(_clientDiagnostics, _pipeline, _tokenCredential, _endpoint, _apiVersion);
```
annelo-msft, line 1135: Why do we cache the ResourceSetRulesClient and not the CollectionsClient?

samples/Azure.Analytics.Purview.Account/Customizations.cs
AlexanderSher: Swagger doesn't give us any information except for the case when we have "factory" top level client that has only method to create subclients based on their operation group names

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/VirtualMachineScaleSetVMExtensionUpdateOperation.cs
```diff
@@ -18,9 +18,9 @@
 namespace Azure.ResourceManager.Sample.Models
 {
     /// <summary> The operation to update the VMSS VM extension. </summary>
-    public partial class VirtualMachineScaleSetVMExtensionUpdateOperation : Operation<VirtualMachineScaleSetVirtualMachineExtension>, IOperationSource<VirtualMachineScaleSetVirtualMachineExtension>
+    public partial class VirtualMachineScaleSetVMExtensionUpdateOperation : Operation<VirtualMachineExtensionVirtualMachineScaleSet>, IOperationSource<VirtualMachineExtensionVirtualMachineScaleSet>
```
m-nash, line 21: why did this renaming happen?

test/TestProjects/MgmtMultipleParentResource/Generated/AnotherParent.cs
```diff
@@ -416,11 +416,11 @@ public virtual AnotherParentUpdateOperation Update(AnotherParentUpdate anotherBo
...
 
-        /// <summary> Gets a list of AnotherParentChildBodies in the AnotherParent. </summary>
-        /// <returns> An object representing collection of AnotherParentChildBodies and their operations over a AnotherParent. </returns>
-        public AnotherParentChildBodyCollection GetAnotherParentChildBodies()
+        /// <summary> Gets a list of ChildBodyAnotherParents in the AnotherParent. </summary>
+        /// <returns> An object representing collection of ChildBodyAnotherParents and their operations over a AnotherParent. </returns>
+        public ChildBodyAnotherParentCollection GetChildBodyAnotherParents()
         {
-            return new AnotherParentChildBodyCollection(this);
+            return new ChildBodyAnotherParentCollection(this);
```
m-nash, line 423: why did renames happen in this project as well?

m-nash (changes requested):

test/TestProjects/MgmtMultipleParentResource/Generated/AnotherParent.cs
AlexanderSher: Fixed

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/VirtualMachineScaleSetVMExtensionUpdateOperation.cs
AlexanderSher: Fixed

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -22,10 +24,11 @@ public static class Options
             public const string HeadAsBoolean = "head-as-boolean";
             public const string SkipCSProjPackageReference = "skip-csproj-packagereference";
             public const string LowLevelClient = "low-level-client";
+            public const string SingleTopLevelClient = "single-top-level-client";
```
AlexanderSher, line 27: Added `single-top-level-client` parameter

docs/features/subclients.md
```diff
@@ -0,0 +1,197 @@
...
+  default:
+    name: MyService
+```
+
+For each operation group, **autorest.csharp** generates individual client type called using `$"{operationGroups.language.default.name}Client"` format when group has a name or `$"{language.default.name}Client"` format when group name is empty.
+
+```cs
+public partial class NamedGroupClient
+{
+    public virtual async Task<Response> Op1Async(RequestOptions options = null);
```
ShivangiReja, line 52: `RequestOptions` got renamed to `RequestContext`. Can we rename it everywhere?

test/TestProjects/SubClients-LowLevel/Generated/RootClient.cs
```diff
@@ -0,0 +1,135 @@
...
+            {
+                scope.Failed(e);
+                throw;
+            }
+        }
+
+        private volatile Parameter _cachedParameter0;
+
+        /// <summary> Initializes a new instance of Parameter. </summary>
+        public virtual Parameter GetParameterClient()
```
ShivangiReja, line 101: Can we add some unit tests to verify that the generated code is correct?

test/TestProjects/SingleTopLevelClientWithOperations-LowLevel/Generated/TopLevelClientWithOperationClient.cs
```diff
@@ -0,0 +1,134 @@
...
+                scope.Failed(e);
+                throw;
+            }
+        }
+
+        private volatile Client1Client _cachedClient1Client;
+        private volatile Client2Client _cachedClient2Client;
+
+        /// <summary> Initializes a new instance of Client1Client. </summary>
+        public virtual Client1Client GetClient1ClientClient()
```
ShivangiReja, line 95: Can we add some unit tests to verify generated methods for single top-level client?

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
```diff
@@ -19,25 +18,15 @@ namespace AutoRest.CSharp.Mgmt.Output
     /// </summary>
     internal class MgmtLongRunningOperation : LongRunningOperation
     {
```
annelo-msft, line 20: Why is it that there are changes to management plane as part of this PR?

src/assets/Generator.Shared/HttpPipelineExtensions.cs
```diff
@@ -57,43 +56,31 @@ public static Response ProcessMessage(this HttpPipeline pipeline, HttpMessage me
...
                 default:
-                    var exception = await clientDiagnostics.CreateRequestFailedExceptionAsync(message.Response).ConfigureAwait(false);
-                    if (statusOption == ErrorOptions.NoThrow)
-                    {
-                        return new ErrorResponse<bool>(message.Response, exception);
-                    }
-
-                    throw exception;
+                    var exception = await clientDiagnostics.CreateRequestFailedExceptionAsync(response).ConfigureAwait(false);
+                    return new ErrorResponse<bool>(response, exception);
```
annelo-msft, line 68: Is it correct to remove the check for status option here?  Would ProcessMessageAsync have handled any errors that come up?

src/assets/Generator.Shared/CodeGenClientAttribute.cs
```diff
@@ -1,13 +1,18 @@
...
+
 using System;
 
 namespace Azure.Core
 {
     [AttributeUsage(AttributeTargets.Class)]
     internal class CodeGenClientAttribute : CodeGenTypeAttribute
     {
+        public Type? ParentClient { get; set; }
+        public bool ForcePublicConstructors { get; set; }
```
annelo-msft, line 14: Remind me what ForcePublicConstructors means?

test/TestProjects/SingleTopLevelClientWithOperations-LowLevel/Generated/TopLevelClientWithOperationClient.cs
```diff
@@ -0,0 +1,134 @@
...
+                return _pipeline.ProcessMessage(message, _clientDiagnostics, context);
+            }
+            catch (Exception e)
+            {
+                scope.Failed(e);
+                throw;
+            }
+        }
+
+        private volatile Client1 _cachedClient1;
```
annelo-msft, line 91: Why volatile?

test/TestProjects/SubClients-LowLevel/SubClients-LowLevel.json
```diff
@@ -0,0 +1,63 @@
...
+    "title": "LlcSubClients"
+  },
+  "host": "localhost:3000",
+  "schemes": [
+    "http"
+  ],
+  "produces": [ "application/json" ],
+  "consumes": [ "application/json" ],
+  "paths": {
+    "/root/{cachedParameter}": {
```
annelo-msft, line 14: What case is this test testing?

docs/features/subclients.md
AlexanderSher: Fixed

test/TestProjects/SubClients-LowLevel/Generated/RootClient.cs
AlexanderSher: Done

test/TestProjects/SingleTopLevelClientWithOperations-LowLevel/Generated/TopLevelClientWithOperationClient.cs
AlexanderSher: Done

src/assets/Generator.Shared/HttpPipelineExtensions.cs
AlexanderSher: Yes, `ProcessMessageAsync` would throw already if `requestContext.ErrorOptions != ErrorOptions.NoThrow`

src/assets/Generator.Shared/CodeGenClientAttribute.cs
AlexanderSher: We generate public constructors for the client no matter if it is a subclient or not.

test/TestProjects/SingleTopLevelClientWithOperations-LowLevel/Generated/TopLevelClientWithOperationClient.cs
AlexanderSher: AFAIK, clients are multithreaded.

test/TestProjects/SubClients-LowLevel/SubClients-LowLevel.json
AlexanderSher: It validates that we avoid naming collision of the fields. Parent client has client parameter `cachedParameter` and `GetParameterClient` method that caches the result. Both would try to create `_cachedParameter` field.

test/TestProjects/SingleTopLevelClientWithOperations-LowLevel/Generated/TopLevelClientWithOperationClient.cs
AlexanderSher: Thinking about it again, we probably should use `syncObj` pattern. let me fix it.

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
AlexanderSher: SubClients have broken the general naming pattern for clients. Previously, all clients had independent names. Now, LLC client may have different name depending on either it is a top level client or subclient.

test/TestProjects/SingleTopLevelClientWithOperations-LowLevel/Generated/TopLevelClientWithOperationClient.cs
annelo-msft: Our clients are threadsafe.  What about using https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazyinitializer.ensureinitialized?view=net-5.0?

AlexanderSher: This is implementation: https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/Threading/LazyInitializer.cs

It is basically a wrapper around `Volatile.Read` and `Interlocked.CompareExchange`. We can generate those as well, it will save us some memory on syncObj. Calling methods from `LazyInitializer` directly would be an overhead - they require factory delegates that may produce closures.

AlexanderSher: There is even an open issue for this: https://github.com/dotnet/runtime/issues/26255

AlexanderSher: Fixed

ShivangiReja (approved):
Generator code looks good to me! @annelo-msft could you please take a look at the generated code?

src/assets/Generator.Shared/CodeGenClientAttribute.cs
annelo-msft: Could we add a comment to document this?

test/TestProjects/SubClients-LowLevel/SubClients-LowLevel.json
annelo-msft: Is there a way we could either rename files or add comments to make this clear?  I would not have guessed this from the test files.

AlexanderSher: Comments to the json input?

annelo-msft: Or rename of SubClients-LowLevel to indicate a more specific case being tested?

AlexanderSher: No, the project itself tests subclients in general. This particular parameter though tests resolution of field names collision.

src/assets/Generator.Shared/CodeGenClientAttribute.cs
AlexanderSher: Removed `ForcePublicConstructors` per https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-subclient-no-construct

annelo-msft (approved):

PR #1565, opened 2021-10-13
Request path pattern for mgmt plane SDK

### LINKED ISSUE #1581: Add TenantExtensions and ManagementGroupsExtensions

ArcturusZhang:
# Description

Fixes #1581
Fixes https://github.com/Azure/autorest.csharp/issues/1609
Fixes https://github.com/Azure/autorest.csharp/issues/1608
Fixes https://github.com/Azure/autorest.csharp/issues/1593

This PR includes the following:
1. Introduce a new class `MgmtTypeProvider`, which inherits from `TypeProvider`. This class includes the information that will be generated as a class with operations, like `Resource`, `ResourceContainer` and the extension classes.
2. Now we write methods using a group of operations under the same request path, instead of `OperationGroup`s.
3. Now we determine the parent-child relationship using request path, instead of relying on `OperationGroup`s or configurations. But this is configurable.
4. The new request path pattern could identify the singleton resources with an enum with only one value. But due to some updates on autorest.modelerfour, we need to add a flag to enforce the legacy behaviour (see the storage sample).
5. Introduce the contextual request paths to solve the issue of incorrect parameters during RestOperation invocation.
6. Support non-stringlike contextual parameters (like `int`)
7. Rewrite all the writers to make sure the generated code will have the same pattern.

Known issues and future works:
1. The item type of pageable methods in extension classes are not wrapped by its corresponding Resource type if it is a resource data. Considering that we might have multiple resources with the same resource data, I do not think this is good approach to do in our SDK which might lead to potential breaking changes.
But `GetAll` methods in the container class will still wrap with its corresponding resource.
2. Configuration to merge operations and resources are not implemented in this PR yet.
3. Configuration flag to suppress some error into warnings and throw errors by default if the generator cannot decide something
4. Some `ResourceType`s are magic string instead of using the constant from the corresponding classes.
5. Despite we can encapsulate multiple operations into the same resource or resource container class, we do not really validate them to have the same parameters and return value. This might introduce compile errors in the generated code.

# Checklist

# Ready to Land?


test/TestProjects/MgmtScopeResource/Generated/PolicyAssignmentContainer.cs
```diff
@@ -270,207 +269,343 @@ public async virtual Task<Response<bool>> CheckIfExistsAsync(string policyAssign
...
+            else if (Id.ResourceType == "Microsoft.ManagementGroups/managementGroups")
+            {
+                Page<PolicyAssignment> FirstPageFunc(int? pageSizeHint)
+                {
+                    using var scope = _clientDiagnostics.CreateScope("PolicyAssignmentContainer.GetAll");
+                    scope.Start();
+                    try
                     {
-                        var response = _restClient.GetForManagementGroup(Id.Name, filter, top, cancellationToken: cancellationToken);
+                        var response = _policyAssignmentsRestClient.GetForManagementGroup(Id.Parent.Name, filter, top, cancellationToken: cancellationToken);
```
fengzhou-msft: The first argument becomes `Id.Parent.Name` which seems wrong.

fengzhou-msft: It looks like the issue is caused by subscription parameter is now in the constructor of PolicyAssignmentsRestOperations as you commented UpdateSubscriptionIdForTenantIdResource in MgmtOutputLibrary.

test/TestProjects/MgmtScopeResource/Generated/PolicyAssignmentContainer.cs
```diff
@@ -270,207 +269,343 @@ public async virtual Task<Response<bool>> CheckIfExistsAsync(string policyAssign
...
+            }
+        }
+
+        /// <summary> Filters the list of <see cref="PolicyAssignment" /> for this resource group represented as generic resources. </summary>
+        /// <param name="nameFilter"> The filter used in this operation. </param>
+        /// <param name="expand"> Comma-separated list of additional properties to be included in the response. Valid values include `createdTime`, `changedTime` and `provisioningState`. </param>
+        /// <param name="top"> The number of results to return. </param>
+        /// <param name="cancellationToken"> A token to allow the caller to cancel the call to the service. The default value is <see cref="CancellationToken.None" />. </param>
+        /// <returns> A collection of resource that may take multiple service requests to iterate over. </returns>
+        public virtual Pageable<GenericResource> GetAllAsGenericResources(string nameFilter, string expand = null, int? top = null, CancellationToken cancellationToken = default)
```
fengzhou-msft: This method only applies for resource group and subscription.

test/TestProjects/MgmtScopeResource/Generated/PolicyAssignmentContainer.cs
```diff
@@ -36,17 +35,11 @@ protected PolicyAssignmentContainer()
...
-
-        /// <summary> Verify that the input resource Id is a valid container for this type. </summary>
-        /// <param name="identifier"> The input resource Id to check. </param>
-        protected override void ValidateResourceType(ResourceIdentifier identifier)
-        {
+            _policyAssignmentsRestClient = new PolicyAssignmentsRestOperations(_clientDiagnostics, Pipeline, ClientOptions, Id.SubscriptionId, BaseUri);
         }
 
         /// <summary> Gets the valid resource type for this object. </summary>
-        protected override ResourceType ValidResourceType => ResourceIdentifier.RootResourceIdentifier.ResourceType;
```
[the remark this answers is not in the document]
ArcturusZhang: Great catch, I will take a look at this to make sure this could work

test/TestProjects/MgmtScopeResource/Generated/PolicyAssignmentContainer.cs
```diff
@@ -118,21 +111,22 @@ public async virtual Task<PolicyAssignmentCreateOperation> CreateOrUpdateAsync(s
             }
         }
 
-        /// <summary> Gets details for this resource from the service. </summary>
+        /// <summary> This operation retrieves a single policy assignment, given its name and the scope it was created at. </summary>
```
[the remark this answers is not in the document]
ArcturusZhang: Well... this comes from the description of this operation in swagger.

test/TestProjects/MgmtScopeResource/Generated/PolicyAssignmentContainer.cs
```diff
@@ -270,207 +269,343 @@ public async virtual Task<Response<bool>> CheckIfExistsAsync(string policyAssign
...
         /// <param name="top"> Maximum number of records to return. When the $top filter is not provided, it will return 500 records. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         /// <returns> A collection of <see cref="PolicyAssignment" /> that may take multiple service requests to iterate over. </returns>
         public virtual Pageable<PolicyAssignment> GetAll(string filter = null, int? top = null, CancellationToken cancellationToken = default)
         {
-            Page<PolicyAssignment> FirstPageFunc(int? pageSizeHint)
+            if (Id.ResourceType == "Microsoft.Resources/resourceGroups")
```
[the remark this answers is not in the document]
ArcturusZhang: Yeah - I listed this as known issue #4. I will take a look at this.

test/TestProjects/MgmtScopeResource/Generated/PolicyAssignmentContainer.cs
ArcturusZhang: This has been fixed

ArthurMa1978 (approved):

PR #1657, opened 2021-11-02
[LLC] Use GetAll<ResourceName> for list method if ResourceName has the same plural and sigular form

fengzhou-msft:
This PR also renames `Get` to `Get<ResourceName>`.

# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
+++ b/src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
@@ -8,12 +8,18 @@ using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using Humanizer;
+using Humanizer.Inflections;
 using Microsoft.CodeAnalysis.CSharp;
 
 namespace AutoRest.CSharp.Utilities
 {
     internal static class StringExtensions
     {
+        static StringExtensions()
+        {
+            Vocabularies.Default.AddUncountable("data");
+        }
+
         public static bool IsNullOrEmpty(this string? text) => String.IsNullOrEmpty(text);
         public static bool IsNullOrWhiteSpace(this string? text) => String.IsNullOrWhiteSpace(text);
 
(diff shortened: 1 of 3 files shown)


src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -8,12 +8,18 @@
...
+using Humanizer.Inflections;
 using Microsoft.CodeAnalysis.CSharp;
 
 namespace AutoRest.CSharp.Utilities
 {
     internal static class StringExtensions
     {
+        static StringExtensions()
+        {
+            Vocabularies.Default.AddUncountable("data");
```
fengzhou-msft, line 20: The library treats `datum` as the singular form for `data`. I think we want to use `data` in both singular and plural cases for .NET SDK, so customizing `add` as uncountable here.

annelo-msft (commented):
LGTM - let's have @AlexanderSher or @ShivangiReja review as well.

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -278,11 +284,29 @@ public static string ToPlural(this string single, bool inputIsKnownToBeSingular
             return single.Pluralize(inputIsKnownToBeSingular);
         }
 
+        private static string LastWordToPlural(this string single, bool inputIsKnownToBeSingular = true)
```
AlexanderSher, line 287: We need tests for this method

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -278,11 +284,29 @@ public static string ToPlural(this string single, bool inputIsKnownToBeSingular
...
+        {
+            var words = single.SplitByCamelCase();
+            var lastWord = words.LastOrDefault();
+            if (lastWord != null)
```
AlexanderSher, line 291: How is it possible that we don't have a single word?

test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs
```diff
@@ -21,6 +21,7 @@ public class StringExtensionsTests
         [TestCase("sheep", "sheep")]
         [TestCase("people", "people", false)]
         [TestCase("child", "children")]
+        [TestCase("data", "data", false)]
```
AlexanderSher, line 24: Can you also add cases for `"datatip"`, `"tipdata"`, `"da_ta"`?

test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs
```diff
@@ -43,10 +44,11 @@ public void ValidatePluralize(string noun, string expected, bool inputIsKnownToB
         [TestCase("sheep", "sheep")]
         [TestCase("people", "person")]
         [TestCase("children", "child")]
+        [TestCase("data", "data", false)]
```
AlexanderSher, line 47: And more cases here as well, including `"datas"`.

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
fengzhou-msft: Added more tests for `SplitByCamelCase` and I believe we will always have a last word. If the input is `_`, the result will be 2 empty strings `{"", ""}`.

test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs
fengzhou-msft: Added. As I only customized for `data`, all these words would use the original rules from the `Humanize` library which does not handle combinations of nouns well. While `RenameGetMethod` and `RenameListToGet` will use the private methods which extract the last word and transform it as expected. I also added tests fro them.

annelo-msft (approved):

AlexanderSher:
LGTM

Issue #1653, opened 2021-10-21
HTTP 202 will throw error in ApplicationGateway.Start


archerzz:
# Description
This needs further investigation. But in the current test case, `ApplicationGateway.Start` will throw an error like below. This could be due to back-end service not strictly follow the LRO spec. But it takes 45 minutes to create an application, so we need further investigation on this case.

```
Error Message:
   Azure.RequestFailedException : Service request failed.
Status: 202 (Accepted)

Content:
null

Headers:
Pragma: no-cache
x-ms-request-id: a38e1ce9-2bed-4d00-a23b-375e19229646
x-ms-client-request-id: 5696b0c46de2a5e742317089beba8b54
Azure-AsyncOperation: REDACTED
x-ms-correlation-request-id: REDACTED
x-ms-arm-service-request-id: REDACTED
Strict-Transport-Security: REDACTED
Content-Length: 4
Cache-Control: no-cache
Content-Type: application/json; charset=utf-8
Expires: -1
Location: REDACTED
Server: Microsoft-HTTPAPI/2.0,Microsoft-HTTPAPI/2.0
x-ms-ratelimit-remaining-subscription-reads: REDACTED
x-ms-routing-request-id: REDACTED
X-Content-Type-Options: REDACTED
Date: Fri, 27 Aug 2021 08:33:27 GMT

  Stack Trace:
     at Azure.Core.OperationInternals.<UpdateStatusAsync>d__23.MoveNext() in C:\Users\mingzhehuang\.nuget\packages\microsoft.azure.autorest.csharp\3.0.0-beta.20210826.1\content\Generator.Shared\OperationInternals.cs:line 132
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at Azure.Core.OperationInternals.<UpdateStatusAsync>d__25.MoveNext() in C:\Users\mingzhehuang\.nuget\packages\microsoft.azure.autorest.csharp\3.0.0-beta.20210826.1\content\Generator.Shared\OperationInternals.cs:line 142
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1675, opened 2021-11-11
feat(lro): honour `retry-after` and enlarge default polling interval

archerzz:
# Description

## Problem

This is found during diagnosis of #1653 Currently we have 2 issues in the LRO logic:
1. The polling interval is 1 second, which is too short. That could: a) make clients more likely to be rate limited by server side; b) generate a lot of LRO records during test recording.
2. We don't honor the `Retry-After` header in previous response, per [spec](https://github.com/Azure/azure-resource-manager-rpc/blob/69e229f04cb6b91a82c97fc6f61d56055107c447/v1.0/async-api-reference.md#call-action-post-asynchronously), we should follow that if available.
3. There is no wait between the operation request and the 1st polling request.

## Solution

 1. enlarge default polling interval from 1 second to 30 seconds
 2. wait the given interval before sending a polling request, instead of after, thus there is an interval between the operation request and the 1st polling request
 3. use `retry-after` header in the last response as the interval before the next polling request, use the last available value, otherwise roll back to default polling interval

# Checklist

# Ready to Land?


AlexanderSher (changes requested):
There is a similar change in `OperationInternal` from Azure.Core. Please wait with this change for couple days, I want to check if we can reuse it.

AlexanderSher:
Supported through `OperationInternal`

archerzz:
Changes were made in https://github.com/Azure/autorest.csharp/pull/1705

PR #1679, opened 2021-11-15
change DateTimeOffset serialization in TypeFormatter class

yukun-dong:
# Description
The Serialization of `DateTimeOffset` is controlled by `TypeFormatters` . For date with timezone 0 it would be serialized as such format "2021-10-27T02:05:15.0290004Z" else it would be "2021-10-27T10:05:15.0290004+08:00".

The current Storage service doesn't support the latter date format and would return 400 error("Request parameters'timetoRestore' should be a UTC DateTime."). Therefore when using SDK, need to call extra `ToUniversalTime()` function to convert local time to UTC time first.

Considering adding extra `ToUniversalTime()` may be unfriendly to users, we might consider modifying `TypeFormatters` class to serialize `DateTimeOffSet` to be UTC time directly.

# Checklist

# Ready to Land?

### DIFF
--- a/src/assets/Generator.Shared/TypeFormatters.cs
+++ b/src/assets/Generator.Shared/TypeFormatters.cs
@@ -26,8 +26,8 @@ namespace Azure.Core
         {
             "D" => value.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture),
             "U" => value.ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture),
-            "O" when value.Offset == TimeSpan.Zero => value.ToString(RoundtripZFormat, CultureInfo.InvariantCulture),
-            "o" when value.Offset == TimeSpan.Zero => value.ToString(RoundtripZFormat, CultureInfo.InvariantCulture),
+            "O" => value.ToUniversalTime().ToString(RoundtripZFormat, CultureInfo.InvariantCulture),
+            "o" => value.ToUniversalTime().ToString(RoundtripZFormat, CultureInfo.InvariantCulture),
             _ => value.ToString(format, CultureInfo.InvariantCulture)
         };
 
--- a/test/AutoRest.Shared.Tests/TypeFormatterTests.cs
+++ b/test/AutoRest.Shared.Tests/TypeFormatterTests.cs
@@ -13,9 +13,9 @@ namespace Azure.Core.Tests
         public static object[] DateTimeOffsetCases =
         {
             new object[] { "O", new DateTimeOffset(2020, 05, 04, 03, 02, 01, 123, default), "2020-05-04T03:02:01.1230000Z" },
-            //new object[] { "O", new DateTimeOffset(2020, 05, 04, 03, 02, 01, 123, new TimeSpan(1, 0, 0)), "2020-05-04T03:02:01.1230000+01:00" },
+            new object[] { "O", new DateTimeOffset(2020, 05, 04, 03, 02, 01, 123, new TimeSpan(1, 0, 0)), "2020-05-04T02:02:01.1230000Z" },
             new object[] { "O", new DateTimeOffset(3155378975999999999, default), "9999-12-31T23:59:59.9999999Z" },
-            //new object[] { "O", new DateTimeOffset(3155378975999999999, new TimeSpan(1, 0, 0)), "9999-12-31T23:59:59.9999999+01:00" },
+            new object[] { "O", new DateTimeOffset(3155378975999999999, new TimeSpan(1, 0, 0)), "9999-12-31T22:59:59.9999999Z" },
 
             new object[] { "o", new DateTimeOffset(2020, 05, 04, 03, 02, 01, 123, default), "2020-05-04T03:02:01.1230000Z" },
 
--- a/test/AutoRest.Shared.Tests/TypeFormatterTests.cs
+++ b/test/AutoRest.Shared.Tests/TypeFormatterTests.cs
@@ -55,5 +55,26 @@ namespace Azure.Core.Tests
             Assert.AreEqual(expected, formatted.ToString());
             Assert.AreEqual(date, formatted.GetDateTimeOffset(format));
         }
+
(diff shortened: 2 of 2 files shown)


test/AutoRest.Shared.Tests/TypeFormatterTests.cs
```diff
@@ -13,9 +13,9 @@ public class TypeFormatterTests
         public static object[] DateTimeOffsetCases =
         {
             new object[] { "O", new DateTimeOffset(2020, 05, 04, 03, 02, 01, 123, default), "2020-05-04T03:02:01.1230000Z" },
-            new object[] { "O", new DateTimeOffset(2020, 05, 04, 03, 02, 01, 123, new TimeSpan(1, 0, 0)), "2020-05-04T03:02:01.1230000+01:00" },
+            //new object[] { "O", new DateTimeOffset(2020, 05, 04, 03, 02, 01, 123, new TimeSpan(1, 0, 0)), "2020-05-04T03:02:01.1230000+01:00" },
```
archerzz, line 16: Just update the new test case, I think here it should be end with `Z`?

AlexanderSher:
Is there any existing client that will be broken by this change?

archerzz:
> Is there any existing client that will be broken by this change?

@AlexanderSher It's hard to search all reference to `TypeFormatter.ToString(DateTimeOffset value, string format)` since it's called back by JSON lib and it's in the shared codes. But we do know it has already broken the preview of track2 storage mgmt SDK.

- In Track1, we use `Datetime` and it's always been transformed to `YYYY-MM-DDZ` format **in requests**.
- In Track2, we use `DatetimeOffset` and it's serialization logic is different. It won't serialize to `YYYY-MM-DDZ` format, unless the time zone is UTC+0, which is not good for users, because they need to ensure time zone conversion before passing the value to SDK.

----

This PR changes `TypeFormatter`, which seems too broad, since we are not sue how it's used elsewhere.
How about this? We change the logic in https://github.com/Azure/autorest.csharp/blob/8d35db16beb918f3a8aecd44c16a35bf7941f9de/src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs#L105
- From: `writer.WriteStringValue(dateTimeOffset, "O");`
- To: `writer.WriteStringValue(dateTimeOffset.ToUniversalTime(), "O");`

This will only impact the serialization of request models. We already specify using `O` format which intends to use `YYYY-MM-DDZ` if time zone is UTC+0. From backend service perspective, it's hard to believe a service will support various time format, but not `YYYY-MM-DDZ` format which is a recommended way when handling date time serialization/deserialization. I think this change should be safe.

src/assets/Generator.Shared/TypeFormatters.cs
```diff
@@ -26,8 +26,8 @@ internal class TypeFormatters
         {
             "D" => value.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture),
             "U" => value.ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture),
-            "O" when value.Offset == TimeSpan.Zero => value.ToString(RoundtripZFormat, CultureInfo.InvariantCulture),
-            "o" when value.Offset == TimeSpan.Zero => value.ToString(RoundtripZFormat, CultureInfo.InvariantCulture),
+            "O" => value.ToUniversalTime().ToString(RoundtripZFormat, CultureInfo.InvariantCulture),
```
m-nash, line 29: Can we add unit tests here to verify that if I deserialize an object, store it some then reserialize it we don't have any data loss when it comes to timezone?

m-nash (changes requested):

archerzz:
@m-nash I added a few test cases to ensure the time millis are the same after conversion.

Besides, I find one API spec: https://github.com/microsoft/api-guidelines/blob/07a1e42e3e33f94df782abfcbaec24b63968c120/Guidelines.md#1131-the-dateliteral-format

> Informally, a DateValue is either an ISO 8601-formatted string or a JSON object containing two properties named kind and value that together define a point in time.

Here we are serializing the `DateTimeOffset` into a datetime string, not into a datetime JSON object which contain both value and kind. So it should follow [ISO 8601](https://262.ecma-international.org/5.1/#sec-15.9.1.15) which use the format `YYYY-MM-DDTHH:mm:ss.sssZ`.

AlexanderSher:
LGTM

test/AutoRest.Shared.Tests/TypeFormatterTests.cs
```diff
@@ -55,5 +55,26 @@ public void FormatsDatesAsJson(string format, DateTimeOffset date, string expect
...
+        {
+            string[] formats = { "O", "o" };
+
+            foreach (string format in formats)
+            {
+                var originalDate = DateTimeOffset.Parse(dateString);
+                var originalTimeMillis = originalDate.ToUnixTimeMilliseconds();
+
+                var formatted = TypeFormatters.ToString(originalDate, format);
+                var utcDate = DateTimeOffset.Parse(formatted);
```
AlexGhiondea, line 75: Do we use the DateTimeOffset.Parse method when deserializing, or are we using the TypeFormatter? If we use the TypeFormatter then I think we should test with that as well..

AlexanderSher: AFAIK, we never deserialize datetime by itself, it is either a part of a json or xml. For json, we call `Value.GetDateTimeOffset("O")`, for xml, we call `XElementExtensions.GetDateTimeOffsetValue`

I've checked both yesterday, both produced no errors.

m-nash: TypeFormatter ends up calling DateTimeOffset.Parse [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Generator.Shared/TypeFormatters.cs#L130).  So I think we are good.

AlexGhiondea (approved):

m-nash (approved):

ArthurMa1978 (approved):

PR #1692, opened 2021-11-17
Fix CodeGen attributes not working when output-folder is not a direct subfolder of the project directory

fengzhou-msft:
# Description
Some CodeGen attributes are used in customized code and will change the generated code accordingly. But if output-folder is not a direct subfolder of the project directory, for instance, when it's configured as `$(this-folder)/SomeFolder/Generated`, the attributes stop working. 

The root cause is that the generator needs to [add existing custom files](https://github.com/Azure/autorest.csharp/blob/93a2861f7d3f383d605b416d72262622664a07e7/src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs#L118-L126) in the project in order to find the attrributes defined in them. But [project directory](https://github.com/Azure/autorest.csharp/blob/93a2861f7d3f383d605b416d72262622664a07e7/src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs#L38) is hardcoded (with a hardcoded [ProjectRelativeDirectory](https://github.com/Azure/autorest.csharp/blob/93a2861f7d3f383d605b416d72262622664a07e7/src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs#L12) value) to always be the parent folder of the output-folder. 

This PR adds a `project-folder` option to give the correct value if ProjectRelativeDirectory is not the default `../`. It supports both absolute path and relative path to output-folder. The `project-folder` option aligns with the [shared flag](https://github.com/Azure/autorest/blob/main/docs/generate/flags.md#shared-flags) for autorest.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -122,7 +123,8 @@ internal static Configuration LoadConfiguration(string basePath, string json)
                 ReadOption(root, Configuration.Options.SkipCSProjPackageReference),
                 ReadOption(root, Configuration.Options.LowLevelClient),
                 ReadOption(root, Configuration.Options.SingleTopLevelClient),
-                MgmtConfiguration.LoadConfiguration(root)
+                MgmtConfiguration.LoadConfiguration(root),
```
AlexanderSher, line 126: Let's keep Mgmt configuration at the end.

fengzhou-msft: Updated.

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -46,6 +46,19 @@ private static void WriteIfNotDefault(Utf8JsonWriter writer, string option, bool
             }
         }
 
+        private static void WriteIfNotDefault(Utf8JsonWriter writer, string option, string? value)
```
AlexanderSher, line 49: Great!

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -35,7 +35,7 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(Task<CodeModel> codeModel
             ValidateConfiguration (configuration);
 
             Directory.CreateDirectory(configuration.OutputFolder);
-            var projectDirectory = Path.Combine(configuration.OutputFolder, Configuration.ProjectRelativeDirectory);
+            var projectDirectory = Path.Combine(configuration.OutputFolder, configuration.ProjectRelativeDirectory!);
```
AlexanderSher, line 38: I wonder if we should just call it `ProjectDirectory` and add somewhere a description that if it relative, it will be resolved from `OutputFolder`. `Path.Combine` will ignore the first argument, so absolute path should just work.

fengzhou-msft: Good suggestion. There is another place I'm not sure whether absolute path works and I will verify it:

https://github.com/Azure/autorest.csharp/blob/5e9f905be1deb76ece2ef6a721d8b09cf1b65501/src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpProj.cs#L122

I also discovered that autorest already has a [shared flag](https://github.com/Azure/autorest/blob/main/docs/generate/flags.md#shared-flags) called `project-folder`, so we can reuse that.

test/TestProjects/TypeSchemaMapping/readme.md
```diff
@@ -0,0 +1,16 @@
...
+
+``` yaml
+require: $(this-folder)/../../../readme.md
+input-file: $(this-folder)/TypeSchemaMapping.json
+namespace: TypeSchemaMapping
+output-folder: $(this-folder)/SomeFolder/Generated
+skip-csproj: true
+project-folder: $(this-folder)
+# relative path to output-folder also works
+# project-folder: ../../
```
fengzhou-msft, line 15: @AlexanderSher I updated the PR to support both absolute path and relative path for `project-folder`. To make things easier, in the constructor of Configuration, I will turn the absolute path to relative path, which requires no changes to existing projectDirectory handling (except the option renaming) and also meets the need that `Configuration.json` should only contain relative path (absolute path are not portable to different machines/environments).

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
fengzhou-msft: Verified that we have to use relative path for the above code snippet.

eng/Generation.psm1
```diff
@@ -25,6 +25,10 @@ function Invoke($command)
 function Invoke-AutoRest($baseOutput, $projectName, $autoRestArguments, $sharedSource, $fast)
 {
     $outputPath = Join-Path $baseOutput "Generated"
+    if ($projectName -eq "TypeSchemaMapping")
```
pakrym, line 28: Why do we need a special case for this project only?

pakrym:
Do we need to set the project folder in https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/build/CodeGeneration.targets#L35 ?

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -108,6 +112,23 @@ private static bool GetOptionValue(IPluginCommunication autoRest, string option)
             }
         }
 
+        private static string GetOptionStringValue(IPluginCommunication autoRest, string option, Func<string, string>? func)
```
pakrym, line 115: Do we need a new method?  Can we use `GetOptionValue(autoRest, Options.ModelNamespace) ?? "default value"` ?

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -108,6 +112,23 @@ private static bool GetOptionValue(IPluginCommunication autoRest, string option)
...
+        {
+            var projectFolder = autoRest.GetValue<string?>(Options.ProjectFolder).GetAwaiter().GetResult();
+            return projectFolder == null ? GetDefaultOptionStringValue(option)! : (func == null ? projectFolder : func(projectFolder));
+        }
+
+        public static string? GetDefaultOptionStringValue(string option)
```
pakrym, line 121: Should the default calculation move into the Configuration.cs? So it's applied to autorest and direct invocation both/

pakrym:
LGTM overall, some nits.

eng/Generation.psm1
fengzhou-msft: I am using this project to test when `outputPath` is not the default value. I tried to simply remove `outputPath`  in the command and use the values in readme.md instead, but I got quite a few errors for that and it would cost me more time to fix all the projects especially for the command when runnning with `-fast`, the changes needed seem tricky. So I chose the easy way to do it as I don't think we need more tests for `outputPath` not being the default value for now.

fengzhou-msft:
> Do we need to set the project folder in https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/build/CodeGeneration.targets#L35 ?

If we can only use the default value in CodeGeneration.targets, there is no need to add it there. Adding a value there has the issue that the config values in autorest.md will be overridden by autorest CLI option values. I'm actually making another PR (#1709) to add the feature to not add the values for output folder and namespace in msbuild so we can use the customized values in autorest.md.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
fengzhou-msft: GetOptionValue returns a bool, it was originally written for bool options only. I did not see a good way to to use it for string options, so I added a new method. And since the string may need extra processing, I added a new signature of `Func<string, string>? func` parameter.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
fengzhou-msft: I actually did the default calculation in Configuration.ctor in the initial code because of the same reason you pointed out. 

Then I realized other optional options are not doing the same. Configuration.json uses a strategy to only write the value if it's not default. Later when reading the values, it will give the bool option the default value if the option is not found. Putting the write and read code with the handling of default values together makes some sense to me. I used the same strategy for `ProjectFolder` and prefer to handle the options consistently, that's why I made the change.

AlexanderSher (approved):

PR #1705, opened 2021-11-19
Operation internal

AlexanderSher:
First part of the https://github.com/Azure/autorest.csharp/issues/1706

This change extracts status update  logic from `OperationInternals` into `NextLinkOperationImplementation` type and makes all autorest `XXXInternals` types use `OperationInternal` and `OperationInternal<T>` from Azure.Core.

This change doesn't modify any generated code, which will be updated in a series of PR's that would follow:
- First, `NonLongRunningOperationWriter` will be updated so that generated operation types will reference `Response<T>` directly.  That will allow us to get rid of `OperationOrResponseInternals` and `OperationOrResponseInternals<T>`
- Then, `MgmtLongRunningOperationWriter` will be updated to generate operations that use  `OperationInternal<T>` and implement `IOperation<T>` instead of `IOperationSource<T>`
- Finally, `LongRunningOperationWriter` will be updated similarly to `MgmtLongRunningOperationWriter` so that `OperationInternals` and `OperationInternals<T>` can be removed.

Current PR has two implications in `azure-sdk-for-net:
- `OperationInternal` and `OperationInternal<T>` are not included in management library projects by default. We would have to either update the `Directory.Build.Common.targets` or add them into every project explicitly (I'd prefer the first one).
- `OperationInternal` supports `retry-after` headers while `OperationInternals` does not. Lots of recorded tests have these headers in the recorded session, which makes tests run for hours. We will have to either change these sessions or add some filtering in the test infrastructure.


src/assets/Generator.Shared/OperationInternalsOfT.cs
```diff
@@ -21,12 +21,12 @@ namespace Azure.Core
     /// </summary>
     /// <typeparam name="T">The final result of the LRO.</typeparam>
 #pragma warning disable SA1649 // File name should match first type name
-    internal class OperationInternals<T> : OperationInternals
+    internal class OperationInternals<T>: IOperation<T>
 #pragma warning restore SA1649 // File name should match first type name
     {
         private readonly IOperationSource<T> _source;
```
pakrym, line 27: In the end of the series of refactorings are we still going to keep the `IOperationSource<T>`?

pakrym (approved):
👍👍

src/assets/Generator.Shared/OperationInternalsOfT.cs
AlexanderSher: No, will be removed as well.

Issue #1689, opened 2021-11-16
[LLC] For protocol methods, call CreateMessage(RequestContext context)


annelo-msft:
Once [this PR](https://github.com/Azure/azure-sdk-for-net/pull/25214) is merged, we will need generated code to call CreateMessage(RequestContext context) to create `HttpMessages` in order to enable the AddPolicy feature.

PR #1708, opened 2021-11-20
[LLC] For protocol methods, call CreateMessage(RequestContext context)

ShivangiReja:
Fixes: https://github.com/azure/autorest.csharp/issues/1689

* Update Azure.Core version to `1.22.0-alpha.20211119.1` and `Azure.Core Experimental` version to `0.1.0-preview.18`
* For protocol methods, call CreateMessage(RequestContext context)
* Remove LowLevelCallbackPolicy from LLC pipelines


samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountCollections.cs
```diff
@@ -709,9 +709,9 @@ internal HttpMessage CreateGetChildCollectionNamesRequest(string skipToken)
             return message;
         }
 
-        internal HttpMessage CreateGetCollectionPathRequest()
+        internal HttpMessage CreateGetCollectionPathRequest(RequestContext context)
```
annelo-msft, line 712: Let's make this an optional parameter by setting it to null in the method signature.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountCollections.cs
```diff
@@ -726,9 +726,9 @@ internal HttpMessage CreateGetCollectionPathRequest()
             return message;
         }
 
-        internal HttpMessage CreateGetChildCollectionNamesNextPageRequest(string nextLink, string skipToken)
+        internal HttpMessage CreateGetChildCollectionNamesNextPageRequest(string nextLink, string skipToken, RequestContext context)
```
annelo-msft, line 729: Same here regarding making context optional.

annelo-msft (commented):
Just a couple of comments on the output.  I'll let @AlexanderSher review the generator code changes.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountCollections.cs
ShivangiReja: We use common method for creating a request for HLC and LLC. For HLC, we don't add the null signature because it's an internal method. That's why I didn't change it. But if you want to me update it, I can do it but it will also update internal RestClient create request methods for HLC.

annelo-msft (approved):
LGTM wrt output.

PR #1709, opened 2021-11-21
Update CodeGeneration.targets so Azure.ResourceManager can use namespace and output-folder values from autorest.md

fengzhou-msft:
I drafted a [PR](https://github.com/Azure/azure-sdk-for-net/pull/25318) to autogen code for ResourceManager where batch tasks in [autorest.md](https://github.com/Azure/azure-sdk-for-net/blob/dd90d84da112ca4a2b3c53cca213b02c3af781e0/sdk/resourcemanager/Azure.ResourceManager/src/autorest.md) is used to generate swaggers for resources in different namespaces and folders.

However, when building the package with `dotnet build /t:GenerateCode`, the GenerateCode target provides values for `--namespace` and `--output-folder`, which override the corresponding values in `autorest.md`. So all generation in ResourceManager end up with the same namespace and generate code in the same `Generated` folder.

Originally I intended to fix the issue by removing these 2 autorest CLI options in `CodeGeneration.targets` and fix the autorest.md files that lack these values, but I found there are too many dataplane packages do not provide these 2 values in autorest.md. So I just added conditional check for `UseDefaultNamespaceAndOutputFolder` whenn running autorest command, and remove the 2 options if it's false. I'm setting the property for resourcemanager in https://github.com/Azure/azure-sdk-for-net/pull/25488. This has no impact with existing package generations and can avoid surprises if the target is used somewhere we're not aware of.

# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/build/CodeGeneration.targets
+++ b/src/AutoRest.CSharp/build/CodeGeneration.targets
@@ -17,6 +17,7 @@
     <ManagementSharedCodeDirectory>$(MSBuildThisFileDirectory)../content/Management.Shared/</ManagementSharedCodeDirectory>
 
     <_GenerateCode Condition="'$(AutoRestInput)' != ''">true</_GenerateCode>
+    <BaseCommand>node $(AutoRestEntryPoint) --max-memory-size=8192 --skip-csproj --skip-upgrade-check --version=$(AutoRestCoreVersion) $(AutoRestInput) $(AutoRestAdditionalParameters) --use=$(MSBuildThisFileDirectory)../tools/netcoreapp3.1/any/ --clear-output-folder=true --shared-source-folders=&quot;$(AzureCoreSharedCodeDirectory);$(AutoRestSharedCodeDirectory)&quot;</BaseCommand>
   </PropertyGroup>
 
   <Target Name="GenerateCode" Condition="'$(_GenerateCode)' == 'true'" >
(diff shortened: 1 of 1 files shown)


src/AutoRest.CSharp/build/CodeGeneration.targets
```diff
@@ -32,7 +33,8 @@
 
     <Error Text="Following GitHub URLs do not contain commit hash: @(GithubUrlsWithoutHash) please use permalinks for code generation inputs (see https://help.github.com/en/github/managing-files-in-a-repository/getting-permanent-links-to-files) " Condition="'@(GithubUrlsWithoutHash)' != ''" />
 
-    <Exec Command="node $(AutoRestEntryPoint) --max-memory-size=8192 --skip-csproj --skip-upgrade-check --version=$(AutoRestCoreVersion) $(AutoRestInput) $(AutoRestAdditionalParameters) --use=$(MSBuildThisFileDirectory)../tools/netcoreapp3.1/any/ --output-folder=$(MSBuildProjectDirectory)/Generated --clear-output-folder=true --namespace=$(RootNamespace) --shared-source-folders=&quot;$(AzureCoreSharedCodeDirectory);$(AutoRestSharedCodeDirectory)&quot;" />
+    <Exec Command="$(AutoRestBaseCommand) --output-folder=$(MSBuildProjectDirectory)/Generated --namespace=$(RootNamespace)" Condition="'$(NoDefaultNamespaceAndOutputFolder)' != 'true'"/>
```
weshaggard, line 36: Is msbuild the correct place to put this configuration? Should it instead be in the autorest.md?

fengzhou-msft: I agree that `output-folder` and `namespace`  should be configured by autorest.md. In this PR, I just tried to avoid breaking changes if we simply remove these configurations in msbuild as I saw lots of SDKs do not have `output-folder` or `namespace` configured in their autorest.md while allowing SDKs using the values in autorest.md with the new property.

weshaggard: I guess another option would be to move the opt-out into the autorest.md file for Azure.ResourceManager. If that opt-out is provided then we essentially ignore these options on the autorest side. 

Although I'm not an expert on how autorest works here so I'm not sure if that is the optimal approach or not.

fengzhou-msft: We'll need to touch autorest (core) as currently in autorest.csharp we just read the final config values from autorest which decides which source to use for the config value. And as a general practice, for command line tools, the priority for config values usually goes like command options > environment variables > config files (it's kind of the easier you can make changes to the values, the higher priority it has). We may not want to break that with some opt-out options for the autorest CLI. Therefore, I prefer to make the change within msbuild.

pakrym (approved):

Issue #1702, opened 2021-11-19
Support the case that in a request path has variables as the resource types


ArcturusZhang:
We have an example in DNS: 
```
/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/dnsZones/{zoneName}/{recordType}/{relativeRecordSetName}
```
Considering that our `ArmResource` has a `ResourceType` property as a constant - therefore it is not acceptable that we get a class from `ArmResource` that has multiple ResourceTypes.

Instead we need to expand this path into multiple paths, and each of them has the ResourceType as a constant.

Issue #1694, opened 2021-11-17
Tuple resource collection always complains no `GetAll` method with no required parameter


ArcturusZhang:
In our current implementation, we require a `GetAll` method in the resource collection class to provide the functionalities of iterating things in this collection.
But if the corresponding resource is a tuple resource, which means that to determine one resource we need a tuple of parameters instead of only one, like `/subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/cloudServiceOsFamilies/{osFamilyName}`.

To solve this problem, we might need to move all the required parameters of the `GetAll` method to its constructor so that the `GetAll` method does not have any required parameters, and we could provide functions to iterate over things.

PR #1711, opened 2021-11-22
[Mgmt] Expand enum in resource types & put extra parameters in `GetAll` method to constructor

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1702
Fixes https://github.com/Azure/autorest.csharp/issues/1694

# Description

This PR introduces the functionality that expand one resource with a resource type that contains variables in it into multiple resources with one solid and constant resource type. This is required since the base class of all resources `ArmResource` can only hold one constant resource type in its `ResourceType` property.

Therefore this PR changes our map for `Resource`s to `Dictionary<RequestPath Resource>`, where the key is the expanded request paths without variables in its corresponding resource type. Same thing happens to `ResourceCollection`s, since in this case we can have multiple resources under the same request path.

The `CreateOrUpdate` (and other methods in the collection class) will automatically omit the parameters that originally appear in the resource type, this variable will be given the corresponding constant value when invoking the corresponding `RestOperations`.

This PR also introduces a new format for the `request-path-to-resource-name` configuration to ensure you could use the configuration to override the resource name in this case. You can do this by using this format as key: `request-path|resource-type` (see the test project for detailed examples).

This PR also refactors the way we initialize resource and resource collection. Now the map of resources and the map of resource collections are guaranteed to initialize together.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -37,12 +38,12 @@ internal class MgmtOutputLibrary : OutputLibrary
         /// <summary>
         /// This is a map from raw request path to the corresponding <see cref="Resource"/>
         /// </summary>
-        private Dictionary<string, Resource>? _rawRequestPathToArmResource;
+        private Dictionary<string, Dictionary<ResourceType, Resource>>? _rawRequestPathToArmResource;
```
m-nash, line 41: Lets discuss this, its not intuitive why all of these types need a dictionary now and I wonder if there is a more clear way to represent this

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
```diff
@@ -24,7 +25,7 @@ public MgmtLongRunningOperation(Input.Operation operation, OperationGroup operat
             DefaultNamespace = $"{context.DefaultNamespace}.Models";
             if (operation.ShouldWrapResultType(ResultType, context))
             {
-                WrapperType = context.Library.GetArmResource(operation.GetHttpPath()).Type;
+                WrapperType = context.Library.GetArmResources(operation.GetHttpPath()).First().Type;
```
m-nash, line 28: is First always safe always?

m-nash (changes requested):

m-nash (approved):

ArthurMa1978 (approved):

PR #1713, opened 2021-11-23
Fix error when generating Servicebus mgmt

yukun-dong:
# Description

The subscription Id in ServiceBus: **/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ServiceBus/namespaces/{namespaceName}/topics/{topicName}/subscriptions/{subscriptionName}**

The current code generator would treat the latter "subscriptions" as the first "subscriptions" and pass the subscriptionId to the subscriptionName, which is incorrect.

```CSHARP
 var response = _subscriptionsRestClient.CreateOrUpdate(Id.ResourceGroupName, Id.Parent.Name, Id.Name, Id.SubscriptionId, parameters, cancellationToken);
```

This PR add a method to convert subscriptionId parameter in all operations to be Implementation.Method to fix the error when generating ServiceBus mgmt. After modification the generated code would be like this:
```csharp
var response = _subscriptionsRestClient.CreateOrUpdate(Id.SubscriptionId, Id.ResourceGroupName, Id.Parent.Name, Id.Name, subscriptionName, parameters, cancellationToken);
```
# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -102,6 +103,26 @@ public MgmtOutputLibrary(CodeModel codeModel, BuildContext<MgmtOutputLibrary> co
...
+                    var newpParams = op.Parameters.ToList();
+                    for (int i = 0; i < newpParams.Count; i++)
+                    {
+                        //updater the first subscriptionId to be 'method'
+                        if (newpParams[i].Language.Default.Name.Equals("subscriptionId", StringComparison.InvariantCultureIgnoreCase))
+                        {
+                            newpParams[i].Implementation = ImplementationLocation.Method;
+                            break;
+                        }
+                    }
```
ArcturusZhang, line 121: Since the class `Parameter` is a reference type, I believe it is valid if we convert this iteration into a `foreach` loop like this:
```suggestion
                    for (var p in op.Parameters)
                    {
                        //updater the first subscriptionId to be 'method'
                        if (p.Language.Default.Name.Equals("subscriptionId", StringComparison.InvariantCultureIgnoreCase))
                        {
                            p.Implementation = ImplementationLocation.Method;
                            break;
                        }
                    }
```

ArcturusZhang (approved):

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -81,6 +81,7 @@ public MgmtOutputLibrary(CodeModel codeModel, BuildContext<MgmtOutputLibrary> co
             OmitOperationGroups.RemoveOperationGroups(codeModel, context);
             _context = context;
             _mgmtConfiguration = context.Configuration.MgmtConfiguration;
+            UpdateSubscriptionIdForAllResource(codeModel);
```
AlexanderSher, line 84: So this method modifies original codeModel?

ArthurMa1978 (approved):

Issue #1680, opened 2021-11-15
feat(generator): `CreateOrUpdate` methods for single resource should return resource class, not data class


archerzz:
## Description

**[Note]**: This is `Request path` generator only, since existing generator will generate different code structure.

Right now for normal resources, the `CreateOrUpdate` methods will return the corresponding resource classes. For example, 
`public partial class CassandraResourceCreateUpdateCassandraKeyspaceOperation : Operation<CassandraKeyspace>, IOperationSource<CassandraKeyspace>` inherits `Operation<CassandraKeyspace>`, **not** `Operation<CassandraKeyspaceData>`.

But for singleton resource, the corresponding method signature is like `public partial class CassandraResourceMigrateCassandraKeyspaceToAutoscaleOperation : Operation<ThroughputSettingsData>, IOperationSource<ThroughputSettingsData>`. Here is the original [REST spec](https://github.com/Azure/azure-rest-api-specs/blob/f780b571abddf2fc99233212a00510d3bef6878d/specification/cosmos-db/resource-manager/Microsoft.DocumentDB/stable/2021-06-15/cosmos-db.json#L3842).

That's not symmetric, and we should try to keep consistent here. Besides, there could be issues in other methods (patch, etc).

PR #1723, opened 2021-11-26
now we find resource in LRO and Fake-LRO from request path instead of using resource data

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1680

# Description

Previously we have a functionality that if a LRO or a fake LRO is returning a resource data, we would check if it corresponds to a resource, if it does, we replace the return type of this LRO or fake LRO.

But we get quite a few cases in other RPs that multiple resources shares the same resource data, the above algorithm cannot determine which resource this LRO or fake LRO belongs.

This PR changes the algorithm to use request path to find its corresponding resource, which solves the above situation.

# Checklist

# Ready to Land?


AlexanderSher:
Can we have some tests for this?

PR #1724, opened 2021-11-26
support class reference type

archerzz:
# Description
- add `ClassReferenceTypeChooser` which will check if  a model/resource data class can be replaced by a reference type
- change model initialization in `MgmtOutputLibaray`, replace `CSharpType` in model/resource data with the reference types
- change code gen logic:
  - support generate `ClassReferenceType` attribute accoring to `x-ms-mgmt-classReferenceType` (resource manager only)
  - do not generate classes for model/resource data which can be replaced by reference types
  - change generated de-serialization logic for model/resource data which can be replaced by reference types

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -102,6 +103,76 @@ public MgmtOutputLibrary(CodeModel codeModel, BuildContext<MgmtOutputLibrary> co
...
+            _resourceModels = new Dictionary<Schema, TypeProvider>();
+
+            // first, construct models and resource data models
+            foreach (var schema in _allSchemas)
+            {
+                if (_resourceDataSchemaNameToOperationSets.ContainsKey(schema.Name))
+                {
+                    var model = BuildResourceModel(schema);
+                    _resourceModels.Add(schema, model);
+                    _nameToTypeProvider.Add(schema.Name, model); // TODO: ADO #5829 create new dictionary that allows look-up with multiple key types to eliminate duplicate dictionaries
```
AlexanderSher, line 119: Wouldn't the keys `schema` and `schema.Name` be semantically the same?

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -241,9 +312,25 @@ private ArmClientExtensions EnsureArmClientExtensions()
 
         private Dictionary<Schema, TypeProvider>? _models;
 
-        public Dictionary<Schema, TypeProvider> ResourceSchemaMap => _resourceModels ??= BuildResourceModels();
+        public Dictionary<Schema, TypeProvider> ResourceSchemaMap
+        {
+            get
+            {
+                if (_resourceModels == null)
```
AlexanderSher, line 319: Do you need these properties to be lazy? Can you initialize them directly?

archerzz: @AlexanderSher I tried to initialize them in constructor, and failed... There is a cyclic reference between `MgmtOutputLibrary<BuildContext>` and `BuildContext`. A significant refactoring is required so that we initialize the models in constructor.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
archerzz: I think so. I moved the codes from other places. `_resourceModels` and `_nameToTypeProvider` are used in different scenarios in which one used `schema` as key and another use `schema.Name` as key. It would be better we refactor them in a separate PR.

archerzz:
changed the annotation name from `x-ms-mgmt-classReferenceType` to `x-ms-typeReferenceType`, because this logic has the potential to replace enum types as well.

fengzhou-msft (approved):

PR #1725, opened 2021-11-29
Recognize the singleton resource if the diff request path is a constant

ArcturusZhang:
# Description

If the diff between the request path of a resource and its parent is constant (without variables in it), we will recognize it as a singleton resource now. In this case, the name of this resource will be a constant, which means this is a singleton resource and should not have a collection.

This PR also introduces a new configuration to restore previous behavior back by assign `singleton-resource-requires-keyword: true` in the configuration

TODO list:
- the List method should go to its parent class (extension class if the parent is outside the RP), now it will just be lost

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
```diff
@@ -105,6 +106,7 @@ public void Write(Utf8JsonWriter writer, string settingName)
             IsArmCore = !IsValidJsonElement(armCore) ? false : Convert.ToBoolean(armCore.ToString());
             DoesResourceModelRequireType = !IsValidJsonElement(resourceModelRequiresType) ? true : Convert.ToBoolean(resourceModelRequiresType.ToString());
             DoesResourceModelRequireName = !IsValidJsonElement(resourceModelRequiresName) ? true : Convert.ToBoolean(resourceModelRequiresName.ToString());
+            DoesSingletonRequiresKeyword = !IsValidJsonElement(singletonRequiresKeyword) ? false : Convert.ToBoolean(singletonRequiresKeyword.ToString());
```
lirenhe, line 109: make name consistent?

ArcturusZhang: Local variables I believe we do not need to make them consistent. If you are referring to the property names, it is consistent.

ArcturusZhang: And this configuration is here to keep backward compatibility

fengzhou-msft (approved):

PR #1741, opened 2021-12-06
[LLC] Null Validations Fix

### LINKED ISSUE #1697: [LLC] Generated LLC does not throw  parameter validation exceptions as claimed in docstrings
When Generating an LLC client, the methods in the client contain docstrings that claim to make validations (for example: asserting that a required parameter is not null). However, these assertions are not actually being made. An example for the docstring as well as the associated client [can be found here](https://github.com/Azure/azure-sdk-for-net/blob/cb7f4c05046fc78177b1a245ed755d20ae4c101d/sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/src/Generated/QuestionAnsweringProjectsClient.cs#L65). Unless I'm missing something, this is a bug.

The most straightforward solution is to ensure that these kinds of validations are being generated within the methods (similar to non-LLCs).

AhmedLeithy:
# Description

Hey all, 

This PR fixes #1697.

Seems like there were a couple of `WriteParameterNullChecks()` calls missing from the `WriteLRO` and `WriteClient` methods. I added them and regenerated the test and sample clients. They now have the correct validations. 

This is my first time contributing to this repo, so apologies if I missed anything!

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
+++ b/src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
@@ -212,6 +212,8 @@ namespace AutoRest.CSharp.Generation.Writers
 
             using (WriteClientMethodDeclaration(writer, clientMethod, clientMethod.OperationSchemas, returnType, async))
             {
+                writer.WriteParameterNullChecks(clientMethod.RestMethod.Parameters);
+
                 using (WriteDiagnosticScope(writer, clientMethod.Diagnostic, client.Fields.ClientDiagnosticsField.Name))
                 {
                     var messageVariable = new CodeWriterDeclaration("message");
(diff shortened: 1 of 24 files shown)


src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -216,6 +216,8 @@ private static void WriteClientMethod(CodeWriter writer, LowLevelRestClient clie
 
             using (WriteClientMethodDeclaration(writer, restMethod, clientMethod.OperationSchemas, returnType, async))
             {
+                writer.WriteParameterNullChecks(clientMethod.RestMethod.Parameters);
```
annelo-msft, line 219: Where do we check whether the parameter is required or optional?

AhmedLeithy: My understanding is as follows: parameter objects have the bool property `ValidateNotNull`, which is used [here](https://github.com/Azure/autorest.csharp/blob/5474a4124a0e64707779e9612e6da4c3fbd81e24/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs#L213) to decide whether to write the code check for this variable or not. Default values for the parameter also seem to be taken into consideration.

Am I answering the right question? Or are you asking about how the properties are set in the first place?

AlexanderSher (approved):

PR #1742, opened 2021-12-06
Preliminary changes for new LRO pattern and resource clients

AlexanderSher:
This PR contains several changes that have minimal impact on generated code but are required for the upcoming work on new LRO pattern and resource clients:
- Method generation is detached from `OperationGroup` and bound to `IEnumerable<Operation>` only, so client can have methods referencing operations from different operation groups
- Special parameters that are added to `CreateXXXRequest` methods now have fixed order and are all added in one place
- Changes in `LowLevelRestClient` have been affecting data plane code generation, so now it is derived from `TypeProvider` itself
- Versions of autorest libraries are bumped


samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineScaleSetExtensionUpdate.Serialization.cs
```diff
@@ -28,10 +28,10 @@ void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
                 writer.WritePropertyName("publisher");
                 writer.WriteStringValue(Publisher);
             }
-            if (Optional.IsDefined(TypePropertiesType))
+            if (Optional.IsDefined(TypeAutoGenerated))
```
AlexanderSher, line 31: This change is from autorest.core update

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -107,198 +167,174 @@ private Response[] BuildResponses(Operation operation, Request request, out CSha
...
+                    case RequestConditionHeaders.IfMatch | RequestConditionHeaders.IfNoneMatch | RequestConditionHeaders.IfModifiedSince | RequestConditionHeaders.IfUnmodifiedSince:
+                        requestConditionParameter = RequestConditionsParameter;
+                        break;
+                    case RequestConditionHeaders.IfMatch:
+                    case RequestConditionHeaders.IfNoneMatch:
+                        requestConditionParameter = BuildParameter(requestConditionRequestParameter, typeof(ETag));
+                        references[requestConditionRequestParameter] = CreateReference(requestConditionRequestParameter, requestConditionParameter);
+                        break;
+                    default:
+                        throw new NotSupportedException($"Behavior for combination of headers '{requestConditionHeaders}' in '{operation.Language.Default.Name}' is undefined");
```
AlexanderSher, line 269: According to https://github.com/Azure/azure-sdk-for-net/issues/24853, only these 3 cases are supported. If we want to support other cases instead of throwing exception, we need to define method signature for remaining cases explicitly.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -107,198 +167,174 @@ private Response[] BuildResponses(Operation operation, Request request, out CSha
...
-            }
+                else if (contentTypeRequestParameter == null && IsContentTypeParameter(requestParameter))
+                {
+                    contentTypeRequestParameter = requestParameter;
+                }
+                else if (IsRequestConditionHeader(requestParameter, out var header))
+                {
+                    if (requestParameter.IsRequired)
+                    {
+                        throw new Exception("Required conditional request headers are not supported.");
```
AlexanderSher, line 210: We haven't discussed where conditional request header parameter should be positioned if it is required. I'm not sure, however, if these header parameters can be required, but IMO it is better to fail instead of generating unexpected signature.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -1416,131 +1416,110 @@ public virtual Operation<BinaryData> StartTranslation(RequestContent content, Re
             }
         }
 
-        internal HttpMessage CreateStartTranslationRequest(RequestContent content)
+        internal HttpMessage CreateGetDocumentStatusRequest(Guid id, Guid documentId)
```
annelo-msft, line 1419: Out of curiosity - why are these methods being reordered?  It make it harder to see what the actual change is to this file.

samples/Azure.AI.FormRecognizer/Generated/CodeModel.yaml
```diff
@@ -3066,6 +3066,8 @@ operationGroups:
           - !Parameter &ref_126
             schema: *ref_124
             implementation: Method
+            extensions:
```
annelo-msft, line 3069: What motivates adding this to FormRecognizer?

samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineScaleSetExtensionUpdate.Serialization.cs
annelo-msft: Is it not a breaking change to rename a property?

samples/CognitiveSearch/Generated/CodeModel.yaml
```diff
@@ -7999,7 +7999,8 @@ schemas: !Schemas
                     description: Parameter group
                 protocol: !Protocols {}
               implementation: Method
-              extensions: {}
+              extensions:
+                x-nullable: false
```
annelo-msft, line 8003: Similarly, will this result in any unintended changes to the Search API?

src/AutoRest.CSharp/Common/Output/Builders/ClientBuilder.cs
```diff
@@ -67,7 +66,7 @@ public static string GetClientPrefix(string name, BuildContext context)
                 foreach (var request in operation.Requests)
                 {
                     RestClientMethod startMethod = restClient.GetOperationMethod(request);
-                    var name = nameOverrider?.Invoke(operationGroup, operation, startMethod) ?? operation.CSharpName();
+                    var name = operation.CSharpName();
```
annelo-msft, line 69: Has this functionality moved to another place?

AlexanderSher: No, `nameOverrider` is always `null`.

samples/Azure.AI.FormRecognizer/Generated/CodeModel.yaml
AlexanderSher: @timotheeguerin , do you know what change has added this?

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
AlexanderSher: That is due to typo :)
Fixed.

samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineScaleSetExtensionUpdate.Serialization.cs
AlexanderSher: Yes, it is. @m-nash, this change has affected only mgmt, do you know anything about it?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -31,6 +31,7 @@ internal class LowLevelClientWriter : ClientWriter
     {
         private static readonly CSharpType RequestContextParameterType = new(typeof(RequestContext), true);
 
+        private static readonly Parameter WaitForCompletionParameter = new("waitForCompletion", "true if method should return after the long-running operation is completed; otherwise, false.", new CSharpType(typeof(bool)), null, false);
```
AlexanderSher, line 34: @annelo-msft , can you think of a better description here?

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
AlexanderSher: Leave the exception here for now.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
AlexanderSher: Updated so that `RequestConditions` will be a default parameter type.

samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineScaleSetExtensionData.Serialization.cs
```diff
@@ -33,10 +33,10 @@ void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
                 writer.WritePropertyName("publisher");
                 writer.WriteStringValue(Publisher);
             }
-            if (Optional.IsDefined(TypePropertiesType))
+            if (Optional.IsDefined(TypeAutoGenerated))
```
m-nash, line 36: Definitely a breaking change that we need to investigate, if its coming from another package, we need to work with them on what the change was intended to do and put in a fix before we adopt the new package

m-nash (changes requested):

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
annelo-msft: How about this?

`true` if the method should wait to return until the long-running operation has completed on the service; `false` if it should return after starting the operation.  For more information on long-running operations, please see [Azure.Core Long-Running Operation samples.](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/LongRunningOperations.md).

src/AutoRest.CSharp/Common/Output/Models/Requests/RestClientMethod.cs
```diff
@@ -36,31 +36,5 @@ public RestClientMethod(string name, string? description, CSharpType? returnType
...
-                var pathSegments = this.Request.PathParameterSegments;
-
-                return this.Parameters.Where(parameter => !pathSegments.Any(
-                    pathSegment => pathSegment.Value.Reference.Type.Name == parameter.Type.Name &&
-                    pathSegment.Value.Reference.Name == parameter.Name)
-                ).ToList();
-            }
-        }
-
-        public List<Parameter> PathParameters
```
m-nash, line 53: I don't think we should move to this mgmt plane only, it applies to any method, there are set of params used on the path and a set of params used elsewhere

AlexanderSher: This property has been added in this PR: https://github.com/Azure/autorest.csharp/pull/1368/files#r660745102

I was against adding it to the common code, but I never had a change to finish that conversation because PR was submitted while I was busy with LLC. Here were my reasons:
- It is never used as a collection, the only usage it to find if parameter is path parameter or not
- It is a duplicated logic. In the `RestClientBuilder`, we have all the information about parameter position, so one can add a `RequestLocation` property to `Parameter` if needed.
- Comparing `CSharpType` instances by `Name` in general case is incorrect. If these types are generic, it will ignore generic parameters. Nullability is also not part of the name. So this property is case-specific.
- It behaves like enumerable (which actually hints that it is, in fact, an extension method), but creates a new List<T> every time in process.
- Not very important in autorest.csharp, but it is also quite slow. For example, in `BuildParameterMapping`, calling this code requires O(P^3) in worst case.

samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineScaleSetExtensionData.Serialization.cs
AlexanderSher: Fixed

annelo-msft (approved):
Per offline discussion, outstanding issues have been addressed.

samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineScaleSetExtensionUpdate.Serialization.cs
AlexanderSher: Fixed

samples/Azure.AI.FormRecognizer/Generated/CodeModel.yaml
AlexanderSher: I've confirmed that this change has no effect on FormRecognizer

m-nash (approved):

PR #1749, opened 2021-12-08
Fix for missing api-version in OperationLocation

fengzhou-msft:
Resolve https://github.com/Azure/azure-sdk-for-net/issues/25411

# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/assets/Generator.Shared/NextLinkOperationImplementation.cs
+++ b/src/assets/Generator.Shared/NextLinkOperationImplementation.cs
@@ -26,7 +26,6 @@ namespace Azure.Core
 
         private string? _lastKnownLocation;
         private string _nextRequestUri;
-        private string? _apiVersion;
 
         public static IOperation Create(HttpPipeline pipeline, RequestMethod requestMethod, Uri startRequestUri, Response response, OperationFinalStateVia finalStateVia)
         {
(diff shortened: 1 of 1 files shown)


src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -26,6 +26,7 @@ internal class NextLinkOperationImplementation : IOperation
 
         private string? _lastKnownLocation;
         private string _nextRequestUri;
+        private string? _apiVersion;
```
annelo-msft, line 29: I'm wondering how this approach will scale to accommodate all of the customizations that could happen in code generated from a swagger file.  For example, there is a recent customer request to handle a 207 response as a request as well - would we handle that here, too?  Or is there a way to add an abstract method (or something similar) to call out to code generated for the library to create the message for polling requests?

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -115,6 +115,17 @@ private void UpdateNextRequestUri(ResponseHeaders headers)
             }
         }
 
+        private static string AppendApiVersion(string uri, Uri startRequestUri)
+        {
+            var apiVersion = startRequestUri.Query.Trim('?').Split('&').Select(s => s.Split('=')).Where(arr => arr[0].Equals("api-version")).Select(arr => arr[1]).FirstOrDefault();
```
annelo-msft, line 120: https://docs.microsoft.com/en-us/dotnet/api/system.web.httputility.parsequerystring?view=net-6.0 ?

annelo-msft (approved):
@AlexanderSher, I am approving this as a fix for the Cognitive library issue -- I will start keeping track of it as something we can refactor to use generated code as I proposed elsewhere.

AlexanderSher:
Conversation goes on in https://github.com/Azure/azure-sdk-for-net/issues/25411, so I think we should wait with merge here.

PR #1751, opened 2021-12-08
[LLC] Call CreateMessage(RequestContext context) instead of CreateMessage()

AlexanderSher:
Rebase of @ShivangiReja 's https://github.com/Azure/autorest.csharp/pull/1708 over the current codebase


src/AutoRest.CSharp/Common/Output/Models/Shared/KnownParameters.cs
```diff
@@ -0,0 +1,29 @@
...
+        public static readonly Parameter Pipeline = new("pipeline", "The pipeline instance to use", new CSharpType(typeof(HttpPipeline)), null, true);
+        public static readonly Parameter KeyAuth = new("keyCredential", "The key credential to copy", new CSharpType(typeof(AzureKeyCredential)), null, false);
+        public static readonly Parameter TokenAuth = new("tokenCredential", "The token credential to copy", new CSharpType(typeof(TokenCredential)), null, false);
+
+        public static readonly Parameter RequestContent = new("content", "The content to send as the body of the request.", RequestContentType, null, true);
+        public static readonly Parameter RequestContentNullable = new("content", "The content to send as the body of the request.", RequestContentNullableType, /*Constant.Default(RequestContentNullableType)*/null, false);
+
+        public static readonly Parameter RequestContext = new("context", "The request context.", RequestContextNullableType, Constant.Default(RequestContextNullableType), false);
```
AlexanderSher, line 25: @annelo-msft , we probably need better description for `RequestContext` (and maybe you want better descriptions for other parameters as well).

annelo-msft: What about "The request context, which can override default behaviors on the request on a per-call basis." ?

annelo-msft (approved):

Issue #1746, opened 2021-12-07
[LLC] Generator creates duplicate when two MIME types are listed in consumes block


annelo-msft:
#### Repro

Generate a client from this swagger file: https://github.com/Azure/autorest.testserver/blob/main/swagger/dpg-update1.json#L43

#### Actual result

Two identical methods are created for:
1. PostParametersAsync
2. PostParameters
3. CreatePostParametersRequest

#### Expected result

One method should be created that has the contentType parameter, allow the user to specify one or the other of the MIME types.

fengzhou-msft:
Modelerfour changes for this issue: https://github.com/Azure/autorest/pull/4405, we will need to update the generator to leverage the new `requestMediaTypes` entry.

lirenhe:
@annelo-msft, it looks this link is broken: https://github.com/Azure/autorest.testserver/blob/main/swagger/llc_update1.json#L43
Could you help to update it?

pshao25:
> @annelo-msft, it looks this link is broken: https://github.com/Azure/autorest.testserver/blob/main/swagger/llc_update1.json#L43 Could you help to update it?

Updated. It changes to https://github.com/Azure/autorest.testserver/blob/main/swagger/dpg-update1.json

annelo-msft:
Thank you, @pshao25! :)

lirenhe:
The original PR was reverted due to the apply of the latest of m4 and autorest.core.  @pshao25,  could you take a look on how to add the fix back?

PR #1763, opened 2021-12-09
Add DPG testserver tests for service driven evolution

fengzhou-msft:
# Description
This PR uses customized code to fix the dpg code of `dpg_update1` and `media_types` due to #1746
Addresses https://github.com/azure/autorest.csharp/issues/1944

Testing cases:
- [x] **params_getNewOperation** - Initially neither path or method exist for this operation. After evolution, this is a **new method in a new path**
- [x] **params_deleteParameters** - Initially the path exists but there is no delete method. After evolution this is a **new method in a known path**
- [x] **params_headNoParams** - Initially has no query parameters. After evolution, a **new optional query parameter** is added
- [x] **params_getRequired** - Initially only has one required Query Parameter. After evolution, a **new optional query parameter** is added
- [x] **params_getOptional** - Initially has one optional query parameter. After evolution, a **new optional query parameter** is added
- [x] **params_putRequiredOptional** - Initially has one required query parameter and one optional query parameter.  After evolution, a **new optional query parameter** is added

Relevant docs:
- https://github.com/Azure/autorest.testserver/wiki/Data-plane-codegen-testing#service-driven-evolution
- https://github.com/Azure/autorest.testserver/blob/main/swagger/dpg_initial.json
- https://github.com/Azure/autorest.testserver/blob/main/swagger/dpg_update1.json

# Checklist

# Ready to Land?


test/TestServerProjectsLowLevel/llc_update1/Custom/ParamsClient.cs
```diff
@@ -0,0 +1,81 @@
...
+    [CodeGenSuppress("CreatePostParametersRequest", typeof(RequestContent), typeof(RequestContext))]
+    public partial class ParamsClient
+    {
+        /// <summary> POST a JSON or a JPEG. </summary>
+        /// <param name="content"> The content to send as the body of the request. </param>
+        /// <param name="contentType"> Upload file type. Allowed values: &quot;application/json&quot; | &quot;image/jpeg&quot;. </param>
+        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="content"/> is null. </exception>
+#pragma warning disable AZC0002
+        public virtual async Task<Response> PostParametersAsync(RequestContent content, ContentType contentType, RequestContext context = null)
```
ShivangiReja, line 25: I believe we want these methods for testing purposes, I wonder why don't we add this method in testserver repo in the [swagger](https://github.com/Azure/autorest.testserver/blob/main/swagger/llc_update1.json) so that other languages can also leverage?

ShivangiReja, line 25: I believe we want this method for testing purposes, I wonder why don't we add this in testserver repo in the [swagger](https://github.com/Azure/autorest.testserver/blob/main/swagger/llc_update1.json) so that other languages can also leverage this?

fengzhou-msft: This method does come from testserver swagger https://github.com/Azure/autorest.testserver/blob/main/swagger/llc_update1.json#L93 and we use custom code here as LLC currently generates duplicate methods.

ShivangiReja (approved):

test/TestServerProjectsLowLevel/llc_update1/Custom/ParamsClient.cs
```diff
@@ -0,0 +1,81 @@
...
+using Azure;
+using Azure.Core;
+
+namespace llc_update1_LowLevel
+{
+    /// <summary> The Params service client. </summary>
+    [CodeGenSuppress("PostParametersAsync", typeof(RequestContent), typeof(RequestContext))]
+    [CodeGenSuppress("PostParameters", typeof(RequestContent), typeof(RequestContext))]
+    [CodeGenSuppress("CreatePostParametersRequest", typeof(RequestContent), typeof(RequestContext))]
+    public partial class ParamsClient
```
annelo-msft, line 17: @fengzhou-msft, I see you've added customization here.  Is the idea that once we've introduced "state in autorest" we'll come back and update this test to handle it fully generated?

fengzhou-msft: yes, that's the plan.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
```diff
@@ -0,0 +1,41 @@
...
+using AutoRest.TestServer.Tests.Infrastructure;
+using dpg_initial_LowLevel;
+using NUnit.Framework;
+
+namespace AutoRest.TestServer.Tests
+{
+    public class DpgInitialTest : TestServerLowLevelTestBase
+    {
+        [Test]
+        public Task DPGAddOptionalInput() => Test(async (host) =>
```
annelo-msft, line 15: nit: please use .NET casing convention consistently.

test/AutoRest.TestServerLowLevel.Tests/dpg_update1.cs
```diff
@@ -0,0 +1,79 @@
...
+            var result = await new ParamsClient(Key, host).GetRequiredAsync("a");
+            var responseBody = JsonData.FromBytes(result.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody["message"]);
+            var result2 = await new ParamsClient(Key, host).GetRequiredAsync("a", default);
+            var responseBody2 = JsonData.FromBytes(result2.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody2["message"]);
+            var result3 = await new ParamsClient(Key, host).GetRequiredAsync("a", context: default);
+            var responseBody3 = JsonData.FromBytes(result3.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody3["message"]);
+            var result4 = await new ParamsClient(Key, host).GetRequiredAsync("a", "b");
```
annelo-msft, line 28: Could you help me understand what's happening here?  How do we know we didn't make breaking changes between the initial test and the update1 test?

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
fengzhou-msft: Updated.

test/AutoRest.TestServerLowLevel.Tests/dpg_update1.cs
fengzhou-msft: Customized code like overload methods are added to make v2(update1) backward compatible for non-breaking changes from swagger. Here all the test cases of initial are copied to update1. Therefore, we can consider update1 is backward compatible with inital.

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
```diff
@@ -0,0 +1,184 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Core;
+
+namespace dpg_update1_LowLevel
+{
+    /// <summary> The Params service client. </summary>
+    [CodeGenSuppress("PostParametersAsync", typeof(RequestContent), typeof(RequestContext))]
```
annelo-msft, line 14: Curious: What do the [CodeGenSuppress] attributes do?

annelo-msft (changes requested):
I've updated the description with the full set of test cases from the swagger -- I'm not sure I see tests that validate each of them, but if they're there, let me know and I'll have another look at this tomorrow.  Thanks!

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
fengzhou-msft: The autogenerated `PostParametersAsync` has the issue of generating duplicate methods, so here we can use `CodeGenSuppress` to suppress the autogenerated `PostParametersAsync` methods whose input parameters are of the types of RequestContent and RequestContext. Then we can write the correct customized code for a `PostParametersAsync` method.

The generator will scan existing customized code (all the files in the project that are not under the Generated folder) for the attributes like CodeGenSuppress, then apply the actions like removing the method to the generated code with Roslyn code analyzer.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerLowLevelTestBase.cs
```diff
@@ -81,7 +81,7 @@ private async Task Test(string scenario, Func<Uri, Task> test, bool ignoreScenar
         private static string GetScenarioName()
         {
             var testName = TestContext.CurrentContext.Test.Name;
-            var indexOfUnderscore = testName.IndexOfAny(new[]{'_', '('});
+            var indexOfUnderscore = testName.LastIndexOfAny(new[]{'_', '('});
```
fengzhou-msft, line 84: Previously we have cases to use multiple tests to test different cases of the same scenario like `DpgNewBodyType_JSON` and `DpgNewBodyType_JPEG` will both count coverage for `DpgNewBodyType`. Therefore, we have this logic to drop the part after the underscore.

In the updated testserver, however, the scenario name itself can have underscores like `DpgAddOptionalInput_OptionalParam`. Therefore, I updated here to drop the part after the last underscore and we can append `_Test` (for instance, use `DpgAddOptionalInput_OptionalParam_Test`) to the test name if the test name contains underscores.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
```diff
@@ -0,0 +1,41 @@
...
+        [Test]
+        public Task DpgAddOptionalInput() => Test(async (host) =>
+        {
+            var result = await new ParamsClient(Key, host).GetRequiredAsync("a");
+            var responseBody = JsonData.FromBytes(result.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody["message"]);
+            var result2 = await new ParamsClient(Key, host).GetRequiredAsync("a", default);
+            var responseBody2 = JsonData.FromBytes(result2.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody2["message"]);
+            var result3 = await new ParamsClient(Key, host).GetRequiredAsync("a", context: default);
```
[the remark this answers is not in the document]
fengzhou-msft: Here I'm trying to list all the ways that a customer can call with the method in `dpg_initial` and then copy the same tests to `dpg_update1`. We can catch that some cases are not supported by `dpg_update1` generated code. After adding customized overload methods to `dpg_update1` , all the cases here in `dpg_initial` are then supported by `dpg_update1`.

fengzhou-msft:
> I've updated the description with the full set of test cases from the swagger -- I'm not sure I see tests that validate each of them, but if they're there, let me know and I'll have another look at this tomorrow. Thanks!

I just noticed that there are now more scenarios supported by the testserver. I have added all the ones listed in the description.

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
annelo-msft: I see, thank you for that explanation!

Is that the final approach we decided for addressing this problem, that we'll recommend not using the generated method and writing the two method overloads for different content types by hand?

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
```diff
@@ -0,0 +1,320 @@
...
+    public partial class ParamsClient
+    {
+        /// <summary>
+        /// Get true Boolean value on path.
+        ///  Initially has one optional query parameter. After evolution, a new optional query parameter is added
+        ///  Overload method to be compatible with dpg_initial where optional parameters become required.
+        /// </summary>
+        /// <param name="optionalParam"> I am an optional parameter. </param>
+        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        public virtual async Task<Response> GetOptionalAsync(string optionalParam, RequestContext context)
```
annelo-msft, line 26: Are we able to ExportAPI.cs for these clients?  It would be helpful for reviewing the public APIs for them.

annelo-msft, line 26: I see, we are writing these manually right now, but in the future, these methods will be generated by the state-aware generator, is that correct?

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
```diff
@@ -0,0 +1,320 @@
...
+        /// </remarks>
+        public virtual async Task<Response> PostParametersAsync(RequestContent content, RequestContext context = null)
+        {
+            Argument.AssertNotNull(content, nameof(content));
+
+            using var scope = ClientDiagnostics.CreateScope("ParamsClient.PostParameters");
+            scope.Start();
+            try
+            {
+                using HttpMessage message = CreatePostParametersRequest(content, ContentType.ApplicationJson, context);
```
annelo-msft, line 219: Did we decide that our convention is that if we don't take a ContentType, we default ContentType to ApplicationJson?  Why is this better than always taking a ContentType and setting the default value of ContentType to ApplicationJson?

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
```diff
@@ -0,0 +1,91 @@
...
+    public class DpgInitialTest : TestServerLowLevelTestBase
+    {
+        [Test]
+        public Task DpgAddOptionalInput_NoParams_Test() => Test(async (host) =>
+        {
+            var result = await new ParamsClient(Key, host).HeadNoParamsAsync();
+            Assert.IsEmpty(result.Content.ToArray());
+            Assert.AreEqual(200, result.Status);
+            Assert.AreEqual(123, result.Headers.ContentLength);
+            var result2 = await new ParamsClient(Key, host).HeadNoParamsAsync(new Azure.RequestContext());
```
annelo-msft, line 22: nit: would it be possible to separate test cases with carriage returns?  It would make them reasier to read.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
```diff
@@ -0,0 +1,91 @@
...
+using AutoRest.TestServer.Tests.Infrastructure;
+using dpg_initial_LowLevel;
+using NUnit.Framework;
+
+namespace AutoRest.TestServer.Tests
+{
+    public class DpgInitialTest : TestServerLowLevelTestBase
+    {
+        [Test]
+        public Task DpgAddOptionalInput_NoParams_Test() => Test(async (host) =>
```
annelo-msft, line 16: What are the tests validating in dpg_initial.cs?  Is it establishing a baseline so we know the same code works on both versions?  Because in this method, we are not validating that you can add an optional parameter -- that doesn't happen until the v2 client in dpg_update1.cs.

My goal for this comment is just to make sure the test is named in a way that other developers can understand.  If we want to keep the name AddOptionalInput, let's add a comment stating where the addition happens so folks aren't confused when it's not here.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var result = await new ParamsClient(Key, host).HeadNoParamsAsync();
```
annelo-msft, line 18: Should we test this passing `default` for context as well?

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
```diff
@@ -0,0 +1,91 @@
...
+            var result = await new ParamsClient(Key, host).GetRequiredAsync("param");
+            var responseBody = JsonData.FromBytes(result.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody["message"]);
+            var result2 = await new ParamsClient(Key, host).GetRequiredAsync("param", default);
+            var responseBody2 = JsonData.FromBytes(result2.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody2["message"]);
+            var result3 = await new ParamsClient(Key, host).GetRequiredAsync("param", context: default);
+            var responseBody3 = JsonData.FromBytes(result3.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody3["message"]);
+            Assert.ThrowsAsync<ArgumentNullException>(async () => await new ParamsClient(Key, host).GetRequiredAsync(null));
```
annelo-msft, line 40: We shouldn't be validating that the method throws on a null query parameters, only if it is a path parameter.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var result = await new ParamsClient(Key, host).GetRequiredAsync("param");
+            var responseBody = JsonData.FromBytes(result.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody["message"]);
+            var result2 = await new ParamsClient(Key, host).GetRequiredAsync("param", default);
+            var responseBody2 = JsonData.FromBytes(result2.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody2["message"]);
+            var result3 = await new ParamsClient(Key, host).GetRequiredAsync("param", context: default);
+            var responseBody3 = JsonData.FromBytes(result3.Content.ToMemory());
+            Assert.AreEqual("An object was successfully returned", (string)responseBody3["message"]);
```
annelo-msft, line 39: Should we add a test case for passing a non-default context?

test/AutoRest.TestServerLowLevel.Tests/dpg_update1.cs
```diff
@@ -0,0 +1,171 @@
...
+            Assert.IsEmpty(result4.Content.ToArray());
+            Assert.AreEqual(200, result4.Status);
+            Assert.AreEqual(123, result4.Headers.ContentLength);
+        });
+
+        [Test]
+        public Task DpgAddOptionalInput() => Test(async (host) =>
+        {
+            // compatible with dpg_initial
+            var result = await new ParamsClient(Key, host).GetRequiredAsync("param");
```
annelo-msft, line 44: Is there a way to reuse the code from the dpg_initial to ensure that it is identical?  This feels a little error-prone if someone modified one test file and not the other.  It may not be a big deal to do it this way for tests, though.

annelo-msft (commented):
Added mostly questions, this is getting close!

annelo-msft:
Do we have any test cases covering these scenarios? https://gist.github.com/annelo-msft/827658b363941fdde699ac3925dd6ccf#2-changing-an-input-that-was-required-in-vi-to-optional-in-vi1

If not, we'll want to create an issue to do this in a separate PR.

test/AutoRest.TestServer.Tests/Infrastructure/TestServerLowLevelTestBase.cs
```diff
@@ -81,8 +81,13 @@ private async Task Test(string scenario, Func<Uri, Task> test, bool ignoreScenar
         private static string GetScenarioName()
         {
             var testName = TestContext.CurrentContext.Test.Name;
-            var indexOfUnderscore = testName.IndexOfAny(new[]{'_', '('});
-            return indexOfUnderscore == -1 ? testName : testName[..indexOfUnderscore];
+            var indexOfDelimiter = testName.LastIndexOf('_');
+            if (indexOfDelimiter == -1)
+            {
+                indexOfDelimiter = testName.IndexOf('(');
```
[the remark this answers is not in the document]
fengzhou-msft: Previously, test names do not contain `_`.  Now the testserver has `DPGAddOptionalInput_NoParams`, `DPGAddOptionalInput_RequiredOptionalParam`, `DPGAddOptionalInput_OptionalParam` as test names. Instead of passing `DPGAddOptionalInput` as the name to report coverage, now we need to pass the full name. 

As a workaround to not break previous tests, I made this change and we will name the test like `DpgAddOptionalInput_NoParams_Test` and then `DpgAddOptionalInput_NoParams` will be passed as the name for coverage to match the full test name.

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
fengzhou-msft: Yes, that's my understanding.

test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj
```diff
@@ -20,4 +20,4 @@
     <Compile Include="../../src/assets/**/*.cs" />
```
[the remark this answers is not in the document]
fengzhou-msft: This looks like to be caused by editor after opening the file, the last empty line is removed. I'll revert the change.

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
fengzhou-msft: This is based on current generated code and I agree that the backward compatible PostParametersAsync method here could cause confusion if a customer never used v1 and started with v2. We can generate the contenType parameter with default value if it only has one content type for LLC.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
fengzhou-msft: The test name has to start with DpgAddOptionalInput_NoParams which matches the name in testserver, otherwise we get a coverage report error, but we can modify the suffix after the last `_`, maybe use V1 and V2 correspondingly. I will add comments as well.

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
annelo-msft: Yes, I see, this was my misunderstanding.

The update1 API looks like this:

```csharp
public virtual Response PostParameters(RequestContent content, RequestContext context = null);
public virtual Response PostParameters(RequestContent content, ContentType contentType, RequestContext context = null);
```

The initial method should make RequestContext be required, since only one overload in a method group should have optional parameters.  Let's discuss in our design meeting today whether contentType should take a default value.

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
annelo-msft: Per our offline conversation, we will handle this as part of a separate PR so we don't block this PR on that work.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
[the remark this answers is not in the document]
annelo-msft: @AlexanderSher, does this make sense to you now?

test/AutoRest.TestServerLowLevel.Tests/dpg_update1.cs
fengzhou-msft: Refactored to share dpg_initial test code.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
fengzhou-msft: Added.

test/AutoRest.TestServerLowLevel.Tests/dpg_initial.cs
fengzhou-msft: Updated.

test/TestServerProjectsLowLevel/dpg_update1/Custom/ParamsClient.cs
fengzhou-msft: Per the discussion, I added `[EditorBrowsable(EditorBrowsableState.Never)]` to the overload methods for v1 that does not have contentType parameter.

annelo-msft (approved):

PR #1764, opened 2021-12-09
Implement new LRO pattern

AlexanderSher:
This PR adds `waitForCompletion` parameter to LRO methods in LLC code

### DIFF
--- a/test/AutoRest.TestServerLowLevel.Tests/lro.cs
+++ b/test/AutoRest.TestServerLowLevel.Tests/lro.cs
@@ -23,7 +23,7 @@ namespace AutoRest.TestServer.Tests
             var options = new AutoRestLongRunningOperationTestServiceClientOptions();
             options.AddPolicy(new CustomClientRequestIdPolicy(), HttpPipelinePosition.PerCall);
             var value = RequestContent.Create(new object());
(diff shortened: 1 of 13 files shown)


test/AutoRest.TestServerLowLevel.Tests/lro.cs
```diff
@@ -23,7 +23,7 @@ public class LroTest : TestServerLowLevelTestBase
             var options = new AutoRestLongRunningOperationTestServiceClientOptions();
             options.AddPolicy(new CustomClientRequestIdPolicy(), HttpPipelinePosition.PerCall);
             var value = RequestContent.Create(new object());
-            var operation = await new LROsCustomHeaderClient(Key, endpoint, options).PostAsyncRetrySucceededAsync(value);
+            var operation = await new LROsCustomHeaderClient(Key, endpoint, options).PostAsyncRetrySucceededAsync(false, value);
```
annelo-msft, line 26: Would it make sense to add some cases where we pass true to an LRO wait method?

annelo-msft (approved):

test/AutoRest.TestServerLowLevel.Tests/lro.cs
AlexanderSher: All lro tests now cover both cases

PR #1778, opened 2021-12-14
Change the way we get plural resource names

ArcturusZhang:
# Description

1. Change the way we get the plural form of a resource name. Now we use `LastWordToPlural()` instead which will first split the string into words, and change the last word to its plural form, then concat them back. This will give us more accurate result, such as, we will get the plural form of `KeyInformation` to be `KeyInformation` instead of `KeyInformations` since `information` is uncountable.
2. Found a [bug](https://github.com/Humanizr/Humanizer/issues/1154) in `Humanizer` that if I call `"S".Singularize()`, it throws exceptions. Unfortunately our test project has this case. I have to make a work around.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -278,34 +280,56 @@ public static bool IsCSharpKeyword(string? name)
...
+        public static string ResourceNameToPlural(this string resourceName)
+        {
+            var pluralResourceName = resourceName.LastWordToPlural(false);
+            var singularResourceName = resourceName.LastWordToSingular(false);
+            return pluralResourceName != singularResourceName ?
+                pluralResourceName :
+                $"All{pluralResourceName}";
+        }
+
+        public static string LastWordToPlural(this string single, bool inputIsKnownToBeSingular = true)
```
AlexanderSher, line 304: Please add tests to these methods since they are public now.

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -278,34 +280,56 @@ public static bool IsCSharpKeyword(string? name)
...
+        /// <param name="single"></param>
+        /// <param name="inputIsKnownToBeSingular"></param>
+        /// <returns></returns>
         public static string ToPlural(this string single, bool inputIsKnownToBeSingular = true)
         {
             return single.Pluralize(inputIsKnownToBeSingular);
         }
 
-        private static string LastWordToPlural(this string single, bool inputIsKnownToBeSingular = true)
+        public static string ResourceNameToPlural(this string resourceName)
```
AlexanderSher, line 295: If this method is used only in mgmt, please move it into mgmt-specific type

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
ArcturusZhang: Test cases added

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
ArcturusZhang: I move this function to the StringExtensions of Mgmt namespace

samples/Azure.ResourceManager.Sample/Generated/Extensions/ArmClientExtensions.cs
```diff
@@ -78,25 +78,25 @@ public static VirtualMachineExtensionImage GetVirtualMachineExtensionImage(this
...
-        #region VirtualMachineExtension
-        /// <summary> Gets an object representing a VirtualMachineExtension along with the instance operations that can be performed on it but with no data. </summary>
+        #region VirtualMachineExtensions
+        /// <summary> Gets an object representing a VirtualMachineExtensions along with the instance operations that can be performed on it but with no data. </summary>
         /// <param name="armClient"> The <see cref="ArmClient" /> instance the method will execute against. </param>
         /// <param name="id"> The resource ID of the resource to get. </param>
-        /// <returns> Returns a <see cref="VirtualMachineExtension" /> object. </returns>
-        public static VirtualMachineExtension GetVirtualMachineExtension(this ArmClient armClient, ResourceIdentifier id)
+        /// <returns> Returns a <see cref="VirtualMachineExtensions" /> object. </returns>
+        public static VirtualMachineExtensions GetVirtualMachineExtensions(this ArmClient armClient, ResourceIdentifier id)
```
fengzhou-msft, line 86: It looks like the resource is named `VirtualMachineExtensions`. Shall we rename it to ``VirtualMachineExtension` with directive?

ArcturusZhang: Is this the case? Let me take a look...

ArcturusZhang: This is caused by a bug in `LastWordToSingular`. I have fixed it and added a test case accordingly

AlexanderSher:
LGTM

fengzhou-msft (approved):

PR #1781, opened 2021-12-14
Fix #1765: Ensure required parameter ordering

### LINKED ISSUE #1765: [LLC] Ensure required parameter ordering
Our current parameter ordering for LLC is as follows:

1. (waitForCompletion)
2. (required parameters)
3. (RequestContent)
4. (ContentType)
5. (optional parameters)
6. ([Match|Request]Conditions)
7. RequestContext

In order to align with management plane, we will ensure that required parameters are ordered as followed:
 - First, required path parameters, sorted to match the order of the URI
 - Second, remaining required query params will come in the order they are defined in swagger after the path parameters

Management plane will move required query parameters before the body parameters in their parameter list.  The resulting ordering is as follows:

1. (waitForCompletion)
2. (required path and query parameters)
    a. Required path parameters, sorted to match the order of the URI
    b. Remaining required query params will come in the order they are defined in swagger after the path parameters
3. (RequestContent)/ Body parameters
4. (ContentType) – not used by management plane
5. (optional path and query parameters)
    a. Required path parameters, sorted to match the order of the URI
    b. Remaining required query params will come in the order they are defined in swagger after the path parameters
6. ([Match|Request]Conditions)
7. RequestContext

AlexanderSher:


test/TestProjects/BodyAndPath-LowLevel/BodyAndPath-LowLevel.json
```diff
@@ -172,6 +172,54 @@
             }
           }
         }
+      },
+      "/{item3}/item4/{item2}/item1": {
```
annelo-msft, line 176: I'm a little confused by this test -- would it be possible add a description field to state what it is testing and what the expected outcome is?

For example, it looks like `item4` is listed as a query parameter, but given its position in the path, isn't it a path parameter?

test/TestProjects/RequestContextAllOptional-LowLevel/Generated/RequestContextAllOptionalClient.cs
```diff
@@ -52,19 +52,19 @@ public RequestContextAllOptionalClient(AzureKeyCredential credential, Uri endpoi
...
         /// <summary> No RequestBody and ResponseBody. </summary>
         /// <param name="id"> Query parameter Id. </param>
-        /// <param name="top"> Query parameter top. </param>
         /// <param name="skip"> Query parameter skip. </param>
+        /// <param name="top"> Query parameter top. </param>
         /// <param name="status"> Query parameter status. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
 #pragma warning disable AZC0002
-        public virtual async Task<Response> NoRequestBodyResponseBodyAsync(int id, int? top = null, int skip = 12, string status = "start", RequestContext context = null)
+        public virtual async Task<Response> NoRequestBodyResponseBodyAsync(int id, int skip = 12, int? top = null, string status = "start", RequestContext context = null)
```
annelo-msft, line 60: Why do these two change positions?  Aren't they both optional?

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
```diff
@@ -1077,26 +1077,26 @@ IEnumerable<Page<BinaryData>> CreateEnumerable(string nextLink, int? pageSizeHin
         /// 
         /// </remarks>
 #pragma warning disable AZC0002
-        public virtual AsyncPageable<BinaryData> GetMultiplePagesFragmentNextLinkAsync(string apiVersion, string tenant, RequestContext context = null)
+        public virtual AsyncPageable<BinaryData> GetMultiplePagesFragmentNextLinkAsync(string tenant, string apiVersion, RequestContext context = null)
```
annelo-msft, line 1080: This feels dangerous to me - existing code will compile but pass the wrong values for `tenant` and `apiVersion`.  Which of our new rules is dicating this?

test/TestProjects/BodyAndPath-LowLevel/BodyAndPath-LowLevel.json
AlexanderSher: Every parameter has a description where it should go. I can add an explanation why it should go there.

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
AlexanderSher: `tenant` is defined in path, `apiVersion` is defined in query.

test/TestProjects/RequestContextAllOptional-LowLevel/Generated/RequestContextAllOptionalClient.cs
AlexanderSher: [Skip](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/RequestContextAllOptional-LowLevel/RequestContextAllOptional.json#L38) is required while [top](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/RequestContextAllOptional-LowLevel/RequestContextAllOptional.json#L23) is not.

It is a good point however that both have default values. 

We can either strengthen the order here, e.g.:
1. Required without default values
2. Required with default values
3. Optional with default values

or use default value instead of `required` as criteria.

annelo-msft: Yes, from the point of view of a developer using our clients, if we provide a default value for something, it's optional to them.  Let's use whether or not it has a default value to indicate whether it's "required" or "optional."

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
annelo-msft: What is our plan to make sure all the relevant code is updated correctly in our generated libraries, since we won't have compiler support for this?

test/TestProjects/RequestContextAllOptional-LowLevel/Generated/RequestContextAllOptionalClient.cs
annelo-msft: > Skip is required while top is not.

This sounds like it's not a standard swagger then - per @tg-msft, it sounds like in real life, top (i.e. specifying the page size) might be required and skip not.  Given that, I'm ok with this change.

test/TestProjects/BodyAndPath-LowLevel/BodyAndPath-LowLevel.json
AlexanderSher: Done

test/AutoRest.TestServer.Tests/TypeAsserts.cs
```diff
@@ -69,6 +69,53 @@ public static void TypeOnlyDeclaresThesePublicMethods(Type type, params string[]
             Assert.Fail($"Type \"{type}\" doesn't have public methods {missingMethods} and has public methods {extraMethods} that aren't expected.");
         }
 
+        public static void HasPublicInstanceMethod(Type type, string methodName, Parameter[] parameters)
```
AlexanderSher, line 72: @annelo-msft , can you please check the messages in this assertion method?

annelo-msft: These look good to me!

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
annelo-msft: I see.  So this shouldn't have any effect on existing public APIs?  If not, then I'm not concerned.

annelo-msft (approved):

PR #1783, opened 2021-12-15
Generate ArmResourceExtensions

### LINKED ISSUE #1757: Support generating ArmResource Extension for ResourceManager
If a resource can be created under any resource, we should add a Get[Resources] extension method to ArmResource and remove the Get[Resources] method in other extension classes (tenant/subscription/reosurce group/management group).

fengzhou-msft:
# Description
Resolve #1757

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/OperationExtensions.cs
```diff
@@ -152,7 +152,7 @@ private static int RankRequestPath(RequestPath requestPath)
 
         public static string GetHttpPath(this Operation operation)
         {
-            return operation.GetHttpRequest()?.Path ??
+            return operation.GetHttpRequest()?.Path.TrimEnd('/') ??
```
ArcturusZhang, line 155:
```suggestion
            return operation.GetHttpRequest()?.Path?.TrimEnd('/') ??
```

fengzhou-msft: Path is defined as `public string Path { get; set; }`, so if `GetHttpRequest()` is not null, `Path` won't/can't be null. The null-conditional operators are [short-circuiting](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-). If `GetHttpRequest()` is null, the expression stops executing, so I guess we don't need the extra `?` here.

test/TestProjects/MgmtScopeResource/Generated/Extensions/ArmResourceExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace MgmtScopeResource
+{
+    /// <summary> A class to add extension methods to ArmResource. </summary>
+    public static partial class ArmResourceExtensions
+    {
+        #region PolicyAssignment
+        /// <summary> Gets an object representing a PolicyAssignmentCollection along with the instance operations that can be performed on it. </summary>
+        /// <param name="armResource"> The <see cref="ArmResource" /> instance the method will execute against. </param>
+        /// <returns> Returns a <see cref="PolicyAssignmentCollection" /> object. </returns>
+        public static PolicyAssignmentCollection GetPolicyAssignments(this ArmResource armResource)
```
AlexanderSher, line 19: Do we control somehow that method names in different extensions don't clash with each other?

Yao725 (approved):
LGTM

ArthurMa1978 (approved):

test/TestProjects/MgmtScopeResource/Generated/Extensions/ArmResourceExtensions.cs
fengzhou-msft: If you mean the method name between the `ArmResourceExtensions` and other extensions, the generator would remove the method from other extensions if it is already in ArmResourceExtensions. For other extensions, we prefer to give them the same name, so user experience is consistent. For instance, we have `subscription.GetAll()` and `managementGroup.GetAll()` instead of the original `subscription.GetAllBySubscription()` and `managementGroup.GetAllByManagementGroup()`.
 
If you mean method names for the same extension across different RPs, I think it equals to the question that do we control resource name conflicts across different RPs as the method names are just derived from resource names. @ArcturusZhang did some work to avoid resource name conflicts in the same RP but I don't think we have the checks for all RPs. One practice in API review is to avoid single word as class names which are likely to cause name conflicts across different RPs.

ArcturusZhang (approved):
LGTM, but in the future we still need some changes if a resource could be extension of some specific resource type from another RP, we might need to validate the resource type of the current ArmResource to see if we are calling it on the correct resource.

Issue #1642, opened 2021-10-26
Wrap non-pageable list method that returns a collection into pageable


ArcturusZhang:
This benefits the breaking change - previously adding a pageable extension to an operation will cause a breaking change in the SDK. This also benefits the user experience, after doing this, our customer will have consistency experience no matter the List method is pageable or not.
Before this change, to iterate something using a pageable list method, they need to do this:
```
await foreach (var something in collection.GetAllAsync())
```
to iterate something using a non-pageable list method, they need to do this:
```
foreach (var something in await collection.GetAllAsync())
```

PR #1791, opened 2021-12-17
Wrap non-pageable list functions into pageable

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1642

# Description

This PR wraps the non-pageable list result into pageable result to produce more consistent APIs in the mgmt plane SDKs.
This PR also makes some refactor to some existing code.

# Checklist

# Ready to Land?


samples/Azure.ResourceManager.Sample/Generated/VirtualMachineScaleSetVirtualMachineExtensionCollection.cs
```diff
@@ -293,20 +294,25 @@ public async virtual Task<Response<bool>> CheckIfExistsAsync(string vmExtensionN
...
+            Page<VirtualMachineScaleSetVirtualMachineExtension> FirstPageFunc(int? pageSizeHint)
             {
-                var response = _virtualMachineScaleSetVMExtensionsRestClient.List(Id.SubscriptionId, Id.ResourceGroupName, Id.Parent.Name, Id.Name, expand, cancellationToken);
-                return Response.FromValue(response.Value.Value.Select(value => new VirtualMachineScaleSetVirtualMachineExtension(Parent, value)).ToArray() as IReadOnlyList<VirtualMachineScaleSetVirtualMachineExtension>, response.GetRawResponse());
-            }
-            catch (Exception e)
-            {
-                scope.Failed(e);
-                throw;
+                using var scope = _clientDiagnostics.CreateScope("VirtualMachineScaleSetVirtualMachineExtensionCollection.GetAll");
```
AlexanderSher, line 302: This code is different from the previous one semantically. Previously, user would've got fully initialized collection after method execution. Now, `FirstPageFunc` will be called only during enumeration. Is that expected?

ArcturusZhang: Well, this is not expected. Do you have a suggestion how we could restore previous behavior? Does ``PageableHelpers` have related APIs to do this? Like I first call the API and get the array then I pass it into some function and it gives me a `Pageable` of something?

AlexanderSher: Sure. You can use `FromPages`:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pageable.cs#L97
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/AsyncPageable.cs#L118
e.g.:
```cs
using var scope = _clientDiagnostics.CreateScope("VirtualMachineScaleSetVirtualMachineExtensionCollection.GetAll");
scope.Start();
try
{
    var response = _virtualMachineScaleSetVMExtensionsRestClient.List(Id.SubscriptionId, Id.ResourceGroupName, Id.Parent.Name, Id.Name, expand, cancellationToken: cancellationToken);
    var page = Page.FromValues(response.Value.Value.Select(value => new VirtualMachineScaleSetVirtualMachineExtension(Parent, value)), null, response.GetRawResponse());
    return Pageable<T>.FromPages(new[]{page});
}
catch (Exception e)
{
    scope.Failed(e);
    throw;
}
```

ArcturusZhang: Thank you! Let me switch to this and confirm with Michael

ArcturusZhang: I suppose this should be expected. We could merge with this behaviour and if this is not good enough, we could revisit this piece of logic. Changing this is not a breaking change for the SDK API.

PR #1794, opened 2021-12-20
feat(mgmt): add `BuildId` for resource class

### LINKED ISSUE #1612: [feature] generate `BuildID` method for each resource type class
# Description

We've spot a typical type of use cases that users want to operate the resource by the context path values (e.g. `SubscriptionID`, `ResourceGroupName`, etc).

It's a little bit challenging to let users compose the corresponding `ResourceIdentifier`. For example, users need to know the `resourceType`.

So we can provide utility method in each resource type class to help build it, like below:
```csharp
PublicIpAddress.BuildID(subscriptionId, resourceGroupName, name);
```

archerzz:
# Description
One feed back from internal customer review is that they want a quick way to compose the `ResourceIdentifier` instances which are shortcuts for several scenario.

This commit will:
- add static `BuildId` methods under each resource class
- each `BuildId` method will accept a series of context path parameters, and generate the corresponding `ResourceIdentifier`

resolve #1612

# Checklist

# Ready to Land?


samples/Azure.Management.Storage/Generated/BlobContainer.cs
```diff
@@ -22,6 +22,12 @@ namespace Azure.Management.Storage
     /// <summary> A Class representing a BlobContainer along with the instance operations that can be performed on it. </summary>
     public partial class BlobContainer : ArmResource
     {
+        /// <summary> Generate the resource identifier of a <see cref="BlobContainer"/> instance. </summary>
+        public static ResourceIdentifier BuildId(string subscriptionId, string resourceGroupName, string accountName, string containerName)
```
m-nash, line 26: I wonder if we should call this CreateResourceIdentifier

m-nash (changes requested):
Before we check this can we make these changes in Azure.ResourceManager and add a section to the readme explaining how to create Ids for resources using this method?

samples/Azure.Management.Storage/Generated/BlobContainer.cs
archerzz: @m-nash I'm fine with the change. `BuildId` is from original discussion thread. I was also wondering if we should use another name.

m-nash (approved):

src/AutoRest.CSharp/Mgmt/Generation/ResourceWriter.cs
```diff
@@ -79,6 +80,30 @@ public override void Write()
...
+        }
+
+        private void WriteBuildIdMethods()
+        {
+            foreach (var requestPath in _resource.RequestPaths)
+            {
+                _writer.Line();
+                _writer.WriteXmlDocumentationSummary($"Generate the resource identifier of a <see cref=\"{TypeOfThis}\"/> instance.");
+                var parameterList = string.Join(", ", requestPath.Where(segment => !segment.IsConstant).Select(segment => $"string {segment.ReferenceName}"));
+                using (_writer.Scope($"public static ResourceIdentifier BuildId({parameterList})"))
```
ArcturusZhang, line 95:
```suggestion
                using (_writer.Scope($"public static {typeof(ResourceIdentifier)} BuildId({parameterList})"))
```

PR #1796, opened 2021-12-23
Changes to handle version override in client options

m-nash:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Generation/ResourceWriter.cs
```diff
@@ -204,12 +204,13 @@ protected virtual void WriteCtors()
...
-                var subscriptionParamString = client.Parameters.Any(p => p.Name.Equals("subscriptionId")) ? ", Id.SubscriptionId" : string.Empty;
-                _writer.Line($"{GetRestClientVariableName(client)} = new {client.Type.Name}({ClientDiagnosticsField}, {PipelineProperty}, {ClientOptionsProperty}{subscriptionParamString}, {BaseUriField});");
-            }
+            Func<MgmtRestClient, string> getSubId = (restClient) => { return restClient.Parameters.Any(p => p.Name.Equals("subscriptionId")) ? ", Id.SubscriptionId" : string.Empty; };
+            WriteRestClientConstructionForResource(_resource, This.RestClients, getSubId, ClientDiagnosticsField, ClientOptionsProperty, PipelineProperty, BaseUriField, "new ", false);
+        }
+
+        private string ConvertToVariableName(string str)
+        {
+            return char.ToLower(str[0]) + str.Substring(1);
```
ArcturusZhang, line 213: Looks like this function is not used anywhere, could we remove it?
And we have similar one in StringExtensions `FirstLetterToUpperCase`

src/AutoRest.CSharp/Mgmt/Generation/ResourceVersionWriter.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            using (writer.Namespace($"{_resource.Type.Namespace}.Models"))
+            {
+                writer.WriteXmlDocumentationSummary($" The versions of {_resource.ResourceName} this client is compatible with. ");
+                using (writer.Scope($"public readonly partial struct {VersionClassName} : IEquatable<{VersionClassName}>"))
+                {
+                    writer.Line($"private readonly string _value;");
+
+                    writer.Line();
+                    var versions = new List<string>() { "2019-10-01" }; //TODO: Get this from somewhere
```
ArcturusZhang, line 38: What is this version for?

src/AutoRest.CSharp/Mgmt/Output/MgmtExtensions.cs
```diff
@@ -41,7 +41,20 @@ private IEnumerable<MgmtClientOperation> EnsureClientOperations()
...
+        private Resource? GetResourceFromResourceType(ResourceTypeSegment resourceType)
+        {
+            var candidates = _context.Library.ArmResources.Where(resource => resource.ResourceType == resourceType);
+            if (candidates.Count() == 0)
+                return null;
+
+            if (candidates.Count() == 1)
+                return candidates.First();
+
+            throw new InvalidOperationException($"Found more than 1 candidate for {resourceType}, results were ({string.Join(',', candidates.Select(r => r.ResourceName))})");
```
ArcturusZhang, line 57: I have some concerns about this implementation here.
We have the case that multiple resources have the same resource type - see the MgmtExtensionResource for example.
This is a case for the same resource type staying under the different scopes (namely subscription, tenant and management group). In this test project, we do not have an operation in extension for them, therefore we did not get an exception during that project. And when the exception happens, we cannot resolve it.

Do we have a way to first escape this finding of candidates from a proper configuration? (a new or an existing one)

src/AutoRest.CSharp/Mgmt/Generation/ResourceVersionWriter.cs
ArthurMa1978: Dummy one?

src/AutoRest.CSharp/Mgmt/Generation/ResourceWriter.cs
ArthurMa1978: `FirstLetterToLowerCase` ~

m-nash: removed as this wasn't needed anymore

src/AutoRest.CSharp/Mgmt/Generation/ResourceVersionWriter.cs
m-nash: Removing as this isn't needed anymore

src/AutoRest.CSharp/Mgmt/Output/MgmtExtensions.cs
m-nash: I believe the check and throw we do will handle the case you describe.  Right now it has never happened but if it does we will need to adjust our implementation to handle it.  As long as this doesn't silently fail I am fine with it.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -185,6 +185,10 @@ public static void WriteParameter(this CodeWriter writer, Parameter clientParame
                     writer.WriteConstant(clientParameter.DefaultValue.Value);
                 }
             }
+            else if (!clientParameter.IsRequired)
+            {
+                writer.Append($" = null");
```
AlexanderSher, line 190: I think we can use `$" = default"` here and merge this condition with the one below

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -257,7 +261,7 @@ public static void WriteVariableAssignmentWithNullCheck(this CodeWriter writer,
         {
             // Temporary check to minimize amount of changes in existing generated code
             var assignToSelf = parameter.Name == variableName;
-            if (parameter.DefaultValue != null && !TypeFactory.CanBeInitializedInline(parameter.Type, parameter.DefaultValue))
+            if (parameter.DefaultValue != null && (!parameter.UseDefaultValueInCtorParam || !TypeFactory.CanBeInitializedInline(parameter.Type, parameter.DefaultValue)))
```
AlexanderSher, line 264: Can we change the name to something like "SetDefaultValueInBody"? In that case, property will have meaning in any signature, not just constructor

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -269,7 +273,7 @@ public static void WriteVariableAssignmentWithNullCheck(this CodeWriter writer,
                 }
 
                 var defaultValue = parameter.DefaultValue.Value;
-                if (defaultValue.IsNewInstanceSentinel || TypeFactory.IsExtendableEnum(parameter.Type) || parameter.DefaultValue.Value.Type.Equals(parameter.Type))
+                if (defaultValue.IsNewInstanceSentinel || TypeFactory.IsExtendableEnum(parameter.Type) || parameter.DefaultValue.Value.Type.Equals(parameter.Type) || parameter.Type.Equals(typeof(string)))
```
AlexanderSher, line 276: Why `parameter.DefaultValue.Value.Type.Equals(parameter.Type)` didn't work here?

src/AutoRest.CSharp/Mgmt/Models/MgmtRestClientBuilder.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public static IEnumerable<RequestParameter> GetMgmtParametersFromOperations(ICollection<Operation> operations) =>
+            operations
+                .SelectMany(op => op.Parameters.Concat(op.Requests.SelectMany(r => r.Parameters)))
+                .Where(p => p.Implementation == ImplementationLocation.Client)
+                .Distinct(new ParameterCompareer());
+
+        public override Parameter BuildConstructorParameter(RequestParameter requestParameter)
+        {
+            bool isApiVersion = requestParameter.Origin == "modelerfour:synthesized/api-version";
```
AlexanderSher, line 52: This maybe not optimal performance-wise, but since the only difference here is that api version parameter has `UseDefaultValueInCtorParam = false`, we can reduce this method to:
```cs
public override Parameter BuildConstructorParameter(RequestParameter requestParameter)
{
    var parameter = base.BuildConstructorParameter(requestParameter);
    return parameter.IsApiVersionParameter ? parameter with { UseDefaultValueInCtorParam = false } : parameter;

}
```

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
m-nash: (ignoreNullable || IsNullable == other.IsNullable)

param.DefaultValue.Value.Type.IsNullable == false
parameter.Type.IsNullable == true

therefore equals fails

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
m-nash: then the property doesn't make sense in the other place its used where we skip setting the actual default in the ctor param.

I can make two properties if you want one with each name and use them in each place, but the usage here should always be the opposite of the usage in the other place

AlexanderSher: Why? It perfectly makes sense to me:
```cs
if (clientParameter.DefaultValue != null && !clientParameter.SetDefaultValueInBody) 
{
```

`WriteParameter` writes parameter **in the signature** of the method. If parameter has default value, but `SetDefaultValueInBody: true`, default value must be set **in the body**. Hence, just `= default` in the signature.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
AlexanderSher: Looks like a bug to me. If parameter is nullable but default value is not, we should still be able to assign it. Maybe something like:
```cs
if (defaultValue.IsNewInstanceSentinel || TypeFactory.IsExtendableEnum(parameter.Type) || defaultValue.Type.Equals(parameter.Type) || defaultValue.Type.Equals(parameter.Type.WithNullable(false))
```

ArcturusZhang (commented):
LGTM

ArthurMa1978 (approved):

PR #1798, opened 2021-12-23
Exception for non supported header

chunyu3:
# Description

# Checklist

# Ready to Land?


test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
```diff
@@ -168,6 +368,97 @@ internal HttpMessage CreateCollapseGetRequest(RequestConditions requestCondition
...
+            var message = _pipeline.CreateMessage(context);
+            var request = message.Request;
+            request.Method = RequestMethod.Get;
+            var uri = new RawRequestUriBuilder();
+            uri.Reset(_endpoint);
+            uri.AppendPath("/RequestConditionCollapse/missIfMatch", false);
+            request.Uri = uri;
+            if (requestConditions != null)
+            {
+                if (requestConditions.IfMatch != null) throw new NotSupportedException($"IfMatch head is not supported here");
```
AlexanderSher, line 400: Supported values are the feature of parameter, not header. Hence this check should happen in `MissIfMatchGet`/`MissIfMatchGetAsync` methods, and exception type will be `ArgumentException`. Ideally, I'd like to have a method in `Azure.Core.Argument`, similar to `Argument.AssertNotNullOrEmpty`, something like:
```cs
public static void AssertHasOnlySupportedHeaders(RequestConditions? requestConditions, string paramName, params string[] supportedHeaders)
{
    if (requestConditions != null)
    {
        AssertHasNoUnsupportedHeader(requestConditions.IfMatch, paramName, supportedHeaders, HttpHeader.Names.IfMatch);
        AssertHasNoUnsupportedHeader(requestConditions.IfNoneMatch, paramName, supportedHeaders, HttpHeader.Names.IfNoneMatch);
        AssertHasNoUnsupportedHeader(requestConditions.IfModifiedSince, paramName, supportedHeaders, HttpHeader.Names.IfModifiedSince);
        AssertHasNoUnsupportedHeader(requestConditions.IfUnmodifiedSince, paramName, supportedHeaders, HttpHeader.Names.IfUnmodifiedSince);
    }
}

private static void AssertHasNoUnsupportedHeader<T>(T? value, string paramName, string[] supportedHeaders, string unsupportedHeader) where T : struct
{
    if (value.HasValue && !supportedHeaders.Contains(unsupportedHeader))
    {
        throw new ArgumentException($"{unsupportedHeader} header is not supported. Supported headers are: {string.Join(", ", supportedHeaders)}.", paramName);
    }
}
```
```cs
public virtual async Task<Response> MissIfMatchGetAsync(RequestConditions requestConditions = null, RequestContext context = null) 
{
    Argument.AssertHasOnlySupportedHeaders(requestConditions, nameof(requestConditions), "If-None-Match", "If-Modified-Since");
```

This approach would have several benefits:
- testability
- less generated code
- no need to regenerate old code if another property is added to `RequestConditions`

AlexanderSher: The question however is how to add validation information to the `AutoRest.CSharp.Output.Models.Shared.Parameter`, since we don't have any support for custom validation. Let me think about it.

test/TestProjects/CollapseRequestCondition-LowLevel/CollapseRequestCondition-LowLevel.json
```diff
@@ -343,18 +450,46 @@
               "name": "modified-access-conditions"
             },
             "description": "Specify an ETag value to operate only on blobs with a matching value."
-          }  
+          },
+	        {
```
AlexanderSher, line 454: Formatting

chunyu3: resolved

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
chunyu3: Yes, currently we have not put parameter validation in the rest Method, and currently we have put some validation in Create****Request, such as check if a header is NULL when try to add the Head to HTTP Request. We can put the validation either location, but we need to consistent.

src/assets/Azure.Core.Shared/Argument.cs
```diff
@@ -212,5 +212,24 @@ public static string CheckNotNullOrEmpty(string value, string name)
             AssertNotNullOrEmpty(value, name);
             return value;
         }
+
+        public static void AssertHasOnlySupportedHeaders(RequestConditions? requestConditions, string paramName, string[] supportedHeaders)
```
annelo-msft, line 216: I believe files under Azure.Core.Shared are copied in from the azure-sdk-for-net repo, so this change may need to be made there.

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/NonCollapseClient.cs
```diff
@@ -88,17 +88,18 @@ public virtual Response CollapsePut(RequestContent content, ETag? ifMatch = null
             }
         }
 
-        /// <param name="requestConditions"> The content to send as the request conditions of the request. </param>
+        /// <param name="content"> The content to send as the body of the request. </param>
+        /// <param name="ifNoneMatch"> Specify an ETag value to operate only on blobs without a matching value. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
 #pragma warning disable AZC0002
-        public virtual async Task<Response> CollapseGetAsync(RequestConditions requestConditions = null, RequestContext context = null)
+        public virtual async Task<Response> IfNonMatchPutAsync(RequestContent content, ETag? ifNoneMatch = null, RequestContext context = null)
```
annelo-msft, line 95: nit: spelling - I called this out in the .json file as well.

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
```diff
@@ -164,6 +364,97 @@ internal HttpMessage CreateCollapseGetRequest(RequestConditions requestCondition
...
+            var message = _pipeline.CreateMessage(context);
+            var request = message.Request;
+            request.Method = RequestMethod.Get;
+            var uri = new RawRequestUriBuilder();
+            uri.Reset(_endpoint);
+            uri.AppendPath("/RequestConditionCollapse/missIfModifiedSince", false);
+            request.Uri = uri;
+            if (requestConditions != null)
+            {
+                if (requestConditions.IfModifiedSince != null) throw new NotSupportedException($"IfModifiedSince head is not supported here");
```
annelo-msft, line 414: Parameter validations should happen inside the client call, before other actions are executed.  (For example, we don't want to start a Diagnostic Scope, if we know the method is going to fail with an ArgumentException.)

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
annelo-msft: I'd like to simplify what goes into Argument.cs so that the generated code looks more like this:

```
public virtual Response DoThing(RequestConditions conditions, RequestContext context = default)
{
  Argument.AssertNull(requestConditions.IfMatch, nameof(requestConditions, "Service does not support the If-Match header for this operation.");
  // the rest of the service call
}

**Review by annelo-msft — COMMENTED** [review]
I left suggested changes.  I'd also like to get @ShivangiReja's review on the code under /src after my comments are addressed.

File: src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs [hunk 6]
```diff
@@ -15,6 +15,7 @@
 using Azure.Core;
 using AutoRest.CSharp.Utilities;
 using Azure;
+using AutoRest.CSharp.Common.Output.Models;
```
ShivangiReja, line 18: Nit: Why is this needed? nothing changed in this file.

test/TestProjects/CollapseRequestCondition-LowLevel/CollapseRequestCondition-LowLevel.json
```diff
@@ -402,4 +565,4 @@
       }
     }
   }
-}
+}
```
ShivangiReja, line 568: Nit: Add new line at the end of the file

test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj
```diff
@@ -1,14 +1,13 @@
...
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>
     <IsPackable>false</IsPackable>
     <Nullable>annotations</Nullable>
-    <NoWarn>SA1649;SA1633</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.22.0" />
+    <PackageReference Include="Azure.Core" Version="1.22.0-alpha.20211207.1" />
```
ShivangiReja, line 10: Why are we using different version of Azure.Core only for tests?

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/MatchConditionCollapseClient.cs
```diff
@@ -267,23 +267,6 @@ internal HttpMessage CreateCollapseGetRequest(MatchConditions matchConditions, R
...
-            uri.AppendPath("/MatchConditionCollapse/multi", false);
-            request.Uri = uri;
-            if (requestConditions != null)
-            {
-                request.Headers.Add(requestConditions, "R");
-            }
-            message.ResponseClassifier = ResponseClassifier200.Instance;
-            return message;
-        }
-
```
ShivangiReja, line 286: Why is this method got removed? I think you will get compile time error because we are calling this method from  [MulticollapseGetAsync()](https://github.com/Azure/autorest.csharp/blob/cf0fce46867d697a0e6fa91ea9525b0c370bfbad/test/TestProjects/CollapseRequestCondition-LowLevel/Generated/MatchConditionCollapseClient.cs#L183) and [MulticollapseGet()](https://github.com/Azure/autorest.csharp/blob/cf0fce46867d697a0e6fa91ea9525b0c370bfbad/test/TestProjects/CollapseRequestCondition-LowLevel/Generated/MatchConditionCollapseClient.cs#L203).

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -886,7 +888,7 @@ private void AddRequestParameter(string name, RequestParameter requestParameter,
                 }
             }
         }
-
+        /*
         [Flags]
         private enum RequestConditionHeaders
```
ShivangiReja, line 893: We are adding `RequestConditionHeaders.cs` for enums, can we remove this commented code?

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
```diff
@@ -128,6 +128,206 @@ public virtual Response CollapseGet(RequestConditions requestConditions = null,
             }
         }
 
+        /// <param name="requestConditions"> The content to send as the request conditions of the request. </param>
+        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+#pragma warning disable AZC0002
+        public virtual async Task<Response> MissIfNonMatchGetAsync(RequestConditions requestConditions = null, RequestContext context = null)
+#pragma warning restore AZC0002
+        {
+            using var scope = _clientDiagnostics.CreateScope("RequestConditionCollapseClient.MissIfNonMatchGet");
```
ShivangiReja, line 137: Shouldn't it generate `_clientDiagnostics` -> `ClientDiagnostics` ?

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
chunyu3: hi @annelo-msft 
When there are more than one request condition need to check, the Argument.AssertNull will be call more than once.
e..g 

public virtual Response DoThing(RequestConditions conditions, RequestContext context = default)
{
    Argument.AssertNull(requestConditions.IfMatch, nameof(requestConditions), "Service does not support the If-Match header for this operation.");
  Argument.AssertNull(requestConditions.IfNoneMatch, nameof(requestConditions), "Service does not support the If-None-Match header for this operation.");
    // the rest of the service call
}

The worst case, the Argument.AssertNull will call three times.
Is it acceptable?

src/assets/Azure.Core.Shared/Argument.cs
chunyu3: Yes, we need to update in azure-sdk-for-net repo, after we merge those changes into azure-sdk-for-net repo, we need first file an PR in autorest.csharp to update the Argument.cs which is copied from azure-sdk-for-net. Then file another PR to include our business logic. We can merge those two PR as one.

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
annelo-msft: I think this is a very edge case where we have these asserts in the first place.  I think simplicity of Azure.Core APIs and readability of the code should be the driving factor, and we can revisit the perf implications if we find we have a problem in the future.  This would give us a strong motivating reason to write more complex APIs into the Assert class.

We will not add properties to RequestConditions often -- this type represents [HTTP conditional requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests), and has [four of the five headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests#conditional_headers) represented.  We haven't yet found a case in Azure that needed the fifth - we would only add it if we did, and then there would be no further headers to add.

And @AlexanderSher, I don't understand the argument about having to regenerate code -- don't we always regenerate code when we make changes to the generator?

chunyu3: When we add a new field into RequestConditions type, then we need to also check if the new field is set or not. In that way, we need to regenerated the existing SDK client libraries to reflect it. 
But if we use AssertHasOnlySupportedHeaders, we just need to update the method in Azure.Core, the existing SDK client libraries does not updated.

chunyu3: Yes, RequestConditions type is for HTTP conditional requests, so it is rare or maybe not to be updated (such as add or remove an field), so we may not need to consider this scenario current. Right? @annelo-msft

annelo-msft: We'll never remove a field, because that would be a breaking change, and there is only one field left to add, so we would only ever regenerate once.  So, yes, I wouldn't worry about the possible addition - we regenerate the library code fairly often anyway.

chunyu3: I will implement it this way.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
chunyu3: removed

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/MatchConditionCollapseClient.cs
chunyu3: resolved

chunyu3: resolved

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
chunyu3: resolved

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
chunyu3: removed

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
chunyu3: removed

test/TestProjects/CollapseRequestCondition-LowLevel/CollapseRequestCondition-LowLevel.json
chunyu3: resolved

test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj
chunyu3: resolved

test/TestProjects/CollapseRequestCondition-LowLevel/Generated/RequestConditionCollapseClient.cs
annelo-msft: Thanks, @chunyu3!

annelo-msft (approved):
The output in RequestConditionCollapseClient looks correct to me.  @AlexanderSher or @ShivangiReja, does the generator source look good to you?

PR #1809, opened 2021-12-30
mgmt testGen: use-create-resourceIdentifier

changlong-liu:
# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
+++ b/src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
@@ -43,26 +43,6 @@ namespace AutoRest.CSharp.MgmtTest.Generation
             }
         }
 
-        public static object? ConvertToStringDictionary(object dict)
-        {
-            if (dict is null)
-            {
-                return dict;
-            }
-            Type t = dict.GetType();
-            bool isDict = t.IsGenericType && t.GetGenericTypeDefinition() == typeof(Dictionary<,>);
-            if (!isDict)
-            {
-                return dict;
-            }
-            var ret = new Dictionary<string, object?>();
-            foreach (KeyValuePair<object, object> entry in (IEnumerable< KeyValuePair<object, object>>)dict)
-            {
-                ret.Add(entry.Key.ToString()!, ConvertToStringDictionary(entry.Value));
-            }
-            return ret;
-        }
-
         public static string FormatResourceId(string resourceId)
         {
             if (resourceId.Length == 0)
--- a/src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
+++ b/src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
@@ -12,16 +12,10 @@ using AutoRest.CSharp.Input;
 using AutoRest.CSharp.Mgmt.Decorator;
 using AutoRest.CSharp.Mgmt.Output;
 using AutoRest.CSharp.Output.Models.Shared;
-using AutoRest.CSharp.Utilities;
 using AutoRest.CSharp.Output.Builders;
 using AutoRest.CSharp.Output.Models.Serialization.Json;
-using System.Text.Json;
 using AutoRest.CSharp.Mgmt.Generation;
 using AutoRest.CSharp.Mgmt.Models;
-using AutoRest.CSharp.Output.Models.Requests;
-using System.Text.Json.Serialization;
-using Azure.Core.Serialization;
-using System.Runtime.InteropServices.WindowsRuntime;
 using System.Text.RegularExpressions;
 using Azure.Core;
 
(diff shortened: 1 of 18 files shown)


src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -309,6 +283,16 @@ public void WriteSchemaObjectExampleValue(CodeWriter writer, ObjectType sot, Exa
                 WriteSchemaObjectExampleProperties(writer, sot, ev, variableName, consumedProperties);
         }
 
+        protected void CreateTagWriterDelegate(string newVariableName, CSharpType valueType, ExampleValue ev)
```
ArcturusZhang, line 286: I was told that CodeWriterDelegate is the last resort in our generator, but I believe this is a proper usage here. I do not cannot think of a better solution given the case here.

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -556,7 +538,7 @@ public void WriteExampleValue(CodeWriter writer, CSharpType cst, ExampleValue ex
             }
         }
 
-        public static string useVariableName(string variableName)
+        public string useVariableName(string variableName)
```
ArcturusZhang, line 541:
```suggestion
        public string UseVariableName(string variableName)
```

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -580,7 +562,7 @@ public static string useVariableName(string variableName)
             throw new Exception($"Can't find a valid variable name start with {variableName}");
         }
 
-        public static void clearVariableNames() {
+        public void clearVariableNames() {
```
ArcturusZhang, line 565:
```suggestion
        public void ClearVariableNames() {
```

src/AutoRest.CSharp/MgmtTest/Generation/MgmtBaseTestWriter.cs
```diff
@@ -556,7 +538,7 @@ public void WriteExampleValue(CodeWriter writer, CSharpType cst, ExampleValue ex
             }
         }
 
-        public static string useVariableName(string variableName)
+        public string useVariableName(string variableName)
         {
             if (Utilities.StringExtensions.IsCSharpKeyword(variableName))
```
ArcturusZhang, line 543: Looks like we are trying to avoid multi-defined variables here. Could you please try to use `CodeWriterDeclaration`? In the `CodeWriter` class when it is writing code, it will do the deduplication for your variables.
Therefore I believe we could remove this function, and from where we are invoking this, we just write `var something = new CodeWriterDeclaration("variableName");`, and writes it using the format symbol `:D`, like
```
writer.Line($"var {something:D} = Id");
```
and code writer could take care of the local variable names in the same scope.
```

File: src/AutoRest.CSharp/MgmtTest/Generation/ResourceTestWriter.cs [hunk 5]
```diff
@@ -34,6 +33,7 @@ internal class ResourceTestWriter : MgmtBaseTestWriter
         public ResourceTestWriter(CodeWriter writer, Resource resource, BuildContext<MgmtOutputLibrary> context): base(writer, resource, context)
         {
             _resource = resource;
+            _getAllOperation = _resource.AllOperations;
```
ArcturusZhang, line 36: Shall we rename this variable? `resource.AllOperations` returns a collection that contains all the operations in this resource, and `getAllOperation` stands for one single operation with the name of `GetAll`.

src/AutoRest.CSharp/MgmtTest/Generation/ResourceTestWriter.cs
```diff
@@ -110,10 +107,12 @@ protected void WriteTesterCtors()
             }
         }
 
-        public string WriteGetResource(Resource resource, string requestPath, ExampleModel exampleModel)
+        public string WriteGetResource(Resource resource, string resourceIdentifierParams, ExampleModel exampleModel)
```
ArcturusZhang, line 110: I believe this parameter `resourceIdentifierParams`, its actual value is like `"0000-0000-...", "myRg", "myAvailabilitySet"`?
I suggest we change this type to `FormattableString` which is a recommended way to pass a intermediate result of the code writer, which leaves some space for the case that we need some special format that only code writer could handle.

changlong-liu: yes, the resourceIdentifierParams does value to string like ("0000-0000-...", "myRg"..).
I will change the type.

ArcturusZhang (approved):
You might need to make a merge from feature/v3 since Michael just merged quite a few changes

PR #1810, opened 2021-12-30
Support RP's user agent in mgmt plane lro

### LINKED ISSUE #1588: [feature] set `User-Agent` to RP SDK info in LRO query requests
# Description

A left-over from preview. Right now the `User-Agent` header in LRO query requests are still bound to `ResourceManager`, like below. The current value should be each RP SDK info, like `azsdk-net-ResourceManager.Network/1.0.0-alpha.20210825.3 (.NET Framework 4.8.4341.0; Microsoft Windows 10.0.19043 )`
```
 "RequestUri": "https://management.azure.com/subscriptions/db1ab6f0-4769-4b27-930e-01e2ef9c123c/providers/Microsoft.Network/locations/eastus/operations/7b67485a-1ab2-4ee2-83a5-7a970219ffbf?api-version=2021-02-01",
      "RequestMethod": "GET",
      "RequestHeaders": {
        "Authorization": "Sanitized",
        "User-Agent": "azsdk-net-ResourceManager/1.0.0-alpha.20210825.3 (.NET Framework 4.8.4341.0; Microsoft Windows 10.0.19043 )",
```

[Note]: Migrated from [ADO 6154](https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/6154/).

Yao725:
# Description
<i>Fix #1588 
Currently we use the user agent `azsdk-net-ResourceManager` in all the mgmt plane SDK, this PR aims to modify the user agent according to the RP SDK info.</i>

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -698,8 +702,8 @@ protected virtual void WriteLROResponse(CSharpType lroObjectType, string diagnos
                 {
                     _writer.Append($"{ContextProperty}, ");
```
archerzz, line 703: So we don't need to pass `_userAgent` for non-long-running operation?

archerzz (commented):
Looks like we are not symmetric when passing `User-Agent` to long running and non-long running operations. Can we unify them?

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
Yao725: My bad, fixed

Yao725: FYI, I decide not to implement this feature in non-long-running operation, the reason is that for non-long-running operation we initialize the [`_operation`](https://github.com/Azure/autorest.csharp/blob/8436f4e9b38bfb6ca00ac70eaf165e59b09e4175/src/AutoRest.CSharp/Mgmt/Generation/NonLongRunningOperationWriter.cs?_pjax=%23js-repo-pjax-container%3Afirst-of-type%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%3Afirst-of-type%2C%20%5Bdata-pjax-container%5D%3Afirst-of-type#L35) using `OperationOrResponseInternals` or `OperationOrResponseInternals<T>`. As we use response to create this `_operation`, everything is not related with LRO.

archerzz:
Can we use a class, instead of a `Dictionary`?

samples/Azure.Management.Storage/Generated/BlobContainer.cs
```diff
@@ -426,8 +426,10 @@ public async virtual Task<BlobContainerObjectLevelWormOperation> ObjectLevelWorm
             scope.Start();
             try
             {
+                var additionalOptions = new ChangeTrackingDictionary<string, string>();
```
m-nash, line 429: Lets make this a readonly class field of type string vs Dictionary and read it from the rest operations since the value is already evaluated there.

As for how to get the value into NextLinkOperationImplementation I prefer if we didn't add a new ctor param everywhere let me explore some ideas.

m-nash (changes requested):

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -23,11 +23,12 @@ internal class NextLinkOperationImplementation : IOperation
         private readonly OperationFinalStateVia _finalStateVia;
         private readonly RequestMethod _requestMethod;
         private readonly HttpPipeline _pipeline;
+        private readonly ChangeTrackingDictionary<string, string> _additionalOptions;
 
         private string? _lastKnownLocation;
         private string _nextRequestUri;
 
-        public static IOperation Create(HttpPipeline pipeline, RequestMethod requestMethod, Uri startRequestUri, Response response, OperationFinalStateVia finalStateVia)
+        public static IOperation Create(HttpPipeline pipeline, RequestMethod requestMethod, Uri startRequestUri, Response response, OperationFinalStateVia finalStateVia, ChangeTrackingDictionary<string, string>? additionalOptions = default)
```
m-nash, line 31: I think what we want to do here is instead replace `RequestMethod requestMethod, Uri startRequestUri` with `Request originalRequest` which will then allow us to pull out the user-agent header in the original request and store in a private field.  Then inside the CreateRequest method if that field is not null set the user-agent to be the same.

m-nash (changes requested):

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -208,6 +208,7 @@ public static void WriteRequestCreation(CodeWriter writer, RestClientMethod clie
 
                 if (writeUserAgentOverride)
                 {
+                    writer.Line($"{request}.Headers.SetValue(HttpHeader.Names.UserAgent, _userAgent);");
```
m-nash, line 211: We shouldn't need this here this is what the MgmtTelemitryPolicy does for us

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -41,19 +43,20 @@ public static IOperation Create(HttpPipeline pipeline, RequestMethod requestMeth
...
+            _requestMethod = request.Method;
             _headerSource = headerSource;
-            _startRequestUri = startRequestUri;
+            _startRequestUri = request.Uri.ToUri();
             _nextRequestUri = nextRequestUri;
             _originalResponseHasLocation = originalResponseHasLocation;
             _lastKnownLocation = lastKnownLocation;
             _finalStateVia = finalStateVia;
             _pipeline = pipeline;
+            _userAgent = request.Headers.TryGetValues("User-Agent", out IEnumerable<string>? userAgent) ? string.Join(" ", userAgent) : default;
```
m-nash, line 59: Does it not work to simply call TryGetValue and let it do the string manipulation?

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -166,6 +169,10 @@ private async ValueTask<Response> GetResponseAsync(bool async, string uri, Cance
         private HttpMessage CreateRequest(string uri)
         {
             HttpMessage message = _pipeline.CreateMessage();
+            if (!string.IsNullOrEmpty(_userAgent))
+            {
+                message.SetProperty("UserAgentOverride", _userAgent!);
```
m-nash, line 174: I think we can just set the header here and not use the override mechanism?

request.Headers.SetValue

m-nash (changes requested):

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
Yao725: Yeah, I understand that, but the reason that we need this line is `MgmtTelemetryPolicy` is going to extract the `User-Agent` information from [message's property](https://github.com/Azure/azure-sdk-for-net/blob/e09905cbba81cd31a4add5bb4425127b9dc1589f/sdk/resourcemanager/Azure.ResourceManager/src/Internal/MgmtTelemetryPolicy.cs#L21), however we just pass the request to the OperationInternals, which doesn't contains this information. So this line is just used to store the 'User-Agent' so that we can pull out it in OperationInternals.

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
Yao725: Because we have the `MgmtTelemetryPolicy`, which will re-write the `User-Agent` value, so this command `request.Headers.SetValue`will not work.

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
Yao725: TryGetValue will do the string manipulation with comma, so the generated `User-Agent` header will like this:
```
"User-Agent": "azsdk-net-ResourceManager.Cdn/1.0.0-alpha.20220119.1,(.NET 6.0.1; Microsoft Windows 10.0.19044)"
```
But for now our 'User-Agent' will be an array and looks like this:
```
"User-Agent": [
          "azsdk-net-ResourceManager.Cdn/1.0.0-alpha.20220119.1",
          "(.NET 6.0.1; Microsoft Windows 10.0.19044)"
        ]
```
That's the reason I use TryGetValues and join the string by myself.

src/assets/Generator.Shared/MgmtPlaneOperationHelpers.cs
```diff
@@ -0,0 +1,53 @@
...
+
+#nullable enable
+
+using System;
+using System.Collections.Concurrent;
+using Azure.Core.Pipeline;
+
+namespace Azure.Core
+{
+    internal static class MgmtPlaneOperationHelpers
```
m-nash, line 12: For these classes that will end up in code lets stick with Arm as the prefix

src/assets/Generator.Shared/MgmtPlaneOperationHelpers.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        /// <summary>
+        /// Store the property values that are shared by current management plane RP.
+        /// </summary>
+        private static ConcurrentDictionary<string, object> _mgmtProperties = new ConcurrentDictionary<string, object>();
```
m-nash, line 17: Lets figure out where this caching belongs.  Need to sync with https://github.com/Azure/azure-sdk-for-net/pull/27079#discussion_r813451688

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -208,7 +212,7 @@ internal static bool TryGetApiVersion(Uri startRequestUri, out ReadOnlySpan<char
             // 1.Original request is a put method;
             // 2.Original request is a management plane patch method. For data plane patch method, the final uri will be determinned by the original response header and "_finalStateVia";
             // 3.Original response header contains "Location" and FinalStateVia is configured to OriginalUri.
-            if (_requestMethod == RequestMethod.Put || _requestMethod == RequestMethod.Patch && _startRequestUri.Scheme == "https" && _startRequestUri.Host == "management.azure.com" || _originalResponseHasLocation && _finalStateVia == OperationFinalStateVia.OriginalUri)
+            if (_requestMethod == RequestMethod.Put || _requestMethod == RequestMethod.Patch && (_isMgmtPlane ?? false) || _originalResponseHasLocation && _finalStateVia == OperationFinalStateVia.OriginalUri)
```
m-nash, line 215: do we know for sure this behavior only applies to mgmt plane?  It doesn't seem like that should be the case, its ok if dataplane will never use this behavior, but the LRO implementation shouldn't vary for data vs mgmt plane in specification. @annelo-msft  can you comment?

m-nash (changes requested):

src/assets/Generator.Shared/MgmtPlaneOperationHelpers.cs
Yao725: Changed

src/assets/Generator.Shared/MgmtPlaneOperationHelpers.cs
Yao725: Done

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
Yao725: I use this to solve [this problem](https://github.com/Azure/autorest.csharp/pull/1720#issue-1062198206) and for now I just want to apply this behavior to mgmt plane. My previous implementation didn't distinguish between mgmt. plane and data plane, which resulted in the test failure in `Azure.Communication.PhoneNumbers`.

PR #1814, opened 2021-12-31
Fix to support generating the four scopes in Azure.ResourceManager

fengzhou-msft:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/ScopeDetection.cs
```diff
@@ -58,10 +59,20 @@ private static RequestPath CalculateScopePath(RequestPath requestPath)
...
+                return new RequestPath(requestPath.Take(indexOfProvider));
+            }
+            // Returns RequestPath.ResourceGroup for the list path "/subscriptions/{subscriptionId}/resourceGroups" as well.
+            if (requestPath.SerializedPath.StartsWith("/subscriptions/{subscriptionId}/resourceGroups", StringComparison.InvariantCultureIgnoreCase))
+                return RequestPath.ResourceGroup;
+            if (requestPath.SerializedPath.StartsWith("/subscriptions", StringComparison.InvariantCultureIgnoreCase))
+                return RequestPath.Subscription;
+            if (requestPath.SerializedPath.Equals("/tenants"))
+                return RequestPath.Tenant;
+            return requestPath;
```
ArcturusZhang, line 75: I do not like hard-coded strings here, which might lead to problems in the future.
Looks like we are testing if these paths are subscription, resource group, mgmt group. How about we get the resource type from these paths, and switch the resource type?

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -242,7 +242,8 @@ private IEnumerable<MgmtClientOperation> EnsureAllOperations()
                     if (IsListOperation(operation, operationSet))
                     {
                         // if this operation is a collection operation, it should be the parent of its corresponding resource request path
-                        var diff = requestTrimmedPath.TrimAncestorFrom(resourceTrimmedPath);
+                        // In the case of /{scope}/providers/Microsoft.Resources/links being a list method for /{linkId}, we need to reverse the trim.
+                        var diff = requestTrimmedPath.SerializedPath.Length < resourceTrimmedPath.SerializedPath.Length ? requestTrimmedPath.TrimPrefixPathFrom(resourceTrimmedPath) : requestTrimmedPath.TrimPrefixPathFrom(requestTrimmedPath);
```
ArcturusZhang, line 246: In our latest changes, we change the implementation around here, you might need to have a look if this change is still needed in the latest code

ArcturusZhang:
Also FYI I am introducing some new configurations which might have impact to your implementations: https://github.com/Azure/autorest.csharp/pull/1808

src/AutoRest.CSharp/Mgmt/Generation/ResourceCollectionWriter.cs
```diff
@@ -100,7 +100,7 @@ protected override void WriteCtors()
 
             // write "parent resource" constructor
             _writer.Line();
-            _writer.WriteXmlDocumentationSummary($"Initializes a new instance of {TypeNameOfThis} class.");
+            _writer.WriteXmlDocumentationSummary($"Initializes a new instance of the <see cref=\"{TypeNameOfThis}\"/> class class.");
```
ArcturusZhang, line 103: We have a duplicated word class here

src/AutoRest.CSharp/Mgmt/Models/RequestPath.cs
```diff
@@ -144,6 +144,11 @@ public RequestPath TrimPrefixPathFrom(RequestPath other)
         {
             if (this == other)
                 return RequestPath.Tenant;
+            // Handle the special case of trim provider from feature
+            if (this.SerializedPath == "/subscriptions/{subscriptionId}/providers/{resourceProviderNamespace}" && other.SerializedPath.StartsWith("/subscriptions/{subscriptionId}/providers/Microsoft.Features/providers/{resourceProviderNamespace}/features"))
+            {
+                return new RequestPath(other._segments.Skip(this.Count + 2));
+            }
```
ArcturusZhang, line 151: Why do we have to have this special case?
I would like to find a better solution instead of hard-coding a special case here

fengzhou-msft: The path `/subscriptions/{subscriptionId}/providers/{resourceProviderNamespace}` and `/subscriptions/{subscriptionId}/providers/Microsoft.Features/providers/{resourceProviderNamespace}` are equivalent in hierarchy.

src/AutoRest.CSharp/Mgmt/Decorator/ScopeDetection.cs
fengzhou-msft: This mainly takes care of the cases of the list path for scopes. For instance, previously `/subscriptions` returns itself as scope, then it does not match the scope of `/subscriptions/{subscriptionId}`,  now it returns `/subscriptions/{subscriptionId}` so the scopes can match. `/subscriptions` is not a valid ResourceIdentifier, I'm not sure how to get its resource type.

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
fengzhou-msft: The change is no longer needed and reverted.

src/AutoRest.CSharp/Mgmt/Models/RequestPath.cs
fengzhou-msft: As this is the handling of trim ancestor path from a request path where the ancestor path is not an exact prefix, no existing config can be used. Improved code with constant string.

ArthurMa1978 (approved):

Issue #1817, opened 2022-01-05
Intermittent OOM in unit test


archerzz:
## Problem

From time to time, we hit the following out of memory error in the unit test on pipeline. Although we could enlarge the memory limit for pipeline, we'd better diagnose the root cause, because previously there is no OOM error. It seems the memory footprint deteriorated during last few months.

```
 Failed ValidateRequiredParamsInCtor [688 ms]
  Error Message:
   System.OutOfMemoryException : Insufficient memory to continue the execution of the program.
  Stack Trace:
     at System.Runtime.InteropServices.Marshal.AllocHGlobal(IntPtr cb)
   at System.Reflection.Internal.NativeHeapMemoryBlock.DisposableData..ctor(Int32 size)
   at System.Reflection.Internal.StreamMemoryBlockProvider.ReadMemoryBlockNoLock(Stream stream, Boolean isFileStream, Int64 start, Int32 size)
   at System.Reflection.PortableExecutable.PEReader..ctor(Stream peStream, PEStreamOptions options, Int32 size)
   at Microsoft.CodeAnalysis.ModuleMetadata.CreateFromStream(Stream peStream, PEStreamOptions options)
   at Microsoft.CodeAnalysis.MetadataReference.CreateFromFile(String path, MetadataReferenceProperties properties, DocumentationProvider documentation)
   at AutoRest.CSharp.AutoRest.Plugins.GeneratedCodeWorkspace.CreateGeneratedCodeProject() in D:\a\1\s\autorest.csharp\src\AutoRest.CSharp\Common\AutoRest\Plugins\GeneratedCodeWorkspace.cs:line 155
   at AutoRest.CSharp.AutoRest.Plugins.GeneratedCodeWorkspace.Create(String projectDirectory, String outputDirectory, String[] sharedSourceFolders) in D:\a\1\s\autorest.csharp\src\AutoRest.CSharp\Common\AutoRest\Plugins\GeneratedCodeWorkspace.cs:line 111
   at AutoRest.TestServer.Tests.Mgmt.OutputLibrary.OutputLibraryTestBase.Generate(String testProject, String subFolder) in D:\a\1\s\autorest.csharp\test\AutoRest.TestServer.Tests\Mgmt\OutputLibrary\OutputLibraryTestBase.cs:line 48
[...]

PR #1820, opened 2022-01-06
Use Ubuntu for CI

AlexanderSher:
Current [OOM issue](https://github.com/Azure/autorest.csharp/issues/1817) prevents most of the PR's to succeed in CI. As a temporary measure, we need to switch to VMs that have more RAM.
1ES machines can be used, but windows machines have a long path limit, hence we try to use ubuntu.


eng/pipelines/build.yml
```diff
@@ -34,10 +34,17 @@ stages:
       - job: Build
         timeoutInMinutes: 120
         pool:
-          vmImage: windows-2022
+          name: azsdk-pool-mms-ubuntu-2004-general
+          vmImage: "MMSUbuntu20.04"
```
ShivangiReja, line 38: Why are we running the build on ubuntu? are there any issues on windows?

AlexanderSher: Yes, Windows images don't support long paths yet: https://github.com/Azure/azure-sdk-tools/issues/2497

eng/Generation.psm1
```diff
@@ -29,7 +29,7 @@ function Invoke-AutoRest($baseOutput, $projectName, $autoRestArguments, $sharedS
     {
         $outputPath = Join-Path $baseOutput "SomeFolder" "Generated"
     }
-    $namespace = $projectName.Replace('-', '_')
+    $namespace = $projectName.Replace('-', '_') -replace "(^|_)[a-z]",{return $_.Value.ToUpper()}
```
annelo-msft, line 32: Will this work on Windows as well?

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -84,7 +84,7 @@ internal static string SaveConfiguration(Configuration configuration)
                     WriteIfNotDefault(writer, Configuration.Options.SkipCSProjPackageReference, configuration.SkipCSProjPackageReference);
                     WriteIfNotDefault(writer, Configuration.Options.DataPlane, configuration.DataPlane);
                     WriteIfNotDefault(writer, Configuration.Options.SingleTopLevelClient, configuration.SingleTopLevelClient);
-                    WriteIfNotDefault(writer, Configuration.Options.ProjectFolder, configuration.ProjectFolder);
+                    WriteIfNotDefault(writer, Configuration.Options.ProjectFolder, configuration.ProjectFolder.Replace("\\", "/"));
```
annelo-msft, line 87: Is it worth investing in making this configurable, or do we plan to update all these same files if/when we move this back to Windows?

src/AutoRest.CSharp/Properties/launchSettings.json
```diff
@@ -2,487 +2,487 @@
   "profiles": {
     "Accessibility": {
       "commandName": "Project",
-      "commandLineArgs": "--standalone $(SolutionDir)\\test\\TestProjects\\Accessibility\\Generated"
+      "commandLineArgs": "--standalone $(SolutionDir)/test/TestProjects/Accessibility/Generated"
```
annelo-msft, line 5: Will folks be able to debug the project locally on Windows machines with this change?

test/AutoRest.TestServer.Tests/NonStringEnumTests.cs
```diff
@@ -3,8 +3,8 @@
 
 using System.Threading.Tasks;
 using AutoRest.TestServer.Tests.Infrastructure;
-using non_string_enum;
-using non_string_enum.Models;
+using Non_String_Enum;
```
annelo-msft, line 6: Is this due to a limitation of .NET Core on Linux?

eng/Generation.psm1
AlexanderSher: Yes, that is exactly to produce identical results on Windows and Linux

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
AlexanderSher: On Windows it works fine with any type of slashes. On Linux it works only with `/`

test/AutoRest.TestServer.Tests/NonStringEnumTests.cs
AlexanderSher: We are using Roslyn formatting to make generated files look readable. For some reason, Roslyn sorts `using` directives differently on Linux and Windows in terms of lower case and upper case characters. And it is not configurable yet: https://github.com/dotnet/roslyn/issues/44052
Since we need a fix ASAP, I've come up with this workaround.

src/AutoRest.CSharp/Properties/launchSettings.json
AlexanderSher: Yes:
image

annelo-msft (approved):
LGTM

AlexanderSher:
Closing in favor of https://github.com/Azure/autorest.csharp/pull/1823

PR #1823, opened 2022-01-08
Use 1ES Windows VM for CI

AlexanderSher:
Current [OOM issue](https://github.com/Azure/autorest.csharp/issues/1817) prevents most of the PR's to succeed in CI. As a temporary measure, we need to switch to VMs that have more RAM.

Since https://github.com/Azure/azure-sdk-tools/issues/2497 is fixed, we can use Windows VM.

### DIFF
--- a/eng/PublishRelease.ps1
+++ b/eng/PublishRelease.ps1
@@ -17,7 +17,7 @@ try {
 
     Write-Host "Publishing $file on GitHub!"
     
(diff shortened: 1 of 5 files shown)


AlexanderSher:
Both public and internal builds have passed.

ArcturusZhang (approved):
Thank you @AlexanderSher!

PR #1831, opened 2022-01-11
fix(mgmt): no leading "/" if 1st segment is not constant

### LINKED ISSUE #1830: [bug] "CreateResourceIdentifier": no leading "/" if 1st segment is not constant
## Problem

When generating `CreateResourceIdentifier()`, we should not add leading "/" for request paths of which the first segment is not a constant.
For example, "{scope}/providers/Microsoft.Resources/deployments/{deploymentName}". Callers can specify the leading "/" through the 1st argument to `CreateResourceIdentifier()`.

archerzz:
# Description

When generating `CreateResourceIdentifier()`, we should not add leading "/" for request paths of which the first segment is not a constant.
For example, "{scope}/providers/Microsoft.Resources/deployments/{deploymentName}". Callers can specify the leading "/" through the 1st argument to `CreateResourceIdentifier()`.

resolve #1830

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Generation/ResourceWriter.cs
```diff
@@ -99,7 +99,8 @@ private void WriteCreateResourceIdentifierMethods()
                     // - https://github.com/Azure/azure-rest-api-specs/blob/719b74f77b92eb1ec3814be6c4488bcf6b651733/specification/storage/resource-manager/Microsoft.Storage/stable/2021-04-01/blob.json#L58
                     // - https://github.com/Azure/azure-rest-api-specs/blob/719b74f77b92eb1ec3814be6c4488bcf6b651733/specification/storage/resource-manager/Microsoft.Storage/stable/2021-04-01/blob.json#L146
                     // so here we have to use `Seqment.BuildSerializedSegments` instead of `RequestPath.SerializedPath` which could be from `RestClientMethod.Operation.GetHttpPath`
-                    _writer.Line($"var resourceId = $\"{Segment.BuildSerializedSegments(requestPath)}\";");
+                    // If first segment is "{var}", then we should not add leading "/". Instead, we should let callers to specify, e.g. "{scope}/providers/Microsoft.Resources/..." v.s. "/subscriptions/{subscriptionId}/..."
+                    _writer.Line($"var resourceId = $\"{Segment.BuildSerializedSegments(requestPath, requestPath.First().IsConstant)}\";");
```
ArcturusZhang, line 103: I believe that a variable could take a leading slash when it has `SkipUrlEncoding = true`.
For now, I believe the first segment is a variable works the same as SkipUrlEncoding, but I suppose we could make it more accurate
```suggestion
                    _writer.Line($"var resourceId = $\"{Segment.BuildSerializedSegments(requestPath, requestPath.First().IsConstant && requestPath.First().SkipUrlEncoding)}\";");
```

ArcturusZhang (approved):

fengzhou-msft (approved):

src/AutoRest.CSharp/Mgmt/Generation/ResourceWriter.cs
archerzz: After discussion, we thought the ideal case might be `requestPath.First().IsConstant || requestPath.First().SkipUrlEncoding`. But since we haven't seen a case that a leading variable needs leading `/`, we thought it would be fine to keep as-is.

PR #1841, opened 2022-01-12
Append or replace api-version for nextRequestUri

fengzhou-msft:
The fix in this PR is based on https://github.com/Azure/azure-sdk-for-net/issues/25411#issuecomment-1005892110.

Resolve https://github.com/Azure/azure-sdk-for-net/issues/25411
Resolve https://github.com/azure/azure-sdk-for-net/issues/26234

# Description

# Checklist

# Ready to Land?


src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -104,17 +106,42 @@ private void UpdateNextRequestUri(ResponseHeaders headers)
...
             }
         }
 
+        private static string AppendOrReplaceApiVersion(string uri, Uri startRequestUri)
+        {
+            NameValueCollection startRequestUriQueries = HttpUtility.ParseQueryString(startRequestUri.Query);
+            var apiVersion = startRequestUriQueries.Get("api-version");
+            if (apiVersion != null)
+            {
+                if (uri.Contains("api-version"))
```
m-nash, line 126: Can you provide test cases which have expected input / output to help understand what we are trying to do here.  I suspect we can make this more performant but will need a better idea of the use cases to know that.

m-nash (changes requested):

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -104,17 +106,49 @@ private void UpdateNextRequestUri(ResponseHeaders headers)
...
+                    var indexOfFirstSignAfterApiVersion = uri.IndexOf("&", apiVersionIndex);
+                    ReadOnlySpan<char> uriBeforeApiVersion = uriSpan.Slice(0, apiVersionIndex + apiVersionQuery.Length + "=".Length);
+                    if (indexOfFirstSignAfterApiVersion == -1)
+                    {
+                        return string.Concat(uriBeforeApiVersion, apiVersion);
+                    }
+                    else
+                    {
+                        ReadOnlySpan<char> uriAfterApiVersion = uriSpan.Slice(indexOfFirstSignAfterApiVersion);
+                        return string.Concat(uriBeforeApiVersion, apiVersion, uriAfterApiVersion);
```
fengzhou-msft, line 140: The test projects use .netstandard 2.0 which doesn't support `string.Concat(ReadOnlySpan<Char>, ReadOnlySpan<Char>, ReadOnlySpan<Char>)`, so this caused build error for test projects.

test/AutoRest.Shared.Tests/AppendOrReplaceApiVersionTests.cs
```diff
@@ -0,0 +1,25 @@
...
+        [TestCase("https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086?api-version=2021-10-30-PREVIEW", "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086?api-version=2021-09-30-PREVIEW", "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086?api-version=2021-09-30-PREVIEW")]
+        [TestCase("xyz.com?api-version=2021-10-01&a=b", "https://abc.com?api-version=2021-11-01", "xyz.com?api-version=2021-11-01&a=b")]
```
annelo-msft, line 17: Let's please add a test case for a Cognitive Service URI, i.e. one with the API Version as a path parameter.  You can see these as examples:
- https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/QnAMaker/stable/v4.0/QnAMaker.json
- https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/TextAnalytics/stable/v3.1/TextAnalytics.json

test/AutoRest.Shared.Tests/AppendOrReplaceApiVersionTests.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.IO;
+using System.Text.Json;
+using NUnit.Framework;
+
+namespace Azure.Core.Tests
+{
+    public class AppendOrReplaceApiVersionTests
+    {
+        [TestCase("/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086", "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086?api-version=2021-09-30-PREVIEW", "/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086?api-version=2021-09-30-PREVIEW")]
```
m-nash, line 13: nit: multiline these so its easier to compare input vs output

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -104,17 +106,49 @@ private void UpdateNextRequestUri(ResponseHeaders headers)
...
-                    _nextRequestUri = location!;
+                    _nextRequestUri = AppendOrReplaceApiVersion(location!, _startRequestUri);
                     return;
             }
         }
 
+        internal static string AppendOrReplaceApiVersion(string uri, Uri startRequestUri)
+        {
+            const string apiVersionQuery = "api-version";
+            NameValueCollection startRequestUriQueries = HttpUtility.ParseQueryString(startRequestUri.Query);
```
m-nash, line 123: Given this is on the hotpath I think we can improve this more.  It appears that you are constructing a collection which will contain all of the query params.  This is a lot of allocation and time that isn't needed since all we want is the apiVersion query param.

As an example here is the benchmark diff
|                     Method |      Mean |    Error |   StdDev |  Gen 0 | Allocated |
|--------------------------- |----------:|---------:|---------:|-------:|----------:|
| GetApiStringWithParseQuery | 981.35 ns | 7.970 ns | 7.065 ns | 0.3529 |   2,216 B |
|       GetApiStringWithSpan |  28.57 ns | 0.254 ns | 0.225 ns | 0.0102 |      64 B |

The ParseQueryString is 34x slower and allocates 34x more vs doing a simple parse of this one value out of the string.
Here is the code I used to test

```csharp
[MemoryDiagnoser]
public class ParseQueryBench
{
    private const string ApiVersionParam = "api-version";
    private static Uri StartRequestUri = new Uri("https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086?something=something&api-version=2021-09-30-PREVIEW&foobar=foobar");

    [Benchmark]
    public string GetApiStringWithParseQuery()
    {
        NameValueCollection startRequestUriQueries = HttpUtility.ParseQueryString(StartRequestUri.Query);
        return startRequestUriQueries.Get(ApiVersionParam);
    }

    [Benchmark]
    public string GetApiStringWithSpan()
    {
        ReadOnlySpan<char> uriSpan = StartRequestUri.Query.AsSpan();
        int startIndex = uriSpan.IndexOf(ApiVersionParam);
        startIndex += ApiVersionParam.Length + 1;
        ReadOnlySpan<char> remaining = uriSpan.Slice(startIndex);
        int endIndex = remaining.IndexOf('&');
        int length = endIndex == -1 ? uriSpan.Length - startIndex : endIndex;
        return uriSpan.Slice(startIndex, length).ToString();
    }
}
```

I didn't find any utilities in dotnet code that does the same thing so I think we should use this and just add a robust set of tests around it to make sure it works in all cases.

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -104,17 +106,49 @@ private void UpdateNextRequestUri(ResponseHeaders headers)
...
+            var apiVersion = startRequestUriQueries.Get("api-version");
+            if (apiVersion != null)
+            {
+                var uriSpan = uri.AsSpan();
+                if (uri.Contains(apiVersionQuery))
```
m-nash, line 128: use uriSpan anywhere you are using uri to do indexOf / comparisons

m-nash (changes requested):

test/AutoRest.Shared.Tests/AppendOrReplaceApiVersionTests.cs
fengzhou-msft: In this case, we don't need to append api-version, right? The location header ends with a guid.

test/AutoRest.Shared.Tests/AppendOrReplaceApiVersionTests.cs
```diff
@@ -0,0 +1,38 @@
...
+                  "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086?api-version=2021-09-30-PREVIEW",
+                  "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/testRG-416/providers/Microsoft.ManagedIdentity/userAssignedIdentities/testRi-6086?api-version=2021-09-30-PREVIEW")]
+        [TestCase("xyz.com?api-version=2021-10-01&a=b",
+                  "https://abc.com?api-version=2021-11-01",
+                  "xyz.com?api-version=2021-11-01&a=b")]
+        [TestCase("/operations/2608e164-7bb4-4012-98f6-9a862b2218a6",
+                  "https://qnamaker-resource-name.api.cognitiveservices.azure.com/qnamaker/v5.0-preview.1/knowledgebases/create",
+                  "/operations/2608e164-7bb4-4012-98f6-9a862b2218a6")]
+        public void TestAppendOrReplaceApiVersion(string uriToProcess, string startRequestUriStr, string expectedUri)
```
m-nash, line 31: can we add test cases where api-version is in the middle and end of the query params?
can we also add one where api-version is not present but there are other query params?
can we also add an api-version that doesn't match the case (this should fail)

m-nash (changes requested):

m-nash (approved):

annelo-msft (dismissed):
Please wait to merge until the guidance regarding api-version from arch board is finalized: https://github.com/Azure/azure-sdk/issues/3817

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -104,17 +107,66 @@ private void UpdateNextRequestUri(ResponseHeaders headers)
...
                 case HeaderSource.Location when hasLocation:
-                    _nextRequestUri = location!;
+                    _nextRequestUri = AppendOrReplaceApiVersion(location!, _startRequestUri);
                     return;
             }
         }
 
+        internal static string AppendOrReplaceApiVersion(string uri, Uri startRequestUri)
+        {
+            if (TryGetApiString(startRequestUri, out ReadOnlySpan<char> apiVersion))
```
m-nash, line 123: Since this is only ever called with _startRequestUri can we do this as a lazy loaded value so that we only get the apiversion once?

m-nash (changes requested):

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
fengzhou-msft: The AppendOrReplaceApiVersion method is first called before the creation of a NextLinkOperationImplementation instance. I added _apiVersion property to store the value in constructor from the result of the first call and also refactored AppendOrReplaceApiVersion to take in apiVersion directly. Therefore, TryGetApiString is only called once now.

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -104,17 +111,66 @@ private void UpdateNextRequestUri(ResponseHeaders headers)
...
+        {
+            if (apiVersion != null)
+            {
+                var uriSpan = uri.AsSpan();
+                var apiVersionParamSpan = ApiVersionParam.AsSpan();
+                var apiVersionIndex = uriSpan.IndexOf(apiVersionParamSpan);
+                if (apiVersionIndex == -1)
+                {
+                    var concatSymbol = uriSpan.IndexOf('?') > -1 ? "&" : "?";
+                    return $"{uri}{concatSymbol}api-version={apiVersion}";
```
heaths, line 135: Since the goal of this method seems to be partly performance, I would double check that the compiler (at least Release builds) effectively turns this into a `string.Concat` call. Seems it should, but if using `string.Format` or related APIs, those can be very expensive (lots of type checks and other conditions on operands).

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -104,17 +111,66 @@ private void UpdateNextRequestUri(ResponseHeaders headers)
...
+                }
+                else
+                {
+                    var lengthToEqualSignAfterApiVersionParam = apiVersionIndex + ApiVersionParam.Length + 1;
```
heaths, line 139: Nit: while incredibly unlikely, params don't necessary have to be followed by `=`. Might want to actually make sure. Again, though, very unlikely. Perhaps causing a service error in such a case is actually desirable to find issues.

test/AutoRest.Shared.Tests/AppendOrReplaceApiVersionTests.cs
heaths: Also something like `/path?api-version`.

heaths (approved):
Approving, though I agree with @m-nash on adding test cases.

m-nash (approved):

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
fengzhou-msft: I think you're right and a benchmark test showed similar result for the two:
| Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
| Format | 30.19 ns | 0.5622 ns | 0.4984 ns | 0.0135 |     - |     - |     112 B |
| Concat | 30.34 ns | 0.3059 ns | 0.2862 ns | 0.0135 |     - |     - |     112 B |

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
fengzhou-msft: Such case is also handled and I've added corresponding tests.

PR #1844, opened 2022-01-13
Improve null validation

AlexanderSher:
This PR restores null validation in method signatures, simplifies null checks and separates parameter validation methods for LLC and HLC to prepare extension method for more complicated validations


AlexanderSher:
Change isn't required for RC1, but it is desired.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -40,14 +40,8 @@ protected DocumentTranslationClient()
...
         {
-            if (endpoint == null)
-            {
-                throw new ArgumentNullException(nameof(endpoint));
-            }
-            if (credential == null)
-            {
-                throw new ArgumentNullException(nameof(credential));
-            }
+            endpoint = endpoint ?? throw new ArgumentNullException(nameof(endpoint));
```
annelo-msft, line 43: Would it make sense to use Argument.AssertNotNull here instead?  (See: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Keys/src/KeyClient.cs#L62)

annelo-msft (approved):
This is nice - thanks!

AlexanderSher:
Closes https://github.com/Azure/autorest.csharp/issues/1697

Issue #1771, opened 2021-12-10
[Management Plane] Move required query parameters in parameter ordering


annelo-msft:
Per offline discussion, update query parameter position in method signature to align with [data plane ordering](https://github.com/azure/autorest.csharp/issues/1765):

1. (waitForCompletion)
2. (required path and query parameters)
    a. Required path parameters, sorted to match the order of the URI
    b. Remaining required query params will come in the order they are defined in swagger after the path parameters
3. (RequestContent)/ Body parameters
4. (ContentType) – not used by management plane
5. (optional path and query parameters)
6. ([Match|Request]Conditions)
7. RequestContext

ArthurMa1978:
@annelo-msft do you think that mgmt. plane also need to follow LLC pattern?  This is issue belong to the LLC Epics now.

annelo-msft:
Hi @ArthurMa1978!  Yes, @m-nash and I discussed that management plane and data plane could align on parameter ordering with this small change to management plane -- we are making a change on the data plane side to converge with this as well as part of https://github.com/azure/autorest.csharp/issues/1765.  The reason we would like to do this is to make it possible for management plane and data plane to eventually converge to use the same generator code -- this wouldn't be possible if we didn't use the same parameter ordering.

You are right that it probably shouldn't be part of an LLC epic - could you recommend a management plane epic that we could add it to instead?

archerzz:
I think this issue should have been resolved in #1843

annelo-msft:
@AlexanderSher, we have item 5 listed as "optional path and query params" -- is there an example of valid optional path parameters?  Wouldn't this result in a malformed URI segment?

annelo-msft:
@m-nash can you confirm that this issue has been closed?

m-nash:
yeah this is completed from mgmt point of view

annelo-msft:
Thank you, @m-nash !

PR #1894, opened 2022-01-20
feat(mgmt): sort method parameters

### LINKED ISSUE #1843: Sort parameters ordering
[1771](https://github.com/Azure/autorest.csharp/issues/1771)

We need to sort the 2, 3 in above issue

archerzz:
# Description
For mgmt method parameters, we've decided the new order is:
1. `waitForCompletion` (LRO only)
2. required parameters
3. body parameters
4. optiona parameters
5. `cancellationToken`

Item 1 and 5 are not swagger definitions, and they will be added by generator automatically. So we only need to implement the sorting for 2, 3, 4 above.
 This commit:
- update the logic of parameter sorting
- refactor to rename `OrderParameters` to `OrderParametersByRequired`, since it won't ensure body parameter position
- add test cases

resolve #1843

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -289,17 +293,38 @@ private Request BuildRequest(HttpRequest httpRequest, RequestMethodBuildContext
...
+            List<Parameter> requiredParameters = new();
+            List<Parameter> optionalParameters = new();
+            List<Parameter> bodyParameters = new();
             foreach (var (requestParameter, parameter) in allParameters)
             {
                 // Grouped and flattened parameters shouldn't be added to methods
                 if (IsMethodParameter(requestParameter))
                 {
-                    methodParameters.Add(parameter);
+                    // sort the parameters by the following sequence:
```
AlexanderSher, line 304: This change will break parameter ordering in all existing HLC clients.

AlexanderSher (dismissed):
This is a heavy breaking change. What issue this PR is addressing?
If this is a fix to https://github.com/Azure/autorest.csharp/issues/1843, then it should be scoped to mgmt plane. You can also try to use `RestClientBuilder.BuildRequestMethod` that does correct sorting for LLC.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
archerzz: Thanks. I thought the ordering is for HLC as well. I'll split the logic.

archerzz:
@AlexanderSher I updated the codes to split the sorting logic for mgmt and data plane. The logic for mgmt is much simpler than LLC, so I still keep its own version of sorting.

AlexanderSher:
I remember that we have discussed that order of host and path parameters should match their order in the URI template. Is it handled in some other place in Mgmt?

archerzz:
> I remember that we have discussed that order of host and path parameters should match their order in the URI template. Is it handled in some other place in Mgmt?

@AlexanderSher Yes. Path parameters are sorted in https://github.com/Azure/autorest.csharp/blob/faf9247db104799f62d518f78942164102d83408/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs#L770

As for `host`, do you mean `x-ms-parameterized-hos` in swagger? That seems not exist in mgmt swagger, so we don't handle it.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
ArthurMa1978: according to https://github.com/Azure/autorest.csharp/issues/1771 this should apply to all .net sdk

m-nash (approved):

PR #1903, opened 2022-01-21
Refactor constructor writing

AlexanderSher:
This refactoring splits `MethodSignature` into several types, simplifying its usage for constructors.
It also addresses several issues that will be listed below.

This PR doesn't introduce any changes in the generated code.


src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeConstructor.cs
```diff
@@ -10,18 +12,18 @@ internal class ObjectTypeConstructor
...
         public ObjectTypeConstructor(string name, string modifiers, Parameter[] parameters, ObjectPropertyInitializer[] initializers, ObjectTypeConstructor? baseConstructor = null)
         {
-            Signature = new MethodSignature(
+            Signature = new ConstructorSignature(
                 name,
                 $"Initializes a new instance of {name}",
                 modifiers,
                 parameters,
-                baseConstructor?.Signature);
+                new(isBase: true, baseConstructor?.Signature.Parameters ?? Array.Empty<Parameter>()));
```
AlexanderSher, line 20: This is an example of a "temporary" code that has existed for several months. In the PR where `MethodSignature` type was introduced, it was just a bit easier to use the same type for initializer instead of a different one, and this place has slipped through the CR.
Initializer should've had its own type from the start.

src/AutoRest.CSharp/Common/Output/Models/ParameterInvocation.cs
```diff
@@ -1,22 +0,0 @@
...
-using System;
-using AutoRest.CSharp.Generation.Types;
-using AutoRest.CSharp.Generation.Writers;
-using AutoRest.CSharp.Output.Models.Shared;
-
-namespace AutoRest.CSharp.Common.Output.Models
-{
-    internal record ParameterInvocation : Parameter
-    {
-        public ParameterInvocation(Parameter parameter, CodeWriterDelegate? Invocation = default) : this(parameter.Name, parameter.Description, parameter.Type, parameter.DefaultValue, parameter.ValidateNotNull, IsApiVersionParameter: parameter.IsApiVersionParameter, IsResourceIdentifier: parameter.IsResourceIdentifier, SkipUrlEncoding: parameter.SkipUrlEncoding, RequestLocation: parameter.RequestLocation, Invocation: Invocation) { }
```
AlexanderSher, line 13: This type has problems in several places at once:
1. Semantically, `Parameter` type is used to represent parameter in method signature. It has some properties that are used in parameter validation logic, but it is still part of the declaration. `Invocation` property has been used only in initializers or constructor calls, making the purpose of the `ParameterInvocation` type completely different, yet dragging all the properties from the base type, even though they are useless
2. `ParameterInvocation(Parameter parameter, CodeWriterDelegate? Invocation = default)` constructor's first argument is used only to initialize unused fields, hence it was also useless. Yet, when `ParameterInvocation` was created, it was always set to some value.
3. `CodeWriterDelegate` is a type for special cases and should be used only as a last resort. In addition to general problems with callbacks (very complicated code analysis for those who are new to the codebase, inverted stack trace during debugging, accidentally captured variables and unexpected caching, etc.), there is one specific to `CodeWriter` - we have no way to limit the scope of the callback call (see examples below).

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -375,49 +375,44 @@ protected PagingMethodWrapper GetPagingMethod(MgmtRestOperation operation)
             return new PagingMethodWrapper(pagingMethod);
         }
 
-        protected void WritePageFunctionBody(CSharpType itemType, PagingMethodWrapper pagingMethod, MgmtRestOperation operation, IEnumerable<ParameterMapping> parameterMappings,
-            bool isAsync, bool isNextPageFunc)
+        protected void WritePageFunctionBody(CSharpType itemType, PagingMethodWrapper pagingMethod, MgmtRestOperation operation, IEnumerable<ParameterMapping> parameterMappings, bool isAsync, bool isNextPageFunc)
         {
             var wrapResource = WrapResourceDataType(itemType, operation);
             var continuationTokenText = pagingMethod.NextLinkName != null ? $"response.Value.{pagingMethod.NextLinkName}" : "null";
+            var response = new CodeWriterDeclaration("response");
```
AlexanderSher, line 382: We should always use `CodeWriterDeclaration` when we declare a variable, otherwise writer won't be able to resolve name collision.

AlexanderSher: The same thing should apply to fields and parameters, though I'm not sure if our helper methods follow this rule.

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -375,49 +375,44 @@ protected PagingMethodWrapper GetPagingMethod(MgmtRestOperation operation)
...
 
+            _writer
+                .Append($"return {typeof(Page)}.FromValues(response.Value")
+                .AppendIf($".{pagingMethod.ItemName}", !pagingMethod.ItemName.IsNullOrEmpty());
+
             // only when we are listing ourselves, we use Select to convert XXXResourceData to XXXResource
-            FormattableString converter = $"";
             if (wrapResource != null)
             {
-                CodeWriterDelegate dataExpression = w => w.Append($"value");
```
AlexanderSher, line 392: This is an example of uncontrolled scope. Variable `value` is used in several places. However, we don't know where it should be declared, cause we don't know when each of the callbacks will be called and how many times it may happen. Hence no control for names collision.

m-nash (approved):

PR #1905, opened 2022-01-21
Add Uri public property

pshao25:
# Description
<i>Add a public readonly property for Uri. The generated client will add a property like.</i>
```
/// <summary> The HTTP Uri. </summary>
public virtual Uri Uri => _endpoint;
```
# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
+++ b/src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
@@ -101,6 +101,15 @@ namespace AutoRest.CSharp.Generation.Writers
                 .WriteXmlDocumentationSummary($"The HTTP pipeline for sending and receiving REST requests and responses.")
                 .Line($"public virtual {typeof(HttpPipeline)} Pipeline => {client.Fields.PipelineField.Name};");
 
+            var uriField = client.Fields.GetFieldByParameter("endpoint", new CSharpType(typeof(Uri)));
+            if (uriField != null)
+            {
+                writer
+                    .Line()
+                    .WriteXmlDocumentationSummary($"The HTTP Uri.")
+                    .Line($"public virtual {typeof(Uri)} Uri => {uriField.Declaration.ActualName};");
+            }
+
             writer.Line();
         }
 


annelo-msft (commented):
Would you please update the tests and samples based on this change so I can look at the output?  Thanks!

pshao25:
> Would you please update the tests and samples based on this change so I can look at the output? Thanks!

I have added the tests and updated the samples.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -101,6 +101,15 @@ private static void WriteClientFields(CodeWriter writer, LowLevelClient client)
                 .WriteXmlDocumentationSummary($"The HTTP pipeline for sending and receiving REST requests and responses.")
                 .Line($"public virtual {typeof(HttpPipeline)} Pipeline => {client.Fields.PipelineField.Name};");
 
+            var uriField = client.Fields.GetFieldByParameter("endpoint", new CSharpType(typeof(Uri)));
```
ShivangiReja, line 104: You are passing only `Uri` type, endpoint can be string type. It is working because we are not using the type in [this](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/LowLevel/Output/ClientFields.cs#L97) method to get the endpoint field. Can we call [this](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/LowLevel/Output/ClientFields.cs#L105) method instead?

annelo-msft: Shivangi and I discussed this offline - she is going to understand the reason why string is allowed for `endpoint` in some cases and whether that's the behavior we want.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -28,6 +28,9 @@ public partial class DocumentTranslationClient
         /// <summary> The HTTP pipeline for sending and receiving REST requests and responses. </summary>
         public virtual HttpPipeline Pipeline => _pipeline;
 
+        /// <summary> The HTTP Uri. </summary>
+        public virtual string Uri => _endpoint;
```
annelo-msft, line 32: The return type of `Uri` should be `Uri` according to this guideline: https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-primitives-uri

annelo-msft: @ShivangiReja FYI, for your investigation.

test/TestProjects/SingleTopLevelClientWithOperations-LowLevel/Generated/Client4.cs
```diff
@@ -27,6 +27,9 @@ public partial class Client4
         /// <summary> The HTTP pipeline for sending and receiving REST requests and responses. </summary>
         public virtual HttpPipeline Pipeline => _pipeline;
 
+        /// <summary> The HTTP Uri. </summary>
+        public virtual Uri Uri => _endpoint;
```
annelo-msft, line 31: This is interesting.  Should [subclients/resource subclients](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-subclients) have `Uri` on them -- will this always be available?

annelo-msft: @pshao25, how do we decide what the value of `_endpoint` will be for a subclient?

pshao25: Are you saying subclients should not have Uri? 
The value of _endpoint is passed from service client when the sub client is created.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
ShivangiReja: There is no real concept of Uri in Swagger, it's all string with the format metadata. [Here](https://github.com/Azure/autorest.csharp/tree/c43fd0f55fbb7cf0b871eeb7154f87ed8d4e6909/samples/Azure.Analytics.Purview.Account#model-endpoint-parameter-as-a-url-not-a-string) is one example of how we change from string to Uri type.

In terms of generator we don't want to change anything now. Maybe in future we will add analyzer for this and we could say something like: "If a client level path parameter is used in a parametrized host, then it's an endpoint and it must be format as url"

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -101,6 +101,20 @@ private static void WriteClientFields(CodeWriter writer, LowLevelClient client)
                 .WriteXmlDocumentationSummary($"The HTTP pipeline for sending and receiving REST requests and responses.")
                 .Line($"public virtual {typeof(HttpPipeline)} Pipeline => {client.Fields.PipelineField.Name};");
 
+            var endpointParameter = client.Parameters.FirstOrDefault(parameter => parameter.Name == "endpoint");
```
ShivangiReja, line 104: I'm still a bit concerned that we are filtering endpoint based on the name because in some services I see we don't use endpoint name. For example: [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/Generated/BlobRestClient.cs#L33) in blob storage it's called url. 

@annelo-msft do we have any guideline around naming endpoint?

pshao25: Isn't BlobRestClient HLC? I didn't look into that. What is our scope? 

In LLC, the uri field is hardcoded as name pipeline at [here](https://github.com/Azure/autorest.csharp/blob/d6bd09d44c4796c0b0be12bcd1f97f526c6311ac/src/AutoRest.CSharp/LowLevel/Output/ClientFields.cs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L71).

ShivangiReja: Yes, BlobStorage is currently HLC but our code generation should work for all the SDKs.

Could you please point me where we are hardcoding uri field?

ShivangiReja: Yes, BlobStorage is currently HLC but our LLC code generation should work for all the SDKs.

Could you please point me where we are hardcoding uri field? If we say that the name should always be endpoint, then I fine with this :)

annelo-msft: > @annelo-msft do we have any guideline around naming endpoint?

I'm not aware of one - if there isn't anything in the [Azure REST API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md), could you reach out to Laurent or Jeffrey?  Is there another aspect of swagger syntax we could use to identify the endpoint?

ShivangiReja: In swagger perspective it's all string with format metadata. Endpoint parameter will be same as any other string property.

After talking to Laurent, maybe we will a linter rule in future like: "If a client level path parameter is used in a parametrized host, then it's an endpoint and it must be format as url". If that's the case then we can filter endpoint parameter based on the name and url type.

ShivangiReja: In swagger perspective it's all string with format metadata. Endpoint parameter will be same as any other string parameter.

After talking to Laurent, maybe we will a linter rule in future like: "If a client level path parameter is used in a parametrized host, then it's an endpoint and it must be format as url". If that's the case then we can filter endpoint parameter based on the name and url type.

pshao25: > Yes, BlobStorage is currently HLC but our LLC code generation should work for all the SDKs.
> 
> Could you please point me where we are hardcoding uri field? If we say that the name should always be endpoint in SDKs, then I fine with this :)

Need to update the link to uri field [here](https://github.com/Azure/autorest.csharp/blob/90322dced7d08016294dafe9bff59a65d92c306e/src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L559).

ShivangiReja: I'm just curious, in this code that you linked we are also hardcoding the type as Uri, then why we have endpoint generated as string [here](https://github.com/Azure/autorest.csharp/blob/45dc0b8d87c3051f6068678a198e9926ac79394c/samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs#L44) for LLC?

AlexanderSher:
Sorry, but I don't think there is a value on adding Uri for the debugging purposes at all. Private fields are easily available under debugger

image

In fact, there is no difference in accessing public properties and private fields.

With adding a public property, we need to consider that we are providing not just a debugger information, but an extension point in a form of virtual property. If that extension point is fake (no internal code uses it), then it is misleading to the users. If it  is a real extension point (requests creation methods will use it), I would like to get an arch board approval for it.

annelo-msft:
> Sorry, but I don't think there is a value on adding Uri for the debugging purposes at all.

We decided to implement this feature based on the following:
- A customer request for this feature: https://github.com/Azure/azure-sdk-for-net/issues/24634
- Multiple requests from @tg-msft for this feature
- Offline discussion with @KrzysztofCwalina, who said this has been discussed in arch board and approved.

annelo-msft:
@AlexanderSher, would making the property non-virtual address your concerns?

tg-msft:
@KrzysztofCwalina - did we recently have a conversation across languages _(possibly in the context of management plane?)_ where we decided we'd use `Endpoint` as the default name for these URIs (in both .ctor params and these properties) and save `Uri` for resource hierarchies where we might have distinctions between the service endpoint and a specific resource URI?

annelo-msft:
@tg-msft, to close the loop here, yes, we confirmed what you've stated above with @KrzysztofCwalina in our DPG design meeting.  Thakns!

test/TestProjects/SingleTopLevelClientWithOperations-LowLevel/Generated/Client4.cs
AlexanderSher: If there is no endpoint, there would be no property at all. The origin of the issue is that property is required for logging purposes. In this perspective, I don't see the difference between client and subclient.

annelo-msft: @pshao25, here is some additional context: https://github.com/Azure/autorest.csharp/pull/1905#issuecomment-1036455832

Given this, we will not put a `Uri` property on subclients.  We will, however, put an `Endpoint` property on subclients, if they have an endpoint that is passed in to the subclient constructor.

samples/Azure.Management.Storage/Generated/RestOperations/BlobContainersRestOperations.cs
```diff
@@ -21,9 +21,11 @@ internal partial class BlobContainersRestOperations
     {
         private readonly string _userAgent;
         private readonly HttpPipeline _pipeline;
-        private readonly Uri _endpoint;
         private readonly string _apiVersion;
 
+        /// <summary> server parameter. </summary>
+        public Uri Endpoint { get; }
```
m-nash, line 27: Id prefer if we did this only in the public RestOperations objects.  If we update the RestClient model to add this property and remove the field only when its constructed with a public accessibility that would keep the code common but at the same time not introduce these types of unecessary changes all over the place

m-nash:
> @KrzysztofCwalina - did we recently have a conversation across languages _(possibly in the context of management plane?)_ where we decided we'd use `Endpoint` as the default name for these URIs (in both .ctor params and these properties) and save `Uri` for resource hierarchies where we might have distinctions between the service endpoint and a specific resource URI?

I don't have any context here so I am not sure what this renaming is intended to solve, where we should adhere to it, how it compares to what is the [ArmEnvironment ](https://github.com/Azure/azure-sdk-for-net/blob/f0718b12ea/sdk/resourcemanager/Azure.ResourceManager/src/ArmEnvironment.cs) etc.

samples/Azure.Management.Storage/Generated/RestOperations/BlobContainersRestOperations.cs
m-nash: If I understand the intent here.  Why wouldn't we simply add an Endpoint property to public Client and leave all of the internal classes alone?  Each of those internal classes knows about the endpoint only because its more efficient for them to cache the value, but the "owner" of that property is either the ClientOptions object for the service or the public Client for the service.  It seems adding a single property there would make more sense versus adding hundreds of properties all over the place in the internal clients that are simply copies of the original property.

KrzysztofCwalina: @m-nash, I tend to agree.

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
```diff
@@ -77,7 +77,7 @@ public ClientFields(IEnumerable<Parameter> parameters, BuildContext? context = n
 
             foreach (Parameter parameter in parameters)
             {
-                var field = parameter.IsResourceIdentifier
+                var field = parameter.IsResourceIdentifier || parameter.Name == "endpoint"
```
chunyu3, line 80: Can we not use hard-coded 'endpoint' here? 
And the code is out-of-date, need to rebase.

pshao25:
Don't close.

pshao25:
samples\Azure.Storage.Tables currently cannot add directives to rename url to endpoint, because of this [issue](https://github.com/Azure/autorest/issues/4573).

pshao25:
Do not close.

pshao25:
Do not close.

PR #1912, opened 2022-01-24
Update generator shared code to use new APIs in Azure.Core 1.22.0

annelo-msft:
Update generator shared code to use new APIs in Azure.Core 1.22.0, specifically:
- `Response.IsError` property
- `RequestFailedException(Response response)` constructor


pakrym (approved):

pakrym:
Time to do the non-shared code :). It's going to be great!

Issue #1840, opened 2022-01-12
Automatically apply abbreviation casing in the generator


ArcturusZhang:
Usually DotNet code requires the abbreviations in variable names using camel case, like `PublicIpAddress`, instead of `PublicIPAddress`.

We could automatically this rule to the code model in our generator to keep the API stable as much as we can.

PR #1928, opened 2022-01-27
Naming transformation

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1840

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,141 @@
...
+            { "dns", "Dns" },
+            { "http", "Http" },
+            { "https", "Https" },
+        };
+
+        internal static string EnsureNameCase(this string name, bool lowerFirst = false)
+        {
+            var words = name.SplitByCamelCase().ToArray();
+            var builder = new StringBuilder();
+            for (int i = 0; i < words.Length; i++)
```
m-nash, line 119: Lets see if we can improve this a bit from perf perspective.  First lets measure how long this would take against an RP like network, if its more than ~5 seconds lets investigate ways to improve

m-nash (changes requested):

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,141 @@
...
+            ApplyAcronymsDictionary(config);
+            foreach (var schema in codeModel.GetAllSchemas())
+            {
+                TransformSchema(schema);
+            }
+        }
+
+        public static IEnumerable<Schema> GetAllSchemas(this CodeModel codeModel)
+        {
+            return codeModel.Schemas.Choices.Cast<Schema>()
```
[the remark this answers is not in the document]
ArcturusZhang: This has been cached

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
```diff
@@ -121,6 +121,7 @@ public void Write(Utf8JsonWriter writer, string settingName)
             JsonElement? requestPathToSingletonResource = default,
             JsonElement? overrideOperationName = default,
             JsonElement? operationPositions = default,
+            JsonElement? acroynmsDictionary = default,
```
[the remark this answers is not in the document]
ArcturusZhang: This has been renamed to `rename-rules`

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,169 @@
...
+            var regexRawString = string.Join("|", RenameRules.Keys);
+            // we are building the regex that matches
+            // 1. it starts with a lower case letter or a digit which is considered as the end of its previous word
+            //    or it is at the beginning of this string
+            //    or we hit a word separator including \W, _, ., @, -, spaces
+            // 2. it either should match one of the candidates in the rename rules dictionary
+            // 3. it should followed by a upper case letter which is considered as the beginning of next word
+            //    or it is the end of this string
+            //    or we hit a word separator including \W, _, ., @, -, spaces
+            _regex = new Regex(@$"([\W|_|\.|@|-|\s|\$\da-z]|^)({regexRawString})([\W|_|\.|@|-|\s|\$A-Z]|$)");
```
archerzz, line 70: `a-z` may fail the case like `VMIp`?

archerzz: Maybe `split then replace each sub-word` is better than `parse and replace`? not sure if we could find a solution like https://lucene.apache.org/core/8_9_0/analyzers-common/org/apache/lucene/analysis/miscellaneous/WordDelimiterFilter.html

test/AutoRest.TestServer.Tests/Mgmt/Unit/CodeModelExtensionTests.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using AutoRest.CSharp.Mgmt.Decorator;
+using NUnit.Framework;
+
+namespace AutoRest.TestServer.Tests.Mgmt.Unit
```
m-nash, line 8: Can we also add a test class that inherits from testproject tests and asserts the specific expected names that we want to change in the mgmtrenamerules project?

m-nash (changes requested):

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: I did not find anything like this.
This regex is trying to catch "full words", there are always exceptions if we allow that, therefore we ensure a word can only start after a digit, lower case letter, or special symbols.
If the case of `VMIp` really happens, we could replace them by adding a new rule: `VMIp: VMIP`

test/AutoRest.TestServer.Tests/Mgmt/Unit/CodeModelExtensionTests.cs
ArcturusZhang: I will add that

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: Let me add these two into test cases
VMIp -> VmIP
VPNIp -> VpnIP

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,179 @@
...
+        public static void UpdateAcronyms(this IEnumerable<Schema> allSchemas, IReadOnlyDictionary<string, string> renameRules)
+        {
+            ApplyRenameRules(renameRules);
+            foreach (var schema in allSchemas)
+            {
+                TransformSchema(schema);
+            }
+        }
+
+        public static IEnumerable<Schema> GetAllSchemas(this CodeModel codeModel)
```
m-nash, line 97: Id prefer if we did this inline in the library ctor since this opens the opportunity to get called multiple times.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,179 @@
...
+            result = originalName.EnsureNameCase();
+            languages.Default.Name = result;
+            _cache.TryAdd(originalName, result);
+        }
+
+        private static void TransformObjectSchema(ObjectSchema objSchema)
+        {
+            // transform the name of this schema
+            TransformBasicSchema(objSchema);
+            foreach (var property in objSchema.Properties)
```
m-nash, line 143: is this necessary since the loop that calls this is from allSchemas?

test/AutoRest.TestServer.Tests/Mgmt/Unit/CodeModelExtensionTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        [TestCase("resourceIdSuffix", "ResourceIdSuffix")]
+        [TestCase("IpsilateralDisablity", "IpsilateralDisablity")]
+        [TestCase("HumanIpsilateralDisablity", "HumanIpsilateralDisablity")]
+        [TestCase("DnsIpAddressIDsForWindowsOs", "DnsIPAddressIdsForWindowsOS")]
+        [TestCase("DnsIpAddressIDsForWindowsOsPlatform", "DnsIPAddressIdsForWindowsOSPlatform")]
+        [TestCase("something_IDs_Ip_Os", "Something_Ids_Ip_OS")]
+        [TestCase("Dynamic365IDs", "Dynamic365Ids")]
+        [TestCase("VirtualWAN", "VirtualWan")]
+        [TestCase("VirtualWANs", "VirtualWans")]
+        public void EnsureNameCaseTest(string name, string expected)
```
m-nash, line 48: Add a list of comments somewhere of use cases that we know this algorithm doesn't cover so we are explicitly calling them out.  for instance myvm we aren't covering since we are relying on Pascal / camel case to split the words.

Are we covering variable names here?  how do our helper methods to convert from variable to property names deal with names that start with one of these acronyms in camelCase
oSPlatform -> osPlatform
oSPlatform -> OSPlatform
vMThingie -> vmThingie
The rule should be that one cap all cap one small all small

m-nash (changes requested):

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: makes sense.
Caching this here might cause concurrency issues while running test cases.
I removed this function and put it inline in outputLibrary

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: it is necessary because this is changing the property's name not the type name of that property. Like in the following schema
```
public class Something {
    public PropertyType PropertyName {get; set;}
}
```
we are changing the property name, not the name of property type.

test/AutoRest.TestServer.Tests/Mgmt/Unit/CodeModelExtensionTests.cs
ArcturusZhang: Unfortunately we are not covering variable names here. This is one of our known issues.
The name in the swagger might start with a upper case letter or a lower case letter. We are doing the transformation to variable names in [this function](https://github.com/Azure/autorest.csharp/blob/ea40037443519fdfd0bccbc4fc10918d90f46d0c/src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs#L76), but it is not actually looking for the words in a dictionary and will give you the result of `oSPlatform`. I already opened an issue to track that here: https://github.com/Azure/autorest.csharp/issues/1937
Considering all the things that go into public API start with an upper case letter, I did not take that into consideration in this PR.

m-nash: I think we will need to incorporate that here otherwise when we call ToVariableName in various places it will only ToLower the first character and not the full initialism

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: @archerzz I tweaked the algorithm so that we could catch cases like this under the condition that the first abbreviation is in the dictionary. For instance, now we can properly transform `VPNIp` to `VpnIP`, or `VPNIP` to `VpnIP` as long as we have `VPN: Vpn` mapping in our dictionary.

archerzz (approved):

ArcturusZhang:
By running the bench mark on network RP, the new implementation is slightly slow comparing from the old implementation:
```
|                     Method |      Mean |     Error |    StdDev |     Gen 0 | Allocated |
|--------------------------- |----------:|----------:|----------:|----------:|----------:|
| OldEnsureCasingInNetworkRp |  5.766 ms | 0.1133 ms | 0.2264 ms | 1828.1250 |      8 MB |
|    EnsureCasingInNetworkRp | 10.659 ms | 0.2056 ms | 0.5270 ms |  265.6250 |      1 MB |
```
But it is still under 1s, therefore i believe it is acceptable.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,169 @@
...
+using System.Text.RegularExpressions;
+using AutoRest.CSharp.AutoRest.Plugins;
+using AutoRest.CSharp.Input;
+using AutoRest.CSharp.Utilities;
+
+namespace AutoRest.CSharp.Mgmt.Decorator
+{
+    internal static class CodeModelExtension
+    {
+        private static readonly ConcurrentDictionary<string, string> _cache = new ConcurrentDictionary<string, string>();
```
AlexanderSher, line 19: We already have problems with static dictionaries. Please don't add another one.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,169 @@
...
+                    foreach (var p in op.Parameters)
+                    {
+                        //updater the first subscriptionId to be 'method'
+                        if (!setSubParam && p.Language.Default.Name.Equals("subscriptionId", StringComparison.OrdinalIgnoreCase))
+                        {
+                            setSubParam = true;
+                            p.Implementation = ImplementationLocation.Method;
+                        }
+                        //updater the first subscriptionId to be 'method'
+                        if (p.Language.Default.Name.Equals("apiVersion", StringComparison.OrdinalIgnoreCase))
```
AlexanderSher, line 52: why ignore case?

test/TestProjects/MgmtRenameRules/MgmtRenameRules.json
```diff
@@ -0,0 +1,11761 @@
+{
+    "swagger": "2.0",
+    "info": {
+        "title": "ComputeManagementClient",
+        "description": "The Compute Management Client.",
+        "version": "2020-06-01"
+    },
+    "host": "management.azure.com",
```
AlexanderSher: Please live only the parts are essential to the test, and remove the rest.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,169 @@
...
+                TransformLanguage(property.Language);
+            }
+        }
+
+        private static IDictionary<string, string> RenameRules = new Dictionary<string, string>();
+
+        internal static string EnsureNameCase(this string name)
+        {
+            if (_regex == null)
+                return name; // this means we do not have any replacement rules
```
AlexanderSher, line 146: It also maybe because method is called out of order. Please differentiate these two cases.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -0,0 +1,169 @@
...
+                            p.Implementation = ImplementationLocation.Client;
+                        }
+                    }
+                }
+            }
+        }
+
+        private static Regex? _regex;
+
+        internal static void ApplyRenameRules(IReadOnlyDictionary<string, string> renameRules)
```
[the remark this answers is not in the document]
ArcturusZhang: only once, but our test case needs to use this as well, therefore I made it internal. Do we have other solution for this? I do not like this as well

test/AutoRest.TestServer.Tests/Mgmt/Unit/CodeModelExtensionTests.cs
m-nash: We will do this in a separate PR

m-nash (approved):

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: Well, this is not introduced by my PR, I just move this function from MgmtOutputLibrary to here, since this function also works as an extension method of codeModel.
@m-nash do you know the answer?

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: Sure, I will introduce a singleton class for the renaming logic.
Out of curiosity, what problems are from the static dictionaries? One of them I know is that static dictionaries will consume a lot of memories.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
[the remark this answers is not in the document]
ArcturusZhang: If we convert this class to a singleton, I believe this function will be guaranteed to be invoked only once.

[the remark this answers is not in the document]
ArcturusZhang: I put the dictionary of rename rules to every function's parameter list, now we no longer need this function any more and it is removed

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: I change the cache to a local variable and pass it to every function in the naming transformation process.
We truly do not need to preserve this cache forever, therefore I believe a local variable will be enough

test/TestProjects/MgmtRenameRules/MgmtRenameRules.json
ArcturusZhang: I made a shim on this test project, removed a lot of things that are irrelevant to the renaming scenarios.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: I changed the signature by adding renameRules dictionary to the function signature, please take another look

ArcturusZhang:
Hi @AlexanderSher I made some changes, could you please have a look again?

m-nash (approved):

PR #1946, opened 2022-02-04
Update schema to to Uri when property name ends with Uri and is type string

m-nash:
# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
+++ b/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
@@ -87,6 +87,8 @@ namespace AutoRest.CSharp.Mgmt.AutoRest
                 .Concat(_codeModel.Schemas.Objects)
                 .Concat(_codeModel.Schemas.Groups);
 
+            UpdateFrameworkTypes(_allSchemas);
+
             // We can only manipulate objects from the code model, not RestClientMethod
             ReorderOperationParameters();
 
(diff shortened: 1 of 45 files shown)


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -97,6 +99,21 @@ public MgmtOutputLibrary(CodeModel codeModel, BuildContext<MgmtOutputLibrary> co
...
+        private void UpdateFrameworkTypes(IEnumerable<Schema> allSchemas)
+        {
+            foreach (var schema in _allSchemas)
+            {
+                if (schema is not ObjectSchema objSchema)
+                    continue;
+
+                foreach (var property in objSchema.Properties)
+                {
+                    if (property.Language.Default.Name.EndsWith("Uri"))
```
ShivangiReja, line 111: Is it possible that the property name can only be "uri"? or is it always ends with "Uri" like "somethingUri"?

m-nash: it can be, but we will rely on human intervention to find the edge cases and add a model rename directive to Uri then it will be picked up.  If we go down the edge case path this logic because very complicated very quickly.

Its also possible that a word ends in uri like the plural for thesaurus is thesauri and we wouldn't want to change the type in that case.

m-nash: This is an example diff of the compute API regenerated with this code https://apiview.dev/Assemblies/Review/bb81bcdd82c0415ab8a34c7fb89d5dea?diffRevisionId=f47dc3d72d224644992d6321ccb0d9cb&doc=False&diffOnly=True

ShivangiReja (approved):

PR #1949, opened 2022-02-06
Normalize Update body parameter

m-nash:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -99,6 +103,69 @@ public MgmtOutputLibrary(CodeModel codeModel, BuildContext<MgmtOutputLibrary> co
...
+                        if (bodyParam is null)
+                            continue;
+
+                        if (!usageCounts.TryGetValue(bodyParam.Schema, out var counts))
+                            continue;
+
+                        if (counts.Count != 1 || !counts.TryGetValue(httpRequest.Method, out var count) || count != 1)
+                            continue;
+
+                        bodyParam.Schema.Language.Default.Name = $"{operationGroup.Key.LastWordToSingular()}UpdateOptions";
```
ArcturusZhang, line 162: My biggest concern here is that in quite a lot cases, the key of operation group is not the same as the resource name in RPs. Could we have some trial runs on the RPs like `appservice`, or `sql` and see the result?
Note that they might already have some directive to do the renamings, we might need to remove those first before we generate code for them

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -525,6 +598,73 @@ public bool TryGetRestClients(string requestPath, [MaybeNullWhen(false)] out Has
...
+
+            var resourcesWithSameName = _resourceDataSchemaNameToOperationSets[candidateName];
+            var resourcesWithSameType = OperationSets.SelectMany(opSet =>
+            {
+                var operation = opSet.Operations.First();
+                return GetRequestPath(opSet[operation], operation).Expand(_mgmtConfiguration);
+            }).Where(rqPath =>
+            {
+                return rqPath.Count % 2 == 0 && rqPath.GetResourceType(_mgmtConfiguration).Equals(resourceType);
+            });
```
ArcturusZhang, line 626: I have some concerns about this assert of "request path must have even segments" - it is not true for a path of a scope resource, like `/{scope}/providers/Microsoft.Resources/policyAssignments/{name}` has odd segments, but it is a path of a resource.
How about we loop over the operation sets that are guaranteed to be a resource's?  Like this
```suggestion
            var resourcesWithSameType = _resourceDataSchemaNameToOperationSets.Values.SelectMany(opSet =>
            {
                var operation = opSet.Operations.First();
                return GetRequestPath(opSet[operation], operation).Expand(_mgmtConfiguration);
            }).Where(rqPath =>
            {
                return rqPath.Count % 2 == 0 && rqPath.GetResourceType(_mgmtConfiguration).Equals(resourceType);
            });
```

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -525,6 +598,73 @@ public bool TryGetRestClients(string requestPath, [MaybeNullWhen(false)] out Has
...
+                var types = resourceType.Types;
+                var name = string.Join("", types.Select(segment => segment.ConstantValue.LastWordToSingular().FirstCharToUpperCase()));
+                if (countOfSameResourceTypeName == 1)
+                    return name;
+
+                var parentOperationSet = requestPath.GetScopePath().GetResourceType(_mgmtConfiguration).Last().ConstantValue;
+                // if countOfSameResourceTypeName > 1, we will have to add the scope as prefix to fully qualify the resource type name
+                // first we try to add the parent name as prefix
+                if (!resourcesWithSameType.All(rqPath => rqPath.Equals(requestPath)))
+                    return $"{parentOperationSet.FirstCharToUpperCase().ToSingular()}{name}";
```
ArcturusZhang, line 648: We should use `LastWordToSingular`
```suggestion
                    return $"{parentOperationSet.LastWordToSingular().FirstCharToUpperCase()}{name}";
```

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -525,6 +598,73 @@ public bool TryGetRestClients(string requestPath, [MaybeNullWhen(false)] out Has
...
+
+                // if we get here, parent prefix is not enough, we try the resource name if it is a constant
+                if (requestPath.Last().IsConstant)
+                    return $"{requestPath.Last().ConstantValue}{name}";
+
+                // if we get here, we have tried all approaches to get a solid resource type name, throw an exception
+                throw new InvalidOperationException($"Cannot determine a resource class name for resource with the request path(s): {requestPath}, please assign a valid resource name in `request-path-to-resource-name` section");
```
ArcturusZhang, line 655: Since we will always have only one path
```suggestion
                throw new InvalidOperationException($"Cannot determine a resource class name for resource with the request path: {requestPath}, please assign a valid resource name in `request-path-to-resource-name` section");
```

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -127,8 +127,8 @@ protected internal Resource(IReadOnlyDictionary<OperationSet, IEnumerable<Operat
             yield return new FieldDeclaration(FieldModifiers, ResourceData.Type, DataFieldName);
         }
 
-        public Resource(IReadOnlyDictionary<OperationSet, IEnumerable<Operation>> allOperations, string resourceName, ResourceTypeSegment resourceType, ResourceData resourceData, BuildContext<MgmtOutputLibrary> context)
-            : this(allOperations, resourceName, resourceType, resourceData, context, ResourcePosition)
+        public Resource(IReadOnlyDictionary<OperationSet, IEnumerable<Operation>> allOperations, string resourceDataSchemaName, ResourceTypeSegment resourceType, ResourceData resourceData, BuildContext<MgmtOutputLibrary> context)
```
ArcturusZhang, line 130: Could you please add a `TODO` here so that we could track the refactor of making a 1-1 map between OperationSet and Resource?
I will make the refactor in another PR once yours merges

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -127,8 +127,8 @@ protected internal Resource(IReadOnlyDictionary<OperationSet, IEnumerable<Operat
...
+            : this(allOperations, resourceDataSchemaName, resourceType, resourceData, context, ResourcePosition)
```
ArcturusZhang, line 131: Considering how this constructor is invoked, this `resourceDataSchemaName` now is really the name of this resource now. Could we change the name back?
```suggestion
            : this(allOperations, resourceName, resourceType, resourceData, context, ResourcePosition)
```

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -525,6 +597,73 @@ public bool TryGetRestClients(string requestPath, [MaybeNullWhen(false)] out Has
...
+            if (_context.Configuration.MgmtConfiguration.RequestPathToResourceName.TryGetValue($"{operationSet.RequestPath}|{resourceType}", out name))
+                return name;
+
+            return null;
+        }
+
+        private string GetResourceName(string candidateName, OperationSet operationSet, RequestPath requestPath)
+        {
+            // read configuration to see if we could get a configuration for this resource
+            var resourceType = requestPath.GetResourceType(_mgmtConfiguration);
```
ArcturusZhang, line 613: I have a feeling that the `requestPath` here is only constructed from the path segments, therefore it must be a "unexpanded" path, therefore the resource type from this path directly will have variables in it.
We might need to find a way to pin this to a specific resource type. But how? I did not think of a way from the information we have here.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -525,6 +612,115 @@ public bool TryGetRestClients(string requestPath, [MaybeNullWhen(false)] out Has
...
+            var defaultNameFromConfig = GetDefaultNameFromConfiguration(operationSet, resourceType);
+            if (defaultNameFromConfig != null)
+                return defaultNameFromConfig;
+
+            var resourcesWithSameName = _resourceDataSchemaNameToOperationSets[candidateName];
+            var resourcesWithSameType = ResourceOperationSets
+                .SelectMany(opSet => opSet.GetRequestPath(_context).Expand(_mgmtConfiguration))
+                .Where(rqPath => rqPath.GetResourceType(_mgmtConfiguration).Equals(resourceType))
+                .ToArray();
```
ArcturusZhang, line 637: Looks like we are only iterating this collection, therefore I believe we do not need to convert this to an array.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
m-nash: this is the expanded resource type

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
m-nash: it made debugging a LOT easier but I can remove that now

fengzhou-msft (approved):

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -8,6 +8,7 @@
 using System.Linq;
 using AutoRest.CSharp.Generation.Types;
 using AutoRest.CSharp.Input;
+using AutoRest.CSharp.Mgmt.AutoRest;
```
AlexanderSher, line 11: Common code shouldn't reference Mgmt namespace

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -149,7 +160,13 @@ public RestClientMethod BuildMethod(Operation operation, HttpRequest httpRequest
                 .Where(rp => !IsIgnoredHeaderParameter(rp))
                 .ToArray();
 
-            return parameters.ToDictionary(rp => rp, requestParameter => BuildParameter(requestParameter));
+            var result = new Dictionary<RequestParameter, Parameter>();
+            foreach (var parameter in parameters)
+            {
+                if (!result.ContainsKey(parameter))
```
AlexanderSher, line 166: It is not normal to have one and the same parameter in the method more than once. It may be, however, a side effect of incorrectly transformed swagger. We should fail here rather than swallow it.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -108,6 +109,16 @@ public RestClientMethod BuildRequestMethod(Operation operation, ServiceRequest s
             );
         }
 
+        public static PathSegment[] BuildRequestPathSegments(Operation operation, ServiceRequest serviceRequest, HttpRequest httpRequest, RestClientBuilder restClientBuilder)
+        {
+            var allParameters = restClientBuilder.GetOperationAllParameters(operation, operation.Parameters);
+            var methodParameters = restClientBuilder.BuildMethodParameters(allParameters);
+            var references = allParameters.ToDictionary(kvp => GetRequestParameterName(kvp.Key), kvp => new ParameterInfo(kvp.Key, restClientBuilder.CreateReference(kvp.Key, kvp.Value)));
+            var buildContext = new RequestMethodBuildContext(methodParameters, references);
+            BuildRequestParameters(httpRequest, buildContext, out var queryParams, out var headerParams, out var uriParams, out var pathParams);
```
AlexanderSher, line 118: This method creates lots of objects that are just thrown away. However, from what I understand, `PathSegment` instances are not needed. `PathSegment.ParsePathSegment` uses only `PathSegment.Value` and `PathSegment.Escape`. The first one is provided by `CreateReference`, the last one is simply `!requestParameter.Extensions!.SkipEncoding`. So something like:

```cs
 public IReadOnlyDictionary<string, ParameterInfo> GetReferencesToOperationParameters(Operation operation, IEnumerable<RequestParameter> requestParameters)
{
    var allParameters = GetOperationAllParameters(operation, requestParameters);
    return allParameters.ToDictionary(kvp => GetRequestParameterName(kvp.Key), kvp => new ParameterInfo(kvp.Key, CreateReference(kvp.Key, kvp.Value)));
}
```

should be enough.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -108,6 +109,16 @@ public RestClientMethod BuildRequestMethod(Operation operation, ServiceRequest s
             );
         }
 
+        public static PathSegment[] BuildRequestPathSegments(Operation operation, ServiceRequest serviceRequest, HttpRequest httpRequest, RestClientBuilder restClientBuilder)
```
AlexanderSher, line 112: Method signature where type instance is passed into a static method of that same type is quite unnatural. This should be a normal instance method.

src/AutoRest.CSharp/Common/Utilities/CachedDictionary.cs
```diff
@@ -25,11 +25,13 @@ public CachedDictionary(Func<Dictionary<K, V>> populateProc)
             _populateProc = populateProc;
         }
 
+        public bool IsPopulated => _values != null;
```
AlexanderSher, line 28: Why do you need this?

src/AutoRest.CSharp/Mgmt/Models/OperationSet.cs
```diff
@@ -103,6 +107,36 @@ public bool Equals([AllowNull] OperationSet other)
             return null;
         }
 
+        public RequestPath GetRequestPath(ResourceTypeSegment? hint = null)
+        {
+            return hint.HasValue ? NonHintRequestPath.ApplyHint(hint.Value) : NonHintRequestPath;
+        }
+
+        private RequestPath? _nonHintRequestPath;
+        public RequestPath NonHintRequestPath => _nonHintRequestPath ??= GetNonHintRequestPath();
```
AlexanderSher, line 116: `Lazy<RequestPath>`?

src/AutoRest.CSharp/Mgmt/Models/RequestPath.cs
```diff
@@ -79,13 +79,40 @@ internal struct RequestPath : IEquatable<RequestPath>, IReadOnlyList<Segment>
...
+            var segments = pathSegments
                 .SelectMany(pathSegment => ParsePathSegment(pathSegment, ref index))
                 .ToList();
             _segments = CheckByIdPath(segments);
-            SerializedPath = method.Operation.GetHttpPath();
+            SerializedPath = httpPath;
+        }
+
+        private bool? _isExpandable = null;
+        public bool IsExpandable => _isExpandable ??= GetIsExpandable();
```
AlexanderSher, line 102: `RequestPath` is a struct, so `GetIsExpandable` will be called for every new copy of this struct.

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -58,5 +58,23 @@ public static string GetDefaultNamespace(string? namespaceExtension, BuildContex
             }
             return result;
         }
+
+        public override bool Equals(object? obj)
```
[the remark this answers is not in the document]
m-nash: we create multiple instances of the same type but they aren't treated as equal so this is used in the csharptype.equals it compares the implementation which is a typeprovider

src/AutoRest.CSharp/Common/Utilities/CachedDictionary.cs
m-nash: to allow us to partially build the parameters to get the path segments for a method before we fully initialize everything

PR #1951, opened 2022-02-06
Update to use ArmOperation implementation

m-nash:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Generation/MgmtLongRunningOperationWriter.cs
```diff
@@ -2,157 +2,127 @@
...
+            _name = $"{context.DefaultNamespace.Split('.').Last()}ArmOperation";
+            _genericString = isGeneric ? "<T>" : string.Empty;
+            Filename = isGeneric ? $"LongRunningOperation/{_name}OfT.cs" : $"LongRunningOperation/{_name}.cs";
+            _waitMethod = isGeneric ? "WaitForCompletionAsync" : "WaitForCompletionResponseAsync";
+            _operationType = isGeneric ? typeof(ArmOperation<>) : typeof(ArmOperation);
+            _responseType = isGeneric ? typeof(Response<>) : typeof(Response);
+            _operationOrResponseType = isGeneric ? typeof(OperationOrResponseInternals<>) : typeof(OperationOrResponseInternals);
+            _operationSourceString = isGeneric ? (FormattableString)$"{typeof(IOperationSource<>)} source, " : (FormattableString)$"";
+            _sourceString = isGeneric ? "source, " : string.Empty;
         }
```
ArcturusZhang, line 49: After the refactor, our writer should only do the writing and do not calculate anything. Therefore should we also introduce a `TypeProvider` to do the calculations and make this writer only writes things?

src/AutoRest.CSharp/Mgmt/Models/MgmtRestOperation.cs
```diff
@@ -93,9 +99,15 @@ internal record MgmtRestOperation
 
         public bool IsLongRunningOperation => _isLongRunning.HasValue ? _isLongRunning.Value : Operation.IsLongRunning;
 
+        public bool IsFakeLro => IsLongRunningOperation && !Operation.IsLongRunning;
```
ArcturusZhang, line 102: I personally prefer `IsFakeLongRunningOperation` for this property comparing to the property of `IsLongRunningOperation`. What do you think? Or we could rename the other to `IsLro`

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -99,6 +99,11 @@ public MgmtOutputLibrary(CodeModel codeModel, BuildContext<MgmtOutputLibrary> co
             DecorateOperationSets();
         }
 
+        public Dictionary<CSharpType, OperationSource> CSharpTypeToOperationSource { get; } = new Dictionary<CSharpType, OperationSource>();
```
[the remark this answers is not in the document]
m-nash: This is only to make sure we don't create duplicates.

src/AutoRest.CSharp/Mgmt/Generation/MgmtLongRunningOperationWriter.cs
m-nash: Since this is a static class meaning the definition will never change I think its ok.  Type providers make it nice for variable signatures / methods / ctors / etc

ArcturusZhang: Emmm, did not get the point. Neither is this class a static class, nor the class that this writer is writing. Could you please be more specific?

m-nash: The definition is static.  The methods, ctors, signatures will never change.  Unlike a resource or restOperations etc where the methods are constantly different.

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -108,6 +110,12 @@ private MgmtClientOperation(IReadOnlyList<MgmtRestOperation> operations, BuildCo
 
         public CSharpType? ListItemType => _operations.First().ListItemType;
 
+        public CSharpType? MgmtReturnType => _operations.First().MgmtReturnType;
+
+        public bool IsFakeLro => _operations.First().IsFakeLongRunningOperation;
```
ArcturusZhang, line 115: We might also rename this property as well

src/AutoRest.CSharp/Mgmt/Generation/MgmtLongRunningOperationWriter.cs
ArcturusZhang: OK, make sense. But I still think we should introduce a `TypeProvider` for this, for consistency. And also we will do not have to concatenating the name of this class again like here: https://github.com/Azure/autorest.csharp/blob/b07f41b5b41e1f2fceef7ca4b0b638aaf0ff1d91/src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs#L68

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -749,44 +750,52 @@ protected virtual void WriteLROMethodBody(MgmtClientOperation clientOperation, D
...
+        protected virtual void WriteLROResponse(string diagnosticsVariableName, string pipelineVariableName, MgmtRestOperation operation, IEnumerable<ParameterMapping> parameterMapping, bool async)
         {
-            _writer.Append($"var operation = new {lroObjectType}(");
-
-            if (operation.Operation.IsLongRunning)
+            _writer.Append($"var operation = new {LibraryArmOperation}");
+            if (operation.ReturnType.Arguments.Length > 0)
+            {
+                _writer.Append($"<{operation.ReturnType.Arguments[0]}>");
+            }
```
ArcturusZhang, line 769: Do you think it makes sense if we could put this type `LibraryArmOperation<operation.ReturnType.Arguments[0]>` as a property of `MgmtRestOperation`?

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/AvailabilitySetOperationSource.cs
```diff
@@ -0,0 +1,40 @@
...
+using Azure.Core;
+using Azure.ResourceManager;
+
+namespace Azure.ResourceManager.Sample
+{
+    internal class AvailabilitySetOperationSource : IOperationSource<AvailabilitySet>
+    {
+        private readonly ArmClient _client;
+
+        internal AvailabilitySetOperationSource(ArmClient client)
```
ArcturusZhang, line 21: I found this class is not used in this package and I suppose it is because that every operation related with this is a fake LRO.
Is it possible that we do not generate the `OperationSource` classes if the corresponding operations are fake LROs?
But considering these classes are internal, it is OK if we generate those.

src/AutoRest.CSharp/Mgmt/Generation/MgmtLongRunningOperationWriter.cs
m-nash: Lets make a future work item for this post GA given our tight timelines right now

ShivangiReja (approved):
LGTM

Issue #1931, opened 2022-01-28
Add request path and `operationId` in the summary of an operation


ArcturusZhang:
Now we only add request path, operationId and contextual path when we enable the flag `mgmt-debug.show-request-path`. Since this information is valuable, we could consider to move request path and operationId to summary of every operation. ContextualPath is really debug purpose, we could remove that. And if we do this, we could also remove the flag `mgmt-debug.show-request-path` since we are doing that by default.

PR #1956, opened 2022-02-08
Add request path and operation id in summary

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1931

# Description

# Checklist

# Ready to Land?


samples/Azure.Management.Storage/Generated/BlobContainer.cs
```diff
@@ -90,10 +90,11 @@ public virtual ImmutabilityPolicy GetImmutabilityPolicy()
             return new ImmutabilityPolicy(Client, new ResourceIdentifier(Id.ToString() + "/immutabilityPolicies/default"));
         }
 
-        /// RequestPath: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices/default/containers/{containerName}
-        /// ContextualPath: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices/default/containers/{containerName}
-        /// OperationId: BlobContainers_Get
-        /// <summary> Gets properties of a specified container. </summary>
+        /// <summary>
+        /// Gets properties of a specified container. 
+        /// Request Path: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices/default/containers/{containerName}
```
m-nash, line 95: I believe this is technically the resource path?  Can we output the request path as well which I believe was originally called the contextualPath?

m-nash (changes requested):

samples/Azure.Management.Storage/Generated/BlobContainer.cs
ArcturusZhang: No, this request path is the path of this operation. Contextual path is the path pattern of the Id in this class. For instance, if this operation is in a resource class, the contextual path will be the request path of this resource (the path of Get operation). If this operation is in a collection class, the contextual path will be the request path of its parent, since the Id in this class is its parent's.
I hide the contextual path since it usually does not help the SDK and not so intuitive. We could always see that if anything bad happens during debug.

m-nash (approved):

PR #1959, opened 2022-02-09
Make Data Plane and Management Plane RestClient fields consistent with LLC Client

ShivangiReja:
# Description
* Make Data Plane and Management Plane RestClient fields consistent with LLC Client
* Add description for ClientDiagnostics field


src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -96,8 +96,8 @@ private static void WriteClientFields(CodeWriter writer, LowLevelClient client)
                 writer.WriteFieldDeclaration(field);
             }
 
+            writer.Line();
```
AlexanderSher, line 99: ?

annelo-msft (approved):
Approved with regard to output files for data plane rest client and llc clients.  Please let @m-nash take a look to approve the management outputs as well.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -79,7 +79,7 @@ public static CodeWriter WriteFieldDeclaration(this CodeWriter writer, FieldDecl
 
             if (field.WriteAsProperty)
             {
-                writer.AppendRaw(modifiers.HasFlag(FieldModifiers.ReadOnly) ? "{ get; }" : "{ get; set; }");
+                writer.LineRaw(modifiers.HasFlag(FieldModifiers.ReadOnly) ? "{ get; }" : "{ get; set; }");
```
AlexanderSher, line 82: I'm not sure about this one. You probably want to change the line below:
```cs
return field.WriteAsProperty ? writer.Line() : writer.Line($";");
```

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
```diff
@@ -107,7 +105,11 @@ private void WriteClientCtor(CodeWriter writer, RestClient restClient, CSharpTyp
             {
                 foreach (Parameter clientParameter in restClient.Parameters)
                 {
-                    writer.WriteVariableAssignmentWithNullCheck($"this.{clientParameter.Name}", clientParameter);
+                    var field = restClient.Fields.GetFieldByParameter(clientParameter);
+                    if (field != null)
+                    {
+                        writer.WriteVariableAssignmentWithNullCheck($"{field.Declaration.ActualName}", clientParameter);
```
AlexanderSher, line 111:
```cs
writer.WriteVariableAssignmentWithNullCheck(field.Name, clientParameter);
```

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
```diff
@@ -99,7 +104,7 @@ public ClientFields(BuildContext<LowLevelOutputLibrary> context, IEnumerable<Par
             {
                 "credential" when _keyAuthField != null && parameterType.EqualsIgnoreNullable(_keyAuthField.Type) => _keyAuthField,
                 "credential" when _tokenAuthField != null && parameterType.EqualsIgnoreNullable(_tokenAuthField.Type) => _tokenAuthField,
-                var name => _parameterNamesToFields.TryGetValue(name, out var field) ? field : null
+                var name => _parameterNamesToFields.TryGetValue(name, out var field) ? parameterType.EqualsIgnoreNullable(field.Type) ? field : null : null
```
AlexanderSher, line 107: Considering the way `FieldDeclaration` instances are created, I'm not sure I understand when `parameterType.EqualsIgnoreNullable(field.Type)` is false.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
ShivangiReja: It's the same as AppendRaw() + Line(). So result will be the same, either we do LineRaw() here or we do AppendRaw() and add an empty line later.

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
ShivangiReja: In case of name conflicts with parameter we need to check the type and name both. For example we have [this](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/NameConflicts/Generated/AutoRestParameterFlatteningRestClient.cs#L37) case where pipeline, clientDiagnostics are string and if we check name only it will replace these with client fields which will be incorrect in this case.

Here is the result if we don't check the type:

image

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
ShivangiReja: Added this to add an empty line before the HttpPineline field. [Here](https://github.com/Azure/autorest.csharp/pull/1959/files#diff-0898eb7a22b4bff2482529f6c3099d51aa3268edfb3db4c42c079ed6fb617044) is the generated result.

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
AlexanderSher: Hm... we can assign not-null value to nullable variable, but not the opposite. The problem is that from inside the method, we don't know if it is parameter assigned to field or field assigned to something.

| | Field type not-nullable  | Field type nullable |
| - | ------------- | ------------- |
| **Parameter type not-nullable** | Can assign both ways  | Can't assign field to something of parameter type |
| **Parameter type nullable** | Can't assign parameter to field  | Can assign both ways  |

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
AlexanderSher: Yes, but before adding a new line, we can assign a default value, e.g.:
```cs
public string MyProp { get; set; } = "abc";
```

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
AlexanderSher: Let's just use `parameterType.Equals(field.Type)`

AlexanderSher: If it doesn't work, we will need a flag (let's call it `assignToField`).

When `assignToField == true`:
| | Field type not-nullable  | Field type nullable |
| - | ------------- | ------------- |
| **Parameter type not-nullable** | true | true |
| **Parameter type nullable** | false | true |

When `assignToField == false`:
| | Field type not-nullable  | Field type nullable |
| - | ------------- | ------------- |
| **Parameter type not-nullable** | true | false |
| **Parameter type nullable** | true | true |


In code it would be something like this:
```cs
parameterType.Equals(field.Type) || (assignToField ? parameterType.WithNullable(true).Equals(field.Type) : parameterType.Equals(field.Type.WithNullable(true))
```

ShivangiReja: Equals works fine but there was an issue with the Credential Types. When we create ctor parameters for Credential Types we set `ValidateNotNull` true meaning `nullable: "false"` (check [here](https://github.com/Azure/autorest.csharp/blob/c2db691bf56576ce98a41a6e34aa24354c1e2d5b/src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs#L764)) and when we create fields we set `nullable: "true"` (check [here](https://github.com/Azure/autorest.csharp/blob/c2db691bf56576ce98a41a6e34aa24354c1e2d5b/src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs#L56))

If we set same #nullable settings, Equals works fine.

m-nash (approved):
I prefer if we couple the removal of ClientDiagnostics in here since this is going to be a lot of file changes in azure-sdk-for-net and we could avoid needing to review a massive PR like that twice

Issue #1960, opened 2022-02-09
`None` should be value 0 in sealed choices


ArcturusZhang:
The enum definition like this
```
public enum ResourceSkuCapacityScaleType {
    Automatic = 0,
    Manual = 1,
    None = 2,
}
```
might not be intuitive enough. We would like to change it automatically to 
```
public enum ResourceSkuCapacityScaleType {
    None = 0,
    Automatic = 1,
    Manual = 2,
}
```

This also applies to the wordings that are equivalent to `None`, such like `NoEffect`, `NotSpecified`, etc

PR #1961, opened 2022-02-09
Put `None` or equivalent items at the first in seal choices

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1960

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -12,12 +12,54 @@
...
+        }
+
+        private static ICollection<ChoiceValue> RearrangeChoices(ICollection<ChoiceValue> originalValues)
+        {
+            IEnumerable<ChoiceValue> whateverLeft = originalValues;
+            var result = new List<ChoiceValue>();
+
+            var words = EnumValuesShouldBePrompted.Concat(MgmtContext.MgmtConfiguration.PromptedEnumValues);
+
+            static Func<ChoiceValue, bool> GetFilter(string word)
```
m-nash, line 46: I wonder if this is simpler / more efficient to get the index of the one we want to move.  (this assumes we only ever want to move one which I think we need to assert) Then do a RemoveAt(index) and InsertAt(0, item)

m-nash (changes requested):

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -12,12 +12,54 @@
...
+                choiceSchema.Choices = RearrangeChoices(choiceSchema.Choices);
+            }
+        }
+
+        private static ICollection<ChoiceValue> RearrangeChoices(ICollection<ChoiceValue> originalValues)
+        {
+            IEnumerable<ChoiceValue> whateverLeft = originalValues;
+            var result = new List<ChoiceValue>();
+
+            var words = EnumValuesShouldBePrompted.Concat(MgmtContext.MgmtConfiguration.PromptedEnumValues);
```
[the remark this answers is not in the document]
ArcturusZhang: Good catch. I will fix this

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: I kind of prefer to keep this. The case is that we might have multiple entries in an enum that falls into the list, if we change this into some indice manipulations, it might be more complicated.
And since the iterations are constructing by linq, those will only execute once when we are iterating on it (the `ToArray` invocation). Therefore I believe there is no performance impact here.

ArcturusZhang:
Hi @m-nash I have made some update, please take a look

m-nash (approved):

Issue #1937, opened 2022-01-29
`ToVariableName` function cannot properly handle the property name with acronyms


ArcturusZhang:
We are using this function [`ToVariableName`](https://github.com/Azure/autorest.csharp/blob/ea40037443519fdfd0bccbc4fc10918d90f46d0c/src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs#L76) to generate the name of a variable (starting with lower case letter), but it cannot properly handle the case of a name starting with acronyms like `OSType`. Now it produces `oSType` which is incorrect.

PR #1966, opened 2022-02-11
Fix `ToVariableName`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1937

# Description

Previously the `ToVariableName` function or the `ToCleanName` function with the `camelCase=false` will only change the first letter to lower case, make the letter after a separator (such as `_`) upper case, do not do anything else. We found some issues in that implementation that if we have the correct casing in our properties, such like `OSProfile`, this function will only give us the result `oSProfile` because it does not identify the full first word.

This PR changes this function to make sure it could convert all letters in the first word to lower case, where the "first word" refers to the word before the last upper case letter or before a separator, whichever comes first

# Checklist

# Ready to Land?


test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs
```diff
@@ -79,6 +79,24 @@ public void ValidateSplitByCamelCase(string camelCase, string[] expected)
...
+        [TestCase("OSProfile", true, "OSProfile")]
+        [TestCase("OSProfile", false, "osProfile")]
+        [TestCase("OSProfileId", true, "OSProfileId")]
+        [TestCase("OSProfileId", false, "osProfileId")]
+        [TestCase("VCenterIP", true, "VCenterIP")]
+        [TestCase("VCenterIP", false, "vCenterIP")]
+        [TestCase("D365IP", true, "D365IP")]
+        [TestCase("D365IP", false, "d365IP")]
+        [TestCase("P2SVpn", true, "P2SVpn")]
+        [TestCase("P2SVpn", false, "p2SVpn")]
```
AlexanderSher, line 99: I'd expect this one to be `P2SVpn -> p2sVpn`.

test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs
```diff
@@ -79,6 +79,24 @@ public void ValidateSplitByCamelCase(string camelCase, string[] expected)
         [TestCase("_propertyName", true, "PropertyName")]
         [TestCase("_property_name", false, "propertyName")]
         [TestCase("_property_name", true, "PropertyName")]
+        [TestCase("_OS_profile", true, "OSProfile")]
```
AlexanderSher, line 82: Can we add cases with 2 underscores?

test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs
```diff
@@ -79,6 +79,24 @@ public void ValidateSplitByCamelCase(string camelCase, string[] expected)
...
+        [TestCase("_OS_profile", false, "osProfile")]
+        [TestCase("_OS_Profile", true, "OSProfile")]
+        [TestCase("_OS_Profile", false, "osProfile")]
+        [TestCase("OS", true, "OS")]
+        [TestCase("OS", false, "os")]
+        [TestCase("DNS", true, "DNS")]
+        [TestCase("DNS", false, "dns")]
```
AlexanderSher, line 89: Can we add cases for something like `PrimaryDNS`?

m-nash: do you expect `primaryDNS`?

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -25,6 +25,8 @@ static StringExtensions()
         public static bool IsNullOrEmpty(this string? text) => String.IsNullOrEmpty(text);
         public static bool IsNullOrWhiteSpace(this string? text) => String.IsNullOrWhiteSpace(text);
 
+        private static bool IsWordSeparator(char c) => !SyntaxFacts.IsIdentifierPartCharacter(c) || c == '_';
+
         [return: NotNullIfNotNull("name")]
         public static string ToCleanName(this string name, bool camelCase = true)
```
m-nash, line 31: is this boolean `isCamelCase` or `wantCamelCase`, can we rename it accordingly?

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -45,19 +47,33 @@ public static string ToCleanName(this string name, bool camelCase = true)
             }
 
             bool upperCase = false;
+            int firstWordLength = 1;
```
m-nash, line 50: Why not set this correctly right here instead of constantly incrementing this?  We should be able to do a linear scan and store the first word length in one go.

test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs
ArcturusZhang: Well, this is unfortunate. This function `ToCleanName` does not have the ability to identify words, therefore we either always deal with digits to be separator of words, or always do not deal with digits to be separator of words.
If we do not treat digits as word separators, we will fail at the case `D365IP`.

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
ArcturusZhang: That could work, but initially I was finding a patch to this function instead of rewriting it. And we only need this when camelCase is false

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
ArcturusZhang: Make sense. I will rename this to `WantCamelCase`

test/AutoRest.TestServer.Tests/Common/StringExtensionsTests.cs
m-nash: Please add a follow up item to address this since this happens today

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
m-nash: Can we add a follow up item to address this?

m-nash (approved):
Approved with the 2 follow up items

PR #1969, opened 2022-02-11
Fix #1858: Correctly validate parameters in protocol methods according to the guidelines

### LINKED ISSUE #1858: Correctly validate parameters in protocol methods according to the guidelines
Guidelines are described here: https://azure.github.io/azure-sdk/general_implementation.html#parameter-validation

We do not currently validate path parameters in the LLC generator, and we should.  @AlexanderSher is following up on whether we should validate the presence of a message body that is marked in swagger as required.

If we keep the changes introduced to the WebPubSub library in [this PR](https://github.com/Azure/azure-sdk-for-net/pull/26349/commits/9d6ba3ffa68254a87dc07f9101da37f6b6500aee#r784984605), we will need arch board approval for this.

AlexanderSher:


AlexanderSher:
@annelo-msft , please take a look at the generated comments

AlexanderSher:
@m-nash , I've merged some of the mgmt code into general helpers cause right now the logic for argument validation in LLC and Mgmt is quite similar. Please take a look at the changes in mgmt.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountsClient.cs
```diff
@@ -1123,9 +1123,10 @@ IEnumerable<Page<BinaryData>> CreateEnumerable(string nextLink, int? pageSizeHin
         /// <summary> Initializes a new instance of PurviewAccountCollections. </summary>
         /// <param name="collectionName"> The String to use. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="collectionName"/> is null. </exception>
```
annelo-msft, line 1125: Could we replace "is null" with "is <code>null</code>"?
See: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs.Batch/src/Shared/StringSegment.cs#L43

(Note: if it is not a small change, then let's not worry about it.)

samples/Azure.AI.FormRecognizer/Generated/Models/CopyRequest.cs
```diff
@@ -16,7 +16,7 @@ public partial class CopyRequest
         /// <param name="targetResourceId"> Azure Resource Id of the target Form Recognizer resource where the model is copied to. </param>
         /// <param name="targetResourceRegion"> Location of the target Azure resource. A valid Azure region name supported by Cognitive Services. </param>
         /// <param name="copyAuthorization"> Entity that encodes claims to authorize the copy request. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="targetResourceId"/>, <paramref name="targetResourceRegion"/>, or <paramref name="copyAuthorization"/> is null. </exception>
+        /// <exception cref="ArgumentNullException"> <paramref name="targetResourceId"/>, <paramref name="targetResourceRegion"/> or <paramref name="copyAuthorization"/> is null. </exception>
         public CopyRequest(string targetResourceId, string targetResourceRegion, CopyAuthorizationResult copyAuthorization)
         {
             if (targetResourceId == null)
```
annelo-msft, line 22: Curious: why aren't we using Argument.Assert below?

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountCollections.cs
```diff
@@ -46,13 +46,14 @@ protected PurviewAccountCollections()
...
+        /// <exception cref="ArgumentNullException"> <paramref name="clientDiagnostics"/>, <paramref name="pipeline"/>, <paramref name="endpoint"/>, <paramref name="collectionName"/> or <paramref name="apiVersion"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="collectionName"/> is an empty string, and was expected to be non-empty. </exception>
         internal PurviewAccountCollections(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, TokenCredential tokenCredential, Uri endpoint, string collectionName, string apiVersion = "2019-11-01-preview")
         {
             Argument.AssertNotNull(clientDiagnostics, nameof(clientDiagnostics));
             Argument.AssertNotNull(pipeline, nameof(pipeline));
             Argument.AssertNotNull(endpoint, nameof(endpoint));
-            Argument.AssertNotNull(collectionName, nameof(collectionName));
+            Argument.AssertNotNullOrEmpty(collectionName, nameof(collectionName));
```
annelo-msft, line 56: Curious: How are we deciding when to assert not null vs. not null or empty?  I was thinking it was whether or not it was a string, but we're not using "or empty" on apiVersion.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountResourceSetRules.cs
```diff
@@ -39,7 +39,7 @@ protected PurviewAccountResourceSetRules()
         /// <param name="tokenCredential"> The token credential to copy. </param>
         /// <param name="endpoint"> The account endpoint of your Purview account. Example: https://{accountName}.purview.azure.com/account/. </param>
         /// <param name="apiVersion"> Api Version. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="clientDiagnostics"/>, <paramref name="pipeline"/>, <paramref name="endpoint"/>, or <paramref name="apiVersion"/> is null. </exception>
+        /// <exception cref="ArgumentNullException"> <paramref name="clientDiagnostics"/>, <paramref name="pipeline"/>, <paramref name="endpoint"/> or <paramref name="apiVersion"/> is null. </exception>
         internal PurviewAccountResourceSetRules(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, TokenCredential tokenCredential, Uri endpoint, string apiVersion = "2019-11-01-preview")
         {
             Argument.AssertNotNull(clientDiagnostics, nameof(clientDiagnostics));
```
annelo-msft, line 45: (3 lines below -- re: apiVersion)
I had thought we didn't validate apiVersion in DPG clients because it would never be a string.  But it looks like it is a string here and we are validating it.  Could you help me understand this?

samples/Azure.Storage.Tables/Generated/Models/CorsRule.cs
```diff
@@ -18,7 +18,7 @@ public partial class CorsRule
         /// <param name="allowedHeaders"> the request headers that the origin domain may specify on the CORS request. </param>
         /// <param name="exposedHeaders"> The response headers that may be sent in the response to the CORS request and exposed by the browser to the request issuer. </param>
         /// <param name="maxAgeInSeconds"> The maximum amount time that a browser should cache the preflight OPTIONS request. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="allowedOrigins"/>, <paramref name="allowedMethods"/>, <paramref name="allowedHeaders"/>, or <paramref name="exposedHeaders"/> is null. </exception>
+        /// <exception cref="ArgumentNullException"> <paramref name="allowedOrigins"/>, <paramref name="allowedMethods"/>, <paramref name="allowedHeaders"/> or <paramref name="exposedHeaders"/> is null. </exception>
```
annelo-msft, line 21: Nit below: Why not Argument.AssertNotNull?

test/TestProjects/BodyAndPath-LowLevel/Generated/BodyAndPathClient.cs
```diff
@@ -55,11 +55,12 @@ public BodyAndPathClient(AzureKeyCredential credential, Uri endpoint = null, Bod
         /// <param name="content"> The content to send as the body of the request. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="itemName"/> or <paramref name="content"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="itemName"/> is an empty string, and was expected to be non-empty. </exception>
 #pragma warning disable AZC0002
         public virtual async Task<Response> CreateAsync(string itemName, RequestContent content, RequestContext context = null)
 #pragma warning restore AZC0002
         {
-            Argument.AssertNotNull(itemName, nameof(itemName));
+            Argument.AssertNotNullOrEmpty(itemName, nameof(itemName));
```
annelo-msft, line 63: Just to make sure I'm understanding correctly: itemName is a path parameter here?

test/TestProjects/ExactMatchFlattenInheritance/Generated/Extensions/ResourceGroupExtensions.cs
```diff
@@ -86,15 +85,12 @@ public static Pageable<AzureResourceFlattenModel2> GetAzureResourceFlattenModel2
         /// <param name="name"> The String to use. </param>
         /// <param name="parameters"> The AzureResourceFlattenModel2 to use. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        /// <exception cref="ArgumentException"> <paramref name="name"/> is empty. </exception>
-        /// <exception cref="ArgumentNullException"> <paramref name="name"/> or <paramref name="parameters"/> is null. </exception>
+        /// <exception cref="System.ArgumentException"> <paramref name="name"/> is an empty string, and was expected to be non-empty. </exception>
```
annelo-msft, line 88: Curious: why remove `using System` and use the full namespace here?

samples/Azure.AI.FormRecognizer/Generated/Models/CopyRequest.cs
AlexanderSher: Models are generated by a different writer. If I update that one, we will get several thousands of changed files in azure sdk repo

test/TestProjects/ExactMatchFlattenInheritance/Generated/Extensions/ResourceGroupExtensions.cs
```diff
@@ -86,15 +85,12 @@ public static Pageable<AzureResourceFlattenModel2> GetAzureResourceFlattenModel2
...
+        /// <exception cref="System.ArgumentNullException"> <paramref name="name"/> or <paramref name="parameters"/> is null. </exception>
         public async static Task<Response<AzureResourceFlattenModel2>> PutAzureResourceFlattenModel2Async(this ResourceGroup resourceGroup, string name, AzureResourceFlattenModel2 parameters, CancellationToken cancellationToken = default)
         {
             Argument.AssertNotNullOrEmpty(name, nameof(name));
-            if (parameters == null)
-            {
-                throw new ArgumentNullException(nameof(parameters));
-            }
+            Argument.AssertNotNull(parameters, nameof(parameters));
```
annelo-msft, line 93: Is `parameters` a path parameter?  What causes us to validate it here?

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountsClient.cs
AlexanderSher: Done.

test/TestProjects/ExtensionClientName/Generated/AutoRestParameterFlatteningRestClient.cs
```diff
@@ -57,7 +57,7 @@ internal HttpMessage CreateRenamedOperationRequest(string renamedPathParameter,
         /// <param name="renamedQueryParameter"> The String to use. </param>
         /// <param name="renamedBodyParameter"> The RenamedSchema to use. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="renamedPathParameter"/>, <paramref name="renamedQueryParameter"/>, or <paramref name="renamedBodyParameter"/> is null. </exception>
+        /// <exception cref="ArgumentNullException"> <paramref name="renamedPathParameter"/>, <paramref name="renamedQueryParameter"/> or <paramref name="renamedBodyParameter"/> is null. </exception>
```
annelo-msft, line 60: It looks like we are validating query and body parameter below.  This is because it's the autorest client?

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountsClient.cs
AlexanderSher: Sorry, had to revert it for now. Change footprint is huge. Let's make it a separate PR.

test/TestProjects/ModelShapes/Generated/Models/InputModel.cs
```diff
@@ -24,7 +24,7 @@ public partial class InputModel
         /// <param name="requiredNullableInt"></param>
         /// <param name="requiredNullableStringList"></param>
         /// <param name="requiredNullableIntList"></param>
-        /// <exception cref="ArgumentNullException"> <paramref name="requiredString"/>, <paramref name="requiredStringList"/>, or <paramref name="requiredIntList"/> is null. </exception>
+        /// <exception cref="ArgumentNullException"> <paramref name="requiredString"/>, <paramref name="requiredStringList"/> or <paramref name="requiredIntList"/> is null. </exception>
```
annelo-msft: Is validating model properties the right thing to do?  I think these are server parameters.

samples/CognitiveSearch/Generated/Models/EncryptionKey.cs
```diff
@@ -16,7 +16,7 @@ public partial class EncryptionKey
         /// <param name="keyVaultKeyName"> The name of your Azure Key Vault key to be used to encrypt your data at rest. </param>
         /// <param name="keyVaultKeyVersion"> The version of your Azure Key Vault key to be used to encrypt your data at rest. </param>
         /// <param name="keyVaultUri"> The URI of your Azure Key Vault, also referred to as DNS name, that contains the key to be used to encrypt your data at rest. An example URI might be https://my-keyvault-name.vault.azure.net. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="keyVaultKeyName"/>, <paramref name="keyVaultKeyVersion"/>, or <paramref name="keyVaultUri"/> is null. </exception>
+        /// <exception cref="ArgumentNullException"> <paramref name="keyVaultKeyName"/>, <paramref name="keyVaultKeyVersion"/> or <paramref name="keyVaultUri"/> is null. </exception>
```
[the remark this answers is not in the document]
annelo-msft: Just remembered these are in autorest client and we're only updating DPG client generator - so, disregard comment.

test/TestProjects/ModelShapes/Generated/Models/InputModel.cs
annelo-msft: Autorest client again - pls disregard.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountCollections.cs
AlexanderSher: `apiVersion` isn't a path parameter. Strictly speaking, we probably should validate that it falls into the enum range of available versions (to avoid calling new API with the old code that can simply fail on type casting or something), but I'd prefer to have a separate story for it, cause we had no approve from archboard for it.

test/TestProjects/RequestContextAllOptional-LowLevel/Generated/RequestContextAllOptionalClient.cs
```diff
@@ -172,13 +172,10 @@ public virtual Response RequestBodyResponseBody(RequestContent content, RequestC
         /// <summary> Delete. </summary>
         /// <param name="resourceName"> name. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="resourceName"/> is null. </exception>
 #pragma warning disable AZC0002
         public virtual async Task<Response> DeleteNoRequestBodyResponseBodyAsync(string resourceName, RequestContext context = null)
 #pragma warning restore AZC0002
         {
-            Argument.AssertNotNull(resourceName, nameof(resourceName));
```
annelo-msft: To make sure I understand: this was a query parameter, which is why we're removing validation?

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountResourceSetRules.cs
AlexanderSher: It is a subclient, which gets its version from the parent.

AlexanderSher: I actually wonder if we should generate any argument checks here at all. This constructor is not meant to be called from anywhere expect factory method of the parent client, which should guarantee that all parameters are correct.

samples/Azure.Storage.Tables/Generated/Models/CorsRule.cs
AlexanderSher: It is a model. Model writer has not been updated yet.

test/TestServerProjectsLowLevel/header/Generated/HeaderClient.cs
```diff
@@ -52,7 +52,6 @@ public HeaderClient(AzureKeyCredential credential, Uri endpoint = null, HeaderCl
         /// <summary> Send a post request with header value &quot;User-Agent&quot;: &quot;overwrite&quot;. </summary>
         /// <param name="userAgent"> Send a post request with header value &quot;User-Agent&quot;: &quot;overwrite&quot;. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="userAgent"/> is null. </exception>
```
annelo-msft: Is `userAgent` something we'd want to special case?  It seems like more of a client parameter than a server parameter.

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
```diff
@@ -1079,8 +1080,7 @@ IEnumerable<Page<BinaryData>> CreateEnumerable(string nextLink, int? pageSizeHin
         public virtual AsyncPageable<BinaryData> GetMultiplePagesFragmentNextLinkAsync(string tenant, string apiVersion, RequestContext context = null)
 #pragma warning restore AZC0002
         {
-            Argument.AssertNotNull(tenant, nameof(tenant));
-            Argument.AssertNotNull(apiVersion, nameof(apiVersion));
```
annelo-msft: We want to always be sending `apiVersion` in requests, so I think it makes sense to validate `apiVersion` here.

samples/CognitiveSearch/Generated/Models/EncryptionKey.cs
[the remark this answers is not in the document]
AlexanderSher: In general, it is quite hard to provide meaningful error to the user if model is not properly initialized. There can be array of objects with dictionaries of arrays of objects. If one of those is missing something, how are we going to point to it? Showing plain JSON is not an option - user has no idea how we have serialized the model, so no way to match model values to json values.

test/TestProjects/BodyAndPath-LowLevel/Generated/BodyAndPathClient.cs
AlexanderSher: Yes.

test/TestProjects/ExactMatchFlattenInheritance/Generated/Extensions/ResourceGroupExtensions.cs
AlexanderSher: This is mgmt logic. I have unified their code with ours (cause now we are doing mostly the same thing), hence the `if` block that is replaced with  `Argument.AssertNotNull`.
@m-nash , what are the validation rules here?

test/TestProjects/ExactMatchFlattenInheritance/Generated/Extensions/ResourceGroupExtensions.cs
AlexanderSher: `using System` has been removed because there is no more code that requires it (previously it was `throw new ArgumentNullException` that required `using System`). I think Roslyn doesn't add `using` directive only for doc comments

test/TestServerProjectsLowLevel/header/Generated/HeaderClient.cs
AlexanderSher: Why?

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
AlexanderSher: Hm... that is an interesting place you have found! I'd expect it to use _apiVersion field, but for some reason it is generated as a method parameter. Need to debug this one.

samples/Azure.ResourceManager.Sample/Generated/Extensions/SubscriptionExtensions.cs
```diff
@@ -32,8 +31,8 @@ private static SubscriptionExtensionClient GetExtensionClient(Subscription subsc
         /// <param name="subscription"> The <see cref="Subscription" /> instance the method will execute against. </param>
         /// <param name="location"> The name of a supported Azure region. </param>
         /// <param name="publisherName"> The String to use. </param>
-        /// <exception cref="ArgumentException"> <paramref name="location"/> or <paramref name="publisherName"/> is empty. </exception>
-        /// <exception cref="ArgumentNullException"> <paramref name="location"/> or <paramref name="publisherName"/> is null. </exception>
+        /// <exception cref="System.ArgumentException"> <paramref name="location"/> or <paramref name="publisherName"/> is an empty string, and was expected to be non-empty. </exception>
```
m-nash, line 34: Can we explicitly add System namespace when we enter the function that is writing these exception comments ?

m-nash (changes requested):

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
AlexanderSher: Ok, so these parameters don't have `origin: modelerfour:synthesized/api-version`, e.g.: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestServerProjectsLowLevel/paging/Generated/CodeModel.yaml#L1739
(compare to https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestServerProjectsLowLevel/body-complex/Generated/CodeModel.yaml#L1972)

@timotheeguerin , do you know why these ones aren't marked as api-version?

m-nash (approved):
mgmt changes look good, but there are a few conflicts now

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountCollections.cs
annelo-msft: Sounds good.  Would you be willing to file an issue to track this so it doesn't get lost?

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountResourceSetRules.cs
annelo-msft: There's a chance at some point in the future ClientDiagnostics migrates to Core (this would take work, so I'm not saying it'll happen any time soon).  If that happened and this constructor became public, would we know to add back validation of arguments?  If so, I'm fine to take them out here.

test/TestProjects/ExactMatchFlattenInheritance/Generated/Extensions/ResourceGroupExtensions.cs
annelo-msft: I see, it's strange it doesn't add the using directive for doc comments.  I am fine taking Roslyn's suggestion here, was just wondering if we could avoid code churn.

test/TestProjects/ExactMatchFlattenInheritance/Generated/Extensions/ResourceGroupExtensions.cs
annelo-msft: If we find issues where management plane isn't in compliance with the guidelines, it would be good to file issues just so we can keep track of these for later.

test/TestServerProjectsLowLevel/header/Generated/HeaderClient.cs
annelo-msft: We can discuss this separately - let's not block this PR on it.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountResourceSetRules.cs
AlexanderSher: Sorry, but I'm not sure why would this constructor become public. Subclients are non-instantiable by design: https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-subclients

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountCollections.cs
AlexanderSher: Let's first decide if we want comments for internal constructors at all.

test/TestProjects/ExtensionClientName/Generated/AutoRestParameterFlatteningRestClient.cs
AlexanderSher: Yes

test/TestProjects/ExactMatchFlattenInheritance/Generated/Extensions/ResourceGroupExtensions.cs
m-nash: Parameters is a body parameter in this case.  I would imagine it validates because its marked as required.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountResourceSetRules.cs
annelo-msft: Good point.

test/TestProjects/ExactMatchFlattenInheritance/Generated/Extensions/ResourceGroupExtensions.cs
annelo-msft: We recently updated the guidelines to make it more clear that we don't validate service parameters: https://azure.github.io/azure-sdk/general_implementation.html#parameter-validation

This is even if they are required - the service should be responsible for reporting that error.  I don't think it's urgent that we fix this, but we should have an issue to track it.

m-nash: For methods like PUT or PATCH this doesn't make much sense.  If we allowed a customer to pass in null what would they expect to have happen?  We literally got this question from a customer who was confused why we would allow it.

It also opens up the issue of the [RestOperations ](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Administration/src/Generated/RoleAssignmentsRestClient.cs#L164)today validate this as well not just the public methods here, and if we remove that then we will get [null reference exceptions](https://github.com/Azure/autorest.csharp/blob/feature%2Fv3/src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs#L30) in the serializer that we will need to fix and test that layer especially against already GAd data plane libraries

ArthurMa1978: Agree with Michael.
We don't need to validate the service data, but if the body parameter is required why don't we throw NULL exception directly?
My understanding is the validation should not be happened on the service data itself, which means the value of body content, properties, but should still check shell.

samples/Azure.Analytics.Purview.Account/Generated/PurviewAccountCollections.cs
AlexanderSher: Fixed. Parameter validation is removed from internal constructors

m-nash (approved):
new mgmt plane changes look good as well in this folder TestProjects/MgmtCollectionParent

PR #1971, opened 2022-02-12
Make context static so it doesn't have to passed around everywhere

m-nash:
Refactor outputlibrary to utilize cached dictionary

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/ParentDetection.cs
```diff
@@ -27,79 +27,79 @@ internal static class ParentDetection
...
+                return MgmtContext.Library.SubscriptionExtensions.AsIEnumerable();
             // the only option left is the tenant. But we have our last chance that its parent could be the scope of this
             var scope = parentRequestPath.GetScopePath();
             // if the scope of this request path is parameterized, we return the scope as its parent
             if (scope.IsParameterizedScope())
             {
                 // we already verified that the scope is parameterized, therefore we assert the type can never be null
-                var types = resourceOperationSet.GetRequestPath(context).GetParameterizedScopeResourceTypes(context.Configuration.MgmtConfiguration)!;
-                return FindScopeParents(types, context);
+                var types = resourceOperationSet.GetRequestPath().GetParameterizedScopeResourceTypes(MgmtContext.MgmtConfiguration)!;
```
ArcturusZhang, line 65: We missed a method here - this parameter is a `MgmtConfiguration`

ArcturusZhang (approved):
Since we have a static context now, it might make more sense we could also move some methods in the decorators to their corresponding classes (such as the methods in `ResourceTypeBuilder`, they are in the decorator class because it requires a configuration). But I believe it is better to do that in a separate PR

PR #1973, opened 2022-02-14
change the way we check if a resource data is taggable

ArcturusZhang:
# Description

We are using different logic for checking if the update is only updating tags and the resource data is taggable. This introduce inconsistencies in RPs (such as network), their resource data is not inheriting from TrackedResourceData, but still is taggable.

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
+++ b/src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
@@ -24,6 +24,11 @@ namespace AutoRest.CSharp.Mgmt.Decorator
             return schema.Parents!.All.OfType<ObjectSchema>().SelectMany(parentSchema => parentSchema.Properties).Concat(schema.Properties);
         }
 
+        private static bool IsTagsProperty(Property property)
+            => property.CSharpName().Equals("Tags")
+                && property.Schema is DictionarySchema dictSchema
+                && dictSchema.ElementType.Type == AllSchemaTypes.String;
+
         public static bool HasTags(this Schema schema)
         {
             if (schema is not ObjectSchema objSchema)
@@ -33,9 +38,7 @@ namespace AutoRest.CSharp.Mgmt.Decorator
 
             var allProperties = objSchema.GetAllProperties();
 
-            var tagsProperties = allProperties.Where(property => property.CSharpName().Equals("Tags")
-                && property.Schema is DictionarySchema dictSchema
-                && dictSchema.ElementType.Type == AllSchemaTypes.String);
+            var tagsProperties = allProperties.Where(property => IsTagsProperty(property));
 
             return tagsProperties.Any();
         }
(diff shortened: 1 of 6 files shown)


src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
```diff
@@ -24,6 +24,22 @@ private static IEnumerable<Property> GetAllProperties(this ObjectSchema schema)
             return schema.Parents!.All.OfType<ObjectSchema>().SelectMany(parentSchema => parentSchema.Properties).Concat(schema.Properties);
         }
 
+        public static bool HasTags(this Schema schema)
```
Yao725, line 27: Can we use this method in below method `IsTagsOnly`?

Yao725 (approved):

src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
ArcturusZhang: These two are different. One of them is identifying if one of its properties is tags, the other one is if its properties contains tags.
I could introduce a helper method to identify if a property is `tags`. This will make some of the code in these two methods shared

fengzhou-msft (approved):

src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
```diff
@@ -24,6 +24,22 @@ private static IEnumerable<Property> GetAllProperties(this ObjectSchema schema)
...
+        {
+            if (schema is not ObjectSchema objSchema)
+            {
+                return false;
+            }
+
+            var allProperties = objSchema.GetAllProperties();
+
+            var tagsProperties = allProperties.Where(property => property.CSharpName().Equals("Tags")
```
ArthurMa1978, line 36: return true immediately when reach the first tag property instead of fetch all then any

src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
```diff
@@ -24,6 +24,25 @@ private static IEnumerable<Property> GetAllProperties(this ObjectSchema schema)
...
+        public static bool HasTags(this Schema schema)
+        {
+            if (schema is not ObjectSchema objSchema)
+            {
+                return false;
+            }
+
+            var allProperties = objSchema.GetAllProperties();
+
+            var tagsProperties = allProperties.Where(property => IsTagsProperty(property));
```
ArthurMa1978, line 41: return objSchema.GetAllProperties().any(property => IsTagsProperty(property));

ArcturusZhang: Make sense

ArthurMa1978 (approved):

Issue #1652, opened 2021-10-29
[feature] A more elegant solution on excluding the `Operations_List` operation


ArcturusZhang:
Every RP should have an operation of listing all the operations that this RP supports, usually it has the `operationId` of `Operations_List` in the swaggers. But I see quite a few counter-cases like: https://github.com/Azure/azure-rest-api-specs/blob/69eacf00a36d565d3220d5dd6f4a5293664f1ae9/specification/web/resource-manager/Microsoft.Web/stable/2021-02-01/Provider.json#L227

It would be more elegant if we could omit this request by detecting its request path, instead of our current implementation which relies on the operation group name: https://github.com/Azure/autorest.csharp/blob/af581b9f3703417f9d9597612fbd6db942d3193d/src/AutoRest.CSharp/Mgmt/Decorator/OmitOperationGroups.cs#L21
it is possible that we could accidentally remove something also in the `Operations` operation group.

PR #1976, opened 2022-02-14
Remove the `Operations_List` operation by its path which is more reliable

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1652

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -33,6 +34,55 @@ public static void UpdateFrameworkTypes(this IEnumerable<Schema> allSchemas)
             }
         }
 
+        public static void ReorderOperationParameters(this CodeModel codeModel)
```
AlexanderSher, line 37: How this method correlates with https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Mgmt/Models/MgmtRestClientBuilder.cs#L57 ?

ArcturusZhang: This method is moved from MgmtOutputLibrary to here, it is not a new method.
This method only sort the path parameters (which are all required) in a order as they appears in the path. We need this because the code model will just create the parameter in the order of definition.
And this method is changing the things in code model, therefore this runs like this:
1. First we modify the code model to make sure the path parameters are in right order.
2. Then we reach the function `BuildParameters` sort by required parameters, body parameter and then optional parameters

AlexanderSher: It feels like it duplicates logic that we have in LLC: https://github.com/Azure/autorest.csharp/blob/0a87c4af88516fb49c3192b55b2aa5f02e6f9011/src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs#L210

ArcturusZhang: Maybe, but since this function in mgmt plane directly changed the code model, the changes will apply to RestOperations and every method in the generated client (the `AutoRest.CSharp.Input.Operation` class). Will this function in LLC do the same?
And this code in mgmt plane was introduced a long time ago, maybe we should refactor this later. I created an issue for this: https://github.com/Azure/autorest.csharp/issues/1985

PR #1977, opened 2022-02-14
feat(mgmt): exponential polling interval

archerzz:
1. refactor to implement various polling strategy for `OperationInternalBase`
2. add property `OperationOrResponseInternals.PollingStrategy` to override default polling strategy
3. update `MgmtLongRunningOperationWriter` to set exponential polling strategy for mgmt plane

resolve #26922

# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/assets/Azure.Core.Shared/OperationInternalBase.cs
+++ b/src/assets/Azure.Core.Shared/OperationInternalBase.cs
@@ -28,6 +28,7 @@ namespace Azure.Core
             _scopeAttributes = scopeAttributes?.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
             RawResponse = rawResponse;
             DefaultPollingInterval = TimeSpan.FromSeconds(1);
+            PollingStrategy = new ConstantPollingStrategy();
         }
 
         /// <summary>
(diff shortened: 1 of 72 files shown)


src/assets/Azure.Core.Shared/OperationInternalBase.cs
```diff
@@ -58,6 +59,22 @@ protected OperationInternalBase(ClientDiagnostics clientDiagnostics, Response ra
         /// </summary>
         public TimeSpan DefaultPollingInterval { get; set; }
```
archerzz, line 60: I need more time to figure out if we can remove `DefaultPollingInterval`. Although it's public, it seems not be explicitly set (e.g. no place really set a non-default value) and the getter is invoked in polling implementation.
If we could remove it, then the logic of `PollingInterval` (see below) can be simplified.

test/TestProjects/ExactMatchFlattenInheritance/Generated/LongRunningOperation/ExactMatchFlattenInheritanceArmOperation.cs
```diff
@@ -29,11 +29,13 @@ protected ExactMatchFlattenInheritanceArmOperation()
         internal ExactMatchFlattenInheritanceArmOperation(Response response)
         {
             _operation = new OperationOrResponseInternals(response);
+            _operation.PollingStrategy = new ExponentialPollingStrategy();
```
archerzz, line 32: Another options is that we add constructors of `OperationOrResponseInternals` for mgmt plane. And invoke the specific constructor here.

src/assets/Azure.Core.Shared/OperationInternalBase.cs
AlexanderSher: This type is defined in Azure.Core, so the change should happen there.

archerzz:
This PR has been covered in https://github.com/Azure/autorest.csharp/pull/2000

PR #1978, opened 2022-02-15
Update name transformer

ArcturusZhang:
# Description

In the previous implementation, we did not transform the name of choice values. This PR also transforms that.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -86,9 +88,24 @@ private static void TransformSchema(Schema schema, NameTransformer transformer,
             }
         }
 
-        private static void TransformBasicSchema(Schema schema, NameTransformer transformer, ConcurrentDictionary<string, string> wordCache)
+        private static void TransformChoiceSchema(ChoiceSchema choiceSchema, NameTransformer transformer, ConcurrentDictionary<string, string> wordCache)
```
AlexanderSher, line 91: What's the difference between `TransformChoiceSchema` and `TransformSealedChoiceSchema`? They look identical to me.

ArcturusZhang: Yes, they are identical. But unfortunately `ChoiceSchema` and `SealedChoiceSchema` does not inherit from each other, they just share the same parent, and the most important property `Choices` is individually defined in these two classes.
I do not want to write too many lines in the switch statement, therefore I create these two functions to do all the things.

AlexanderSher: Will this work:

```cs
private static void TransformChoiceSchema(Languages languages, ICollection<ChoiceValue> choices, NameTransformer transformer, ConcurrentDictionary<string, string> wordCache) 
{
    TransformLanguage(languages, transformer, wordCache);
    TransformChoices(choices, transformer, wordCache);
}
```

```cs
case ChoiceSchema choiceSchema:
    TransformChoiceSchema(choiceSchema.Language, choiceSchema.Choices, transformer, wordCache);
    break;
case SealedChoiceSchema sealedChoiceSchema:
    TransformChoiceSchema(sealedChoiceSchema.Language, sealedChoiceSchema.Choices, transformer, wordCache);
    break;
```

Yao725 (approved):

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
ArcturusZhang: Let me do this change

PR #1979, opened 2022-02-15
Fix create result failure when consuming same content twice

Yao725:
Current implementation of method `CreateResult` will have json serialization failure when the customer sets the `waitForCompletion` parameter to true but still call the method `WaitForCompletion` by themselves. The reason is that some lro operations like `Vault Update` will directly return 200 and so the `CreateResult` consumes same content twice. This PR is going to reset the position of the content back to 0 to avoid this failure.

# Description

# Checklist

# Ready to Land?

### DIFF
--- a/samples/Azure.Management.Storage/Generated/LongRunningOperation/BlobRestoreStatusOperationSource.cs
+++ b/samples/Azure.Management.Storage/Generated/LongRunningOperation/BlobRestoreStatusOperationSource.cs
@@ -18,14 +18,28 @@ namespace Azure.Management.Storage
     {
         BlobRestoreStatus IOperationSource<BlobRestoreStatus>.CreateResult(Response response, CancellationToken cancellationToken)
         {
-            using var document = JsonDocument.Parse(response.ContentStream);
-            return BlobRestoreStatus.DeserializeBlobRestoreStatus(document.RootElement);
+            try
+            {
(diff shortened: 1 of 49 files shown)


AlexanderSher:
Does it mean that JSON parsing also happens twice?

Yao725:
> Does it mean that JSON parsing also happens twice?

Yes, Actually this is an edge case and will happen when the customer sets the `waitForCompletion` parameter to true but still call the method WaitForCompletion by themselves.

Yao725:
> Not quite an edge case. Can you point to a specific operation where this problem reproduced?

@AlexanderSher , [here](https://github.com/Yao725/azure-sdk-for-net/blob/c4f5862269659d79814f6cb241db6a15da3af25a/sdk/keyvault/Azure.ResourceManager.KeyVault/tests/VaultIssueTests.cs#L27) is how to reproduce this problem. When you try to update an existing vault, the backend service will return [200 and the provisioningState is Succeeded](https://github.com/Yao725/azure-sdk-for-net/blob/c4f5862269659d79814f6cb241db6a15da3af25a/sdk/keyvault/Azure.ResourceManager.KeyVault/tests/SessionRecords/VaultIssueTests/KeyVaultManagementVaultCreate()Async.json#L138-L196), so the second call of `WaitForCompeletion` throws exception.

Issue #1958, opened 2022-02-09
[Refactor] Introduce a new TypeProvider for our new `[RPName]ArmOperation` classes (non-generic and generic) version


ArcturusZhang:
This writer is writing a class without a TypeProvider, which introduces a little bit duplicated calculations.
Also I believe we should rename this class, maybe call it `RpArmOperationWriter`?

PR #1980, opened 2022-02-15
Refactor to introduce a `TypeProvider` for the long running operation class

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1958

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Generation/MgmtLongRunningOperationWriter.cs
```diff
@@ -2,85 +2,78 @@
...
         }
 
         public void Write()
         {
             using (_writer.Namespace(MgmtContext.Context.DefaultNamespace))
             {
                 _writer.Line($"#pragma warning disable SA1649 // File name should match first type name");
-                _writer.Line($"internal class {_name}{_genericString} : {_operationType}");
+                var genericString = _lro.IsGeneric ? "<T>" : string.Empty;
+                _writer.Line($"internal class {_lro.Type.Name}{genericString} : {_lro.BaseType}");
```
AlexanderSher, line 34: If we introduce a type for LRO's, there should be no type manipulations in the writer. This should be something like 
```cs
using (writer.Scope($"internal partial class {_lro.Type:D}")) {
```

Right now `CodeWriter` doesn't support `D` for the type, I'll add it in separate PR

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
```diff
@@ -0,0 +1,95 @@
...
+{
+    internal class MgmtLongRunningOperation : TypeProvider
+    {
+        public bool IsGeneric { get; }
+
+        protected override string DefaultName { get; }
+
+        protected override string DefaultAccessibility => "internal";
+
+        public MgmtLongRunningOperation(bool isGeneric) : base(MgmtContext.Context)
```
AlexanderSher, line 29: Let's have two factory methods instead: `CreateGeneric` and `CreateNonGeneric`

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            IsGeneric = isGeneric;
+            DefaultName = $"{MgmtContext.Context.DefaultNamespace.Split('.').Last()}ArmOperation";
+            BaseType = isGeneric ? typeof(ArmOperation<>) : typeof(ArmOperation);
```
AlexanderSher, line 33: Let's split this into `if/else` block instead of checking `isGeneric` for every single parameter

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
```diff
@@ -0,0 +1,95 @@
...
+            WaitMethod = isGeneric ? "WaitForCompletionAsync" : "WaitForCompletionResponseAsync";
+            OperationType = isGeneric ? typeof(ArmOperation<>) : typeof(ArmOperation);
+            ResponseType = isGeneric ? typeof(Response<>) : typeof(Response);
+            OperationOrResponseType = isGeneric ? typeof(OperationOrResponseInternals<>) : typeof(OperationOrResponseInternals);
+            OperationSourceString = isGeneric ? (FormattableString)$"{typeof(IOperationSource<>)} source, " : (FormattableString)$"";
```
AlexanderSher, line 38: Is it used anywhere?

samples/Azure.Management.Storage/Generated/LongRunningOperation/StorageArmOperation.cs
```diff
@@ -31,9 +31,9 @@ internal StorageArmOperation(Response response)
             _operation = new OperationOrResponseInternals(response);
         }
 
-        internal StorageArmOperation(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, Request request, Response response, OperationFinalStateVia finalStateVia)
+        internal StorageArmOperation(ClientDiagnostics clientDiagnostic, HttpPipeline pipeline, Request request, Response response, OperationFinalStateVia finalStateVia)
```
AlexanderSher, line 34: Is it intentional?

src/AutoRest.CSharp/Mgmt/Generation/MgmtLongRunningOperationWriter.cs
ArcturusZhang: Shall we also introduce something in `CSharpType` to represent a generic type with a generic type argument placeholder? 
So that this could write to `SampleArmOperations<T>`

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
ArcturusZhang: Well, no, this is introduced during the refactor. I will remove this shortly

samples/Azure.Management.Storage/Generated/LongRunningOperation/StorageArmOperation.cs
ArcturusZhang: Ah, I already noticed this change, and has changed it. I did not regenerate this project yet. If you take a look at the Azure.ResourceManager.Sample project, this change is not there.
I will refresh all the test project shortly

src/AutoRest.CSharp/Mgmt/Generation/MgmtLongRunningOperationWriter.cs
AlexanderSher: `CSharpType` can represent a generic type, I've updated CodeWriter so that it can correctly write it: https://github.com/Azure/autorest.csharp/pull/1994

ArcturusZhang: Thank you! But currently we are still missing a piece that `TypeProvider` never generate its `Type` as a generic type: 
https://github.com/Azure/autorest.csharp/blob/2e5b4431de73126a7f4b493bfe25cd5c6a9e824e/src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs#L22
And we might rarely have a case that we need to write a declaration of a framework type. Should we also have this fixed in a separate PR?

ArcturusZhang: @AlexanderSher I added some attempt of doing that in [the latest commit](https://github.com/Azure/autorest.csharp/pull/1980/commits/9f865a65d47ff5472796654764c477ea20da14e8), please take a look.

And BTW do we have a handy way to instantiate a generic type placeholder (like `T`) of type `CSharpType`? I have to use the generic type parameter from a system type, and then convert them to CSharpType.

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -19,7 +19,9 @@ protected TypeProvider(BuildContext context)
             _existingType = new Lazy<INamedTypeSymbol?>(() => Context.SourceInputModel?.FindForType(DefaultNamespace, DefaultName));
         }
 
-        public CSharpType Type => new(this, TypeKind is TypeKind.Struct or TypeKind.Enum);
+        protected virtual CSharpType[]? Arguments { get; } = null;
```
AlexanderSher, line 22: `TypeProvider` is too general type to add `Arguments` property. Let's make `public CSharpType Type` property `virtual` and add arguments in `MgmtLongRunningOperation` directly.

ArcturusZhang:
Hi @AlexanderSher I have made some changes, please have a look again, thanks

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        protected Parameter GenericResponseParameter => new Parameter("response", null, ResponseType, null, false);
+
+        protected Parameter? SourceParameter => IsGeneric ? new Parameter("source", null, typeof(IOperationSource<>), null, false) : null;
+        protected Parameter ClientDiagnosticParameter => new Parameter("clientDiagnostics", null, typeof(ClientDiagnostics), null, false);
+        protected Parameter PipelineParameter => new Parameter("pipeline", null, typeof(HttpPipeline), null, false);
+        protected Parameter RequestParameter => new Parameter("request", null, typeof(Request), null, false);
+        protected Parameter ResponseParameter => new Parameter("response", null, typeof(Response), null, false);
+        protected Parameter FinalStateParameter => new Parameter("finalStateVia", null, typeof(OperationFinalStateVia), null, false);
```
AlexanderSher, line 107: `MgmtLongRunningOperation` has no derived types, so there properties can be made private. Also, they can be static readonly, similar to https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Output/Models/Shared/KnownParameters.cs (parameter instances can be shared between method signatures).

src/AutoRest.CSharp/Mgmt/Output/MgmtLongRunningOperation.cs
```diff
@@ -0,0 +1,109 @@
...
+                Modifiers: "internal",
+                Parameters: new[] { GenericResponseParameter });
+        }
+
+        private ConstructorSignature? _requestCtor;
+        public ConstructorSignature? RequestCtor => _requestCtor ??= EnsureRequestCtor();
+        private ConstructorSignature? EnsureRequestCtor()
+        {
+            var parameters = new List<Parameter> { ClientDiagnosticParameter, PipelineParameter, RequestParameter, ResponseParameter, FinalStateParameter };
+            if (SourceParameter is not null)
```
AlexanderSher, line 91: It is better to check `IsGeneric` here directly.

AlexanderSher:
@ArcturusZhang I've added some more comments, but the main question I have is why do we generate completely identical internal types for every project?

ArcturusZhang:
> @ArcturusZhang I've added some more comments, but the main question I have is why do we generate completely identical internal types for every project?

Unfortunately I do not know the exact answer of this question because this new type (both versions, generic or non-generic) is introduced by @m-nash in [this PR](https://github.com/Azure/autorest.csharp/pull/1951).
This PR only focuses on the refactoring of that writer.

ArcturusZhang:
Since the LRO objects have been changed quite a bit, I would like to redo all the works here when it is necessary

PR #1982, opened 2022-02-16
RequestPath refactoring

AlexanderSher:
Simplification of the logic from https://github.com/Azure/autorest.csharp/pull/1949 + some optimizations
This PR doesn't introduce any changes in generated code.

### DIFF
--- a/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
+++ b/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
@@ -41,6 +41,11 @@ namespace AutoRest.CSharp.Mgmt.AutoRest
         private CachedDictionary<string, OperationSet> RawRequestPathToOperationSets { get; }
 
         /// <summary>
+        /// This is a map from operation to its request path
+        /// </summary>
+        private CachedDictionary<Operation, RequestPath> OperationsToRequestPaths { get; }
+
+        /// <summary>
         /// This is a map from raw request path to the corresponding <see cref="MgmtRestClient"/>
         /// The type of values is a HashSet of <see cref="MgmtRestClient"/>, because we might get the case that multiple operation groups might share the same request path
         /// </summary>
@@ -77,7 +82,7 @@ namespace AutoRest.CSharp.Mgmt.AutoRest
         /// <summary>
         /// This is a map from <see cref="OperationGroup"/> to the list of raw request path of its operations
         /// </summary>
-        private Dictionary<OperationGroup, IEnumerable<string>> _operationGroupToRequestPaths;
+        private readonly Dictionary<OperationGroup, IEnumerable<string>> _operationGroupToRequestPaths;
 
         public MgmtOutputLibrary()
         {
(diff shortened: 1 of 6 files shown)


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -167,18 +169,17 @@ public MgmtOutputLibrary()
                         if (count != 1)
                             continue;
 
-                        RequestPath requestPath = GetRequestPath(operationGroup, operation);
+                        RequestPath requestPath = RequestPath.FromOperation(operation, operationGroup);
                         var operationSet = RawRequestPathToOperationSets[requestPath];
                         var resourceDataModelName = ResourceDataSchemaNameToOperationSets.FirstOrDefault(kv => kv.Value.Contains(operationSet));
-                        var name = bodyParam.Schema.Language.Default.Name;
                         if (resourceDataModelName.Key is not null)
```
AlexanderSher, line 175: This is very non-obvious though. `ResourceDataSchemaNameToOperationSets` doesn't have null keys, but `FirstOrDefault` may return default value for `KeyValuePair`, hence it is null. We need better code here.

src/AutoRest.CSharp/Mgmt/Models/RequestPath.cs
```diff
@@ -70,50 +68,39 @@ internal struct RequestPath : IEquatable<RequestPath>, IReadOnlyList<Segment>
...
+            foreach (var request in operation.Requests)
+            {
+                var httpRequest = request.Protocol.Http as HttpRequest;
+                if (httpRequest is null)
+                    continue;
 
-        public static RequestPath FromPathSegments(PathSegment[] pathSegments, string httpPath)
-        {
-            return new RequestPath(pathSegments, httpPath);
+                var references = new MgmtRestClientBuilder(operationGroup).GetReferencesToOperationParameters(operation, request.Parameters);
```
AlexanderSher, line 81: `MgmtRestClientBuilder` can be shared between `operationGroup` instances, but `operationGroup.Key` is used in the exception below.

src/AutoRest.CSharp/Mgmt/Models/RequestPath.cs
```diff
@@ -325,39 +312,61 @@ public IEnumerable<RequestPath> Expand()
...
                 {
-                    var pieces = ((string)value).Split('/', StringSplitOptions.RemoveEmptyEntries);
-                    segmentIndex += pieces.Length;
-                    return pieces.Select(s => new Segment(s));
+                    // in this case, we have a constant in this path segment, which might contains slashes
+                    // we need to split it into real segments
+                    // For instance we might get "/providers/Microsoft.Storage/blobServices/default" here
+                    // we will never get null in this constant since it comes from request path
+                    var literalSpan = span;
+                    while (!literalSpan.IsEmpty)
```
AlexanderSher, line 326: This begs to be a `Split` extension method. Will take a look into it tomorrow.

src/AutoRest.CSharp/Mgmt/Models/RequestPath.cs
```diff
@@ -376,13 +385,7 @@ public bool Equals(RequestPath other)
...
         }
 
-        public override bool Equals(object? obj)
-        {
-            if (obj == null)
-                return false;
-            var other = (RequestPath)obj;
-            return other.Equals(this);
-        }
+        public override bool Equals(object? obj) => obj is RequestPath other && Equals(other);
```
AlexanderSher, line 388: Old implementation would fail on type cast if `obj` is not `RequestPath`.

m-nash (approved):

ArcturusZhang (approved):

PR #1983, opened 2022-02-16
Support ManagedServiceIdentity v3

fengzhou-msft:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -108,6 +109,18 @@ protected virtual ObjectTypeProperty CreatePropertyType(ObjectTypeProperty objec
...
+                            && (typeToReplace.Properties.First(p => p.Declaration.Name == "Type")?.SchemaProperty?.Schema is SealedChoiceSchema sealedChoiceSchema
+                                && sealedChoiceSchema.Choices.Any(c => c.Value == ManagedServiceIdentityTypeV3Converter.systemAssignedUserAssignedV3Value)
+                                || typeToReplace.Properties.First(p => p.Declaration.Name == "Type")?.SchemaProperty?.Schema is ChoiceSchema choiceSchema
+                                && choiceSchema.Choices.Any(c => c.Value == ManagedServiceIdentityTypeV3Converter.systemAssignedUserAssignedV3Value)))
+                        {
+                            if (objectTypeProperty.SchemaProperty!.Schema.Extensions == null)
+                                objectTypeProperty.SchemaProperty!.Schema.Extensions = new DictionaryOfAny() { { "x-ms-mgmt-modelVersion", "v3" } };
+                            else
+                                objectTypeProperty.SchemaProperty!.Schema.Extensions?.TryAdd("x-ms-mgmt-modelVersion", "v3");
+                        }
```
ArcturusZhang, line 123: Could we add some comments to describe what this is doing?

ArcturusZhang (commented):
A high level question is that does the replacement requires the identity in this RP to have all four kinds of identity type?

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -217,7 +218,14 @@ public static void ToSerializeCall(this CodeWriter writer, JsonSerialization ser
                                 var systemObjectType = objectType as SystemObjectType;
                                 if (systemObjectType != null && IsCustomJsonConverterAdded(systemObjectType))
                                 {
-                                    writer.Append($"JsonSerializer.Serialize(writer, {name});");
+                                    var optionalSerializeOptions = string.Empty;
+                                    if (systemObjectType.SystemType == typeof(ManagedServiceIdentity) && valueSerialization.Version == "v3")
+                                    {
+                                        writer.UseNamespace("Azure.ResourceManager.Models");
+                                        writer.Line($"var serializeOptions = new JsonSerializerOptions {{ Converters = {{ new {nameof(ManagedServiceIdentityTypeV3Converter)}() }} }};");
```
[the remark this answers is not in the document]
fengzhou-msft: This could work, but the generated code becomes:
`var serializeOptions = new JsonSerializerOptions { Converters = { new ResourceManager.Models.ManagedServiceIdentityTypeV3Converter() } };`

`Azure.ResourceManager.Models` is added automatically, but not sure why it's using `ResourceManager.Models.ManagedServiceIdentityTypeV3Converter`.

fengzhou-msft:
> A high level question is that does the replacement requires the identity in this RP to have all four kinds of identity type?

Yes, it requires that. There is a `SystemAssignedServiceIdentityType` which has less types and service RP may match the corresponding`SystemAssignedServiceIdentity` if possible.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -108,6 +109,20 @@ protected virtual ObjectTypeProperty CreatePropertyType(ObjectTypeProperty objec
                     if (match != null)
                     {
                         propertyType = ReferenceTypePropertyChooser.GetObjectTypeProperty(objectTypeProperty, match);
+                        // If the matched type is ManagedServiceIdentity and the type to replace is a sealed enum or extensible enum containing the v3 "SystemAssigned,UserAssigned" value,
+                        // we shall mark "x-ms-mgmt-modelVersion" to be "v3". The serializer will then use the v3 serializer for ManagedServiceIdentity later.
+                        if (match.Implementation is SystemObjectType systemObjectType
```
m-nash, line 114: I wonder if this should be an option on the serializer vs an x-ms flag.  This technically breaks the idempotent requirements on if i get model from the output library it should the same every time I ask, but here we are adding a flag only when it has been accessed by this class

m-nash (changes requested):

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
fengzhou-msft: Updated to add a JsonSerializationOptions property for JsonSerialization which can be used to to determine whether the serializer should use an option with the v3 converter.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
fengzhou-msft: Added and refactored.

m-nash (approved):

PR #1986, opened 2022-02-17
Transform parameter names

ArcturusZhang:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -59,18 +59,46 @@ public static void UpdateSubscriptionIdForAllResource(this CodeModel codeModel)
...
+                }
+            }
+        }
+
+        private static void TransformOperation(Operation operation, NameTransformer transformer, ConcurrentDictionary<string, string> wordCache)
+        {
+            TransformLanguage(operation.Language, transformer, wordCache);
+            foreach (var parameter in operation.Parameters)
+            {
+                TransformLanguage(parameter.Language, transformer, wordCache);
```
Yao725, line 98: I think for a parameter name, we should also keep the camel case. But if I understand correctly, the transformation here might break it. For example, if a parameter has a name like `osProfile`, it will be changed to `OSProfile`, right? .

ArcturusZhang: Not really. The code here does change the name of this parameter in the code model from `osProfile` to `OSProfile`. But when we are parsing the code model, [this line of code](https://github.com/Azure/autorest.csharp/blob/c4dd241e65f767b32c6d35eeda93dc3b52f78ff6/src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs#L72) will be invoked, change the name of the parameter back to `osProfile`.

Yao725 (approved):

PR #1991, opened 2022-02-17
Generate protocol methods in RestClient

ShivangiReja:
Generate protocol methods in RestClient if they are configured in autorest.md.

Here is an example of the configuration:
```yaml
protocol-method-list:
  - Create
  - Delete
```


test/TestProjects/ProtocolMethodsInRestClient/Generated/TestServiceRestClient.cs
```diff
@@ -0,0 +1,404 @@
...
+                throw;
+            }
+        }
+
+        /// <summary> Delete resource. </summary>
+        /// <param name="resourceId"> The id of the resource. </param>
+        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="resourceId"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="resourceId"/> is an empty string, and was expected to be non-empty. </exception>
+#pragma warning disable AZC0002
```
annelo-msft, line 289: Curious: do we still need these warning disables?

test/AutoRest.TestServer.Tests/ProtocolMethodsInRestClientTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        [TestCase("Delete", true)]
+        [TestCase("Create", true)]
+        [TestCase("Get", false)]
+        public void ProtocolMethodGeneratedInRestClient(string methodName, bool isProtocolMethodGenerated)
+        {
+            var methods = typeof(TestServiceRestClient).GetMethods();
+            Assert.IsNotNull(methods);
+
+            var restClientMethods = methods.Where(m => m.Name.Equals(methodName));
+            if (isProtocolMethodGenerated)
```
annelo-msft, line 22: It feels to me like we should be asserting that isProtocolMethodGenerated is equal to a value we learn from the generated client, rather than conditionally asserting based on its value.

ShivangiReja: [Here](https://github.com/Azure/autorest.csharp/blob/7d3f3ac43b97f4921f5b3f9c38202f11275ea60d/test/AutoRest.TestServer.Tests/ProtocolMethodsInRestClientTests.cs#L13-L15) I added the method names and an expected flags to verify that the protocol method should generate for this or not. [Here](https://github.com/Azure/autorest.csharp/blob/7d3f3ac43b97f4921f5b3f9c38202f11275ea60d/test/AutoRest.TestServer.Tests/ProtocolMethodsInRestClientTests.cs#L35) I'm asserting in [two of the methods](https://github.com/Azure/autorest.csharp/blob/7d3f3ac43b97f4921f5b3f9c38202f11275ea60d/test/AutoRest.TestServer.Tests/ProtocolMethodsInRestClientTests.cs#L24), one is generated as protocol method.

test/TestProjects/ProtocolMethodsInRestClient/Generated/TestServiceRestClient.cs
ShivangiReja: I just shared the same code we generate for DPG. Please let me know if we want to remove this warning for RestClient.

ShivangiReja: just want to clarify, are you thinking​ to remove the warning for all the DPG libraries or just for RestClient?

ShivangiReja: This warning is for:
`"DO ensure all service methods, both asynchronous and synchronous,  take an optional CancellationToken parameter called cancellationToken.", "Client method should have cancellationToken as the last optional parameter (both name and it being optional matters"` ([here](https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs#L17-L23))

So I don't think we should remove this warning because we don't take CancellationToken parameter for protocol methods.

annelo-msft: @KrzysztofCwalina told me he had turned off this analyzer, but I haven't looked into it yet.  Would you be willing to look in the azure-sdk-tools repo to see if this happened?

test/AutoRest.TestServer.Tests/ProtocolMethodsInRestClientTests.cs
annelo-msft: I guess my concern is that if the wrong value was passed in for isProtocolMethodGenerated, we would check the wrong test case.  It also seems strange [here](https://github.com/Azure/autorest.csharp/blob/7d3f3ac43b97f4921f5b3f9c38202f11275ea60d/test/AutoRest.TestServer.Tests/ProtocolMethodsInRestClientTests.cs#L40), that we're effectively saying

```
if (isProtocolMethodGenerated == false)
{
    Assert.IsFalse(isProtocolMethodGenerated);
}
```

ShivangiReja: `isProtocolMethodGenerated` values refers to the method names added in the config and that's how I'm verifying that these methods are generated as protocol methods. Otherwise, how would we know which methods we wants to verify?

I've removed:
` Assert.IsFalse(isProtocolMethodGenerated);`

ShivangiReja: `isProtocolMethodGenerated` values refer to the method names added in the config(true, if method name added in the config, false, if not) to verify that these methods are generated as protocol methods. Otherwise how do we verify these methods?

I've removed:
`Assert.IsFalse(isProtocolMethodGenerated);`

annelo-msft: Would it make sense to split this into two separate test cases, rather than branching on `isProtocolMethodGenerated`?

ShivangiReja: Updated!

test/TestProjects/ProtocolMethodsInRestClient/Generated/TestServiceRestClient.cs
ShivangiReja: Seems like we have turned off this analyzer because the code is compiling even if I remove `#pragma warning disable AZC0002`. I tested in [DeviceUpdate](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/deviceupdate/Azure.IoT.DeviceUpdate/src/Generated/DeviceUpdateClient.cs#L122) in azure-sdk-for-net repo.

ShivangiReja: @annelo-msft This PR will create a lot of noise if I do this change. Hence, I created a separate PR to remove this https://github.com/Azure/autorest.csharp/pull/1993. Could you please take a look?

annelo-msft: Approved!

annelo-msft (commented):
I am happy with the output of these changes; @AlexanderSher, would you please have a look at the generator implementation?

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -124,6 +125,20 @@ private static string ReadStringOption(JsonElement root, string option)
             }
         }
 
+        private static void WriteNonEmptyArray(Utf8JsonWriter writer, string name, string[] values)
```
AlexanderSher, line 128: This probably should be in some general `Utf8JsonWriter` extension/helper type.

src/AutoRest.CSharp/DataPlane/Output/DataPlaneRestClient.cs
```diff
@@ -1,23 +1,74 @@
...
 using AutoRest.CSharp.Output.Models.Types;
+using AutoRest.CSharp.Utilities;
 
 namespace AutoRest.CSharp.Output.Models
 {
     internal class DataPlaneRestClient : RestClient
     {
         private BuildContext<DataPlaneOutputLibrary> _context;
 
+        public IReadOnlyList<LowLevelClientMethod>? ProtocolMethods => _context.Configuration.ProtocolMethodList.Length > 0 && GetProtocolMethods() != null
```
AlexanderSher, line 19: Do we always have only one `*RestClient` per project?

ShivangiReja: No, it can be more than one RestClient per project and with this change if we have "Delete" `OperationId` in 2 RestClients, it will generate protocol method for both of them. Since the protocol method generated internally and we want to make it simple, we are taking only operationId for now. In future if it will create a lot of noise in RestClient, we will change and take key, value pair(<`OperationGroup`, List<<`OperationId`>>>).

AlexanderSher: Honestly, I'd prefer this all to be configured in partial types. But we can discuss it when/if we need more granulated configuration.

ShivangiReja: Yes, but in partial RestClient type, devs would have to write protocol method manually. We added this config, so that this method will be auto generated.

annelo-msft: We can have that discussion when we get there -- I'd like to understand how it would work in partial types, and what that looks like.  I think the solution we have is good for the current problem we're solving.

annelo-msft (approved):

PR #1992, opened 2022-02-18
Hide properties that have types that only have a single property inside of them

m-nash:
# Description
To see how this would look in azure-sdk-for-net see this PR https://github.com/Azure/azure-sdk-for-net/pull/27090

We want to automatically hide properties of objects that only have a single property inside of them. We do this making the original property internal and then exposing a wrapper to the inner property. This allows us to not have to worry about the serialization / deserialization since each class in the hierarchy still exists, but at the same time customers won't have to new an object just to set a single property.

```c#
        /// <summary> The AddressSpace that contains an array of IP address ranges that can be used by subnets. </summary>
        internal AddressSpace AddressSpace { get; set; }
        /// <summary> A list of address blocks reserved for this virtual network in CIDR notation. </summary>
        public IList<string> AddressPrefixes
        {
            get
            {
                if (AddressSpace is null)
                    AddressSpace = new AddressSpace();
                return AddressSpace.AddressPrefixes;
            }
        }
```

If a second property is added after GA we will mark the wrapper property with browsable never and make the internal property public.

There are only two cases for ctor one is default like the example above and the other is a single param.  Both cases are handled by the model writer.

# Checklist

# Ready to Land?


AlexanderSher:
What will happen if another property is added to `AddressSpace`?

samples/Azure.Management.Storage/Generated/Models/SmbSetting.cs
```diff
@@ -31,7 +31,19 @@ internal SmbSetting(Multichannel multichannel, string versions, string authentic
...
+        internal Multichannel Multichannel { get; set; }
+        /// <summary> Indicates whether multichannel is enabled. </summary>
+        public bool? MultichannelEnabled
+        {
+            get => Multichannel is null ? default : Multichannel.Enabled;
+            set
+            {
+                if (Multichannel is null)
+                    Multichannel = new Multichannel();
+                Multichannel.Enabled = value;
```
AlexanderSher, line 43: Do we treat adding required property to `Multichannel` as a breaking change?

m-nash: If you add a required parameter to any method including ctor thats a breaking change and shouldn't be allowed

ShivangiReja (approved):

PR #2009, opened 2022-02-23
Omit orphaned models using compilation functionality

### LINKED ISSUE #1585: Omit orphaned models
Models can be orphaned by skipping an entire operation group like Operations or by replacing the type with a reference type therefore the original is no longer referenced anywhere

example there are two resource objects so we have to explicitly specify one

https://github.com/Azure/autorest.csharp/pull/1427/files#diff-73c11042786cf1e5a8d918e0ee37dcb8934f391ae7dd17dd3d6cb84bb766481bR15

ArcturusZhang:
Fixes #1585 
Supersedes #2004 

# Description

Preview the changes here: https://github.com/Azure/azure-sdk-for-net/pull/27204

# Checklist

# Ready to Land?


AlexanderSher:
What is a performance impact of this change for `Azure.ResourceManager.Sample`?

ArcturusZhang:
> What is a performance impact of this change for `Azure.ResourceManager.Sample`?

This is WIP, I will give you a comparison after everything is done.

ArcturusZhang:
> What is a performance impact of this change for `Azure.ResourceManager.Sample`?

I am running this command to approximate the time during generating the sample project: `dotnet build && Measure-Command { .\eng\Generate.ps1 Azure.ResourceManager.Sample }`.
But I only run once for each case.
This is the result of the implementation in this PR:
```
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 58
Milliseconds      : 748
Ticks             : 587488205
TotalDays         : 0.000679963200231481
TotalHours        : 0.0163191168055556
TotalMinutes      : 0.979147008333333
TotalSeconds      : 58.7488205
TotalMilliseconds : 58748.8205
```
This is the result on branch `feature/v3`:
```
Days              : 0
Hours             : 0
Minutes           : 1
Seconds           : 18
Milliseconds      : 721
Ticks             : 787216563
TotalDays         : 0.00091113028125
TotalHours        : 0.02186712675
TotalMinutes      : 1.312027605
TotalSeconds      : 78.7216563
TotalMilliseconds : 78721.6563
```
I would say the performance impact should be very minor.

m-nash:
Please remove this class and verify this holistic solution covers all those cases SinglePropertyHider

ArcturusZhang:
This is the result of running this command:
```
dotnet build && Measure-Command {
    For ($i=0; $i -lt 20; $i++) {
        Write-Host Loop $i
        .\eng\Generate.ps1 Azure.ResourceManager.Sample
    }
}
```
On this PR:
```
Days              : 0
Hours             : 0
Minutes           : 17
Seconds           : 34
Milliseconds      : 124
Ticks             : 10541247874
TotalDays         : 0.0122005183726852
TotalHours        : 0.292812440944444
TotalMinutes      : 17.5687464566667
TotalSeconds      : 1054.1247874
TotalMilliseconds : 1054124.7874
```
On feature/v3 branch:
```
Days              : 0
Hours             : 0
Minutes           : 17
Seconds           : 7
Milliseconds      : 557
Ticks             : 10275579394
TotalDays         : 0.0118930317060185
TotalHours        : 0.285432760944444
TotalMinutes      : 17.1259656566667
TotalSeconds      : 1027.5579394
TotalMilliseconds : 1027557.9394
```
Therefore I believe the performance impact of this change is very minor

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -170,5 +171,236 @@ public async Task<CSharpCompilation> GetCompilationAsync()
         }
 
         public static bool IsGeneratedDocument(Document document) => document.Folders.Contains(GeneratedFolder);
+        private static bool IsMgmtRootDocument(Document document) => IsGeneratedDocument(document) && (!document.Name.Contains('/') || document.Name.Contains("Extensions/"));
```
AlexanderSher, line 174: `Path.GetDirectoryName(document.Name) is "Extensions" or string.Empty`

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -170,5 +171,236 @@ public async Task<CSharpCompilation> GetCompilationAsync()
...
+                {
+                    var root = await document.GetSyntaxRootAsync();
+                    classVisitor.Visit(root);
+                }
+            }
+
+            return classVisitor.ModelDeclarations;
+        }
+
+        private async IAsyncEnumerable<SyntaxNode> TraverseAllPublicModels()
```
AlexanderSher, line 282: This method misses lots of cases, for example:
- If model 'A' is publicly referenced by model 'B' and model 'B' is publicly referenced by rest operation, both should be kept public
- If collection of models is publicly referenced by rest operation, it should be kept public
- If generic parameter constraint publicly references model, it should be kept public

and so on, and any combination of this, e.g.: `GetSomething<T>() where T: IEnumerable<SomeModel>`

It also works with syntax nodes rather than symbols, which is incorrect. I'll look a bit more into this today.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -170,5 +171,236 @@ public async Task<CSharpCompilation> GetCompilationAsync()
...
+                    break;
+                case IPropertySymbol propertySymbol:
+                    await ProcessSymbol(propertySymbol.Type, result);
+                    break;
+                default:
+                    throw new InvalidOperationException($"Not implemented for symbol {symbol.GetType()}");
+            }
+        }
+
+        public async Task RemoveOrphanedModels(ImmutableHashSet<string> modelsToKeep)
```
fengzhou-msft, line 369: Rename the method to something like `InternalizeOrphanedModels`?

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
ArcturusZhang: I believe the first two cases are covered by the current implementation, since this is also adding types in the type arguments recursively.
I did not find an example of the third case in mgmt generated code - do we have this case? I have generated all the existing RPs on this PR: https://github.com/Azure/azure-sdk-for-net/pull/27204
And a case of discriminator is also missing in the current implementation, like we have a model A returned by an operation, and it has multiple derived types from the discriminator and those derived types should be marked as public as well.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
ArcturusZhang: Unfortunately pattern matching can only apply to constants, and string.Empty is not a constant, therefore I have to use `""` instead of `string.Empty`.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
ArcturusZhang: Make sense since we are not removing anything

m-nash (approved):

PR #2010, opened 2022-02-23
Simplify LRO

AlexanderSher:
View with navigation: https://github.dev/Azure/autorest.csharp/pull/2010

This PR simplifies generation of  LRO's:
- `MgmtLongRunningOperationWriter` has been generating identical types with different names. They all are replaced with `RealArmOperation` and `RealArmOperation<T>` shared types
- Instantiation of `RealArmOperation` and `RealArmOperation<T>` is delegated to `ArmOperationHelpers.ProcessMessage` / `ArmOperationHelpers.ProcessMessageAsync` that accepts `HttpMessage`.
- For fake LRO's `ArmOperationHelpers` would create instances of `FakeArmOperation` / `FakeArmOperation<T>` 
- `MgmtRestClientWriter` is decoupled from `RestClientWriter` to minimize dependency on legacy HLC code.
- `OperationOrResponseInternals` / `OperationOrResponseInternals<T>` is not required anymore.


AlexanderSher:
This PR is a draft. If it is going to be merged, `ArmOperationHelpers` and supporting types should be added to the `Azure.ResourceManager` rather than directly to the `Autorest.CSharp`.

m-nash:
I don't think we want to incorporate these changes at this time.  Given this is all internal code and not affecting any functionality or performance I prefer to pick this conversation back up after GA if we still feel this is necessary.

PR #2012, opened 2022-02-25
Cleanup

AlexanderSher:
This PR doesn't change any generated code

- Make `IsEnum` feature of type rather than parameter of that type
- Add `MethodSignatureModifiers` enum to implement `MethodSignature.WithAsync` that helps creating async signatures from sync and visa versa
- Remove arm-specific code from `WriteMethodDeclaration`

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
@@ -7,7 +7,7 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using AutoRest.CSharp.Generation.Types;
-using AutoRest.CSharp.Mgmt.Decorator;
+using AutoRest.CSharp.Input;
 using AutoRest.CSharp.Output.Models;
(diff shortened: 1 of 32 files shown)


src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -92,74 +93,45 @@ public static CodeWriter WriteFieldDeclaration(this CodeWriter writer, FieldDecl
...
-                var firstParam = method.Parameters.FirstOrDefault();
-                bool isExtensionMethod = firstParam is not null && firstParam.IsExtensionParameter;
-
-                if (method.Modifiers.Contains("public") && !isExtensionMethod)
-                    writer.Append($"virtual ");
-
-                if (isExtensionMethod)
-                    writer.Append($"static ");
+                writer
+                    .AppendRawIf("async ", methodBase.Modifiers.HasFlag(Async) && Configuration.AzureArm)
```
AlexanderSher, line 111: ARM has been writing `async` before `virtual`. Canonical order is when `async` goes after `virtual`, but it can be fixed in a separate PR.

ShivangiReja (approved):

PR #2017, opened 2022-02-28
Replace ClientDiagnostics.CreateRequestFailedExceptionAsync with new RequestFailedException(Response)

AlexanderSher:
This PR removes all dependencies on `ClientDiagnostics.CreateRequestFailedExceptionAsync` from generate code.


samples/Azure.ResourceManager.Sample/Generated/RestOperations/AvailabilitySetsRestOperations.cs
```diff
@@ -105,7 +105,7 @@ public async Task<Response<AvailabilitySetData>> CreateOrUpdateAsync(string subs
                         return Response.FromValue(value, message.Response);
                     }
                 default:
-                    throw await ClientDiagnostics.CreateRequestFailedExceptionAsync(message.Response).ConfigureAwait(false);
+                    throw new RequestFailedException(message.Response);
```
m-nash, line 108: Can we remove the ClientDiagnostics property / ctor parameter since its not needed any more

ShivangiReja: We need ClientDiagnostics property and ctor parameter when we generate protocol method in RestClient for some cases for example: [ProtocolMethodsInRestClient/Generated/TestServiceRestClient.cs](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/ProtocolMethodsInRestClient/Generated/TestServiceRestClient.cs#L136)

m-nash: is it possible to detect that and omit if there are none of those?  I believe its probably the majority case that it isn't needed

test/TestProjects/ProtocolMethodsInRestClient/Generated/TestServiceRestClient.cs
```diff
@@ -296,74 +365,6 @@ public virtual Response Delete(string resourceId, RequestContext context = null)
...
-
-        /// <summary> Retrieves information about the resource. </summary>
-        /// <param name="resourceId"> The id of the resource. </param>
-        /// <param name="cancellationToken"> The cancellation token to use. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="resourceId"/> is null. </exception>
-        public async Task<Response<Resource>> GetAsync(string resourceId, CancellationToken cancellationToken = default)
-        {
-            if (resourceId == null)
-            {
-                throw new ArgumentNullException(nameof(resourceId));
```
ShivangiReja, line 321: Why did we move all the protocol methods at the end? I think it is more convenient if we have this structure:
* Create request method for HLC
* Async HLC method
* Sync HLC method
* Create request method for protocol method
* Async version of  protocol method
* Sync version of protocol method

Instead of first all the HLC methods and then all the protocol methods.

samples/Azure.Management.Storage/Generated/RestOperations/BlobContainersRestOperations.cs
```diff
@@ -112,10 +99,9 @@ public async Task<Response<ListContainerItems>> ListAsync(string subscriptionId,
                         return Response.FromValue(value, message.Response);
                     }
                 default:
-                    throw await ClientDiagnostics.CreateRequestFailedExceptionAsync(message.Response).ConfigureAwait(false);
+                    throw new RequestFailedException(message.Response);
             }
         }
-
```
annelo-msft, line 118: Was it intentional to remove this line break?  I like it for readability.

test/TestProjects/ProtocolMethodsInRestClient/Generated/TestServiceRestClient.cs
AlexanderSher: I don't know if we have any preferences. This file is codegened, it won't be edited by hand. Intellisence sorts everything alphabetically. From the before/after comparison perspective, it is easier to check added methods if they are in one place rather than spread across the file. 

@annelo-msft , what do you think?

annelo-msft (approved):
Data-plane looks great, thanks for this!

Curious why there are no changes to TestServerProjectsLowLevel?

AlexanderSher:
> Curious why there are no changes to TestServerProjectsLowLevel?

For LLC, exception is thrown by HttpPipelineExtensions, which has been updated already: https://github.com/Azure/autorest.csharp/blob/feature/v3/src%2Fassets%2FGenerator.Shared%2FHttpPipelineExtensions.cs#L32

samples/Azure.Management.Storage/Generated/RestOperations/BlobContainersRestOperations.cs
AlexanderSher: No, accidental. Thanks for noticing!

test/TestProjects/ProtocolMethodsInRestClient/Generated/TestServiceRestClient.cs
AlexanderSher: Fixed

samples/Azure.Management.Storage/Generated/RestOperations/BlobContainersRestOperations.cs
AlexanderSher: Fixed

src/AutoRest.CSharp/Mgmt/Output/MgmtRestClient.cs
```diff
@@ -2,30 +2,55 @@
...
+        public static readonly Parameter ApplicationIdParameter = new("applicationId", "The application id to use for user agent", new CSharpType(typeof(string)), null, false);
+
+        private readonly RestClientBuilder _clientBuilder;
         private IReadOnlyList<Resource>? _resources;
 
-        public MgmtRestClient(OperationGroup operationGroup)
-            : base(operationGroup, MgmtContext.Context, operationGroup.Language.Default.Name, new MgmtRestClientBuilder(operationGroup))
+        public ClientFields Fields { get; }
+
+        public MgmtRestClient(OperationGroup operationGroup, RestClientBuilder clientBuilder)
```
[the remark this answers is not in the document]
AlexanderSher: Fixed

PR #2043, opened 2022-03-09
Rename Type property

fengzhou-msft:
# Description
This PR uses an Azure.ResourceManager version that also removed ErrorResponse common types.

# Checklist

# Ready to Land?


test/TestProjects/ReferenceTypes/Generated/Models/ErrorAdditionalInfo.cs
```diff
@@ -21,17 +20,17 @@ public ErrorAdditionalInfo()
...
+        internal ErrorAdditionalInfo(string errorAdditionalInfoType, object info)
         {
-            Type = type;
+            ErrorAdditionalInfoType = errorAdditionalInfoType;
             Info = info;
         }
 
         /// <summary> The additional info type. </summary>
-        public ResourceType Type { get; }
+        public string ErrorAdditionalInfoType { get; }
```
fengzhou-msft, line 33: Previously, a property with name `Type` in ReferenceType will use ResourceType type, here it's not correct. Change back to string with the changes in this PR.

test/TestProjects/ReferenceTypes/Generated/Models/CheckNameAvailabilityRequest.cs
```diff
@@ -22,17 +21,17 @@ public CheckNameAvailabilityRequest()
...
             Name = name;
-            Type = type;
+            CheckNameAvailabilityRequestType = checkNameAvailabilityRequestType;
         }
 
         /// <summary> The name of the resource for which availability needs to be checked. </summary>
         public string Name { get; set; }
         /// <summary> The resource type. </summary>
-        public ResourceType Type { get; set; }
+        public string CheckNameAvailabilityRequestType { get; set; }
```
fengzhou-msft, line 35: This model only has name and type, no id. Therefore, the result of the auto-renaming is not what we want. To correct it, we can set x-ms-client-name to `ResourceType`. The result will become `public ResourceType ResourceType`.

samples/Azure.Management.Storage/Azure.Management.Storage.csproj
```diff
@@ -11,11 +11,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.ResourceManager" Version="1.0.0-alpha.20220310.5" />
+    <PackageReference Include="Azure.ResourceManager" Version="1.0.0-alpha.20220312.1" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.23.0-alpha.20220310.5" />
+    <PackageReference Include="Azure.Core" Version="1.23.0-alpha.20220312.1" />
```
Yao725, line 18: Why do we need to upgrade the version of Azure.Core? Is it because the definition of `ResourceType` live in Azure.Core?

Yao725 (approved):

samples/Azure.Management.Storage/Azure.Management.Storage.csproj
fengzhou-msft: I believe it's not because of some feature change. I have not figured out the root cause, but now if you publish an alpha version of Azure.ResourceManager, it relies on Azure.Core with the same alpha version number. And if we don't update the Azure.Core version here, compiler will complain that there is a downgrade of Azure.Core. We can probably remove the explicit dependency of Azure.Core here since Azure.ResourceMananger depends on Azure.Core.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -72,10 +72,38 @@ public static void VerifyAndUpdateFrameworkTypes(this IEnumerable<Schema> allSch
...
+                    // With UpdateAcronyms processing, the case of the name may be changed, so use ignore case comparison.
+                    // Do not use property.SerializedName=="type" so that we can still use x-ms-client-name to override the auto-renaming here if there is some edge case.
+                    if (property.Language.Default.Name.Equals("type", StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (property.Schema.Type == AllSchemaTypes.String
+                            && objSchema.ContainsStringProperty("id")
+                            && objSchema.ContainsStringProperty("name"))
+                        {
+                            property.Language.Default.Name = "resourceType";
+                        }
```
ArcturusZhang, line 84: Here we are checking if a schema is a schema of a resource, we should have similar things in `Decorator/ResourceDetection`

samples/Azure.Management.Storage/Azure.Management.Storage.csproj
Yao725: I have taken a look at recently changes. This is probably because that we change the dependency of Azure.Core from package reference to project reference.

fengzhou-msft: Ah, that explains it.

ArthurMa1978:
https://github.com/azure/azure-sdk-for-net/issues/27297

PR #2047, opened 2022-03-09
Fix #2042: Update Generation for new LRO signatures

### LINKED ISSUE #2042: Update Generation for new LRO signatures
New LRO method signatures are:

```csharp
public virtual Operation<BinaryData> DoThing(WaitUntil waitUntil, ..., RequestContext = default)
```

We will need to update the generator to make this change.

AlexanderSher:


ShivangiReja (approved):
LGTM

PR #2051, opened 2022-03-10
Extend x-ms-format to support ResourceIdentitifer, ResourceType

fengzhou-msft:
# Description
| x-ms-format value | C# Type | 
| --------------- | --------------- |
| arm-id | ResourceIdentifier |
| resource-type | ResourceType |
| duration-constant | TimeSpan |

Previously, for "x-ms-format": "duration-constant", you must add "format": "duration" as well. This PR removed that requirement, so that you only need to add "x-ms-format": "duration-constant".

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -62,9 +62,13 @@ public static Constant ParseConstant(object? value, CSharpType type)
...
             DateTimeSchema {Format: DateTimeSchemaFormat.DateTimeRfc1123} => SerializationFormat.DateTime_RFC1123,
             DateSchema _ => SerializationFormat.Date_ISO8601,
-            DurationSchema _ => schema.Extensions?.Format?.Equals("duration-constant") == true ? SerializationFormat.Duration_Constant : SerializationFormat.Duration_ISO8601,
+            DurationSchema _ => schema.Extensions?.Format?.Equals(XMsFormat.DurationConstant) == true ? SerializationFormat.Duration_Constant : SerializationFormat.Duration_ISO8601,
             TimeSchema _ => SerializationFormat.Time_ISO8601,
-            _ => SerializationFormat.Default
+            _ => schema.Extensions?.Format switch
+                {
+                    XMsFormat.DurationConstant => SerializationFormat.Duration_Constant,
+                    _ => SerializationFormat.Default
```
fengzhou-msft, line 70: This supports the case that only `"x-ms-format": "duration-constant"` is added to the property and we omit `"format":"duration"` which seems redundant. The schema will be treated as string type by modelerfour, therefore add a check here.

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
```diff
@@ -133,6 +133,13 @@ public string[] Formats
         public string? Format => TryGetValue("x-ms-format", out object? value) ? value?.ToString() : null;
     }
 
+    internal static class XMsFormat
```
ArcturusZhang, line 136: Should this be `XmsFormat`?
```suggestion
    internal static class XmsFormat
```

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
```diff
@@ -133,6 +133,13 @@ public string[] Formats
...
+    {
+        public const string ArmId = "armId";
```
ArcturusZhang, line 138: So why not just simple value `resourceId` like `resourceType` in the next line? Is this avoiding some conflicts with something else?

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
```diff
@@ -133,6 +133,13 @@ public string[] Formats
...
+        public const string ResourceType = "resourceType";
+        public const string DurationConstant = "duration-constant";
```
ArcturusZhang, line 140: The value here are using different styles, how about we unify it to use hypens or camel cases? So that in the future we will not get confused

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -62,9 +62,13 @@ public static Constant ParseConstant(object? value, CSharpType type)
             DateTimeSchema {Format: DateTimeSchemaFormat.DateTime} => SerializationFormat.DateTime_ISO8601,
             DateTimeSchema {Format: DateTimeSchemaFormat.DateTimeRfc1123} => SerializationFormat.DateTime_RFC1123,
             DateSchema _ => SerializationFormat.Date_ISO8601,
-            DurationSchema _ => schema.Extensions?.Format?.Equals("duration-constant") == true ? SerializationFormat.Duration_Constant : SerializationFormat.Duration_ISO8601,
+            DurationSchema _ => schema.Extensions?.Format?.Equals(XMsFormat.DurationConstant) == true ? SerializationFormat.Duration_Constant : SerializationFormat.Duration_ISO8601,
             TimeSchema _ => SerializationFormat.Time_ISO8601,
-            _ => SerializationFormat.Default
+            _ => schema.Extensions?.Format switch
+                {
+                    XMsFormat.DurationConstant => SerializationFormat.Duration_Constant,
```
ArcturusZhang, line 69: Why we have this again since we already checked this in the section of duration schema? Is it by design that this `x-ms-format: duration-constant` will only take effect together with `format: duration`? Or the `format duration` is actually optional?

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
fengzhou-msft: This comes from x-ms-format, so I use the camel case XMsFormat.

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
fengzhou-msft: You mean name it resourceId instead of armId? I thought armId is more commonly used in terms of what its format should look like.

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
fengzhou-msft: I was using a format like `theFormat-variant`, when you have only one variant or has a default one, you use `theFormat`. I had a check of the datetime format, it's using `date-time` or variants like `date-time-rfc1123`. Therefore, I will change it to `the-format-variant`.

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
fengzhou-msft: modelerfour will only convert the schema to duration type if we have `format: duration`. If we only add `format: duration-constant`, modelerfour will treat it as string. By adding this check here and also in TypeFactory to return TimeSpan when we have `format: duration-constant` for a string schema, we no longer require `format: duration` when we already set `format: duration-constant`, which is simpler for users.

m-nash (approved):

PR #2053, opened 2022-03-10
Revert ClientDiagnostics removal from DataPlane clients

AlexanderSher:


samples/AppConfiguration/Generated/AppConfigurationRestClient.cs
```diff
@@ -24,14 +23,19 @@ internal partial class AppConfigurationRestClient
...
         /// <param name="endpoint"> The endpoint of the App Configuration instance to send requests to. </param>
         /// <param name="syncToken"> Used to guarantee real-time consistency between requests. </param>
         /// <param name="apiVersion"> Api Version. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="pipeline"/>, <paramref name="endpoint"/> or <paramref name="apiVersion"/> is null. </exception>
-        public AppConfigurationRestClient(HttpPipeline pipeline, string endpoint, string syncToken = null, string apiVersion = "1.0")
+        /// <exception cref="ArgumentNullException"> <paramref name="clientDiagnostics"/>, <paramref name="pipeline"/>, <paramref name="endpoint"/> or <paramref name="apiVersion"/> is null. </exception>
+        public AppConfigurationRestClient(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, string endpoint, string syncToken = null, string apiVersion = "1.0")
         {
+            ClientDiagnostics = clientDiagnostics ?? throw new ArgumentNullException(nameof(clientDiagnostics));
```
ShivangiReja, line 38: @AlexanderSher why are we reverting these changes?

AlexanderSher: `ClientDiagnostics` type has a virtual method that can be overridden: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs#L52

AlexanderSher: If `ExtractFailureContent` is overridden in derived type, there is no way to use it when we call `RequestFailedException` constructor directly.

christothes: Would you mind adding this to the PR description and also linking the issue that will resolve this in the future?

AlexanderSher: https://github.com/Azure/azure-sdk-for-net/issues/27490

PR #2057, opened 2022-03-11
fix the tenant parent issue

ArcturusZhang:
# Description

This PR adds a condition in `ParentDetection` - if the parent request path is the same as the current one, we break the detection by return an empty list.

If the path of a resource's parent is the same as itself, the parent will also be itself. In this case, we will have a method to GetMySelf inside its own class.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/ParentDetection.cs
```diff
@@ -44,6 +44,13 @@ private static IEnumerable<MgmtTypeProvider> GetParent(this Resource resource)
         private static IEnumerable<MgmtTypeProvider> GetParent(this OperationSet resourceOperationSet)
         {
             var parentRequestPath = resourceOperationSet.ParentRequestPath();
+
+            if (parentRequestPath.Equals(resourceOperationSet.GetRequestPath()))
+            {
+                // my parent is myself? Only tenant has this attribute, return empty
+                return Enumerable.Empty<MgmtTypeProvider>();
```
ArthurMa1978, line 51: Consider to add a test

ArcturusZhang: It is very unfortunate that this is very hard to add unit test because this involves quite a few entities from the code model, and building the code model in a unit test is not very practical.
I will talk to Feng to see if we have a test project that generates azure.resourcemanager, we could use that as a test for this.

archerzz: Right. In the future we should focus on the testability. I think some of our codes are too coupled, which makes it hard to do unit testing.

archerzz:
Just curious, why do we need a `GetMySelf`?

ArcturusZhang:
> Just curious, why do we need a `GetMySelf`?

We do not, and this is why we have this PR, to make sure we do not have that

ArthurMa1978 (approved):

Yao725 (approved):

PR #2058, opened 2022-03-11
feat(mgmt): add `Resource` suffix to resource types

### LINKED ISSUE #2056: add `Resource` suffix to resource types

archerzz:
1. change the `Resource.DefaultName` to return the `ResourceName` with suffix `Resource` if it doesn't end with `Resource`
2. update samples and test projects
3. update test cases

Here are the naming rule changes:
|                             | Before                        | After                                 |
| --------------------------- | ----------------------------- | ------------------------------------- |
| Resource                    | {ResourceName}                | {ResourceName}Resource                |
| GetResourceMethod           | Get{ResourceName}             | Get{ResourceName}                     |
| GetResourceCollectionMethod | Get{ResourceName}s            | Get{ResourceName}Resources            |
| Operation Source            | {ResourceName}OperationSource | {ResourceName}ResourceOperationSource |

resolve #2056

# Description

# Checklist

# Ready to Land?


test/AutoRest.TestServer.Tests/Mgmt/Unit/StringExtensionsTests.cs
```diff
@@ -17,5 +17,15 @@ public void ValidateResourceNameToPlural(string resourceName, string expected)
             var result = resourceName.ResourceNameToPlural();
             Assert.AreEqual(expected, result);
         }
+
+        [TestCase("MetadataRole", "MetadataRoleResource")]
+        [TestCase("PrivateResource", "PrivateResource")]
+        [TestCase("PrivateResource2", "PrivateResource2Resource")]
```
archerzz, line 23: Here is an open question: how do we deal with `XXXResourceYYY`? I think `PrivateResource2` is an extreme example which should not happen in reality. So maybe we just stick to simply appending a `Resource` suffix if not exists?

ArcturusZhang: This particular case is because we are making up names in our test project. Feel free to change the name of `PrivateResource2` if it bothers you

But it could happen in real life cases because we have the case of "extension resource" - multiple resources with same definition but different parents, and in this case, the parent's name will prepend to the name of this resource. We should make sure in this case, the `Resource` suffix on parent's resource name does not appear in the middle of child resource.

The RP `appservice` has this kind of cases a lot, please try to generate appservice to see what will happen

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -170,8 +170,8 @@ public Resource(IReadOnlyDictionary<OperationSet, IEnumerable<Operation>> allOpe
 
         public virtual Resource GetResource() => this;
 
-        //private string? _defaultName;
-        protected override string DefaultName => ResourceName;
+        // name after `{ResourceName}Resource`, unless the `ResourceName` already ends with `Resource`
+        protected override string DefaultName => ResourceName.AddResourceSuffixToResourceName();
```
ArcturusZhang, line 174: If we append the `Resource` suffix here, we might see the case when a resource name prepends to its child's resource name, the `Resource` suffix appears in the middle of a name.

The function calculating the resource name is here: https://github.com/Azure/autorest.csharp/blob/52af1748e404411abbc53504e5abc5149f76694a/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs#L517
What do you think we first make sure the name returned by the above function does not end with the word `Resource` and append the `Resource` suffix here?
We do not need to worry the name between two resources will collide - which will happen if we do not do this in this way, and we have insurance on this because we will throw exceptions when two files are written with the same name

samples/Azure.Management.Storage/Generated/Extensions/ResourceGroupExtensionClient.cs
```diff
@@ -32,9 +32,9 @@ private string GetApiVersionOrNull(ResourceType resourceType)
             return apiVersion;
         }
 
-        /// <summary> Gets a collection of StorageAccounts in the StorageAccount. </summary>
-        /// <returns> An object representing collection of StorageAccounts and their operations over a StorageAccount. </returns>
-        public virtual StorageAccountCollection GetStorageAccounts()
+        /// <summary> Gets a collection of StorageAccountResources in the StorageAccountResource. </summary>
+        /// <returns> An object representing collection of StorageAccountResources and their operations over a StorageAccountResource. </returns>
+        public virtual StorageAccountCollection GetStorageAccountResources()
```
ArcturusZhang, line 37: We also need a decision on this: Do we need the Resource suffix after this function? I personally suppose `GetStorageAccounts` is more readable, especially this method is not returning a `StorageAccountResource` instance.

samples/Azure.Management.Storage/Generated/Extensions/ArmClientExtensions.cs
```diff
@@ -13,193 +13,193 @@ namespace Azure.Management.Storage
...
-        #region BlobService
-        /// <summary> Gets an object representing a BlobService along with the instance operations that can be performed on it but with no data. </summary>
+        #region BlobServiceResource
+        /// <summary> Gets an object representing a BlobServiceResource along with the instance operations that can be performed on it but with no data. </summary>
         /// <param name="client"> The <see cref="ArmClient" /> instance the method will execute against. </param>
         /// <param name="id"> The resource ID of the resource to get. </param>
-        /// <returns> Returns a <see cref="BlobService" /> object. </returns>
-        public static BlobService GetBlobService(this ArmClient client, ResourceIdentifier id)
+        /// <returns> Returns a <see cref="BlobServiceResource" /> object. </returns>
+        public static BlobServiceResource GetBlobServiceResource(this ArmClient client, ResourceIdentifier id)
```
ArcturusZhang, line 21: This method name needs discussion as well, but this makes more sense with this name comparing with the occurrence that returns a `Collection`, after all this is returning a `BlobServiceResource` instance.

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
ArcturusZhang: We might need to ensure `ResourceName` never ends with the word `Resource`, and we can change here like this
```suggestion
        // name after `{ResourceName}Resource`, unless the `ResourceName` already ends with `Resource`
        protected override string DefaultName => ResourceName + _suffix; // where _suffix is `Resource`
```

PR #2065, opened 2022-03-14
Improve type renaming

fengzhou-msft:
# Description

# Checklist

# Ready to Land?

### DIFF
--- a/samples/Azure.Management.Storage/Generated/Models/StorageAccountCheckNameAvailabilityParameters.cs
+++ b/samples/Azure.Management.Storage/Generated/Models/StorageAccountCheckNameAvailabilityParameters.cs
@@ -23,12 +23,12 @@ namespace Azure.Management.Storage.Models
             }
 
             Name = name;
-            Type = "Microsoft.Storage/storageAccounts";
+            ResourceType = "Microsoft.Storage/storageAccounts";
         }
 
         /// <summary> The storage account name. </summary>
         public string Name { get; }
         /// <summary> The type of resource, Microsoft.Storage/storageAccounts. </summary>
-        public string Type { get; }
+        public string ResourceType { get; }
     }
 }
--- a/src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
+++ b/src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
@@ -76,11 +76,7 @@ namespace AutoRest.CSharp.Mgmt.Decorator
                     // Do not use property.SerializedName=="type" so that we can still use x-ms-client-name to override the auto-renaming here if there is some edge case.
                     if (property.Language.Default.Name.Equals("type", StringComparison.OrdinalIgnoreCase))
                     {
-                        // TODO: we may reuse the IsResourceModel method from SchemaExtension, but the result for flattened properties is different as although models with matched flattened properties are not treated as Resource but they still inherit from ResourceData. Here we align with the inheritance match to include flattened properties for match.
-                        if (property.Schema.Type == AllSchemaTypes.String
-                            && objSchema.ContainsStringProperty("id")
-                            && objSchema.ContainsStringProperty("name")
-                            || objSchema.Language.Default.Name.Contains("NameAvailability", StringComparison.Ordinal))
+                        if (objSchema.IsResourceData() || objSchema.Language.Default.Name.Contains("NameAvailability", StringComparison.Ordinal))
                         {
                             property.Language.Default.Name = "resourceType";
                         }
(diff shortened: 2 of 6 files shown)


src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -76,17 +76,19 @@ public static void VerifyAndUpdateFrameworkTypes(this IEnumerable<Schema> allSch
...
                         if (property.Schema.Type == AllSchemaTypes.String
                             && objSchema.ContainsStringProperty("id")
-                            && objSchema.ContainsStringProperty("name"))
+                            && objSchema.ContainsStringProperty("name")
+                            || objSchema.Language.Default.Name.Contains("NameAvailability", StringComparison.Ordinal))
                         {
                             property.Language.Default.Name = "resourceType";
                         }
-                        else if (property.Schema.Name.EndsWith("Type", StringComparison.Ordinal))
+                        else if (property.Schema.Name.EndsWith("Type", StringComparison.Ordinal) && property.Schema.Name.Length != 4)
```
fengzhou-msft, line 87: There are cases when an enum is named `Type` which becomes the schema name and we should exclude that.

samples/Azure.Management.Storage/Generated/Models/StorageAccountCheckNameAvailabilityParameters.cs
```diff
@@ -23,12 +23,12 @@ public StorageAccountCheckNameAvailabilityParameters(string name)
...
             Name = name;
-            Type = "Microsoft.Storage/storageAccounts";
+            ResourceType = "Microsoft.Storage/storageAccounts";
         }
 
         /// <summary> The storage account name. </summary>
         public string Name { get; }
         /// <summary> The type of resource, Microsoft.Storage/storageAccounts. </summary>
-        public string Type { get; }
+        public string ResourceType { get; }
```
fengzhou-msft, line 32: This is the case where an enum is named `Type`, then previous code bug renamed `Type` to `Type`. Therefore it didn't throw exception. And the generated type is string because the enum only has one value and generator transformed it to a string with hard-coded value.

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -76,17 +76,19 @@ public static void VerifyAndUpdateFrameworkTypes(this IEnumerable<Schema> allSch
                     // Do not use property.SerializedName=="type" so that we can still use x-ms-client-name to override the auto-renaming here if there is some edge case.
                     if (property.Language.Default.Name.Equals("type", StringComparison.OrdinalIgnoreCase))
                     {
+                        // TODO: we may reuse the IsResourceModel method from SchemaExtension, but the result for flattened properties is different as although models with matched flattened properties are not treated as Resource but they still inherit from ResourceData. Here we align with the inheritance match to include flattened properties for match.
                         if (property.Schema.Type == AllSchemaTypes.String
                             && objSchema.ContainsStringProperty("id")
-                            && objSchema.ContainsStringProperty("name"))
+                            && objSchema.ContainsStringProperty("name")
+                            || objSchema.Language.Default.Name.Contains("NameAvailability", StringComparison.Ordinal))
```
ArcturusZhang, line 83: Shall we add some parenthesis to emphasize the priority of this logical expression?
Or introduce another function to increase the readability here?
I believe we are changing this in this branch, if either
1. this object also has `id` and `name`, which means this is a resource model
2. this object has `NameAvailability` in it, which means this is a check name availability request
Is this correct?

ArcturusZhang (approved):

src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -76,11 +76,7 @@ public static void VerifyAndUpdateFrameworkTypes(this IEnumerable<Schema> allSch
                     // Do not use property.SerializedName=="type" so that we can still use x-ms-client-name to override the auto-renaming here if there is some edge case.
                     if (property.Language.Default.Name.Equals("type", StringComparison.OrdinalIgnoreCase))
                     {
-                        // TODO: we may reuse the IsResourceModel method from SchemaExtension, but the result for flattened properties is different as although models with matched flattened properties are not treated as Resource but they still inherit from ResourceData. Here we align with the inheritance match to include flattened properties for match.
-                        if (property.Schema.Type == AllSchemaTypes.String
-                            && objSchema.ContainsStringProperty("id")
-                            && objSchema.ContainsStringProperty("name")
-                            || objSchema.Language.Default.Name.Contains("NameAvailability", StringComparison.Ordinal))
+                        if (objSchema.IsResourceData() || objSchema.Language.Default.Name.Contains("NameAvailability", StringComparison.Ordinal))
```
ArcturusZhang, line 79: A minor comment is that: will it be safer if we call `CSharpName()` function when we are checking the name?
```

fengzhou-msft: Updated.

PR #2067, opened 2022-03-14
feat(mgmt): add `Resource` suffix to resource types

### LINKED ISSUE #2056: add `Resource` suffix to resource types

archerzz:
1. strip `Resource` suffix from `Resource.ResourceName` during model initialization.
2. change the `Resource.DefaultName` to `ResourceName` with suffix `Resource` if it doesn't end with `Resource`
3. change the `ResourceCollection.DefaultName` to
   `{ResourceName}Collection`
4. update samples and test projects
5. update test cases

Here are the naming rule changes:
- Prerequisite: `EntityName` is the `ResourceName` without suffix
  `Resource`, e.g. `PrivateEndpointResource` => `PrivateEndpoint`
- Below are the name changes:

|                             | Before                        | After                                 |
| --------------------------- | ----------------------------- | ------------------------------------- |
| Resource                    | {ResourceName}                | {EntityName}Resource                |
| Resource Data                 | {ResourceName}Data                | {EntityName}ResourceData           |
| Resource Collection              | {ResourceName}Collection                | {EntityName}Collection                |
| GetResourceMethod           | Get{ResourceName}             | Get{EntityName}Resource                     |
| GetResourceCollectionMethod | Get{ResourceName}s            | Get{EntityName}Resources            |
| Operation Source            | {ResourceName}OperationSource | {EntityName}ResourceOperationSource |

resolve #2056

Strip `Resource` suffix when generating resource types

# Description

# Checklist

# Ready to Land?


test/TestProjects/MgmtResourceName/readme.md
```diff
@@ -0,0 +1,18 @@
...
+``` yaml
+title: MgmtResourceName
+require: $(this-folder)/../../../readme.md
+azure-arm: true
+model-namespace: false
+input-file: $(this-folder)/MgmtResourceName.json
+namespace: MgmtResourceName
+
+
+request-path-to-resource-name:
```
m-nash, line 16: Add a test case that does a client rename and verify in test that it still adds Resource

src/AutoRest.CSharp/Mgmt/Decorator/StringExtensions.cs
```diff
@@ -20,5 +21,25 @@ public static string ResourceNameToPlural(this string resourceName)
...
+        }
+
+        /// <summary>
+        /// Trim `Resource` suffix of a resource name
+        /// </summary>
+        /// <param name="resourceName"></param>
+        /// <returns></returns>
+        public static string TrimResourceSuffixFromResourceName(this string resourceName)
+        {
+            return resourceName.EndsWith(ResourceSuffix) ? resourceName.Substring(0, resourceName.Length - ResourceSuffix.Length) : resourceName;
```
m-nash, line 42: In our discussion I thought there was a config allowing you to not do this for things like GenericResource.  Where is that configuration?

m-nash (changes requested):

src/AutoRest.CSharp/Mgmt/Decorator/StringExtensions.cs
ArcturusZhang: `request-path-to-resource-name` could do this

test/TestProjects/MgmtResourceName/Generated/MemoryResourceCollection.cs
```diff
@@ -0,0 +1,353 @@
...
+using Azure.Core;
+using Azure.Core.Pipeline;
+using Azure.ResourceManager;
+using Azure.ResourceManager.Core;
+using Azure.ResourceManager.Resources;
+
+namespace MgmtResourceName
+{
+    /// <summary> A class representing collection of MemoryResource and their operations over its parent. </summary>
+    public partial class MemoryResourceCollection : ArmCollection, IEnumerable<MemoryResource>, IAsyncEnumerable<MemoryResource>
```
ArcturusZhang, line 25: Hey @m-nash in this test project, we have the `request-path-to-resource-name` configuration which assigned the resource name of this to `MemoryResource` (which ends with `Resource`) and as you can see here, the collection keeps the `Resource` suffix from the config. Same thing needs to be done on `GenericResource` and we can have `GenericResourceCollection`.

src/AutoRest.CSharp/Mgmt/Decorator/StringExtensions.cs
archerzz: It's covered in the test project below. Thanks Dapeng for pointing it out.

test/TestProjects/MgmtResourceName/readme.md
archerzz: added a test case for `rename-model`

m-nash:
Before `ResourceName` equaled `EntityName`, now that we are appending "Resource" to the end they no longer equal.  For simplicity I am using `EntityName` everywhere which as an example would be `VirtualMachine`

Item | Before | After
-- | -- | --
Resource | {EntityName} | {EntityName}Resource
Resource Data | {EntityName}Data | {EntityName}Data
Resource Collection | {EntityName}Collection | {EntityName}Collection
GetResourceMethod | Get{EntityName} | Get{EntityName}
GetResourceCollectionMethod | Get{EntityName}s | Get{EntityName}s
Operation Source | {EntityName}OperationSource | {EntityName}OperationSource

archerzz:
## change back to only append `Resource` suffix for resource type and get resource extension method

1. roll back the following changes:
  1.1 `MgmtOutputLibrary`, we do not trim `Resource` suffix from `ResourceName`
  1.2 `StringExtension` (and corresponding test cases), we do not need the `TrimResourceSuffixFromResourceName`
  1.3 `ResourceData` and `MgmtObjectType`, resource data will use `{ResourceName}Data` as previous
2. replace `resource.Type.Name` with `resource.ResourceName` for the following names:
  2.1 get collection method name
  2.2 get single resource method name (service call)
  2.3 get single child resource method name (service call)
  2.4 operation source name
  2.5 get list resource method name (service call)
  2.6 client diagnstics name (otherwise it will be like `{ResourceName}Resource{ResourceName}ClientDiagnostics
  2.7 test case generation
3. update samples and test cases

archerzz:
Created a pre-check PR on `azure-sdk-for-net`: https://github.com/Azure/azure-sdk-for-net/pull/27588

test/AutoRest.TestServer.Tests/Mgmt/TestProjects/TestProjectTests.cs
```diff
@@ -157,8 +158,11 @@ private static void VerifyMethodReturnType(Type type, Type expectedType, MethodI
             Assert.AreEqual(expectedType, genericArgument, $"Return type did not match for {type.Name}.{method.Name}");
         }
 
-        private Type? GetResourceFromCollection(Type collection) => MyTypes().FirstOrDefault(t => t.Name == GetResourceNameFromCollectionName(collection.Name));
-        private string GetResourceNameFromCollectionName(string collectionName) => collectionName.Substring(0, collectionName.IndexOf("Collection"));
+        private Type? GetResourceFromCollection(Type collection)
+        {
+            var baseName = collection.Name.Substring(0, collection.Name.IndexOf("Collection"));
+            return MyTypes().FirstOrDefault(t => t.Name == baseName.AddResourceSuffixToResourceName() || t.Name == baseName);
```
archerzz, line 164: Since we have `no-resource-suffix` directive, here we cannot simply get the resource name by trimming `Collection` from the collection name.

samples/Azure.ResourceManager.Sample/Generated/Extensions/ResourceGroupExtensionClient.cs
```diff
@@ -32,50 +32,50 @@ private string GetApiVersionOrNull(ResourceType resourceType)
             return apiVersion;
         }
 
-        /// <summary> Gets a collection of AvailabilitySets in the AvailabilitySet. </summary>
-        /// <returns> An object representing collection of AvailabilitySets and their operations over a AvailabilitySet. </returns>
+        /// <summary> Gets a collection of AvailabilitySetResources in the AvailabilitySetResource. </summary>
+        /// <returns> An object representing collection of AvailabilitySetResources and their operations over a AvailabilitySetResource. </returns>
         public virtual AvailabilitySetCollection GetAvailabilitySets()
```
m-nash, line 37: what happened to GetAvailabilitySet(string availabilitySetName)?

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -467,11 +467,11 @@ protected FormattableString GetResourceTypeExpression(ResourceTypeSegment resour
             if (resourceType == ResourceTypeSegment.ResourceGroup)
                 return $"{typeof(ResourceGroup)}.ResourceType";
```
ArcturusZhang, line 468: Therefore we decide not to add `Resource` suffix to `ResourceGroup`?

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -170,8 +170,8 @@ public Resource(IReadOnlyDictionary<OperationSet, IEnumerable<Operation>> allOpe
 
         public virtual Resource GetResource() => this;
 
-        //private string? _defaultName;
-        protected override string DefaultName => ResourceName;
+        // name after `{ResourceName}Resource`, unless the `ResourceName` already ends with `Resource`
+        protected override string DefaultName => Configuration.MgmtConfiguration.NoResourceSuffix.Contains(ResourceName) ? ResourceName : ResourceName.AddResourceSuffixToResourceName();
```
ArcturusZhang, line 174: I am confused by this new configuration and think it might be redundant.
Which scenario is this new configuration for? Could you please explain how to use this new configuration in detail?
For instance, the word `GenericResource` is in this list, in this case, you will get the result `GenericResource` (the first branch is executed). If `GenericResource` is not in the list, you will still get the result `GenericResource` (the second branch is executed). Therefore which scenario this new configuration is designed for?

ArcturusZhang: Also this new configuration does not work with the name of a resource data (because I did not see any changes in the implementation of resource data), therefore the "trimming resource out of resource data's name" is still achieved by the `rename-model` directive. Is this statement correct?

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -37,7 +37,7 @@ public MgmtObjectType(ObjectSchema objectSchema, string? name = default, string?
 
         internal ObjectTypeProperty[] MyProperties => _myProperties ??= BuildMyProperties().ToArray();
 
-        private static string GetDefaultName(ObjectSchema objectSchema, bool isResourceType)
+        protected static string GetDefaultName(ObjectSchema objectSchema, bool isResourceType)
```
ArcturusZhang, line 40: Where do we use this outside this class? I did not see any

ArcturusZhang (commented):
Some comments and questions

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
archerzz: Yes and now.

- Previously it's decided that we will not add `Resource` suffix to the following 3 types:
  - `ResourceGroup`
  - `ResourceProvider` (aka `Provider`, but renamed to `ResourceProvider`)
  - `ResourceLink`
- After the latest discussion, we decided to unify the naming pattern, so all resource types should end with `Resource`.

I'll update this PR to change according to the latest disucssion.

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
archerzz: That's correct. The latest discussion is that service team still need to `rename-model` to remove the `Resource` from their swagger definition. So:
- If you have `Tenant` in swagger, then you'll get `TenantData`, `TenantResource`, `TenantCollection`.
- If you have `PrivateLinkResource` in swagger, then you'll get `PrivateLinkResourceData`, `PrivateLinkResource`, and `PrivateLinkResourceCollection`.

The new configuration is for `ResourceGroup`, `ResourceLink` and `ResourceProvider` (previously `Provider`). When we do not want to append `Resource` suffix. So if I add `no-resource-suffix: ResourceGroup`, I'll get `ResourceGroupData`, `ResourceGroup`, and `ResourceGroupCollection`.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
archerzz: I remember it's extension name or maybe rest operation name.

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
archerzz: In the latest discussion, we agree to keep `no-resource-suffix` for a while. If no usage, I'll move it from `autorest.csharp`.

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
ArcturusZhang: So the conclusion is that we will still have
- ResourceGroupResource
- ResourceProviderResource
- ResourceLinkResource

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
ArcturusZhang: I see the purpose - and it is useful in some cases. But I suggest we change the key of this configuration, instead of using the name of the resource as keys, we should use the request path as the key of this configuration.
The name of a resource could be generated programmatically, and it might change if new swaggers are added to the tag - but the request path of a resource is unique.

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
archerzz: @ArcturusZhang  That's correct. All now ends with `Resource`.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
archerzz: @ArcturusZhang  You're right. When I rolled back the `protected virtual` change, I forgot to change it back to `private`.

m-nash (approved):

PR #2072, opened 2022-03-16
[DPG] Fix methods with content types

### LINKED ISSUE #1746: [LLC] Generator creates duplicate when two MIME types are listed in consumes block

fengzhou-msft:
# Description
Resolve #1746

# Checklist

# Ready to Land?


test/TestProjects/NameConflicts/Generated/AutoRestParameterFlatteningRestClient.cs
```diff
@@ -233,9 +233,7 @@ internal Azure.Core.HttpMessage CreateAnalyzeBodyRequest(string stringBody)
             if (stringBody != null)
             {
                 request.Headers.Add("Content-Type", "application/json");
-                var content = new Utf8JsonRequestContent();
-                content.JsonWriter.WriteStringValue(stringBody);
-                request.Content = content;
+                request.Content = new StringRequestContent(stringBody);
```
annelo-msft, line 236: What made this change?

test/TestServerProjects/body-byte/Generated/ByteClient.cs
```diff
@@ -142,7 +142,7 @@ public virtual Response<byte[]> GetNonAscii(CancellationToken cancellationToken
         /// <summary> Put non-ascii byte string hex(FF FE FD FC FB FA F9 F8 F7 F6). </summary>
         /// <param name="byteBody"> Base64-encoded non-ascii byte string hex(FF FE FD FC FB FA F9 F8 F7 F6). </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual async Task<Response> PutNonAsciiAsync(byte[] byteBody, CancellationToken cancellationToken = default)
+        public virtual async Task<Response> PutNonAsciiAsync(string byteBody, CancellationToken cancellationToken = default)
```
annelo-msft, line 145: Why did `byte[]` change to string?

test/TestServerProjects/body-date/Generated/DateClient.cs
```diff
@@ -176,7 +176,7 @@ public virtual Response<DateTimeOffset> GetUnderflowDate(CancellationToken cance
         /// <summary> Put max date value 9999-12-31. </summary>
         /// <param name="dateBody"> date body. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual async Task<Response> PutMaxDateAsync(DateTimeOffset dateBody, CancellationToken cancellationToken = default)
+        public virtual async Task<Response> PutMaxDateAsync(string dateBody, CancellationToken cancellationToken = default)
```
annelo-msft, line 179: Why did `DateTimeOffset` change to string?

test/TestServerProjects/body-string/Generated/EnumClient.cs
```diff
@@ -75,7 +75,7 @@ public virtual Response<Colors> GetNotExpandable(CancellationToken cancellationT
         /// <summary> Sends value &apos;red color&apos; from enumeration of &apos;red color&apos;, &apos;green-color&apos;, &apos;blue_color&apos;. </summary>
         /// <param name="stringBody"> string body. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual async Task<Response> PutNotExpandableAsync(Colors stringBody, CancellationToken cancellationToken = default)
+        public virtual async Task<Response> PutNotExpandableAsync(string stringBody, CancellationToken cancellationToken = default)
```
annelo-msft, line 78: Why did the enum change to string?

test/TestServerProjectsLowLevel/dpg_update1/Generated/ParamsClient.cs
```diff
@@ -187,6 +187,68 @@ public virtual Response PutRequiredOptional(string requiredParam, string optiona
...
+        /// Schema for <c>Request Body</c>:
+        /// <code>{
+        ///   url: string (required)
+        /// }
+        /// </code>
+        /// 
+        /// </remarks>
+        public virtual async Task<Response> PostParametersAsync(RequestContent content, ContentType contentType, RequestContext context = null)
+        {
+            Argument.AssertNotNull(content, nameof(content));
```
annelo-msft, line 205: Why are we validating a body parameter here?  We should only be validating path parameters.

test/TestServerProjects/body-date/Generated/DateClient.cs
fengzhou-msft: Some of the type changes are caused by a bug in modelerfour and I updated its version to use the one in this fix [PR](https://github.com/Azure/autorest/pull/4477). They are fixed now.

test/TestProjects/MgmtKeyvault/tests/Generated/Mock/DeletedManagedHsmCollectionTest.cs
```diff
@@ -22,16 +21,5 @@ public DeletedManagedHsmCollectionMockTests(bool isAsync) : base(isAsync, Record
             ServicePointManager.ServerCertificateValidationCallback += (sender, cert, chain, sslPolicyErrors) => true;
             Environment.SetEnvironmentVariable("RESOURCE_MANAGER_URL", $"https://localhost:8443");
         }
-
-        [RecordedTest]
-        public async Task Get()
```
fengzhou-msft, line 27: Ignore all the changes in MgmtKeyvault. Mock tests in it has some issues and will be fixed later.

test/TestServerProjectsLowLevel/dpg_update1/Generated/ParamsClient.cs
AlexanderSher: No, in the latest discussion we have agreed that if entire body object is exposed in a public API and that object is marked as required like this we should assert not null on that object, but not its properties. 

@m-nash , correct me if I'm wrong

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
```diff
@@ -161,6 +161,55 @@ public static void SetRequestsToClients(IEnumerable<ClientInfo> clientInfos)
             {
                 foreach (var operation in clientInfo.Operations)
                 {
+                    if (operation.Requests.All(r => operation?.RequestMediaTypes?.Values.Contains(r) == true))
```
AlexanderSher, line 164: We should not modify `CodeModel` tree. While this practice is popular in mgmt plane, I don't want to move it into the DPG. `CodeModel` tree must remain unchanged source of truth.

test/TestServerProjectsLowLevel/dpg_update1/Generated/ParamsClient.cs
annelo-msft: @AlexanderSher, we updated the guidelines to reflect the new guidance: https://azure.github.io/azure-sdk/general_implementation.html#general-params-server-validation

Can you please share where the "latest discussion" information happened?

test/TestProjects/MgmtKeyvault/tests/Generated/Mock/DeletedManagedHsmCollectionTest.cs
fengzhou-msft: testmodeler is updated and MgmtKeyvault Mock tests are fixed.

samples/CognitiveServices.TextAnalytics/Generated/CognitiveServicesTextAnalyticsClient.cs
```diff
@@ -26,6 +27,27 @@ protected CognitiveServicesTextAnalyticsClient()
         {
         }
 
+        /// <summary> Initializes a new instance of CognitiveServicesTextAnalyticsClient. </summary>
+        /// <param name="endpoint"> Supported Cognitive Services endpoints (protocol and hostname, for example: https://westus.api.cognitive.microsoft.com). </param>
+        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
+        /// <param name="options"> The options for configuring the client. </param>
+        public CognitiveServicesTextAnalyticsClient(string endpoint, AzureKeyCredential credential, CognitiveServicesTextAnalyticsClientOptions options = null)
```
annelo-msft, line 34: Why does updating content type add this constructor?

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
```diff
@@ -68,6 +68,24 @@ private ClientFields(IEnumerable<Parameter> parameters, BuildContext? context)
...
+                        case KeySecurityScheme keySecurityScheme:
+                            AuthorizationHeaderConstant = new(Private | Const, typeof(string), "AuthorizationHeader", $"{keySecurityScheme.Name:L}");
+                            _keyAuthField = new(Private | ReadOnly, KnownParameters.KeyAuth.Type.WithNullable(false), "_" + KnownParameters.KeyAuth.Name);
+
+                            fields.Add(AuthorizationHeaderConstant);
+                            fields.Add(_keyAuthField);
+                            credentialFields.Add(_keyAuthField);
+                            parameterNamesToFields[KnownParameters.KeyAuth.Name] = _keyAuthField;
+                            break;
+                        case OAuth2SecurityScheme oAuth2SecurityScheme:
```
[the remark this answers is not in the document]
pshao25: They are same. But I couldn't find a way to match two types in the same case scope.

test/TestProjects/NameConflicts/Generated/AutoRestParameterFlatteningRestClient.cs
pshao25: The request content type is decided by Request.protocol.http.knownMediaType in M4. 
Previously, this request has protocol:
```
protocol: !Protocols 
  http: !HttpWithBodyRequest 
    path: /conflictingLROOverloads
    method: post
    knownMediaType: json
    mediaTypes:
      - application/json
    uri: '{$host}'
```
After m4 change, the protocol changes to:
```
protocol: !Protocols 
  http: !HttpBinaryRequest 
    path: /conflictingLROOverloads
    method: post
    binary: true
    knownMediaType: text
    mediaTypes:
      - application/json
    uri: '{$host}'
```
It looks like a regression. If you think so, I'll create an issue to Tim.

test/TestServerProjectsLowLevel/dpg_update1/Generated/ParamsClient.cs
pshao25: @AlexanderSher do you have any more information about this?

samples/CognitiveServices.TextAnalytics/Generated/CognitiveServicesTextAnalyticsClient.cs
pshao25: This is because of m4 change. The code model adds a new security schema. @timotheeguerin could you comment why this is added?
image

test/TestServerProjectsLowLevel/media_types/Generated/MediaTypesClient.cs
```diff
@@ -51,7 +51,7 @@ public MediaTypesClient(AzureKeyCredential credential, Uri endpoint = null, Medi
 
         /// <summary> Analyze body, that could be different media types. </summary>
         /// <param name="content"> The content to send as the body of the request. </param>
-        /// <param name="contentType"> Upload file type. Allowed values: &quot;application/pdf&quot; | &quot;image/jpeg&quot; | &quot;image/png&quot; | &quot;image/tiff&quot;. </param>
+        /// <param name="contentType"> Upload file type. Allowed values: &quot;application/json&quot; | &quot;application/pdf&quot; | &quot;image/jpeg&quot; | &quot;image/png&quot; | &quot;image/tiff&quot;. </param>
```
annelo-msft, line 54: Why is Json added here?  Did the swagger file change?

annelo-msft (commented):
The changes to dpg-update1 look good to me, but I don't understand why there are so many API changes to other clients when the API specs haven't changed.  Are these intended?

samples/CognitiveServices.TextAnalytics/Generated/CognitiveServicesTextAnalyticsClient.cs
timotheeguerin: https://github.com/Azure/autorest/blob/main/docs/releases/march-2022.md#modelerfour-security-scheme-generalization

test/TestProjects/NameConflicts/Generated/AutoRestParameterFlatteningRestClient.cs
timotheeguerin: So I have a fix for this case here, please give it a try and confirm it is resolving the issue https://github.com/Azure/autorest/pull/4506

However this bring another case that is not handle perfection due to the fact that `knownMediaType` is the same regardless of the content type passed.

In the case you have 
```yaml
consumes:
- application/json
- text/plain
- application/pdf
parameters: 
  - in: body
    schema:
      type: string
```

This would produce 2 overload

```
function foo(contentType: "application/pdf", bytes: Stream);
function foo(contentType: "application/json" | "text/plain", content: string);
```

In theory if you pass `application/json` as the content type for the other overload it should the content as a json string but if you pass text/plain it should pass it raw.

This means we should really need a map from content-Type to knownMediaType here which can be used to serialize differently depending on the passed type.

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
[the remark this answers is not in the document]
ShivangiReja: As per [this](https://github.com/Azure/autorest/blob/main/docs/releases/march-2022.md#modelerfour-security-scheme-generalization), `AADTokenSecurityScheme` got renamed to `OAuth2SecurityScheme` and `AzureKeySecurityScheme` got renamed to `KeySecurityScheme`, then why do we need `case AzureKeySecurityScheme` and `case AADTokenSecurityScheme` here?

samples/CognitiveServices.TextAnalytics/Generated/CognitiveServicesTextAnalyticsClient.cs
ShivangiReja: @annelo-msft - As per the changes in Modelerfour that @timotheeguerin shared, previously it was restricted with security scheme names and in this case it was looking for the `AzureKey` name and TextAnalytics swagger has [apim_key](https://github.com/Azure/azure-rest-api-specs/blob/c00cf4ad1e2c44ccddb08b98a472cf9822b4d687/specification/cognitiveservices/data-plane/TextAnalytics/stable/v3.0/TextAnalytics.json#L14) name. Now it supports ["type": "apiKey"](https://github.com/Azure/azure-rest-api-specs/blob/c00cf4ad1e2c44ccddb08b98a472cf9822b4d687/specification/cognitiveservices/data-plane/TextAnalytics/stable/v3.0/TextAnalytics.json#L15). Hence this constructor is generated with `AzureKeyCredential`.

ShivangiReja: @annelo-msft - As per the changes in Modelerfour that @timotheeguerin shared, previously it was restricted with security scheme names(AADToken and AzureKey) and in this case it was looking for the `AzureKey` name and TextAnalytics swagger has [apim_key](https://github.com/Azure/azure-rest-api-specs/blob/c00cf4ad1e2c44ccddb08b98a472cf9822b4d687/specification/cognitiveservices/data-plane/TextAnalytics/stable/v3.0/TextAnalytics.json#L14) name. Now it supports ["type": "apiKey"](https://github.com/Azure/azure-rest-api-specs/blob/c00cf4ad1e2c44ccddb08b98a472cf9822b4d687/specification/cognitiveservices/data-plane/TextAnalytics/stable/v3.0/TextAnalytics.json#L15), hence this constructor is generated with `AzureKeyCredential`.

pshao25: @annelo-msft looks by design to me.

test/TestProjects/NameConflicts/Generated/AutoRestParameterFlatteningRestClient.cs
pshao25: @timotheeguerin Where do you think we should store this map?

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -1279,15 +1280,15 @@ IEnumerable<Page<BinaryData>> CreateEnumerable(string nextLink, int? pageSizeHin
         /// </code>
         /// 
         /// </remarks>
-        public virtual async Task<Operation<BinaryData>> StartTranslationAsync(WaitUntil waitUntil, RequestContent content, RequestContext context = null)
+        public virtual async Task<Operation<BinaryData>> StartTranslationAsync(WaitUntil waitUntil, RequestContent content, ContentType contentType, RequestContext context = null)
```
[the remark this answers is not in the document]
pshao25: This one is DPG.

test/TestServerProjects/media_types/Generated/MediaTypesClient.cs
```diff
@@ -265,7 +265,7 @@ public virtual Response<string> BinaryBodyWithTwoContentTypes(ContentType1 conte
         /// <param name="contentType"> Upload file type. </param>
         /// <param name="message"> The payload body. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual async Task<Response<string>> BinaryBodyWithThreeContentTypesAsync(ContentType1 contentType, Stream message, CancellationToken cancellationToken = default)
+        public virtual async Task<Response<string>> BinaryBodyWithThreeContentTypesAsync(ContentType1AutoGenerated contentType, Stream message, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
pshao25: This comes from M4 change. 
image
@timotheeguerin do you have any ideas of this change.

test/TestServerProjects/media_types/Generated/MediaTypesRestClient.cs
```diff
@@ -461,33 +461,34 @@ public Response<string> BinaryBodyWithThreeContentTypes(ContentType1 contentType
...
-        internal HttpMessage CreateBinaryBodyWithThreeContentTypesRequest(string message)
+        internal HttpMessage CreatePutTextAndJsonBodyRequest(ContentType1AutoGenerated2 contentType, string message)
         {
             var message0 = _pipeline.CreateMessage();
             var request = message0.Request;
             request.Method = RequestMethod.Post;
             var uri = new RawRequestUriBuilder();
             uri.Reset(_endpoint);
-            uri.AppendPath("/mediatypes/binaryBodyThreeContentTypes", false);
+            uri.AppendPath("/mediatypes/textAndJson", false);
```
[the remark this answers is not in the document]
pshao25: No, this is not a path change.
Previously, we have two `binaryBodyThreeContentTypes`.
Now, after M4 change and enable `textAndJson` scenario, we have one `binaryBodyThreeContentTypes` and one `textAndJson`.
It just looks like we change the path. But it is actually removing one and adding another.

test/TestServerProjectsLowLevel/media_types/Generated/MediaTypesClient.cs
pshao25: No. It is because:
Previously, we have two methods: json and others.
Now after the M4 change, we only have one method: containing json and others.
So the change here is actually: remove the json one and add the json type to the remaining one.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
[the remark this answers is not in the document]
annelo-msft: You are correct, thanks!

test/TestServerProjects/media_types/Generated/MediaTypesRestClient.cs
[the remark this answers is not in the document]
annelo-msft: Would you please provide a link to where the swagger changed?

samples/CognitiveServices.TextAnalytics/Generated/CognitiveServicesTextAnalyticsClient.cs
annelo-msft: I see, so this is new m4 functionality that was added separately from the contentType changes?

pshao25: Yes, I think so.

test/TestServerProjects/media_types/Generated/MediaTypesRestClient.cs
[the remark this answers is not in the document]
pshao25: No, swagger does not change.
But in readme.md, we enabled the disabled operation.
image
In M4, this url segment (operation) `binaryBodyThreeContentTypes` put all the content types into one request. https://github.com/Azure/autorest.csharp/blob/406c1daf2faa5f0576604145c2f07a9efd9b1c18/test/TestServerProjects/media_types/Generated/CodeModel.yaml#L765-L768
So, currently, we have one method for `binaryBodyThreeContentTypes` and one method for `textAndJson`.

Previously, the M4 has two requests for `binaryBodyThreeContentTypes`, one is for https://github.com/Azure/autorest.csharp/blob/2688d3d9ffced52ef328328eb6c955826d252ce5/test/TestServerProjects/media_types/Generated/CodeModel.yaml#L711-L713
and other is for https://github.com/Azure/autorest.csharp/blob/2688d3d9ffced52ef328328eb6c955826d252ce5/test/TestServerProjects/media_types/Generated/CodeModel.yaml#L767-L768
So, previously, we have two methods for `binaryBodyThreeContentTypes`.

In summary, this change is just removing one and adding another.

test/TestServerProjectsLowLevel/dpg_update1/Generated/ParamsClient.cs
pshao25: I think we should not validate content as https://azure.github.io/azure-sdk/dotnet_introduction.html#parameter-validation. Created issue here: https://github.com/Azure/autorest.csharp/issues/2155

test/TestProjects/NameConflicts/Generated/AutoRestParameterFlatteningRestClient.cs
pshao25: @annelo-msft I tried this scenario, if we have type 
- application/json
- text/plain
Then the content will be `StringRequestContent`, which is only for `text/plain`.
```C#
internal Azure.Core.HttpMessage CreateAnalyzeBodyRequest(Models.ContentType contentType, string stringBody)
{
    var message = _pipeline.CreateMessage();
    var request = message.Request;
    request.Method = RequestMethod.Post;
    var uri = new RawRequestUriBuilder();
    uri.Reset(_endpoint);
    uri.AppendPath("/conflictingLROOverloads", false);
    request.Uri = uri;
    if (stringBody != null)
    {
        request.Headers.Add("Content-Type", contentType.ToSerialString());
        request.Content = new StringRequestContent(stringBody);
    }
    return message;
}
```
We might need to change the implementation to 
```C#
internal Azure.Core.HttpMessage CreateAnalyzeBodyRequest(Models.ContentType contentType, string stringBody)
{
    var message = _pipeline.CreateMessage();
    var request = message.Request;
    request.Method = RequestMethod.Post;
    var uri = new RawRequestUriBuilder();
    uri.Reset(_endpoint);
    uri.AppendPath("/conflictingLROOverloads", false);
    request.Uri = uri;
    if (stringBody != null)
    {
        if (contentType == Models.ContentType.TextPlain)
        {
            request.Headers.Add("Content-Type", contentType.ToSerialString());
            request.Content = new StringRequestContent(stringBody);
        }
        else if (contentType == Models.ContentType.ApplicationJson)
        {
            var content = new Utf8JsonRequestContent();
            content.JsonWriter.WriteStringValue(stringBody);
            request.Content = content;
        }
    }
    return message;
}
```

pshao25:
> The changes to dpg-update1 look good to me, but I don't understand why there are so many API changes to other clients when the API specs haven't changed. Are these intended?

I summarize the common pattern of API changes here:
**1. Newly-added client constructor, lists are:** samples/CognitiveServices.TextAnalytics/Generated/CognitiveServicesTextAnalyticsClient.cs
samples/Azure.Network.Management.Interface/Generated/*Client.cs
**Root cause:** M4 now supports auth type: oauth2 and apiKey which does not support before, so the new constructures come up. It is by design to me.

**2. Mode name suffixed by "AutoGenerated", lists are:**
test/TestServerProjects/media_types/Generated/MediaTypesRestClient.cs	
**Root cause:** we generate content type with name "ContentType". When there are more than one content type enum, its name will be suffixed by "AutoGenerated", like https://github.com/Azure/autorest.csharp/pull/2072#discussion_r845839482. 
We need to fix from M4 side if we want.

**3. Type Utf8JsonRequestContent changes to StringRequestContent, lists are:**
https://github.com/Azure/autorest.csharp/pull/2072#discussion_r829279491
This issue could be fixed by update M4 with Tim's fix. But it raises new problem as Tim mentioned at https://github.com/Azure/autorest.csharp/pull/2072#discussion_r845548226 and I put samples below his comments.

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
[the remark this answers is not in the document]
ShivangiReja: @pshao25 Now why do we need `case AzureKeySecurityScheme` and `case AADTokenSecurityScheme` here?

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -874,6 +915,14 @@ public void AddRequestContext()
                 _parameters.Add(KnownParameters.RequestContext);
             }
 
+            private void AddRequestParameterWithMediaTypes(RequestParameter requestParameter, ICollection<string> requestMediaTypes)
+            {
+                var parameter = _parent.BuildContentTypeParameterWithMediaTypes(requestParameter, requestMediaTypes);
```
AlexanderSher, line 920: `BuildContentTypeParameterWithMediaTypes` can be declared in `RequestMethodParametersBuilder`

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -874,6 +915,14 @@ public void AddRequestContext()
...
+                var reference = _parent.CreateReference(requestParameter, parameter, true);
```
AlexanderSher, line 921: Isn't it enough to just cast `(ReferenceOrConstant)parameter` here?

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -613,6 +613,22 @@ private static bool IsRequestConditionHeader(RequestParameter requestParameter,
...
+        {
+            CSharpType type = new CSharpType(typeof(ContentType), requestParameter.IsNullable || !requestParameter.IsRequired);
+
+            return new Parameter(
+                requestParameter.CSharpName(),
+                CreateDescriptionWithMediaTypes(requestParameter, requestMediaTypes),
+                TypeFactory.GetInputType(type),
+                null,
+                requestParameter.IsRequired,
+                IsApiVersionParameter: requestParameter.Origin == "modelerfour:synthesized/api-version",
```
AlexanderSher, line 626: ContentType parameter can't be an `api-version` parameter.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -672,6 +688,24 @@ private static string CreateDescription(OperationGroup operationGroup, string cl
...
 
+        protected static string CreateDescriptionWithMediaTypes(RequestParameter requestParameter, ICollection<string> requestMediaTypes)
+        {
+            var description = string.IsNullOrWhiteSpace(requestParameter.Language.Default.Description) ?
+                $"The {requestParameter.Schema.Name} to use." :
+                BuilderHelpers.EscapeXmlDescription(requestParameter.Language.Default.Description);
+
+            return AddAllowedValues(description, requestMediaTypes);
+
+            static string AddAllowedValues(string description, ICollection<string> choices)
```
AlexanderSher, line 699: What's the point of the local function here? You can simply write:
```cs
var allowedMediaTypes = string.Join(" | ", requestMediaTypes.Select(v => $"\"{v}\""));
return $"{description}{(description.EndsWith(".") ? "" : ".")} Allowed values: {BuilderHelpers.EscapeXmlDescription(allowedValues)}";
```

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
AlexanderSher: Even better, you can just inline it here.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
AlexanderSher: But I wonder why the following code doesn't work:
```cs
var parameter = _parent.BuildParameter(requestParameter, typeof(ContentType)) with { Description = CreateDescriptionWithMediaTypes(requestParameter, requestMediaTypes)};
AddRequestParameter(name, requestParameter, parameter);
```

src/AutoRest.CSharp/DataPlane/Generation/DataPlaneClientWriter.cs
```diff
@@ -142,7 +142,7 @@ private void WriteClientCtors(CodeWriter writer, DataPlaneClient client, BuildCo
             var clientOptionsName = ClientBuilder.GetClientPrefix(context.DefaultLibraryName, context);
             foreach (var scheme in schemes)
             {
-                if (scheme is AzureKeySecurityScheme azureKeySecurityScheme)
+                if (scheme is AzureKeySecurityScheme || scheme is KeySecurityScheme)
```
AlexanderSher, line 145: I think we should throw exception on `AzureKeySecurityScheme` and `AADTokenSecurityScheme` cause we don't expect old types to be produced by M4.

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
```diff
@@ -164,6 +164,11 @@ public static void SetRequestsToClients(IEnumerable<ClientInfo> clientInfos)
                     foreach (var request in operation.Requests)
                     {
                         SetRequestToClient(clientInfo, request, operation);
+
+                        if (operation.Requests.All(r => operation?.RequestMediaTypes?.Values.Contains(r) == true))
```
AlexanderSher, line 168: This condition doesn't depend on `request` variable, why is it inside the loop?

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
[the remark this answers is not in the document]
pshao25: Which one do you mean? I have removed the `AzureKeySecurityScheme `and `AADTokenSecurityScheme`, only leaving `KeySecurityScheme` and `OAuth2SecurityScheme` .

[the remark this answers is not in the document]
ShivangiReja: Yup, that's what I meant.

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
pshao25: Because I want to stop processing following requests if this condition is satisfied.

AlexanderSher: But its result is the same every time this expression is executed, so it can be checked before foreach.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -874,6 +909,24 @@ public void AddRequestContext()
                 _parameters.Add(KnownParameters.RequestContext);
             }
 
+            private void AddRequestParameterWithMediaTypes(RequestParameter requestParameter, ICollection<string> requestMediaTypes)
```
AlexanderSher, line 912: Let's rename this to `AddContentTypeRequestParameter`

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -874,6 +909,24 @@ public void AddRequestContext()
...
+            {
+                CSharpType type = new CSharpType(typeof(ContentType), requestParameter.IsNullable || !requestParameter.IsRequired);
+                var parameter = new Parameter(
+                    requestParameter.CSharpName(),
+                    CreateDescriptionWithMediaTypes(requestParameter, requestMediaTypes),
+                    TypeFactory.GetInputType(type),
+                    null,
+                    requestParameter.IsRequired,
```
AlexanderSher, line 920: @annelo-msft, I think we should always validate this parameter, so we need to pass `true` here.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -874,6 +909,24 @@ public void AddRequestContext()
...
+                    IsApiVersionParameter: false,
+                    IsResourceIdentifier: requestParameter.IsResourceParameter,
+                    SkipUrlEncoding: requestParameter.Extensions?.SkipEncoding ?? false,
+                    RequestLocation: GetRequestLocation(requestParameter));
```
AlexanderSher, line 924: `RequestLocation` here is always `header`

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -874,6 +909,24 @@ public void AddRequestContext()
...
+                CSharpType type = new CSharpType(typeof(ContentType), requestParameter.IsNullable || !requestParameter.IsRequired);
+                var parameter = new Parameter(
+                    requestParameter.CSharpName(),
+                    CreateDescriptionWithMediaTypes(requestParameter, requestMediaTypes),
+                    TypeFactory.GetInputType(type),
+                    null,
+                    requestParameter.IsRequired,
+                    IsApiVersionParameter: false,
+                    IsResourceIdentifier: requestParameter.IsResourceParameter,
+                    SkipUrlEncoding: requestParameter.Extensions?.SkipEncoding ?? false,
```
AlexanderSher, line 923: We can use default values for `IsResourceIdentifier` and `SkipUrlEncoding`, so whole constructor call should look like this:
```cs
var name = requestParameter.CSharpName();
var description = CreateDescriptionWithMediaTypes(requestParameter, requestMediaTypes);
var parameter = new Parameter(name, description, typeof(ContentType), null, true, RequestLocation: RequestLocation.Body);
```

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
AlexanderSher: In fact, we can assume that if `RequestMediaTypes` is not `null`, all requests are added to this dictionary: https://github.com/Azure/autorest/blob/61aff363307502ff422f51e74b3e2edeaa2ff0b3/packages/extensions/modelerfour/src/modeler/modelerfour.ts#L2314

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -613,6 +613,22 @@ private static bool IsRequestConditionHeader(RequestParameter requestParameter,
             return requestParameter.In == HttpParameterIn.Header && ConditionRequestHeader.TryGetValue(GetRequestParameterName(requestParameter), out header);
         }
 
+        private Parameter BuildContentTypeParameterWithMediaTypes(RequestParameter requestParameter, ICollection<string> requestMediaTypes)
```
AlexanderSher, line 616: I think this method isn't used anymore.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
annelo-msft: We can always remove validation without a breaking change, so let's go ahead with this and we'll deal with it later if it's a problem.

annelo-msft: If it's not marked required, it doesn't seem like we should validate it.

Why is this important to us, @AlexanderSher ?

pshao25: @AlexanderSher we will not validate `ContentType`, because it is a structure (value type).

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
pshao25: Is there any possibility that `requestParameter.IsRequired` is false to `ContentType`?

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -874,6 +893,16 @@ public void AddRequestContext()
                 _parameters.Add(KnownParameters.RequestContext);
             }
 
+            private void AddContentTypeRequestParameter(RequestParameter requestParameter, ICollection<string> requestMediaTypes)
+            {
+                var name = requestParameter.CSharpName();
+                var description = CreateDescriptionWithMediaTypes(requestParameter, requestMediaTypes);
```
AlexanderSher, line 899: Let's move `CreateDescriptionWithMediaTypes` next to `AddContentTypeRequestParameter`, make it private and rename into `CreateContentTypeRequestParameterDescription`

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
AlexanderSher: It doesn't really matter. We need to specify content type in our protocol methods. And since `ContentType` is a struct, we don't have to validate this parameter for null.

The question, however, is should we validate specific values here. Should we check if user specifies content type that is not supported by the service.

tg-msft: > Should we check if user specifies content type that is not supported by the service.

No, it'd be better to let the service validate.  Ideally they'd send back a clear/obvious 415, but most of Azure is JSON so it I don't expect serious discoverability problems here.

AlexanderSher: Then we don't need any additional validation here at all.

test/TestServerProjects/media_types/Generated/MediaTypesClient.cs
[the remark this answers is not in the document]
annelo-msft: Could we preview the changes this generator update will have in the azure-sdk-for-net repo before we merge this change?  I'm interested to know whether it will make any unexpected API changes there.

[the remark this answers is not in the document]
pshao25: @annelo-msft I put at PR https://github.com/Azure/azure-sdk-for-net/pull/28363. I reviewed all the changes and all look good to me. I have comments to explain the changes in PR. Please check out.

AlexanderSher (approved):

PR #2077, opened 2022-03-17
Merge extension classes

ArcturusZhang:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Output/MgmtTypeProvider.cs
```diff
@@ -34,7 +34,7 @@ protected MgmtTypeProvider(string resourceName) : base(MgmtContext.Context)
         {
             ResourceName = resourceName;
             IsArmCore = Configuration.MgmtConfiguration.IsArmCore;
-            IsStatic = !IsArmCore && BaseType is null && this is MgmtExtensions extension && extension.ArmCoreType != typeof(ArmResource) && extension.ArmCoreType != typeof(ArmClient);
```
ArcturusZhang, line 37: I have to change this because this line is inside the constructor of the base class of `MgmtExtensions`, and the property `ArmCoreType` is initialized in the constructor of `MgmtExtensions`, therefore here the property `ArmCoreType` will always be null - these two conditions will always be false.

m-nash (approved):
Looks good other than adding the test case we discussed to verify each method on the private client has a public extension calling it.

Issue #2078, opened 2022-03-17
Add support for "protocol-method-list" config to take full operationId including operationGroup_operationId


ShivangiReja:
"protocol-method-list" config only takes operationId list(method names) and generates all the protocol methods with the names mentioned in "protocol-method-list".

Add support for "protocol-method-list" config to take full operationId including operationGroup_operationId. If there is no operationGroup present, then use the default(title name) which is used to generate default RestClientName.

PR #2082, opened 2022-03-18
Add support for "protocol-method-list" config to take full operationId including operationGroup_operationId

ShivangiReja:
Today "protocol-method-list" config only takes operationId list(method names) and generates all the protocol methods with the names mentioned in "protocol-method-list" regardless of what operationGroup is.

This PR adds support for "protocol-method-list" config to take full operationId based on swagger including operationGroup_operationId. If there is no operationGroup present, then we use the default(title name) which is used to generate default RestClientName.

With these changes we can add config like
```yaml
protocol-method-list:
  - Create
  - Delete
  - firstTemplate_Create
  - firstTemplate_Get
  - secondTemplate_Get
```

Fixes: https://github.com/Azure/autorest.csharp/issues/2078


annelo-msft (approved):
Generated output LGTM.  I defer to @AlexanderSher on implementation.

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
```diff
@@ -207,17 +211,70 @@ public DataPlaneRestClient FindRestClient(OperationGroup operationGroup)
...
+                    // when operationGroup is not present use the default name which is used to create client name
+                    var defaultClientName = ClientBuilder.GetClientPrefix(string.Empty, _context);
+                    AddToProtocolMethodsDictionary(protocolMethodsDictionary, defaultClientName, operationId);
+                }
+            }
+
+            return protocolMethodsDictionary;
+        }
+
+        private void AddToProtocolMethodsDictionary(Dictionary<string, List<string>> protocolMethodsDictionary, string operationGroupKey, string methodName)
```
AlexanderSher, line 241: Can this be static?

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
```diff
@@ -207,17 +211,70 @@ public DataPlaneRestClient FindRestClient(OperationGroup operationGroup)
...
+                methodList.Add(methodName);
+            }
+        }
+
+        private IEnumerable<string>? GetProtocolMethodsByOperationGroup(OperationGroup operationGroup)
+        {
+            _protocolMethodsDictionary.TryGetValue(operationGroup.Key, out var methodList);
+            if (methodList == null)
+            {
+                var defaultClientName = ClientBuilder.GetClientPrefix(string.Empty, _context);
```
AlexanderSher, line 261: Why are we doing this fallback?

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
ShivangiReja: Using the private `_context` in [here](https://github.com/Azure/autorest.csharp/blob/66398eee793fc828588cd56079af2c42abcb5704/src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs#L233).

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
ShivangiReja: If we pass only operationId without operationGroup in `protocol-method-list` like this case:
```
protocol-method-list:
  - Create
  ```
then we use default name by calling [ClientBuilder.GetClientPrefix()](https://github.com/Azure/autorest.csharp/blob/66398eee793fc828588cd56079af2c42abcb5704/src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs#L233) method for operationGroup key. I used the same logic as create client name for the methods where there is no operation group present with the operationId.

This way I get the correct client name where the method belongs.

File: src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs [hunk 3]
```diff
@@ -207,17 +211,70 @@ public DataPlaneRestClient FindRestClient(OperationGroup operationGroup)
             return restClients;
         }
 
-        private static IEnumerable<LowLevelClientMethod> GetProtocolMethods(OperationGroup operationGroup, RestClientBuilder restClientBuilder, BuildContext<DataPlaneOutputLibrary> context)
+        // Get a Dictionary<operationGroupName, List<methodNames>> based on the "protocol-method-list" config
+        private Dictionary<string, List<string>> GetProtocolMethodsDictionary()
```
AlexanderSher, line 215: Does it fail somehow if we have a naming conflict?

ShivangiReja: Could you please give me an example for this?

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
AlexanderSher: But that is a different method. Also, `ClientBuilder.GetClientPrefix(string.Empty, _context)` can be acquired once for the whole method.

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
AlexanderSher: Something like

```
protocol-method-list:
  - Create
  - Update
  - Create
```

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
AlexanderSher: Correct me if I'm wrong, but wouldn't that mean that `Create` will be used in every client?

ShivangiReja: No, Create method will not be generated in every client if we pass only `Create` in the config. Create method will be generated in the default RestClient.

Methods without operation groups will be generated in the default RestClient(in case when we pass only operationId in the config ex. `Create`) and methods with operation group will be generated in the configured operation group RestClient(in case when we pass operationGroup_operationId in the config ex. `firstTemplate_Create`, this will generate method in FirstTemplateRestClient)

AlexanderSher: Can you add a test for it? E.g., add `secondTemplate_Create` operation to `ProtocolMethodsInRestClient.json` and validate that protocol method isn't generated for it.

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
ShivangiReja: Updated! 

Used static for `AddToProtocolMethodsDictionary` method and made `_defaultClientName` private class field.

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
ShivangiReja: Already added all the possible scenarios in the test.

[Here](https://github.com/ShivangiReja/autorest.csharp/blob/protocol-method-list/test/TestProjects/ProtocolMethodsInRestClient/readme.md) is config, [here](https://github.com/ShivangiReja/autorest.csharp/tree/protocol-method-list/test/TestProjects/ProtocolMethodsInRestClient/Generated) is generated code and [here](https://github.com/ShivangiReja/autorest.csharp/blob/protocol-method-list/test/AutoRest.TestServer.Tests/ProtocolMethodsInRestClientTests.cs#L14-L56) are tests that validates where the methods are generated.

As an example you can see that `Create` method is not generated as protocol method in SecondTemplateRestClient.

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs
ShivangiReja: In this case generator will not fail and it will generate `Create` method only once in the default RestClient because in order to make `IEnumerable<LowLevelClientMethod>`, [here](https://github.com/ShivangiReja/autorest.csharp/blob/protocol-method-list/src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneOutputLibrary.cs#L277) we are iterating all the operations only once and checking that if the operation name exists in the config or not.

AlexanderSher (approved):

Issue #2063, opened 2022-03-14
Eliminate duplicate names that span across different packages including between data plane and mgmt. plane - For GA


ArthurMa1978:

Yao725:
https://microsoftapc-my.sharepoint.com/:x:/g/personal/yaokou_microsoft_com/ERM4rgb_JP9Glhbl6eft0v0BNKAjPouynFDwA9MK6ETg0Q?e=51mEX1

ArthurMa1978:
https://github.com/Azure/azure-sdk-for-net/pull/27599

ArthurMa1978:
https://github.com/Azure/azure-sdk-for-net/pull/27731

PR #2108, opened 2022-03-25
feat(mgmt): append namespace to some private endpoint resource names

archerzz:
To avoid naming conflicts of some common private endpoint resources across different RPs, we've decided to append namespace to those resource names.
This commit will:
1. add those common private endpoint resources into the collection of which each resource name will be pretended by its namespace
  1.1 The selected names are from the intersection of a) swagger definitions in `common-types/privatelinks.json`; b) conflicted definitions in each RP
3. regenerate samples and test projects
4. update tests

part of https://github.com/Azure/autorest.csharp/issues/2063

# Description

# Checklist

# Ready to Land?


archerzz:
Corresponding name changes in `azure-sdk-for-net`: https://github.com/Azure/azure-sdk-for-net/pull/27754

samples/Azure.Management.Storage/Generated/Models/PrivateEndpointConnectionListResult.Serialization.cs
```diff
@@ -16,7 +16,7 @@ internal partial class PrivateEndpointConnectionListResult
     {
         internal static PrivateEndpointConnectionListResult DeserializePrivateEndpointConnectionListResult(JsonElement element)
         {
-            Optional<IReadOnlyList<PrivateEndpointConnectionData>> value = default;
```
m-nash, line 19: Can we add *ListResult as well?  I know its internal now but can be public later?

archerzz: I've added `PrivateEndpointConnectionListResult` and `PrivateLinkResourceListResult` into the list. I think even if they are not public later, it's good for consistency.

m-nash (approved):
Try to find a way to know that we have prepended the RP name and in those cases keep the method name the shorter version.

i.e. GetPrivateLinkResource vs GetStoragePrivateLinkResource

archerzz:
> Try to find a way to know that we have prepended the RP name and in those cases keep the method name the shorter version.
> 
> i.e. GetPrivateLinkResource vs GetStoragePrivateLinkResource

The reason we have the following generated codes (e.g. method name is shorter than return type name) is because the corresponding swagger API is a special case.
> public virtual Pageable<[KeyVaultPrivateLinkResource](https://apiview.dev/Assemblies/Review/d0f33ca793e44ab6a5c87a81e817ef1c/013d0dd62a684acf8ee5a3a55edaada9?diffRevisionId=a616b1ac0c7a4a7dadc068d79b574d3a&doc=False&diffOnly=True#Azure.ResourceManager.KeyVault.Models.KeyVaultPrivateLinkResource)> [GetPrivateLinkResources](https://apiview.dev/Assemblies/Review/d0f33ca793e44ab6a5c87a81e817ef1c/013d0dd62a684acf8ee5a3a55edaada9?diffRevisionId=a616b1ac0c7a4a7dadc068d79b574d3a&doc=False&diffOnly=True#)(CancellationToken cancellationToken = default);

See the swagger definition [here](https://github.com/Azure/azure-rest-api-specs/blob/63a4b37b66740087b113e00c1b32b2b9c98fd9db/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2021-10-01/keyvault.json#L982).
```
"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}/privateLinkResources": {
      "get": {
        "tags": [
          "PrivateLinkResources"
        ],
        "operationId": "PrivateLinkResources_ListByVault",
```

That's a single list method under `Vault` resource. So, the code generation logic is different.
- `methodName` is determined [here](https://github.com/Azure/autorest.csharp/blob/430cdb47a88dcb5b4a8af23101cdb5a6a275c5d6/src/AutoRest.CSharp/Mgmt/Output/Resource.cs#L304).
- For this case, it will invoke [GetOperationName](https://github.com/Azure/autorest.csharp/blob/430cdb47a88dcb5b4a8af23101cdb5a6a275c5d6/src/AutoRest.CSharp/Mgmt/Output/MgmtTypeProvider.cs#L207), since it's not a list method.
- which will finally fall into [CalculateOperationName](https://github.com/Azure/autorest.csharp/blob/430cdb47a88dcb5b4a8af23101cdb5a6a275c5d6/src/AutoRest.CSharp/Mgmt/Output/MgmtTypeProvider.cs#L223).
- more specifically, the `methodName` will come from `OperationGroup`. See here: https://github.com/Azure/autorest.csharp/blob/430cdb47a88dcb5b4a8af23101cdb5a6a275c5d6/src/AutoRest.CSharp/Mgmt/Output/MgmtTypeProvider.cs#L238

### Conclusion

The shorter name is from a special case. Generally the return type name and the method name are from the same resource name.

PR #2123, opened 2022-03-29
Add scope checks for PageableHelpers tests

AlexanderSher:


src/assets/TestFramework/ClientDiagnosticListener.cs
```diff
@@ -0,0 +1,264 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+
+namespace Azure.Core.Tests
+{
+    public class ClientDiagnosticListener : IObserver<KeyValuePair<string, object>>, IObserver<DiagnosticListener>, IDisposable
```
annelo-msft, line 13: @christothes, @JoshLove-msft, do you know whether we have anything like this in the test framework already?

AlexanderSher: yes, that is actually from the Azure.Core. Let me update the copy script

annelo-msft: Oh awesome, thanks, @AlexanderSher!

ShivangiReja (approved):

Issue #2106, opened 2022-03-24
[DPG] [Design Issue]: Determine how to handle global parameters in DPG Client constructors


annelo-msft:
### Problem

We have a test case in autorest.csharp, where the client constructors are not generated following the .NET guidelines.  This has been described in this issue: https://github.com/azure/autorest.csharp/issues/2055

We need to determine how best to handle global parameters in generated client constructors.  Since there are no [Data Plane Generated libraries](https://grep.app/search?q=data-plane%3A&filter%5Brepo%5D%5B0%5D=Azure/azure-sdk-for-net) that currently have this case in their OpenAPI specs, we will not block DPG GA on this.

### Success Criteria

Redmond team provides a spec indicating how to handle global parameters in generated client constructors.

Once we have determined the design we want for the generated client constructors on clients that have global parameters, we will complete work on #2055.

Note: We have had a number of discussions about this offline, and as time allows, that information will be added to this issue so that it contains the full context.

pshao25:
For optional global parameter, we have decided https://github.com/Azure/autorest.csharp/issues/3706#issuecomment-1752374728

Issue #2055, opened 2022-03-11
Client constructor does not follow guideline


pshao25:
Pick any client generated by DPG, e.g., [DGPClient](https://github.com/pshao25/autorest.csharp/blob/9aec4ef9696e3be7f004bc3305a2cef872c62225/test/TestServerProjectsLowLevel/dpg-customization/Generated/DPGClient.cs#L43). It looks like:
```C#
public DPGClient(AzureKeyCredential credential, Uri endpoint = null, DPGClientOptions options = null);
```
According to [the guidelines for client constructors](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-client-constructor-minimal), it should be
```C#
public DPGClient(AzureKeyCredential credential);
public DPGClient(AzureKeyCredential credential, Uri endpoint = null, DPGClientOptions options = null);
```

Same issue in HLC client if it has public constructor, e.g., [PublicClientCtorClient](https://github.com/pshao25/autorest.csharp/blob/e7dfd8b6c5d12803f40c3652c2ac940a91105f22/test/TestProjects/PublicClientCtor/Generated/PublicClientCtorClient.cs#L36) as:
```C#
public PublicClientCtorClient(Uri endpoint, AzureKeyCredential credential, string param1 = "value1", string param2 = null, PublicClientCtorClientOptions options = null)
public PublicClientCtorClient(Uri endpoint, TokenCredential credential, string param1 = "value1", string param2 = null, PublicClientCtorClientOptions options = null)
```

annelo-msft:
@pshao25, thank you for filing this issue!

We should definitely update this in the DPG Generator.  We will want to add the "minimal constructor" to DPG libraries, and when we do, we'll need to make the ClientOptions parameter non-optional to avoid an ambiguous call site.

@ShivangiReja, is the HLC PublicClientCtorClient case @pshao25 has mentioned above something that we use in HLC's today?

ShivangiReja:
>@ShivangiReja, is the HLC PublicClientCtorClient case @pshao25 has mentioned above something that we use in HLC's today?

[...]

PR #2128, opened 2022-03-30
Add minimal client constructor to meet guidelines

### LINKED ISSUE #2105: [DPG] [BUG] Azure PurviewCatalogClient constructor and other DPG Client Constructors are generated incorrectly
### Problem

In the [Azure.Analytics.Purview.Catalog API](https://apiview.dev/Assemblies/Review/bdfd1e265f30435b8bd38e982710ad15), the constructor is generated as follows:

```csharp
public PurviewCatalogClient(Uri endpoint, TokenCredential credential, PurviewCatalogClientOptions options = null);
```

This is case for all the [DPG Clients](https://grep.app/search?q=data-plane%3A&filter[repo][0]=Azure/azure-sdk-for-net), except WebPubSubServiceClient.

### Success Criteria

All Data Plane Generated libraries have client constructors that follow [the guidelines](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-client-ctor), specifically:

```csharp
public PurviewCatalogClient(Uri endpoint, TokenCredential credential);
public PurviewCatalogClient(Uri endpoint, TokenCredential credential, PurviewCatalogClientOptions options);
```

pshao25:
Fix #2105


samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -35,16 +35,22 @@ protected DocumentTranslationClient()
         {
         }
 
+        /// <summary> Initializes a new instance of DocumentTranslationClient. </summary>
+        /// <param name="endpoint"> Supported Cognitive Services endpoints (protocol and hostname, for example: https://westus.api.cognitive.microsoft.com). </param>
+        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="endpoint"/> or <paramref name="credential"/> is null. </exception>
+        public DocumentTranslationClient(string endpoint, AzureKeyCredential credential) : this(endpoint, credential, new DocumentTranslationClientOptions())
+        {
+        }
```
annelo-msft, line 44: Could we add a carriage return between methods, or would this be too big a change?

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -35,16 +35,22 @@ protected DocumentTranslationClient()
...
         /// <param name="endpoint"> Supported Cognitive Services endpoints (protocol and hostname, for example: https://westus.api.cognitive.microsoft.com). </param>
         /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
         /// <param name="options"> The options for configuring the client. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="endpoint"/> or <paramref name="credential"/> is null. </exception>
-        public DocumentTranslationClient(string endpoint, AzureKeyCredential credential, DocumentTranslationClientOptions options = null)
+        public DocumentTranslationClient(string endpoint, AzureKeyCredential credential, DocumentTranslationClientOptions options)
         {
             Argument.AssertNotNull(endpoint, nameof(endpoint));
             Argument.AssertNotNull(credential, nameof(credential));
-            options ??= new DocumentTranslationClientOptions();
```
annelo-msft, line 47: Why is this line removed?  It seems possible that someone could pass null to the constructor and we would want to create an options instance.

test/TestProjects/Accessibility-LowLevel-NoAuth/Generated/AccessibilityClient.cs
```diff
@@ -25,17 +25,17 @@ public partial class AccessibilityClient
         /// <summary> The HTTP pipeline for sending and receiving REST requests and responses. </summary>
         public virtual HttpPipeline Pipeline => _pipeline;
 
-        /// <summary> Initializes a new instance of AccessibilityClient for mocking. </summary>
-        protected AccessibilityClient()
```
annelo-msft, line 29: Why is the protected constructor for mocking removed?

test/TestProjects/HeadAsBooleanTrue-LowLevel/Generated/HttpSuccessClient.cs
```diff
@@ -32,15 +32,21 @@ protected HttpSuccessClient()
...
+        {
+        }
         /// <summary> Initializes a new instance of HttpSuccessClient. </summary>
         /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
         /// <param name="endpoint"> server parameter. </param>
         /// <param name="options"> The options for configuring the client. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="credential"/> is null. </exception>
-        public HttpSuccessClient(AzureKeyCredential credential, Uri endpoint = null, HttpSuccessClientOptions options = null)
+        /// <exception cref="ArgumentNullException"> <paramref name="credential"/> or <paramref name="endpoint"/> is null. </exception>
+        public HttpSuccessClient(AzureKeyCredential credential, Uri endpoint, HttpSuccessClientOptions options = null)
```
annelo-msft, line 46: Curious: why provide a default value for `options` here?

annelo-msft (commented):
Left some questions.  We will need @AlexanderSher and @ShivangiReja to approve this one as well.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
pshao25: This line is removed because `options` is required parameter now. If they want to pass null, they should call `public DocumentTranslationClient(string endpoint, AzureKeyCredential credential)`, right?

test/TestProjects/Accessibility-LowLevel-NoAuth/Generated/AccessibilityClient.cs
pshao25: This client constructor does not have required parameters. So we have to choose between 

image

test/TestProjects/HeadAsBooleanTrue-LowLevel/Generated/HttpSuccessClient.cs
pshao25: As we agreed in https://github.com/azure/autorest.csharp/issues/2055#issuecomment-1076868591

> If ClientOptions is not the only additional parameter for the overload, we should not make it required.

In this case, we have two optional parameters: `endpoint` and `options`. So I left `options` as optional.

test/TestProjects/Accessibility-LowLevel-NoAuth/Generated/AccessibilityClient.cs
annelo-msft: This is ok for a test case, but it would be a [violation of the guidelines](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-client-constructor-for-mocking) if it ever appeared in a public API.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
annelo-msft: This is incorrect functionality.  The client won't function correctly if `options` is null.  If a customer passes null for options, we should instantiate a default instance of the client options for the library.

test/TestProjects/HeadAsBooleanTrue-LowLevel/Generated/HttpSuccessClient.cs
annelo-msft: I see.  https://github.com/Azure/autorest.csharp/issues/2055#issuecomment-1076868591 was specifically regarding the case where we have global parameters in the swagger, and we [have not finalized the guidance](https://github.com/Azure/autorest.csharp/issues/2106) for that case.

But I understand now - this is because it is a test case and has an optional endpoint.

test/TestProjects/Accessibility-LowLevel-NoAuth/Generated/AccessibilityClient.cs
pshao25: So, we remove the public one?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -121,33 +122,96 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
                 yield break;
             }
 
-            var clientOptionsType = ClientOptions.Type.WithNullable(true);
-            var clientOptionsParameter = new Parameter("options", "The options for configuring the client.", clientOptionsType, Constant.NewInstanceOf(clientOptionsType), false);
-
             if (Fields.CredentialFields.Count == 0)
             {
-                yield return BuildPublicConstructor(null, clientOptionsParameter);
+                foreach (var constructor in BuildPublicConstructor(null))
```
AlexanderSher, line 140: We probably need to split constructors with bodies and without bodies into two separate properties and generate them separately: first, produce all constructors with bodies. Then, use them to generate constructors without bodies that either call other constructors or just protected constructor for mocking.

test/TestProjects/Accessibility-LowLevel-NoAuth/Generated/AccessibilityClient.cs
annelo-msft: It is ok for a test case and this is a test case, so no change is required.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: @AlexanderSher I though about this before to make MinimalPublicConstructors a new property. But the parameter list of these constructors are so related that I need to add more private fields for these parameters, otherwise I need to calculate the same parameter list every time I produce the constructors. Let me know your preference.

AlexanderSher: Common part of parameters can be determined in constructor before creation of constructors starts, e.g.:
```cs
if (!IsSubClient)
{
    var requiredParameters = parameters.Where(p => p.IsRequired).ToArray();
    var optionalParameters = parameters
        .Where(p => !p.IsRequired && !p.IsApiVersionParameter)
        .Append(CreateClientOptionsParameter())
        .ToArray();

    PrimaryConstructors = BuildPrimaryConstructors(requiredParameters, optionalParameters).ToArray();
    SecondaryConstructors = BuildSecondaryConstructors(requiredParameters, optionalParameters).ToArray();
}
else
{
    PrimaryConstructors = Array.Empty<ConstructorSignature>();
    SecondaryConstructors = new[]{ CreateMockingConstructor() };
}
```
where 
```cs
private Parameter CreateClientOptionsParameter()
{
    var clientOptionsType = ClientOptions.Type.WithNullable(true);
    return new Parameter("options", "The options for configuring the client.", clientOptionsType, Constant.NewInstanceOf(clientOptionsType), false);
}
```

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -121,33 +122,96 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
+        {
+            var initializer = constructorParameters.Select(parameter =>
+            {
+                if (parameter.DefaultValue == null)
+                {
+                    return FormattableStringFactory.Create("{0:I}", parameter.Name);
+                }
+
+                var defaultValue = parameter.DefaultValue.Value;
+                if (defaultValue.IsNewInstanceSentinel)
```
AlexanderSher, line 167: Logic for constant writing is [more complicated](https://github.com/Azure/autorest.csharp/blob/eab7a29d375f7928066cacf19c8eb2e9f21f1dcf/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs#L387). We probably need to add helper method to [FormattableStringHelpers](https://github.com/Azure/autorest.csharp/blob/eab7a29d375f7928066cacf19c8eb2e9f21f1dcf/src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs), something like this:
```cs
public static FormattableString GetFormattable(this Constant constant)
{
    if (constant.Value == null)
    {
        // Cast helps the overload resolution
        return $"({constant.Type}){null:L}";
    }

    if (constant.IsNewInstanceSentinel)
    {
        return $"new {constant.Type}()";
    }

    if (!constant.Type.IsFrameworkType && constant.Value is EnumTypeValue enumTypeValue)
    {
        return $"{constant.Type}.{enumTypeValue.Declaration.Name}";
    }

    if (!constant.Type.IsFrameworkType && constant.Value is string enumValue)
    {
        return $"new {constant.Type}({enumValue:L})";
    }

    if (constant.Type.EqualsIgnoreNullable(typeof(DateTimeOffset)))
    {
        var d = (DateTimeOffset)constant.Value;
        d = d.ToUniversalTime();
        return $"new {typeof(DateTimeOffset)}({d.Year:L}, {d.Month:L}, {d.Day:L} ,{d.Hour:L}, {d.Minute:L}, {d.Second:L}, {d.Millisecond:L}, {typeof(TimeSpan)}.{nameof(TimeSpan.Zero)})";
    } 
    
    if (constant.Type.EqualsIgnoreNullable(typeof(byte[])))
    {
        var bytes = (byte[])constant.Value;
        return $"new byte[] {{{string.Join(", ", bytes)}}}";
    }
    
    if (constant.Type.Equals(typeof(ResourceType)))
    {
        return $"{((ResourceType)constant.Value).ToString():L}";
    }

    return $"{constant.Value:L}";
}
```

Then `WriteConstant` can be reduced to 
```cs
WriteConstant(this CodeWriter writer, Constant constant) => writer.Append(constant.GetFormattable());
```

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
AlexanderSher: With this implementation, `BuildPrimaryConstructors` and `BuildSecondaryConstructors` will be something like this:
```cs
private IEnumerable<ConstructorSignature> BuildPrimaryConstructors(IReadOnlyList<Parameter> requiredParameters, IReadOnlyList<Parameter> optionalParameters)
{
    var optionalAsRequired = optionalParameters.Select(p => p with {DefaultValue = null}).ToArray();
    if (Fields.CredentialParameters.Count == 0)
    {
        yield return CreatePrimaryConstructor(requiredParameters.Concat(optionalAsRequired).ToArray());
    }
    else
    {
        foreach (var credentialParameter in Fields.CredentialParameters)
        {
            yield return CreatePrimaryConstructor(requiredParameters.Append(credentialParameter).Concat(optionalAsRequired).ToArray());
        }
    }
}

private IEnumerable<ConstructorSignature> BuildSecondaryConstructors(IReadOnlyList<Parameter> requiredParameters, IReadOnlyList<Parameter> optionalParameters)
{
    if (requiredParameters.Any() || Fields.CredentialParameters.Any())
    {
        yield return CreateMockingConstructor();
    }

    var optionalParametersArguments = optionalParameters
        .Select(p => p.Type.EqualsIgnoreNullable(typeof(Uri)) ? $"new {typeof(Uri):I}({p.DefaultValue!.Value.GetFormattable()})" : p.DefaultValue!.Value.GetFormattable())
        .ToArray();

    if (Fields.CredentialParameters.Count == 0)
    {
        yield return CreateSecondaryConstructor(requiredParameters, optionalParametersArguments);
    }
    else
    {
        foreach (var credentialParameter in Fields.CredentialParameters)
        {
            yield return CreateSecondaryConstructor(requiredParameters.Append(credentialParameter).ToArray(), optionalParametersArguments);
        }
    }
}

private ConstructorSignature CreateMockingConstructor()
    => new(Declaration.Name, $"Initializes a new instance of {Declaration.Name} for mocking.", Protected, Array.Empty<Parameter>());

private ConstructorSignature CreatePrimaryConstructor(IReadOnlyList<Parameter> parameters)
    => new(Declaration.Name, $"Initializes a new instance of {Declaration.Name}", Public, parameters);

private ConstructorSignature CreateSecondaryConstructor(IReadOnlyList<Parameter> parameters, FormattableString[] optionalParametersArguments)
{
    var arguments = parameters
        .Select<Parameter, FormattableString>(p => $"{p.Name}")
        .Concat(optionalParametersArguments)
        .ToArray();
    return new(Declaration.Name, $"Initializes a new instance of {Declaration.Name}", Public, parameters, new ConstructorInitializer(false, arguments));
}
```

test/TestProjects/Accessibility-LowLevel-TokenAuth/Generated/AccessibilityClient.cs
```diff
@@ -32,15 +32,22 @@ protected AccessibilityClient()
...
+        }
+
         /// <summary> Initializes a new instance of AccessibilityClient. </summary>
         /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
         /// <param name="endpoint"> server parameter. </param>
         /// <param name="options"> The options for configuring the client. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="credential"/> is null. </exception>
-        public AccessibilityClient(TokenCredential credential, Uri endpoint = null, AccessibilityClientOptions options = null)
+        /// <exception cref="ArgumentNullException"> <paramref name="credential"/> or <paramref name="endpoint"/> is null. </exception>
+        public AccessibilityClient(TokenCredential credential, Uri endpoint, AccessibilityClientOptions options = null)
```
ShivangiReja, line 47: Why `options` is default in here? I see in some projects it is required.

pshao25: I followed this https://github.com/Azure/autorest.csharp/issues/2055#issuecomment-1076868591, but it seems still under discussion.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: Yes, logic for constant is complicated. But here we only have constructor parameter constant. I didn't see conditions other than https://github.com/pshao25/autorest.csharp/blob/7c4faea10402aa45490cd461e26233294e1775ee/src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs#L61
Do you have examples?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: This implementation does not follow https://github.com/Azure/autorest.csharp/issues/2055#issuecomment-1076868591. Currently I still fix by following this guidance until we have a conclusion about the outcome.

annelo-msft (commented):
This PR meets requirements for DPG GA.  Consider this my approval -- @AlexanderSher will give the official approval to unblock merging the PR.

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -58,6 +60,62 @@ public static FormattableString GetIdentifiersFormattable(this IEnumerable<strin
...
+            {
+                result = $"new byte[] {{";
+                var value = (byte[])constant.Value;
+                foreach (byte b in value)
+                {
+                    result = $"{result}{b}, ";
+                }
+
+                result = $"{result}}}";
+                return result;
```
AlexanderSher, line 104: Since we know the type of argument upfront (it is `byte`), we can create string upfront:

```cs
var bytes = (byte[])constant.Value;
var joinedBytes = string.Join(", ", bytes);
return $"new byte[] {{{joinedBytes}}}";
```

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -58,6 +60,62 @@ public static FormattableString GetIdentifiersFormattable(this IEnumerable<strin
...
+            }
+            else if (frameworkType == typeof(ResourceType))
+            {
+                var value = ((ResourceType)constant.Value).ToString();
```
AlexanderSher, line 108: We can use literal formatting here: `return $"{((ResourceType)constant.Value).ToString():L}";`

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -99,76 +99,85 @@ private static void WriteClientFields(CodeWriter writer, LowLevelClient client)
             writer.Line();
         }
 
+        private static bool IsEmptyConstructorExisted(LowLevelClient client)
+        {
+            foreach (var constructor in client.SecondaryConstructors.Concat(client.PrimaryConstructors))
```
AlexanderSher, line 104: You don't need to check `PrimaryConstructors` - all primary constructors have `ClientOptions` parameter

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -121,33 +125,97 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
         }
 
-        private IEnumerable<ConstructorSignature> BuildPublicConstructors()
+
+        private (ConstructorSignature[] PrimaryConstructors, ConstructorSignature[] SecondaryConstructors) BuildPublicConstructors()
         {
-            if (!_hasPublicConstructors)
+            if (!IsSubClient)
+            {
+                var requiredParameters = RestClientBuilder.GetRequiredParameters(Parameters).ToArray();
```
AlexanderSher, line 133: `Parameters` property is irrelevant and misleading at this point - we don't have any constructor that has set of parameters that match `Parameters` collection, so we need to remove it completely. To determine `requiredParameters` and `optionalParameters`, we need to use builder directly, e.g.:
```cs
var parameters = builder.GetOrderedParametersByRequired();
var requiredParameters = RestClientBuilder.GetRequiredParameters(parameters);
var optionalParameters = RestClientBuilder.GetRequiredParameters(parameters).Append(CreateOptionsParameter());
```

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -99,76 +99,85 @@ private static void WriteClientFields(CodeWriter writer, LowLevelClient client)
             writer.Line();
         }
 
+        private static bool IsEmptyConstructorExisted(LowLevelClient client)
```
AlexanderSher, line 102: method isn't used anywhere

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -121,33 +125,97 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
+
+        private ConstructorSignature CreateMockingConstructor()
+            => new(Declaration.Name, $"Initializes a new instance of {Declaration.Name} for mocking.", Protected, Array.Empty<Parameter>());
+
+        private Parameter CreateOptionsParameter()
+        {
+            var clientOptionsType = ClientOptions.Type.WithNullable(true);
+            return new Parameter("options", "The options for configuring the client.", clientOptionsType, Constant.NewInstanceOf(clientOptionsType), false)
+            {
+                forceInitializeValue = Constant.NewInstanceOf(clientOptionsType)
```
AlexanderSher, line 217: Initializer for the parameter should've been written by this line: https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs#L253

Why `parameter.DefaultValue != null && !TypeFactory.CanBeInitializedInline(parameter.Type, parameter.DefaultValue)` returns `false` for this parameter?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -99,76 +99,85 @@ private static void WriteClientFields(CodeWriter writer, LowLevelClient client)
...
+        }
+
         private static void WriteConstructors(CodeWriter writer, LowLevelClient client)
         {
-            WriteEmptyConstructor(writer, client);
-            foreach (var constructor in client.PublicConstructors)
+            foreach (var constructor in client.SecondaryConstructors)
             {
-                WritePublicConstructor(writer, client, constructor);
+                WritePublicConstructor(writer, client, constructor, false);
```
AlexanderSher, line 117: I think this should be enough to write secondary constructor:

```cs
writer.WriteMethodDocumentation(constructor);
using (writer.WriteMethodDeclaration(constructor))
{
}
writer.Line();
```

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -99,76 +99,85 @@ private static void WriteClientFields(CodeWriter writer, LowLevelClient client)
...
-                FormattableString perRetryPolicies = $"Array.Empty<{typeof(HttpPipelinePolicy)}>()";
+                    FormattableString perCallPolicies = $"Array.Empty<{typeof(HttpPipelinePolicy)}>()";
+                    FormattableString perRetryPolicies = $"Array.Empty<{typeof(HttpPipelinePolicy)}>()";
 
-                var credentialParameter = signature.Parameters.FirstOrDefault(p => p.Name == "credential");
-                if (credentialParameter != null)
-                {
-                    var credentialField = client.Fields.GetFieldByParameter(credentialParameter);
-                    if (credentialField != null)
+                    var credentialParameter = signature.Parameters.FirstOrDefault(p => p.Name == "credential");
```
AlexanderSher, line 147: I may be wrong, but I think it can be simplified:
```cs
foreach (var parameter in signature.Parameters)
{
    var field = client.Fields.GetFieldByParameter(parameter);
    if (field != null)
    {
        if (field.Type.Equals(typeof(AzureKeyCredential)))
        {
            writer.Line($"{field.Name:I} = {parameter.Name:I};");
            perRetryPolicies = $"new {typeof(HttpPipelinePolicy)}[] {{new {typeof(AzureKeyCredentialPolicy)}({field.Name:I}, {client.Fields.AuthorizationHeaderConstant!.Name})}}";
        }
        else if (field.Type.Equals(typeof(TokenCredential)))
        {
            writer.Line($"{field.Name:I} = {parameter.Name:I};");
            perRetryPolicies = $"new {typeof(HttpPipelinePolicy)}[] {{new {typeof(BearerTokenAuthenticationPolicy)}({field.Name:I}, {client.Fields.ScopesConstant!.Name})}}";
        }
        else if (parameter.IsApiVersionParameter)
        {
            writer.Line($"{field.Name:I} = {clientOptionsParameter.Name:I}.Version;");
        }
        else
        {
            writer.Line($"{field.Name:I} = {parameter.Name:I};");
        }
    }
}

writer.Line($"{client.Fields.PipelineField.Name:I} = {typeof(HttpPipelineBuilder)}.{nameof(HttpPipelineBuilder.Build)}({clientOptionsParameter.Name:I}, {perCallPolicies}, {perRetryPolicies}, new {typeof(ResponseClassifier)}());");
```

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: I don't want to calculate same parametes twice. How about adding an `orderedParameter` parameter to the method.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: I added this because options is optional because of this: https://github.com/Azure/autorest.csharp/pull/2128#discussion_r840860585

`options` is required so the condition returns `false`.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
pshao25: In this case, we only go through the signature parameters, but we still have client parameters. And they have overlap. E.g., clinet parameter might contains apiversion parameter, while signature not.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
AlexanderSher: Where do you need those for the second time?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
AlexanderSher: `options` isn't required - you set `DefaultValue` to `Constant.NewInstanceOf(clientOptionsType)`. Here is the test that passes on my machine:
```cs
[Test]
public void ValidateOptions()
{
    var clientOptionsType = new CSharpType(typeof(AccessibilityClientOptions)).WithNullable(true);
    var parameter = new Parameter("options", "The options for configuring the client.", clientOptionsType, Constant.NewInstanceOf(clientOptionsType), false);
    Assert.IsTrue(parameter.DefaultValue != null && !TypeFactory.CanBeInitializedInline(parameter.Type, parameter.DefaultValue));
}
```

If you need to avoid setting it to null in constructor declaration, you can set `UseDefaultValueInCtorParam: false` (name is misleading though, I need to update it).

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
AlexanderSher: Ok, let's leave `Parameters` as they are right now. We can clean it up after GA.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
AlexanderSher: Ok, let's leave it as it is for now

lirenhe:
@pshao25, could you resolve the conflicts from this PR?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: This is not going to work. Here "`options` is required" I mean we force it as required by 
```C#
var optionalToRequired = optionalParameters.Select(parameter => parameter with { DefaultValue = null }).ToArray();
```
when we `BuildPrimaryConstructors`. We lost default value information here. So we need any other places to store this information, otherwise we don't know what type we should use to initialize the `options`

pshao25:
@annelo-msft I put the SDK changes at https://github.com/Azure/azure-sdk-for-net/pull/28365. The problem is we need to manually change some customizations. E.g., `PurviewResourceSetRule` suppress his original constructor by https://github.com/pshao25/azure-sdk-for-net/blob/4cd85e6dd23569269ccb6db92b813e0a394ee69d/sdk/purview/Azure.Analytics.Purview.Administration/src/Customizations/PurviewResourceSetRule.cs#L11. Therefore, some generated [constructors ](https://github.com/pshao25/azure-sdk-for-net/blob/f38268874237893ee519360f88f29379cb2e1511/sdk/purview/Azure.Analytics.Purview.Administration/src/Generated/PurviewResourceSetRule.cs#L39)will fail compilation because the constructor they are calling is suppressed. We need to fix this in several places.

pshao25:
> > We need to fix this in several places.
> 
> Yes, I expected that as part of this change. Given this output in the azure-sdk-for-net directory, this change is sufficient for GA. Please file issues for any remaining items.

I created https://github.com/Azure/autorest.csharp/pull/2128#issuecomment-1108727218

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
AlexanderSher: You can simply skip it:
```cs
optionalParameters.Select(p => ClientOptions.Type.EqualsIgnoreNullable(p.Type) ? p : p with { DefaultValue = null }).ToArray()
```

Or you can even pass `ClientOptions` parameter into the method. Something like:
```cs
var requiredParameters = RestClientBuilder.GetRequiredParameters(orderedParameters).ToArray();
var optionalParameters = RestClientBuilder.GetOptionalParameters(orderedParameters).ToArray();
var clientOptionsParameter = CreateOptionsParameter();
return (
    BuildPrimaryConstructors(requiredParameters, optionalParameters, clientOptionsParameter).ToArray(),
    BuildSecondaryConstructors(requiredParameters, optionalParameters, clientOptionsParameter).ToArray()
);
```
and
```cs
private IEnumerable<ConstructorSignature> BuildPrimaryConstructors(IReadOnlyList<Parameter> requiredParameters, IReadOnlyList<Parameter> optionalParameters, Parameter clientOptionsParameter)
{
    var optionalToRequired = optionalParameters
        .Select(parameter => parameter with { DefaultValue = null })
        .Append(clientOptionsParameter)
        .ToArray();
```

```cs
private IEnumerable<ConstructorSignature> BuildSecondaryConstructors(IReadOnlyList<Parameter> requiredParameters, IReadOnlyList<Parameter> optionalParameters, Parameter clientOptionsParameter)
{
    if (requiredParameters.Any() || Fields.CredentialFields.Any())
    {
        yield return CreateMockingConstructor();
    }

    var optionalParametersArguments = optionalParameters
        .Append(clientOptionsParameter)
        .Select(p => p.Type.EqualsIgnoreNullable(typeof(Uri)) ? $"new {typeof(Uri):I}({p.DefaultValue!.Value.GetConstantFormattable()})" : p.DefaultValue!.Value.GetConstantFormattable())
        .ToArray();
```

pshao25: This cannot work either. In this case `options` will be optional.
Here is the problem:
1. We want `options` to be required, therefore the `DefaultValue` should be `null`.
2. We want `options` can be initialized if user passes in `null` as `options ??= new ClientOptions();`. But we don't know what type should be intialized bacause we get this type from `DefaultValue` which is `null` now.

AlexanderSher: It will work with `options` being optional if you set `UseDefaultValueInCtorParam` to false:
https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs#L203

This way there will be no `= default` generated in the signature.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -16,6 +15,9 @@
 using Azure.Core;
 using Operation = AutoRest.CSharp.Input.Operation;
 using static AutoRest.CSharp.Output.Models.MethodSignatureModifiers;
+using System.Runtime.CompilerServices;
```
lirenhe, line 18: For coding style, do we need to sort the 'using'?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: If you want `DefaultValue` not to be `null`, then it will goes to https://github.com/Azure/autorest.csharp/blob/d51e8d41dffbcb1071dfec5b65b24f99194cc464/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs#L224, we still have the `= default` in signature.
If you don't want `DefaultValue` not to be `null`, then we need some place to store the initialization type.

AlexanderSher:
We also need a test where one of the optional parameters is an enumerable or dictionary

annelo-msft:
> We also need to add one of the optional parameters that is an enumerable or dictionary to our test constructor.

This is not blocking DPG GA because we have no current DPG libraries with this pattern.  Let's please file an issue to track this and assign it to the next sprint.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
AlexanderSher: Ok, so I think it is time to separate `DefaultValue` constant and `Initializer` expression, make `CodeWriterExtensions` just write `DefaultValue` in the signature and `Initializer` in method body without any heuristics. Code that instantiates `Parameter` should fully responsible for what is needed, so there is no need to "adapt" to `CodeWriterExtensions` internal logic.
Here is the [issue](https://github.com/Azure/autorest.csharp/issues/2208) and the [PR](https://github.com/Azure/autorest.csharp/pull/2200).

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
AlexanderSher: Fixed

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -229,7 +229,7 @@ public static CodeWriter WriteParametersValidation(this CodeWriter writer, IEnum
 
         private static CodeWriter WriteParameterValidation(this CodeWriter writer, Parameter parameter)
         {
-            if (parameter.Initializer != null)
+            if (parameter.Validation == ValidationType.None && parameter.Initializer != null)
```
AlexanderSher, line 232: Why do you need this change?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -142,33 +144,98 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
+        }
 
-            var clientOptionsType = ClientOptions.Type.WithNullable(true);
-            var clientOptionsParameter = new Parameter("options", "The options for configuring the client.", clientOptionsType, Constant.Default(clientOptionsType), ValidationType.None, Constant.NewInstanceOf(clientOptionsType).GetConstantFormattable());
+        private IEnumerable<ConstructorSignature> BuildPrimaryConstructors(IReadOnlyList<Parameter> requiredParameters, IReadOnlyList<Parameter> optionalParameters)
+        {
+            var optionalToRequired = optionalParameters.Select(parameter =>
+                ClientOptions.Type.EqualsIgnoreNullable(parameter.Type) ?
+                parameter with { DefaultValue = null, Validation = ValidationType.None } :
+                Parameter.FromOptionalToRequired(parameter)).ToArray();
```
AlexanderSher, line 171: I'd rather make `Parameter.GetValidation` public and inline `FromOptionalToRequired` , cause this method has meaning only in this case, e.g.:
```cs
var optionalToRequired = optionalParameters
    .Select(parameter => ClientOptions.Type.EqualsIgnoreNullable(parameter.Type)
        ? parameter with { DefaultValue = null, Validation = ValidationType.None }
        : parameter with
        {
            DefaultValue = null,
            Validation = parameter.Initializer != null ? Parameter.GetValidation(parameter.Type, parameter.RequestLocation, parameter.SkipUrlEncoding) : parameter.Validation,
            Initializer = null
        }
    )
    .ToArray();
```

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
pshao25: E.g., 
```
public AccessibilityClient(AzureKeyCredential credential, Uri endpoint, AccessibilityClientOptions options)
{
    Argument.AssertNotNull(credential, nameof(credential));
    Argument.AssertNotNull(endpoint, nameof(endpoint));
    options ??= new AccessibilityClientOptions();
}
```
`options` and `endpoint` are both required, but we want to validate `endpoint` notnull while initialize `options` by default value. So when I constructor the parameters, I make the `validation` of `options` as `none`.

AlexanderSher (approved):
Approved

Issue #2141, opened 2022-04-05
Introduce a `Update` method to a resource using the PUT if it does not have a patch


ArcturusZhang:
This is a convenient method for those resources that do not have a patch, if we do not have this method, we will not have any way to update it inside the resource instance, we have to do it in the collection, which involves its parent.

m-nash:
Today if you have a resource object and it only provides a PUT method and no PATCH which acts as a CreateOrUpdate in order to update it you must do it from the parent since there will be no Update method generated on the resource class.

```
var resource = armClient.GetResource(id);
//this method won't exist
resource.Update(new ResourceData());

//instead customers must do something like this
var parent = armClient.GetParent(id.Parent);
var resourceCollection = parent.GetResources();
resourceCollection.CreateOrUpdate(id.Name, new ResourceData());
```

Since PUT works as a CreateOrUpdate then we should provide a duplicate copy of this method named Update on the resource class directly to avoid the inconvenience above.

PR #2163, opened 2022-04-14
Always emit an Update method on the Resource class if there is only a PUT but no PATCH

m-nash:
# Description
Implements this issue https://github.com/Azure/autorest.csharp/issues/2141

# Checklist

# Ready to Land?


samples/Azure.ResourceManager.Sample/Generated/Models/ProximityPlacementGroupPatch.Serialization.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Azure.ResourceManager.Sample.Models
 {
-    internal partial class ProximityPlacementGroupPatch : IUtf8JsonSerializable
+    public partial class ProximityPlacementGroupPatch : IUtf8JsonSerializable
```
ShivangiReja, line 13: Why did we change serialization/deserialization class to public?

m-nash: this model is now used on a public API so it became public.  We have an internalizer that runs and hides all models that aren't publicly exposed to minimize the public api footprint but after this change this model is no longer internal

ShivangiReja: Not in this case but we also generate deserialization method in this class, so it will also be public?

ShivangiReja: And we have [ProximityPlacementGroupPatch](https://github.com/Azure/autorest.csharp/blob/7f88dec51a06eff8dc3780d299441904116a64d6/samples/Azure.ResourceManager.Sample/Generated/Models/ProximityPlacementGroupPatch.cs#L11) public.

ShivangiReja: I was talking about [this](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/NameConflicts/Generated/Models/HttpMessage.Serialization.cs#L26) case but nm we generate an internal deserialization method.

ShivangiReja (approved):

Issue #2167, opened 2022-04-18
[DPG] Set the default suppress nested diagnostic scope flag to true in DPG clients


ShivangiReja:
After we merge https://github.com/Azure/azure-sdk-for-net/pull/28066 PR, set the default suppress nested diagnostic scope flag to true in DPG clients.

Check here for more details - https://github.com/Azure/azure-sdk-for-net/pull/28066#issuecomment-1102989858

PR #2184, opened 2022-04-23
[DPG] Set the default SuppressNestedSpans flag to true for DPG clients

ShivangiReja:
Fixes: https://github.com/Azure/autorest.csharp/issues/2167

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
@@ -10,7 +10,7 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal class ClientOptionsWriter
     {
-        public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvider clientOptions)
+        public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvider clientOptions, bool suppressNestedClientSpans)
         {
             using (writer.Namespace(clientOptions.Type.Namespace))
             {
@@ -36,6 +36,10 @@ namespace AutoRest.CSharp.Generation.Writers
                     writer.WriteXmlDocumentationSummary($"Initializes new instance of {clientOptions.Type.Name}.");
                     using (writer.Scope($"public {clientOptions.Type.Name}(ServiceVersion version = LatestVersion)"))
                     {
+                        if (suppressNestedClientSpans)
+                        {
+                            writer.Line($"Diagnostics.SuppressNestedClientSpans = true;");
+                        }
                         writer.Append($"Version = version ");
                         using (writer.Scope($"switch", end: "};"))
                         {
--- a/src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
@@ -10,7 +10,7 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal class ClientOptionsWriter
     {
-        public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvider clientOptions, bool suppressNestedClientSpans = false)
+        public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvider clientOptions)
         {
             using (writer.Namespace(clientOptions.Type.Namespace))
             {
@@ -36,10 +36,6 @@ namespace AutoRest.CSharp.Generation.Writers
                     writer.WriteXmlDocumentationSummary($"Initializes new instance of {clientOptions.Type.Name}.");
                     using (writer.Scope($"public {clientOptions.Type.Name}(ServiceVersion version = LatestVersion)"))
                     {
-                        if (suppressNestedClientSpans)
-                        {
-                            writer.Line($"Diagnostics.SuppressNestedClientSpans = true;");
-                        }
                         writer.Append($"Version = version ");
                         using (writer.Scope($"switch", end: "};"))
                         {
--- a/src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
@@ -10,7 +10,7 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal class ClientOptionsWriter
     {
-        public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvider clientOptions, bool suppressNestedClientSpans)
+        public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvider clientOptions, bool suppressNestedClientSpans = false)
         {
             using (writer.Namespace(clientOptions.Type.Namespace))
             {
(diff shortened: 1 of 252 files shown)


ShivangiReja:
This PR is failing for [this](https://github.com/Azure/autorest.csharp/blob/55e99dc32f8941cb8e5826286e3224f66750447c/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L32) test case [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1525229&view=logs&j=b5cdfb1e-c180-561f-918e-a169e451719e&t=d557719e-b1de-5962-364a-0820efa4a6bb&l=48), so I created an issue - https://github.com/Azure/azure-sdk-for-net/issues/28354 to fix this.

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -24,7 +24,14 @@ public class DpgCustomizationTest : TestServerLowLevelTestBase
         [Test]
         public Task GetHandwrittenModel() => Test(async (host) =>
         {
+            using var diagnosticListener = new ClientDiagnosticListener("dpg_customization_LowLevel", asyncLocal: true);
```
annelo-msft, line 27: Would it be possible to also add a test that shows us not suppressing the nested scope and seeing `scopes.Count == 2`?

ShivangiReja: Test case added.

* [Without suppressing the nested scope](https://github.com/Azure/autorest.csharp/blob/a8d3964566b6dbed5d144665ab3dbd328887d9d3/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L39)
* [With suppressing the nested scope](https://github.com/Azure/autorest.csharp/blob/a8d3964566b6dbed5d144665ab3dbd328887d9d3/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L25)

src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
```diff
@@ -10,7 +10,7 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal class ClientOptionsWriter
     {
-        public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvider clientOptions)
+        public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvider clientOptions, bool suppressNestedClientSpans)
```
lirenhe, line 13: Can we have the default value of 'false' so we don't need to change the code in DataPlaneTarget.cs

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -24,10 +24,34 @@ public class DpgCustomizationTest : TestServerLowLevelTestBase
...
+            var scopes = diagnosticListener.Scopes;
+            CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+
             Response<Product> result = await new DPGClient(Key, host).GetModelValueAsync("model");
+            Assert.True(scopes.Count == 1);
```
AlexanderSher, line 32: `Assert.AreEqual(1, scopes.Count);`

annelo-msft (approved):
Generator output looks great.  Thanks, @ShivangiReja!

test/TestServerProjectsLowLevel/dpg-customization/Generated/DPGClientOptions.cs
```diff
@@ -27,6 +27,7 @@ public enum ServiceVersion
         /// <summary> Initializes new instance of DPGClientOptions. </summary>
         public DPGClientOptions(ServiceVersion version = LatestVersion)
         {
+            Diagnostics.SuppressNestedClientSpans = true;
```
chunyu3, line 30: We will set the SuppressNestedClientSpans true to suppressNestedClientSpans, but I saw the ClientDiagnostics constructor in ClientDiagnostics.cs. The Diagnostics.SuppressNestedClientSpans is not used at all. It is hard-set false.  Is there any mismatch between Azure.Core and autorest.csharp?


image

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -24,10 +24,34 @@ public class DpgCustomizationTest : TestServerLowLevelTestBase
         [Test]
         public Task GetHandwrittenModel() => Test(async (host) =>
         {
+            using var diagnosticListener = new ClientDiagnosticListener("dpg_customization_LowLevel", asyncLocal: true);
+            var scopes = diagnosticListener.Scopes;
+            CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+
             Response<Product> result = await new DPGClient(Key, host).GetModelValueAsync("model");
+            Assert.AreEqual(1, scopes.Count);
```
chunyu3, line 32: Hi @ShivangiReja Why can we according to the scopes.Count to validate if the scope is suppressed?  I could not find any code in Azure.Core that diagnosticListener will not add the scope which is suppressed.

test/TestServerProjectsLowLevel/dpg-customization/Generated/DPGClientOptions.cs
ShivangiReja: As I mentioned [here](https://github.com/Azure/autorest.csharp/pull/2184#issuecomment-1107117271), we are tracking this issue [here](https://github.com/Azure/azure-sdk-for-net/issues/28354). We also have a [test case](https://github.com/Azure/autorest.csharp/blob/308f082d8f98f707a11f781021068b9524f57c97/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L32) to check that the nested spans are being suppressed are not and currently it is failing because of this issue.

ShivangiReja: As I mentioned [here](https://github.com/Azure/autorest.csharp/pull/2184#issuecomment-1107117271), this PR is blocked and we are tracking this issue [here](https://github.com/Azure/azure-sdk-for-net/issues/28354). We also have a [test case](https://github.com/Azure/autorest.csharp/blob/308f082d8f98f707a11f781021068b9524f57c97/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L32) to check that the nested spans are being suppressed are not and currently it is failing because of this issue.

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
ShivangiReja: `diagnosticListener` is not used to suppress the scope, it is used to listen how many scopes are created, count of the scopes and the scopes are completed or not. We are using it to test the scopes.  The code for ClientDiagnosticListener is in autorest.csharp repo [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/TestFramework/ClientDiagnosticListener.cs). And, the logic to suppress the scope is in the [DiagnosticScope](https://github.com/Azure/azure-sdk-for-net/pull/28066/files#diff-2917d0ddf4987f75fc3ef2ac6445df3150798ecdf01edb8c1a24b2e64cbc72de)

Inside `GetModelValueAsync` method, we are creating two scopes but we are suppressing the nested scopes [here](https://github.com/Azure/autorest.csharp/blob/308f082d8f98f707a11f781021068b9524f57c97/test/TestServerProjectsLowLevel/dpg-customization/Generated/DPGClientOptions.cs#L30). So, in [this](https://github.com/Azure/autorest.csharp/blob/308f082d8f98f707a11f781021068b9524f57c97/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L32-L34) test case we are validating the scope count is one, the scope name and whether or not it is completed using `diagnosticListener`.

AlexanderSher (approved):

PR #2200, opened 2022-05-01
Refactor parameter initialization and validation

AlexanderSher:
This change separates parameter default value in signature, type of validation and initialization expression inside method body into separate fields and moves logic from `CodeWriterExtension` into `Parameter` itself. 
Fixes https://github.com/Azure/autorest.csharp/issues/2208


annelo-msft (commented):
@AlexanderSher, what issue is this responding to?

pshao25:
I thought whether a parameter is required would be decoupled from DefaultValue. If this is still the case, why remove the IsRequired and add a IsOptional? What is the difference from before?

AlexanderSher:
From [docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments#optional-arguments): "Each optional parameter has a default value as part of its definition. If no argument is sent for that parameter, the default value is used."

We can call property "IsSignatureOptional" to make the difference clear

pshao25 (approved):

PR #2213, opened 2022-05-06
Remove dependency on OperationOrResponseInternals

AlexanderSher:
With the updated `OperationInternal` and `OperationInternalOfT`, we can get rid of `OperationOrResponseInternals` and `OperationOrResponseInternals<T>`. That minimizes amount of duplicated code and helps minimize amount of service calls when `WaitForCompletionAsync` is called concurrently by multiple callers


m-nash (commented):
The code looks fine but before we merge this can we run generation / testing against azure-sdk-for-net to validate that everything still works as expected?

AlexanderSher:
Tests have passed

m-nash (approved):

Issue #1647, opened 2021-10-28
[feature] provide code generation report


archerzz:
# Description

Since .net SDK adopts a hierarchical model which is very different from swagger, it would be good that the generator could provide an overall report at the end of generation. That would benefit both users and the generation development.

The contents of the report can include but not limited to:
- Request path hierarchy
- Identified resources
  - Resource definitions under multiple context path
- Identified non-resources

PR #2228, opened 2022-05-11
A prototype for mgmt generation report

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/1647

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/CodeModelExtension.cs
```diff
@@ -167,11 +186,15 @@ public static void UpdateSubscriptionIdForAllResource(this CodeModel codeModel)
...
                             p.Implementation = ImplementationLocation.Method;
+                            // TODO -- change the operation name to `op.OperationId` when it is ready
+                            MgmtContext.Report.Add(ReportLevel.Information, $"The implementation location of parameter {p.Language.Default.Name} in operation `{operationGroup.Key}_{op.Language.Default.Name}` has been changed to Method");
                         }
                         // update the apiVersion parameter to be 'client' method
                         if (p.Language.Default.Name.Equals("apiVersion", StringComparison.OrdinalIgnoreCase))
                         {
                             p.Implementation = ImplementationLocation.Client;
+                            // TODO -- change the operation name to `op.OperationId` when it is ready
+                            MgmtContext.Report.Add(ReportLevel.Information, $"The implementation location of parameter {p.Language.Default.Name} in operation `{operationGroup.Key}_{op.Language.Default.Name}` has been changed to Client");
```
fengzhou-msft, line 197: The original location for apiVersion is Client for most cases, only in a few cases like [ResourcesRestOperations](https://github.com/Azure/azure-sdk-for-net/blob/1e11c266329079133cdb62c7fe9738068002c529/sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/RestOperations/ResourcesRestOperations.cs#L253), the location was Method (The customized code actually moves back the location for apiVersion to Method to offset the change here. There should be a better example for why the change is made here, but I cannot remember now).

ArcturusZhang: GenericResource CRUD need the api-version to be a method parameter, I believe this should be the only use case for this.

I will add a if-check here only to output the log line when there is a change.
Or I could introduce a method in the `Report` class, to determine whether we need the output or not.

fengzhou-msft:
Shall we have report for common type replacement?

ArcturusZhang:
> Shall we have report for common type replacement?

I believe we should have everything that we have changed (comparing with the modeler) in the report, therefore this should be included.

PR #2229, opened 2022-05-11
Introduce partial resources for cross RP resource references

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/28185
Regen: https://github.com/Azure/azure-sdk-for-net/pull/34807 (no public API changes)

# Description

When we want to extend a resource from other RP, we will create a virtual resource in this pattern: `{ResourceName}{RPName}Resource`, and we use this resource as the instance of that resource from other RP, and provide hierarchy structure on this virtual resource.

Design information please refer [this comment](https://github.com/Azure/azure-sdk-for-net/issues/28185#issuecomment-1116659864)

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -392,7 +392,7 @@ private ArmClientExtensions EnsureArmClientExtensions()
         }
 
         private IEnumerable<ResourceData>? _resourceDatas;
-        public IEnumerable<ResourceData> ResourceData => _resourceDatas ??= RawRequestPathToResourceData.Values.Distinct();
+        public IEnumerable<ResourceData> ResourceData => _resourceDatas ??= RawRequestPathToResourceData.Values.Where(data => data is not VirtualResourceData).Distinct();
```
archerzz, line 395: Maybe rename this property to `NonVirtualResourceData`?

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -512,30 +512,45 @@ public bool TryGetRestClients(string requestPath, [MaybeNullWhen(false)] out Has
...
-                    foreach (var resourcePath in resourcePaths)
-                    {
-                        var resourceType = resourcePath.GetResourceType();
-                        var resource = new Resource(operationSet, operations, GetResourceName(resourceDataSchemaName, operationSet, resourcePath), resourceType, resourceData);
-                        var collection = isSingleton ? null : new ResourceCollection(operationSet, operations, resource);
-                        resource.ResourceCollection = collection;
-
-                        requestPathToResources.Add(resourcePath, new ResourceObjectAssociation(resourceType, resourceData, resource, collection));
-                    }
+                    if (operationSet.Count > 0)
```
archerzz, line 521: Why will there be an empty operation set?

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
ArcturusZhang: I am struggling renaming this, or leave this as it is, and change the output part in `MgmtTarget` only to write a file out when it is not `VirtualResourceData`.
What do you think?

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
ArcturusZhang: Yes, the operationSet of a `VirtualResource` is usually empty.

src/AutoRest.CSharp/Mgmt/Output/VirtualResource.cs
```diff
@@ -0,0 +1,39 @@
...
+using AutoRest.CSharp.Mgmt.Decorator;
+using AutoRest.CSharp.Mgmt.Models;
+using AutoRest.CSharp.Output.Models;
+
+namespace AutoRest.CSharp.Mgmt.Output
+{
+    /// <summary>
+    /// A virtual resource stands for a resource from another SDK, and it plays a role of anchor of some operations that belong to this resource in another SDK
+    /// </summary>
+    internal class VirtualResource : Resource
```
m-nash, line 17: Can we call this a `PartialResource`.  Initially I was reviewing this in the context of how a publisher or location is a virtual resource since its not really an entity that you can create or manage but it is an organizational tool to group and find things in the same group.

The concept here is that this will act like partial class across libraries.

test/TestProjects/MgmtVirtualResource/Generated/VirtualMachineMgmtVirtualResource.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// <summary>
+    /// A Class representing a VirtualMachine along with the instance operations that can be performed on it.
+    /// If you have a <see cref="ResourceIdentifier" /> you can construct a <see cref="VirtualMachineMgmtVirtualResource" />
+    /// from an instance of <see cref="ArmClient" /> using the GetVirtualMachineMgmtVirtualResource method.
+    /// Otherwise you can get one from its parent resource <see cref="ResourceGroupResource" /> using the GetVirtualMachine method.
+    /// </summary>
+    public partial class VirtualMachineMgmtVirtualResource : ArmResource
+    {
+        /// <summary> Generate the resource identifier of a <see cref="VirtualMachineMgmtVirtualResource"/> instance. </summary>
+        public static ResourceIdentifier CreateResourceIdentifier(string subscriptionId, string resourceGroupName, string vmName)
```
m-nash, line 28: I think this should be omitted.

m-nash (changes requested):

src/AutoRest.CSharp/Mgmt/Output/VirtualResource.cs
ArcturusZhang: I renamed this to PartialResource

test/TestProjects/MgmtVirtualResource/Generated/VirtualMachineMgmtVirtualResource.cs
ArcturusZhang: I did some research and we cannot remove this method. We need this method when we are trying to do the id re-assemblying (the operation-id-mapping configuration) in LRO.
The operations on the partial resource might also need this configuration to correct things, therefore we need to keep this method.
We could make this internal so that our customer do not have access to this method.

docs/mgmt/readme.md
```diff
@@ -625,6 +626,42 @@ list-exception:
...
+public partial class VirtualMachineNetworkResource
+{
+    public virtual Pageable<PublicIPAddress> GetPublicIPs()
+    {
+        /* the implementation */
+    }
+}
+```
+
+The naming pattern of a partial resource is like `[ResourceName][RPName]Resource`, where `ResourceName` is the value you assigned in the configuration, `RPName` is the name of this RP (for instance, `Network`). The name of a partial resource is also configurable using the `request-path-to-resource-name` configuration, please see [change resource name](#change-resource-name) section for more details.
```
ArthurMa1978, line 660: Not sure, but is there case that the `ResourceName` = `RPName` ? Or similar, then the name of partial resource will be weird, so how about have a conjunction between them?

ArcturusZhang: Considering that the occurrence of partial resource will be not quite common, I believe it would be even more rarely that your edge case could happen. Therefore I did not introduce any "automatic logic" when getting a name for this partial resource.
We could always use `request-path-to-resource-name` to customize the name of a partial resource.
Here we have an example in the test project: https://github.com/Azure/autorest.csharp/blob/55ca9538793cfc0691aecd0b01e1dfd713ad9593/test/TestProjects/MgmtPartialResource/readme.md

ArthurMa1978: Ok, this is LGTM

ArthurMa1978 (approved):

test/TestProjects/MgmtPartialResource/Generated/PartialVmssResource.cs
```diff
@@ -0,0 +1,147 @@
...
+namespace MgmtPartialResource
+{
+    /// <summary>
+    /// A class extending from the VirtualMachineScaleSetResource in MgmtPartialResource along with the instance operations that can be performed on it.
+    /// You can only construct a <see cref="PartialVmssResource" /> from a <see cref="ResourceIdentifier" /> with a resource type of Microsoft.Compute/virtualMachineScaleSets.
+    /// </summary>
+    public partial class PartialVmssResource : ArmResource
+    {
+        /// <summary> Generate the resource identifier of a <see cref="PartialVmssResource"/> instance. </summary>
+        internal static ResourceIdentifier CreateResourceIdentifier(string subscriptionId, string resourceGroupName, string virtualMachineScaleSetName)
```
Yao725, line 27: I'd like to know why this is internal. The internal here means if we want to get a resource identifier, we need to use the corresponding `CreateResourceIdentifier` method in VmssResource of compute, which will result in that we need to add the package `Azure.ResourceManager.Compute` as an reference, or we need to directly use the ctor of `ResourceIdentifier`. And I think both of the options are not friendly for the user.

Yao725: Another finding that makes me think maybe we should change this method to public method: In Consumption we have a partial resource called [BillingAccountConsumptionResource](https://github.com/Azure/azure-sdk-for-net/blob/7d344e2ccdd6626a1dce393240027480527696bd/sdk/consumption/Azure.ResourceManager.Consumption/src/Customized/BillingAccountConsumptionResource.cs#L22), but unfortunately we don't have a BillingAccountResource in the RP `Billing`, in this case it will be difficult for the customer to get the correct resource identifier.

ArcturusZhang:
This is still WIP... will update this in a few days

m-nash (approved):

m-nash:
Please add regen PR before merging

PR #2240, opened 2022-05-18
Enable `x-ms-xxx` extensions for method parameter

ArthurMa1978:
Currently the `x-ms-xxx` extensions only take effect on model property, but not work properly for method parameter.
The issue is coming from the difference in the output from m4:
- For property, the extension is applied on both property and the framework type, 
- For parameter, the extension is only applied on the parameter but not affect framework type.

For example, if config `['x-ms-format'] = 'azure-location'` for a string property named `location`, then both the string type and property have the 'azure-location' attribute:
```
- !Property 
  schema: *ref_3
  required: true
  serializedName: location
  extensions:
	x-ms-format: azure-location
  language: !Languages 
	default:
	  name: location
	  description: The supported Azure location where the key vault should be created.
  protocol: !Protocols {}
```
```	  
- !StringSchema &ref_3
  type: string
  apiVersions:
	- !ApiVersion 
	  version: '2021-10-01'
  extensions:
	x-ms-format: azure-location
  language: !Languages 
	default:
	  name: VaultCreateOrUpdateParametersLocation
	  description: The supported Azure location where the key vault should be created.
  protocol: !Protocols {}
```
if config `['x-ms-format'] = 'azure-location'` for a string parameter named `location`, then only the parameter have the 'azure-location' attribute:
```
- !Parameter &ref_260
schema: *ref_1
implementation: Method
required: true
extensions:
  x-ms-format: azure-location
language: !Languages 
  default:
	name: location
	description: The location of the deleted vault.
	serializedName: location
protocol: !Protocols 
  http: !HttpParameter 
	in: path
```
```
- !StringSchema &ref_1
  type: string
  apiVersions:
	- !ApiVersion 
	  version: '2021-10-01'
  language: !Languages 
	default:
	  name: String
	  description: ''
	  header: Azure-AsyncOperation
  protocol: !Protocols {}
```

So here the fixing is adding an overload `CreateType` which will consume `Extensions` data separately (not from the `Schema`), and then the `RestClientBuilder.BuildParameter` can create the type with the right extensions.


ArthurMa1978:
Here is the draft pr for KeyVault to change type of `location` parameter from `string` to `AzureLocation` https://github.com/Azure/azure-sdk-for-net/pull/28862

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -28,7 +28,9 @@ public TypeFactory(OutputLibrary library)
             _library = library;
         }
 
-        public CSharpType CreateType(Schema schema, bool isNullable) => schema switch
+        public CSharpType CreateType(Schema schema, bool isNullable) => CreateType(schema, schema.Extensions, isNullable);
+
+        public CSharpType CreateType(Schema schema, RecordOfStringAndAny? schemaExtensions, bool isNullable) => schema switch
```
ArcturusZhang, line 33: The parameter name `schemaExtensions` here is quite misleading which indicates this extension is from the schema, and the schema has a property `Extensions`.
But in fact this method is for the case that we are using an extension from other things (like the `requestParameter` instead of the schema
How about we rename this parameter to `extensions`, and add some xml document on this method?

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -177,7 +179,9 @@ internal static bool IsOperationOfAsyncPageable(CSharpType type)
         internal static bool IsOperationOfPageable(CSharpType type)
             => type.IsFrameworkType && type.FrameworkType == typeof(Operation<>) && type.Arguments.Length == 1 && IsPageable(type.Arguments[0]);
 
-        internal static Type? ToFrameworkType(Schema schema) => schema.Type switch
+        internal static Type? ToFrameworkType(Schema schema) => ToFrameworkType(schema, schema.Extensions);
+
+        internal static Type? ToFrameworkType(Schema schema, RecordOfStringAndAny? schemaExtensions) => schema.Type switch
```
ArcturusZhang, line 184: Same comment here, but since this is an internal method, we do not need to add an xml document, but it is always a good thing to add comments describing things

samples/Azure.ResourceManager.Sample/Generated/RestOperations/VirtualMachinesRestOperations.cs
```diff
@@ -87,12 +86,11 @@ public async Task<Response<VirtualMachineListResult>> ListByLocationAsync(string
...
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="subscriptionId"/> or <paramref name="location"/> is null. </exception>
-        /// <exception cref="ArgumentException"> <paramref name="subscriptionId"/> or <paramref name="location"/> is an empty string, and was expected to be non-empty. </exception>
-        public Response<VirtualMachineListResult> ListByLocation(string subscriptionId, string location, CancellationToken cancellationToken = default)
+        /// <exception cref="ArgumentNullException"> <paramref name="subscriptionId"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="subscriptionId"/> is an empty string, and was expected to be non-empty. </exception>
+        public Response<VirtualMachineListResult> ListByLocation(string subscriptionId, AzureLocation location, CancellationToken cancellationToken = default)
         {
             Argument.AssertNotNullOrEmpty(subscriptionId, nameof(subscriptionId));
-            Argument.AssertNotNullOrEmpty(location, nameof(location));
```
fengzhou-msft, line 95: This is a path parameter and cannot be null, shall we add `Argument.AssertNotNull(location, nameof(location));`?

ArthurMa1978: An enum is not allow null by default

fengzhou-msft (approved):

samples/Azure.ResourceManager.Sample/Generated/RestOperations/VirtualMachinesRestOperations.cs
ArthurMa1978: the AzureLocation is a struct which is value type, so the assert is removed as it's can't be null by design

Issue #2099, opened 2022-03-23
[data-plane] when enable 'single top-level client', the name of service client is incorrect


chunyu3:
There are two scenarios which will generate the wrong service client name:
1. When we try to combine 'single top-level client' with 'client hierarchy' with an service which has no top-level method
swagger json:
```
{
  "swagger": "2.0",
  "info": {
    "version": "2021-10-19",
    "title": "SubClients"
  },
  "x-ms-parameterized-host": {
    "hostTemplate": "{endpoint}",
    "useSchemePrefix": true,
    "parameters": [
      {
        "$ref": "#/parameters/EndpointParameter"
      }
    ]
  },
  "schemes": [
    "https"
  ],
  "produces": [ "application/json" ],
  "consumes": [ "application/json" ],
  "paths": {
    "/client1/{cachedParameter}": {
      "get": {
        "operationId": "Client1_GetCachedParameter",
        "parameters": [
          {
            "name": "cachedParameter",
            "type": "string",
            "in": "path",
            "required": true,
            "x-ms-parameter-location": "client"
          }
        ],
        "responses": {
          "200": {
            "description": "Success",
            "schema": {
              "type":"string"
            }
          }
        }
      }
    },
    "/client2/{subParameter}": {
      "get": {
        "operationId": "Client2_GetSubParameter",
        "parameters": [
          {
            "name": "subParameter",
            "in": "path",
            "type": "string",
            "required": true
          }
        ],
        "responses": {
          "200": {
            "description": "Success",
            "schema": {
              "type":"string"
            }
          }
        }
      }
    }
  },
  "definitions": { },
  "parameters": { 
    "EndpointParameter": {
      "name": "endpoint",
      "in": "path",
      "required": true,
      "type": "string",
      "x-ms-skip-url-encoding": true,
      "description": "Account endpoint.",
      "x-ms-parameter-location": "client",
      "x-ms-parameter-grouping": {
        "name": "account-options"
      }
    }
  },
  "securityDefinitions": {},
  "security": [],
  "tags": []
}
```

set 'single top-level client' 

``` yaml
namespace: Azure.Service.SubClients
require: $(this-folder)/../../../readme.md
input-file: $(this-folder)/SubClients-LowLevel.json
data-plane: true
security: AzureKey
security-header-name: Fake-Subscription-Key
single-top-level-client: true
```

And add customization.cs to define 'client hierarchy'
``` c#
using Azure.Core;

namespace SubClient_LowLevel
{
[...]

PR #2256, opened 2022-05-25
[LLC] single-top-level-client configuration

chunyu3:
# Description

# Checklist

# Ready to Land?

- Resolve the service client name incorrect issue when set 'single-top-level-client'. related issue: https://github.com/Azure/autorest.csharp/issues/2099
- Make 'single-top-level-client' as default behavior, related issue: https://github.com/Azure/autorest.csharp/issues/2153


pshao25 (approved):

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
```diff
@@ -110,10 +110,19 @@ private static IReadOnlyList<ClientInfo> SetHierarchy(IReadOnlyDictionary<string
             }
 
             var topLevelClients = clientInfosByName.Values.Where(c => c.Parent == null).ToList();
-            if (!Configuration.SingleTopLevelClient || topLevelClients.Count == 1)
+            if (!Configuration.SingleTopLevelClient)
             {
                 return topLevelClients;
             }
+            if ( topLevelClients.Count == 1)
```
archerzz, line 117: remove the extra space before `topLevelClients`

test/AutoRest.TestServerLowLevel.Tests/RequestContextTests.cs
```diff
@@ -21,14 +21,14 @@ public class RequestContextTests : TestServerLowLevelTestBase
         [Test]
         public Task RequestThrowsByDefault () => Test(host =>
         {
-            PathsClient paths = new PathsClient(Key, host, null);
+            Paths paths = new AutoRestUrlTestServiceClient(Key, host, null).GetPathsClient();
```
archerzz, line 24: so if `single-top-level-client : true`, there will always be a top level client?

samples/Azure.Analytics.Purview.Account/readme.md
```diff
@@ -6,6 +6,7 @@ namespace: Azure.Analytics.Purview.Account
 data-plane: true
 security: AADToken
 security-scopes:  https://purview.azure.net/.default
+single-top-level-client: false
```
archerzz, line 9: Normally it's a convention that the default value of not-defined property should be zero value (e.g., false, 0, null). It's a little bit weird that the default value is true.
It would be better if the property name can be changed to `multiple-top-level-clients`.

archerzz (approved):

samples/Azure.Analytics.Purview.Account/readme.md
pshao25: Why this readme.md still has configuration `data-plane: true`? Shouldn't it be removed as we already make it default?

test/AutoRest.TestServerLowLevel.Tests/RequestContextTests.cs
chunyu3: yes

samples/Azure.Analytics.Purview.Account/readme.md
chunyu3: Remove all data-plane: true

annelo-msft:
@chunyu3, would you please add a link to a PR on the azure-sdk-for-net repo showing the output when the generator update is applied to all clients?

test/AutoRest.TestServerLowLevel.Tests/RequestContextTests.cs
annelo-msft: @chunyu3, why is `PathsClient` not the top level client?

annelo-msft (changes requested):
I'm concerned about this change introducing a different top-level client for all the test clients and having to retrieve a subclient from them to implement the test.

@chunyu3, could you help us understand what the swagger input is that causes this to happen?  I believe all of the languages are using the same test server swagger files, and none of the other languages are considering these subclients.

test/AutoRest.TestServerLowLevel.Tests/RequestContextTests.cs
chunyu3: In Url RP in the test-server, there are more than one operation groups (paths, queries,pathItems). So when set 'single-top-level-client', it will create a new service client and use the title as the client name, all the operation groups will be subclient.

chunyu3:
> @chunyu3, would you please add a link to a PR on the azure-sdk-for-net repo showing the output when the generator update is applied to all clients?

As we discussed before, we will set 'single-top-level-client' as default behavior in autorest.csharp and will update all the samples and tests in autorest.csharp. But we will maintain the RPs in azure-sdk-for-net as it is. That is we will set 'single-top-level-client: false' for RPs in azure-sdk-for-net. So there will be another PR to set flag 'single-top-level-client: false' in azure-sdk-for-net.

chunyu3:
> I'm concerned about this change introducing a different top-level client for all the test clients and having to retrieve a subclient from them to implement the test.
> 
> @chunyu3, could you help us understand what the swagger input is that causes this to happen? I believe all of the languages are using the same test server swagger files, and none of the other languages are considering these subclients.

Hi @annelo-msft 
.Net now implement the subclient feature by 'single-top-level-client' in this way:
When 'single-top-level-client' flag is set, the service will have only one service client even it has more than one operation groups.
1.  If there only one operation group, the operation group will be the service client.
2. If there are more than one operation groups:
  2.1  If there is an operation group with no/empty groupName, the operation group will be the service client and use the title as the client name. The other operation groups will be subclient
  2.2 if there is no such operation group whose name is empty, a new service client will be created and use the title as the client name, all the operation groups will be subclient.

By the way, .NET also provide 'client Hierarchy' customization to support subclient feature.

annelo-msft:
@AlexanderSher, can you please review this implementation and decide whether or not to approve it?

@mikekistler, do you have any concerns with making this change to autorest?

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
```diff
@@ -229,7 +238,7 @@ private static IEnumerable<LowLevelClient> CreateClients(IEnumerable<ClientInfo>
         public class ClientInfo
         {
             public string OperationGroupKey { get; }
-            public string Name { get; }
+            public string Name { get; set; }
```
AlexanderSher, line 241: This property should always remain read-only because it is used as a key in a dictionary:
```cs
var clientInfosByName = codeModel.OperationGroups
    .Select(og => CreateClientInfo(og, context))
    .ToDictionary(ci => ci.Name);
```

In general, it is a good idea to check how read-only property is used before making it writable.

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
```diff
@@ -110,10 +110,19 @@ private static IReadOnlyList<ClientInfo> SetHierarchy(IReadOnlyDictionary<string
...
             var topLevelClients = clientInfosByName.Values.Where(c => c.Parent == null).ToList();
-            if (!Configuration.SingleTopLevelClient || topLevelClients.Count == 1)
+            if (!Configuration.SingleTopLevelClient)
             {
                 return topLevelClients;
             }
+            if (topLevelClients.Count == 1)
+            {
+                var topClient = topLevelClients[0];
+                if (!topClient.ResourceParameters.Any() && !string.IsNullOrEmpty(topClient.OperationGroupKey))
```
AlexanderSher, line 120: This logic is kind of duplicated. We are checking the same thing in `CreateClientInfo`.
I'd say that we should remove this check both from `SetHierarchy` and from `CreateClientInfo` and add client suffix only in the `CreateClients`, right when `LowLevelClient` isntance is instantiated.

mikekistler:
@annelo-msft I have questions about how this versions.
- What happens when a second operation group is added in a new version?  Do I get a client that is compatible with the prior version?
- If the service has multiple operation groups, then I think for compatibility we need it to remain in 2A or 2B in all subsequent versions, so that the client name doesn't change.  I think that means that a service has multiple operation groups and all have a group name, then all new operations must also be in a named operation group.  Does that sound right?

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
chunyu3: Hi @AlexanderSher  If we delayed to add 'client' suffix in `CreateClients`, it may have problem to apply the 'Client Hierarchy' customization because in the customization it defined the 'ParentClient' attribute, e.g. 'ParentClient = typeof(ResourceServiceClient)'. It may cause apply customization fail. We may need to think about the order this is and the way to use this attribute to construct the client hierarchy.

src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs
chunyu3: Thank @AlexanderSher But now we have problem, usually we need to modify the clientName if we want to support both 'single-top-level-client' and `client hierarchy`. If Name is readonly, we need to create an new object instance when we want to update. Maybe we need to define other readonly field to use as key of dictionary.

lirenhe:
Let's keep this open

lirenhe:
@chunyu3, could you help to resolve all those conflicts?  Thanks

lirenhe:
@m-nash, @AlexanderSher, for context we got the feedback that dotnet should also make 'single-top-level-client: true' as the default behaviour in the recent archboard review of argifood.  So could you help to review this PR again?

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -295,9 +295,9 @@ private IEnumerable<LowLevelClient> CreateClients(IEnumerable<ClientInfo> client
                     : BuilderHelpers.EscapeXmlDescription(clientInfo.Description);
 
                 var subClients = new List<LowLevelClient>();
-
+                var clientSuffix = ClientBuilder.GetClientSuffix();
                 var client = new LowLevelClient(
-                    clientInfo.Name,
+                    (parentClient != null || clientInfo.Name.EndsWith(clientSuffix))? clientInfo.Name : clientInfo.Name + clientSuffix,
```
AlexanderSher, line 300: let's extract this line into the variable

m-nash:
I would like to talk about how we represent hierarchies in general in autorest before we merge this.  This might suffice for a short term solution, however we might be able to get to the more generic long term solution in relatively the same amount of time.

PR #2257, opened 2022-05-25
feat(dpg): resolve ambiguous methods with existing codes

### LINKED ISSUE #2204: Make RequestContext required on protocol methods in the event of an ambiguous call site with grow-up method
When we are growing up an DPG client into a Convenience Client, we have two different cases:

1. The protocol client API has GA'ed, so we can't change the protocol methods (this will be the case for most new libraries)
2. The convenience client API has GA'ed, so we can't change the convenience methods (this will be the case for many existing HLC libraries)

We have documented the process for 1, growing up protocol methods, at https://github.com/Azure/azure-sdk-for-net/blob/main/doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md.

For the second case, if we have a method with an ambiguous call site, we will [disambiguate the methods signatures by making RequestContext and other parameters required](https://github.com/Azure/azure-sdk-for-net/issues/28123#issuecomment-1106838980).

The generator will need to be updated to detect this case and generate the protocol method in the appropriate way.

archerzz:
- add clone method in `Paramter`, `MethodSignature` and `LowLevelClientMethod` to clone the object and make parameters required
- add `MethodAmbiguityResolver` which will detect and resolve ambiguous method
- add new logic in the constructor of `LowLevelClient` to detect and resolve ambigous `LowLevelClientMethod`
- add test cases
- a minor refectoring

resolve #2204

# Description

# Checklist

# Ready to Land?


archerzz:
I tried on the `MetricsAdvisor`, and there are more ambiguous methods found under `MetricsAdvisorAdministrationClient`.

image

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
```diff
@@ -0,0 +1,169 @@
...
+using System.Linq;
+using System.Text;
+using AutoRest.CSharp.Generation.Types;
+using AutoRest.CSharp.Input.Source;
+using AutoRest.CSharp.Output.Models;
+using AutoRest.CSharp.Output.Models.Shared;
+using AutoRest.CSharp.Output.Models.Types;
+using Microsoft.CodeAnalysis;
+
+namespace AutoRest.CSharp.LowLevel.Decorator
```
pshao25, line 15: Just curious, why it is a decorator?

pshao25, line 15: Just curious, why it is a decorator?

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <summary>
+        /// Reslove ambiguity between a DPG client method and client methods under the client with the same name as DPG client in existing codes.
+        /// </summary>
+        /// <param name="clientMethod">Client method belonging  to a DPG client.</param>
+        /// <param name="clientSourceModel">Roslyn model for a client with the same name of DPG client in existing codes.</param>
+        /// <returns>A LowLevelClientMethod which have been resolved to avoid ambiguity with the methods in the given client in existing codes.</returns>
+        private static LowLevelClientMethod ResolveClientMethod(LowLevelClientMethod clientMethod, INamedTypeSymbol clientSourceModel)
+        {
+            var clientMethodSourceModels = clientSourceModel.GetMembers().Where(s => s.Kind == SymbolKind.Method &&
+                s.Name == clientMethod.Signature.Name && s.DeclaredAccessibility == Accessibility.Public).Select(s => (IMethodSymbol)s).ToList();
```
pshao25, line 52: Does the methos has to be a Public one?

pshao25, line 52: Does the methos has to be a Public one?

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            return true;
+        }
+
+        private static bool IsSameParameter(Parameter parameter, IParameterSymbol parameterSourceModel)
+        {
+            //if (parameter.Name != parameterSourceModel.Name)
+            //{
+            //    return false;
+            //}
```
pshao25, line 150: I don't think we need to consider name either.

pshao25, line 150: I don't think we need to consider name either.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (!IsSameParameter(clientMethod.Signature.Parameters[i], clientMethodSourceModel.Parameters[i]))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private static bool IsSameParameter(Parameter parameter, IParameterSymbol parameterSourceModel)
```
pshao25, line 145: IsParameterSignatureSame?

pshao25, line 145: IsParameterSignatureSame?

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// Check if a clientMethod has the same paramters as a method from source model has.
+        /// Both methods should:
+        /// - have the same number of parameters
+        /// - each parameter has the same type
+        /// - each parameter has the same default value (e.g. optional or required)
+        /// </summary>
+        /// <param name="clientMethod"></param>
+        /// <param name="clientMethodSourceModel"></param>
+        /// <returns></returns>
+        private static bool IsSameMethod(LowLevelClientMethod clientMethod, IMethodSymbol clientMethodSourceModel)
```
pshao25, line 128: IsMethodSignatureSame?

pshao25, line 128: IsMethodSignatureSame?

pshao25:
The test case location is a little bit confusing. In my opinion, it should be in the HLC folder.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
archerzz: It's a common naming pattern when we want to modify the original model. It has something to do with the `decorator` pattern.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
archerzz: I think here we only care about public methods.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
archerzz: Ah, yes, I forgot to delete the commented coede.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
archerzz: That's a better name. I'll make the change. Thanks.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
pshao25: I didn't see its relationship with `decorator` pattern. We don't have and additional behavior added to the initial object. Your implementation is more like a helper I think.

pshao25:
> > The test case location is a little bit confusing. In my opinion, it should be in the HLC folder.
> 
> I tend not to be restricted in HLC, since theoretically the `SourceInputModel` will include any code file except for those under `Generated` folder. If you check the `MetricsAdvisor`, the HLC clients are located at the project root folder. Besides, as I mentioned before, developers could have written customization codes. So here I just put the existing codes where the Roslyn can find and parse.

I mean, for example, the namespace should not end with `_LowLevel` and the project name should not end with `_LowLevel`. In my understanding, our functionality is: starting from HLC, we generate such LLC. So it seems the test project should start from "without LLC".

ArcturusZhang:
Could mgmt plane borrow this when we need overload? For instance, for optional bag. And this should help us to automatically resolve breaking changes.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <param name="clientSourceModel">Roslyn model for a client with the same name of DPG client in existing codes.</param>
+        /// <returns>A LowLevelClientMethod which have been resolved to avoid ambiguity with the methods in the given client in existing codes.</returns>
+        private static LowLevelClientMethod ResolveClientMethod(LowLevelClientMethod clientMethod, INamedTypeSymbol clientSourceModel)
+        {
+            var clientMethodSourceModels = clientSourceModel.GetMembers().Where(s => s.Kind == SymbolKind.Method &&
+                s.Name == clientMethod.Signature.Name && s.DeclaredAccessibility == Accessibility.Public).Select(s => (IMethodSymbol)s).ToList();
+            if (clientMethodSourceModels.Count > 0)
+            {
+                return ResolveClientMethod(clientMethod, clientMethodSourceModels);
+            }
```
ArcturusZhang, line 56: We actually do not need to convert this to a list (you did not use indices).
if you want to check if this is empty, you could use `clientMethodSourceModels.Any()` which is more efficient.

Also I would suggest we could wrap the condition here to a method to increase the readibility

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
ArcturusZhang: Does the parameter name part of public API? Because our customer could always use the "named parameter pattern" even if the parameter is not optional, like this:
```csharp
var resp = client.CreateSomething(requestContent: requestContent, soemthingElse: something);
```

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
pshao25: But if there is a private method, the ambiguity still exists, right?

test/TestProjects/AmbiguousMethods-LowLevel/readme.md
```diff
@@ -0,0 +1,26 @@
...
+This test project focuses on testing the generator's capability to resolve ambiguous methods in low level.
+
+The scenario is that when growing a DPG client into an exsiting HLC client, there could be ambiguous methods between existing client and the DPG client to generate. The generator should have the capability to detect and resolve some ambiguous caes automatically.
+
+See MethodAmbiguityResolver.cs for details.
+
+
+## Structure
+
+- The two clients under the root folder are simulation of HLC clients.
```
annelo-msft, line 12: Do you mean these are the partial classes that add the convenience APIs to each of the clients?  (I was at first confused thinking these were the legacy-generated RestClients used by HLC libraries)

annelo-msft, line 12: nit: 
```suggestion
- The two clients under the root folder simulate existing HLC clients by providing "GA'ed" convenience APIs we don't want to change
```

test/TestProjects/AmbiguousMethods-LowLevel/readme.md
```diff
@@ -0,0 +1,26 @@
...
+- The two clients under `Generated` folder are the LLC codes.
```
annelo-msft, line 13:
```suggestion
- The two clients under `Generated` folder are the protocol clients
```

^^ Is this correct?

test/TestProjects/AmbiguousMethods-LowLevel/UnambiguousClient.cs
```diff
@@ -0,0 +1,203 @@
...
+namespace AmbiguousMethods_LowLevel
+{
+    /// <summary> Existing Unambiguous service client. Methods are changed to intentionally have no ambiguity with DPG. </summary>
+    public partial class UnambiguousClient
+    {
+        private readonly HttpPipeline pipeline = HttpPipelineBuilder.Build(ClientOptions.Default);
+
+        /// <param name="a"> The String to use. </param>
+        /// <param name="cancellationToken"> CancellationToken to use. </param>
+        public virtual async Task<Response> NoParamAsync(string a, CancellationToken cancellationToken = default)
```
annelo-msft, line 24: It's hard for me to imagine why we would have a convenience API on an HLC client that added an extra string parameter to the method signature.  Can you think of a reason?

test/TestProjects/AmbiguousMethods-LowLevel/AmbiguousClient.cs
```diff
@@ -0,0 +1,194 @@
...
+
+namespace AmbiguousMethods_LowLevel
+{
+    /// <summary> Existing Ambiguous service client. Methods are changed to intentionally have ambiguity with DPG. </summary>
+    public partial class AmbiguousClient
+    {
+        private readonly HttpPipeline pipeline = HttpPipelineBuilder.Build(ClientOptions.Default);
+
+        /// <param name="cancellationToken"> CancellationToken to use. </param>
+        public virtual async Task<Response> NoParamAsync(CancellationToken cancellationToken = default)
```
annelo-msft, line 23: Would it be possible to make the return values of convenience methods return Response<T> so it's easy to spot which are the convenience methods and which are the protocol methods in an APIView?

test/TestProjects/AmbiguousMethods-LowLevel/AmbiguousClient.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+            return await pipeline.ProcessMessageAsync(null, null).ConfigureAwait(false);
+        }
+
+        /// <param name="a"> The string to use. </param>
+        /// <param name="cancellationToken"> CancellationToken to use. </param>
+        public virtual Response NoParam(string a = null, CancellationToken cancellationToken = default)
```
annelo-msft, line 30: I don't understand why a parameter is being added to the `NoParam` method in the convenience layer.  We should at least also have a test case that takes no parameters.

test/TestProjects/AmbiguousMethods-LowLevel/AmbiguousClient.cs
```diff
@@ -0,0 +1,194 @@
...
+        /// <param name="optionalParamA"> The String to use. </param>
+        /// <param name="cancellationToken"> CancellationToken to use. </param>
+        public virtual Response OneOptionalParam(string a = null, string optionalParamA = null, CancellationToken cancellationToken = default)
+        {
+            return pipeline.ProcessMessage(null, null);
+        }
+
+        /// <param name="optionalParamB"> The Integer to use. </param>
+        /// <param name="cancellationToken"> CancellationToken to use. </param>
+        public virtual async Task<Response> TwoOptionalParamAsync(int? optionalParamB = null, CancellationToken cancellationToken = default)
```
annelo-msft, line 52: Similarly, shouldn't the two optional parameters method take two optional parameters?  I realize we can't change the method name, but could we add a comment to describe what we're validating with this method?

annelo-msft (commented):
I'm a little confused by some of your test cases.  What I would expect to see in an [APIView for AmbiguousClient](https://apiview.dev/Assemblies/Review/27b455955ece4499b0268e4eca4ffd02#AmbiguousMethods_LowLevel.AmbiguousClient) would be:

public virtual Response NoParam(CancellationToken cancellationToken = default);
public virtual Response NoParam(RequestContext context);

public virtual Response OneParam(string a, CancellationToken cancellationToken = default);
public virtual Response OneParam(string a, RequestContext context);

This output is also making me realize that if we require RequestContext, we will have to make optional parameters required in some cases:

public virtual Task<Response> OneOptionalParamAsync(string optionalParamA = null, CancellationToken cancellationToken = default);
public virtual Task<Response> OneOptionalParamAsync(string optionalParamA, RequestContext context);

I'm not sure we want to do this -- let me raise this with architects before we more forward with this solution.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
archerzz: I think `LowLevelClientMethod` should only be generated as `public`?

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
archerzz: @ArcturusZhang In this case, we plan to make sure the generated new method is different from the existing one. So if the new method breaks the codes, it's actually a good thing since the code above will invoke the old method.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
archerzz: If you check the other classes under `decorator` folders, you'll realize they don't strictly follow the classical `decorator pattern`. But the purpose of those classes is to decorate the existing models.

Here I follow the convention to put this class under `decorator` folder. Otherwise it could be put under the existing folder.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
archerzz: I'd prefer `IsParameterTypeSame`, since we don't check the name of the parameter.

archerzz: > > > The test case location is a little bit confusing. In my opinion, it should be in the HLC folder.
> > 
> > 
> > I tend not to be restricted in HLC, since theoretically the `SourceInputModel` will include any code file except for those under `Generated` folder. If you check the `MetricsAdvisor`, the HLC clients are located at the project root folder. Besides, as I mentioned before, developers could have written customization codes. So here I just put the existing codes where the Roslyn can find and parse.
> 
> I mean, for example, the namespace should not end with `_LowLevel` and the project name should not end with `_LowLevel`. In my understanding, our functionality is: starting from HLC, we generate such LLC. So it seems the test project should start from "without LLC".

- If the original codes are not in the `_LowLevel` namespace, then there won't be ambiguity. I could have a better name for this, but I think that won't affect the readiness.
- The `_LowLevel` is to identify the test scenario, not to describe a real customer library. I think it would be easier to locate the test project.

archerzz:
> Could mgmt plane borrow this when we need overload? For instance, for optional bag. And this should help us to automatically resolve breaking changes.

I think these are two cases:
- The case we're currently resolving in DPG is to check if I add a new method, will it have some ambiguity with the old one? Both methods will exist.
- In your case, it sounds like I want to replace the old method with the new one, will it have breaking change?

I think the algorithm for those two cases is different.

test/TestProjects/AmbiguousMethods-LowLevel/readme.md
archerzz: @annelo-msft No. Those are the simulation of existing codes.

test/TestProjects/AmbiguousMethods-LowLevel/UnambiguousClient.cs
archerzz: My fault. The intention is to add diversity in the test case. For example, if I want to add a protocol method `Test(string a, string b = null, RequestContext context = null)`, what existing methods will create ambiguity or unambiguity? That's why I intentionally add/remove/change parameters. You comment (and some below) reminds me that I give the names like `NoParam` to methods, so such changes will be confusing.

I'll update the test cases to make it more aligned to the names.

test/TestProjects/AmbiguousMethods-LowLevel/AmbiguousClient.cs
archerzz: sure.

archerzz:
> > > The test case location is a little bit confusing. In my opinion, it should be in the HLC folder.
> > 
> > 
> > I tend not to be restricted in HLC, since theoretically the `SourceInputModel` will include any code file except for those under `Generated` folder. If you check the `MetricsAdvisor`, the HLC clients are located at the project root folder. Besides, as I mentioned before, developers could have written customization codes. So here I just put the existing codes where the Roslyn can find and parse.
> 
> I mean, for example, the namespace should not end with `_LowLevel` and the project name should not end with `_LowLevel`. In my understanding, our functionality is: starting from HLC, we generate such LLC. So it seems the test project should start from "without LLC".

The project name is derived from the folder name by the `\eng\Generate.ps1`. I prefer the `-LowLevel` suffix because it describes the scenario of test cases, e.g. test ambiguity resolution during llc code gen. HLC client is just part of the assumption in the test cases.

test/TestProjects/AmbiguousMethods-LowLevel/AmbiguousClient.cs
archerzz: I change the return type from `Response` to `Response<SuccessResponse>`.

test/TestProjects/AmbiguousMethods-LowLevel/AmbiguousClient.cs
archerzz: See my comment above. I've made changes to make sure the parameter list aligned with the method name, except for cases that the method names only define the count of required parameters (e.g., `OneRequiredParam`). In those cases, I added optional parameters to test if our generator can still detect the correct ambiguity.

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
AlexanderSher: Yeah, this is one of the things in mgmt that I don't want to bring into data plane. This isn't a decorator, and what is more important, this logic isn't optional. It has to be executed for every protocol signature that is going to be created, meaning that there is no reason to execute it after `LowLevelClientMethod` is instantiated. Instead, I'd create signature with correct set of parameters from the start, something linke:
```cs
var parameters = operation.IsLongRunning
    ? requestMethod.Parameters.Prepend(KnownParameters.WaitForCompletion).ToArray()
    : requestMethod.Parameters;

if (existingType is not null && existingType.GetMembers(requestMethod.Name).Any(s => s is IMethodSymbol method && IsAmbiguousButNotSame(method.Parameters, parameters)))
{
    parameters = parameters.Select(p => p with { DefaultValue = null }).ToArray();
}

var methodSignature = new MethodSignature(requestMethod.Name, requestMethod.Description, requestMethod.Accessibility | Virtual, returnType, null, parameters);
```

src/AutoRest.CSharp/LowLevel/Decorator/MethodAmbiguityResolver.cs
```diff
@@ -0,0 +1,170 @@
...
+            return clientMethod;
+        }
+
+        private static IEnumerable<IMethodSymbol> GetPublicMethodsFromSourceModelByName(string methodName, INamedTypeSymbol clientSourceModel)
+        {
+            return clientSourceModel.GetMembers().Where(s => s.Kind == SymbolKind.Method &&
+                s.Name == methodName && s.DeclaredAccessibility == Accessibility.Public).Select(s => (IMethodSymbol)s);
```
AlexanderSher, line 62: `DeclaredAccessibility` can have any value here, not just `Public`

archerzz: @AlexanderSher Correct. But we're generating a public method (e.g. `ClientMethod`), and the ambiguity should happen from customers codes which invoke the public method. If you have an internal/protected/private ambiguous method, it won't cause compilation error in customers codes.

Like this:　https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYZRwBsmALOgLYCGAlgHYAUZADANoC669AJwDOASnyFikgG6D0AI3QBedMwCmAd3QAhViIDcEyUTkA6ALKs4+w+gC+ye0mRpMMbcgJJJLqDQssAF34AGnRAgH55JRUAVwAbOLFPSUdJGwB6dP5aOUYAcxiAexihMIAzdABjAAt6ZjzVYHQAwpIbAAcBQoDVSp6m33R/ZiD6UM4o5WZ4xPEvYkdbIA===

archerzz: I think removing `DeclaredAccessbility == Accessiblity.Public` do no harm here. I just want to avoid over-resolving the ambiguity.

PR #2270, opened 2022-05-31
Rename body parameter using roslyn

ArcturusZhang:
# Description

Previously we are renaming some body parameter type name  during the generation. But it happens after all the models have been generated - which are `TypeProvider`s and their type cannot be changed once determined, therefore we have to do some manually replacement which I do not like very much (see the code snippet that gets deleted in `MgmtOutputLibrary`).

Since the name of models cannot be changed once we determined them in the `BuildModel` method, we need to change them in the post process.

This PR moves the renaming mechanism of body parameter type names to post process, and also modifies some code in test gen to make sure test generation could catch the modifications in source code generation.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -42,7 +42,10 @@ private ModelFactoryTypeProvider(BuildContext context, IEnumerable<MethodSignatu
 
         private static MethodSignature CreateMethod(SchemaObjectType modelType)
         {
-            var ctor = modelType.SerializationConstructor.Signature;
+            var ctor = modelType.SerializationConstructor.Signature with
+            {
+                Description = $"Initializes a new instance of <see cref=\"{modelType.Declaration.Namespace}.{modelType.Declaration.Name}\"/>"
```
AlexanderSher, line 47: Why `<see cref=\"{modelType}\"/>` doesn't work?

ArcturusZhang: About this, could you please see the changes here in this PR? https://github.com/Azure/autorest.csharp/pull/2268
The changes about this in this PR is only prototype, I also have the same question, but we have quite a few exceptions about them.
You should see them in the other PR in which I extracted this changes out.
But since it changes quite a few in the old data plane packages, which causes some concerns, I am planning to limit this change to mgmt plane only.

ArcturusZhang: One of the counter case of why the simple one does not work is that it does not work for primitive types, for instance, it does not compile when you have `cref="Integer"`. If we implement that in this way, we will have the correct result: `cref="int"` (Rolsyn should be doing this transformation).
But the cref with namespace has other issues, for instance, when you have a namespace and a class with the same name, the `cref=SingletonResource.Something` will not compile.
I believe these issues are all resolved in the other PR.

Issue #2275, opened 2022-06-01
Our inheritance chooser might not be working correctly


ArcturusZhang:
Our inheritance chooser works like this:
If this model has a base class (which is a model) as defined in the swagger, we use the "exact match" mechanism to see if this base class could be replaced by one of our type-reference types.
If this model does not have base class, we use the "superset match" mechanism to get if there is a suitable base class from our candidates of type-reference types.

This mechanism has an issue, like [here](https://github.com/Azure/azure-sdk-for-net/blob/a7fa54c08c58f2fb14640f7677e82b9736957aac/sdk/compute/Azure.ResourceManager.Compute/src/Generated/VirtualMachineScaleSetVmExtensionData.cs#L15)
The `ComputeSubResourceData` only has one property `Id`, and this model has its own `type` and `name`. We would expect that this model could extends the `Azure.ResourceManager.Models.ResourceData` class. But instead, this class only inherits from the original `ComputeSubResourceData`.

I suppose we might need to make sure we also check the superset match if the exact match does not match anything. Or more aggressively, we remove exact match, only check superset match to always get the maximized base class.

PR #2276, opened 2022-06-01
Always use "proper" superset match when determining inherits of MgmtObjectType

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/2275
Regeneration: https://github.com/Azure/azure-sdk-for-net/pull/29063

# Description

If we have the following in **swagger** (demonstrated by the unmodified generated code)
```csharp
class SubResource
{
public string Id { get; }
}
class A : SubResource
{
public string Name { get; }
public string Type { get; }
public string Something { get; set; }
}
class B : A
{
public string SomethingElse {get; set;}
}
```
We will get this in our **generated code**
```csharp
// SubResource is removed because it is never used
class A : ResourceData
{
public string Something { get; set; }
}
class B : ResourceData
{
public string Something { get; set; }
public string SomethingElse {get; set;}
}
```
We no longer keep the inheritance relationship between `A` and `B` according to @m-nash 's comment in internal discussion.

And exception of removal of these inheritance relationships is the case of discriminators. We have to keep those inheritance relationship to keep the polymorphism.

This PR also introduces the ability to use customized code to set the base type to avoid future breaking changes.

# Checklist

# Ready to Land?


m-nash (approved):
Approved as long as we can resolve the breaking changes here with custom code for Azure.ResourceManager and Azure.ResourceManager.Resources https://github.com/Azure/azure-sdk-for-net/pull/29063

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -124,22 +124,65 @@ public bool ShouldNotReplaceForProperty()
...
             var typeToReplace = inheritedType?.Implementation as MgmtObjectType;
             if (typeToReplace != null)
             {
                 var match = InheritanceChooser.GetExactMatch(typeToReplace, typeToReplace.MyProperties);
                 if (match != null)
                 {
-                    inheritedType = match;
+                    return match;
                 }
             }
```
ArcturusZhang, line 178: If we always prefer the superset match result, do we still need this check? The superset match should always cover this case.

samples/Azure.Management.Storage/Generated/BlobContainerData.cs
```diff
@@ -14,7 +14,7 @@
 namespace Azure.Management.Storage
 {
     /// <summary> A class representing the BlobContainer data model. </summary>
-    public partial class BlobContainerData : AzureEntityResource
+    public partial class BlobContainerData : ResourceData
```
fengzhou-msft, line 17: In this case, `AzureEntityResource` has one more property (etag) than `ResourceData`. Just want to double check that we are fine with such inheritance changes? [AzureEntityResource](https://github.com/Azure/azure-rest-api-specs/blob/33753f31627bd92d2473e4ce49a3656e392d2a47/specification/common-types/resource-management/v3/types.json#L38-L55) is actually a common type (storage refers it from commom type as well), but we suppressed it in ResourceManager. Assuming we want to add it back, then this would inherit from `AzureEntityResource` again, which seems whimsical to me. Are we able to change the behavior to only replace the original base type when the original one has less properties than (or a subset of ) the common type?

fengzhou-msft (approved):

samples/Azure.Management.Storage/Generated/BlobContainerData.cs
fengzhou-msft: `AzureEntityResource` is suppressed as a common type to avoid too many resource base types and keep the resource inheritance relatively simple. Here the change actually lines with that purpose.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
ArcturusZhang: Since the superset match is checking "proper" superset, therefore the exact match is still needed

Issue #1706, opened 2021-11-19
[MQ] Unify OperationInternal from Azure.Core and autorest


AlexanderSher:
Right now, we have 6 different implementations of the `OperationInternal` logic. Prototyping grow-up story for [LRO in LLC](https://github.com/Azure/autorest.csharp/pull/1661) has shown that we may require two more. 

Also, implementation of `OperationInternal` and `OperationInternals` aren't synchronized and users may get different results while switching from handwritten to codegen'ed internals.

There is also a naming problem as well.

PR #2277, opened 2022-06-01
Remove dependencies on OperationInternals and OperationInternals<T>

AlexanderSher:
Part of the work to [unify LRO](https://github.com/Azure/autorest.csharp/issues/1706)
Uses `OperationInternal` and `OperationInternal<T>` instead of `OperationInternals` and `OperationInternals<T>`.

### DIFF
--- a/test/TestServerProjects/lro/Generated/LRORetrysDelete202Retry200Operation.cs
+++ b/test/TestServerProjects/lro/Generated/LRORetrysDelete202Retry200Operation.cs
@@ -26,7 +26,7 @@ namespace lro
 
         internal LRORetrysDelete202Retry200Operation(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, Request request, Response response)
         {
-            var nextLinkOperation = NextLinkOperationImplementation.Create(pipeline, request.Method, request.Uri.ToUri(), response, OperationFinalStateVia.Location);
+            IOperation nextLinkOperation = NextLinkOperationImplementation.Create(pipeline, request.Method, request.Uri.ToUri(), response, OperationFinalStateVia.Location);
             _operation = new OperationInternal(clientDiagnostics, nextLinkOperation, response, "LRORetrysDelete202Retry200Operation");
         }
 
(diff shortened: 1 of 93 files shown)


annelo-msft (commented):
1. Should we delete the unused types?
2. Could you please provide a PR that shows the output of the generator on the azure-sdk-for-net repo?

AlexanderSher:
1. That would be a separate PR.
2. https://github.com/Azure/azure-sdk-for-net/pull/29071

AlexanderSher:
1. To avoid any issues that can be caused by references to these types in azure-sdk-for-net . When all generated types are updated, I can safely clean-up hand-written code in `azure-sdk-for-net` and then remove these types from `autorest.csharp` repo.
2. Converted to draft

test/TestServerProjects/lro/Generated/LRORetrysDelete202Retry200Operation.cs
```diff
@@ -26,17 +26,20 @@ protected LRORetrysDelete202Retry200Operation()
...
         {
-            _operation = new OperationInternals(clientDiagnostics, pipeline, request, response, OperationFinalStateVia.Location, "LRORetrysDelete202Retry200Operation");
+            IOperation nextLinkOperation = NextLinkOperationImplementation.Create(pipeline, request.Method, request.Uri.ToUri(), response, OperationFinalStateVia.Location);
+            _operation = new OperationInternal(clientDiagnostics, nextLinkOperation, response, "LRORetrysDelete202Retry200Operation");
         }
 
         /// <inheritdoc />
-        public override string Id => _operation.Id;
+#pragma warning disable CA1822
+        public override string Id => throw new NotImplementedException();
```
pshao25, line 35: Why not throw from `OperationInternal`? If we throw here, after implementing the rehydration, we need to uptake again?

archerzz: because if you look at the `OperationInternals` which is the previous implementation we use, you'll find this is the exact codes. This PR replaces `OperationInternals` with `OperationInternal`, basically unfold the codes of `OperationInternals` here.

The previous relationship is `{RP_Action}Operation` --> `OperationInternals` --> `OperationInternal`.
Now the relationship is `{RP_Action}Operation` --> `OperationInternal`. Some codes in `OperationInternals` are not generated in `{RP_Action}Operation`.

archerzz:
LGTM. Previously I was confused of the purpose of `OperationInternals`. Its value is just to set the 1s polling interval. Now we have the default 1s interval in `RetryAfterDelayStrategy`, then we don't need `OperationInternals`.

archerzz (approved):

test/TestServerProjects/lro/Generated/LRORetrysDelete202Retry200Operation.cs
pshao25: I mean, why not add this line to `OperationInternal`
```C#
public override string Id => throw new NotImplementedException();
```
And leave the code in XXXOperation as 
```C#
public override string Id => _operation.Id;
```

PR #2289, opened 2022-06-07
Update basic method signature in reference doc

### LINKED ISSUE #2285: doc(dpg): Updates Basic Method Signature in Reference Docs
## Updates to Reference Docs

- [ ] Change the `<param name="content">`  tag:

    From: `<param name="content"> The content to send as the body of the request. </param>`
    To: `<param name="content"> The content to send as the body of the request. Details of the request body schema are in the Remarks section below.</param>`

- [ ] Change the `<param name="context">` tag:
    From: `<param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>`
    To: `<param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>`

- [ ] Add a `<returns>` tag describing the method's return value.  If the return value is `Response` or `Task<Response>`, the text should be as follows:

    - `The response returned from the service. Details of the request body schema are in the Remarks section below.`

    - Please add `Details of the response body schema are in the Remarks section below` to the response description only when the Response contains a response body with a defined schema.

- [ ] Add an `<exception>` tag:

    `<exception cref="RequestFailedException">Service returned a non-success status code.</exception>`

pshao25:
Update SDK here: https://github.com/Azure/azure-sdk-for-net/pull/29179
Fix #2285


src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -498,13 +498,21 @@ private static CodeWriter.CodeWriterScope WriteClientMethodDeclaration(CodeWrite
...
 
-            writer.WriteMethodDocumentation(methodSignature);
+            WriteMethodDocumentation(writer, methodSignature);
             WriteSchemaDocumentationRemarks(writer, clientMethod.OperationSchemas);
             var scope = writer.WriteMethodDeclaration(methodSignature);
             writer.WriteParametersValidation(methodSignature.Parameters);
             return scope;
         }
 
+        private static void WriteMethodDocumentation(CodeWriter codeWriter, MethodSignature methodSignature)
```
chunyu3, line 508: Do we need to add a new function? Can we update writer.WriteMethodDocumentation function to include the additional part? writer.WriteMethodDocumentation is to write basic document for a method, all the content of basic document will be implemented in that function.

archerzz: `CodeWriter` is commonly shared by mgmt, HLC, andn LLC. The change here is only for LLC, we'd better keep it off the `CodeWriter`.

archerzz (approved):

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
pshao25: I'm afraid not. The original method is used in a lot of places. I believe this change is only for DPG.

annelo-msft (commented):
@pshao25, would you please provide a link to a PR that shows the output of these changes on the azure-sdk-for-net repo?

pshao25:
> @pshao25, would you please provide a link to a PR that shows the output of these changes on the azure-sdk-for-net repo?

Yes, it is here: https://github.com/Azure/azure-sdk-for-net/pull/29179

annelo-msft (commented):
The output of this is correct on the azure-sdk-for-net repo (https://github.com/Azure/azure-sdk-for-net/pull/29179).  I will let someone from the Shanghai team confirm the generator implementation.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -63,7 +63,9 @@ public DocumentTranslationClient(string endpoint, AzureKeyCredential credential,
         /// <summary> Returns the translation status for a specific document based on the request Id and document Id. </summary>
         /// <param name="id"> Format - uuid.  The batch id. </param>
         /// <param name="documentId"> Format - uuid.  The document id. </param>
-        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <response> The response returned from the service. Details of the request body schema are in the Remarks section below. </response>
```
annelo-msft, line 68:
```suggestion
        /// <response> The response returned from the service. Details of the response body schema are in the Remarks section below. </response>
```

annelo-msft, line 68: As discussed in yesterday's meeting, please add `Details of the response body schema are in the Remarks section below.` to the response description only when the Response contains a response body with a defined schema.

annelo-msft (commented):
I was incorrect in my comment here: https://github.com/Azure/autorest.csharp/pull/2289#pullrequestreview-1000522406

I've left a couple more comments that will need to be resolved prior to completion of this PR.

archerzz (approved):

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -1252,10 +1284,11 @@ IEnumerable<Page<BinaryData>> CreateEnumerable(string nextLink, int? pageSizeHin
...
-        /// <param name="content"> The content to send as the body of the request. </param>
+        /// <param name="content"> The content to send as the body of the request. Details of the request body schema are in the Remarks section below. </param>
         /// <param name="contentType"> Body Parameter content-type. Allowed values: &quot;application/*+json&quot; | &quot;application/json&quot; | &quot;text/json&quot;. </param>
-        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="content"/> is null. </exception>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
```
annelo-msft, line 1291: This is missing a \<param\> tag for `response`.

annelo-msft (commented):
Is there a way to add tests to validate that the requirements have been addressed to prevent regressions?  For example, in the last change, the `<response>` tag was removed, even though it is listed as one of the requirements in the issue.  It would be ideal to have tests that catch this so humans don't have to.

annelo-msft (changes requested):
`<response>` tag is missing in some cases.

pshao25:
> `<response>` tag is missing in some cases and should always be there, per the description in #2285

This is because of this https://github.com/Azure/autorest/issues/4561. Modelfour generate `Response` instead of `SchemaResponse`.

annelo-msft (approved):
Agree that `<response>` block is not needed in the missing case.

lirenhe (approved):

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
```diff
@@ -60,7 +60,9 @@ public PagingClient(AzureKeyCredential credential, Uri endpoint, PagingClientOpt
         }
 
         /// <summary> A paging operation that must return result of the default &apos;value&apos; node. </summary>
-        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns cref="Response"> The response returned from the service. Details of the response body schema are in the Remarks section below. </returns>
```
annelo-msft, line 65: `GetNoItemNamePages` returns an `AsyncPageable<BinaryData>` rather than a `Response`.

test/TestServerProjectsLowLevel/lro/Generated/LROsClient.cs
```diff
@@ -58,8 +58,10 @@ public LROsClient(AzureKeyCredential credential, Uri endpoint, AutoRestLongRunni
...
-        /// <param name="content"> The content to send as the body of the request. </param>
-        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <param name="content"> The content to send as the body of the request. Details of the request body schema are in the Remarks section below. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns cref="Response"> The response returned from the service. Details of the response body schema are in the Remarks section below. </returns>
```
annelo-msft, line 64: This method returns a `Task<Operation<BinaryData>>`, not a Response.

annelo-msft (dismissed):
We should correctly indicate the return value of the method in the `<returns>` block.

samples/Azure.AI.DocumentTranslation/Generated/DocumentTranslationClient.cs
```diff
@@ -803,7 +827,9 @@ public virtual Response GetSupportedStorageSources(RequestContext context = null
         /// <param name="createdDateTimeUtcStart"> the start datetime to get items after. </param>
         /// <param name="createdDateTimeUtcEnd"> the end datetime to get items before. </param>
         /// <param name="orderBy"> the sorting query for the collection (ex: &apos;CreatedDateTimeUtc asc&apos;, &apos;CreatedDateTimeUtc desc&apos;). </param>
-        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns cref="AsyncPageable{BinaryData}"> The response returned from the service. Details of the response body schema are in the Remarks section below. </returns>
```
annelo-msft, line 832: The text should reflect the return value of the method.  I recommend:

The <see cref="AsyncPageable{T}"/> from the service containing a list of <see cref="BinaryData"/> objects.  Details of the body schema for each item in the collection are in the Remarks section below.

test/TestServerProjectsLowLevel/paging/Generated/PagingClient.cs
```diff
@@ -60,7 +60,9 @@ public PagingClient(AzureKeyCredential credential, Uri endpoint, PagingClientOpt
         }
 
         /// <summary> A paging operation that must return result of the default &apos;value&apos; node. </summary>
-        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns> The <![CDATA[AsyncPageable<BinaryData>]]> from the service containing a list of <![CDATA[BinaryData]]> objects. Details of the body schema for each item in the collection are in the Remarks section below. </returns>
```
annelo-msft, line 65: We should not use CDATA.  This should only be used in rare circumstances where we cannot use standard tags.  If we cannot use the `<see/>` tag for these, let's remove those tags and file an issue to come back to it later.

test/TestServerProjectsLowLevel/lro/Generated/LROsClient.cs
```diff
@@ -58,8 +58,10 @@ public LROsClient(AzureKeyCredential credential, Uri endpoint, AutoRestLongRunni
...
-        /// <param name="content"> The content to send as the body of the request. </param>
-        /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <param name="content"> The content to send as the body of the request. Details of the request body schema are in the Remarks section below. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns> The <![CDATA[Operation<BinaryData>]]> from the service that will contain a <![CDATA[BinaryData]]> object once the asynchronous operation on the service has completed. Details of the body schema for the operation's final value are in the Remarks section below. </returns>
```
annelo-msft, line 64: Same comment re: CDATA.

chunyu3 (approved):

chunyu3:
https://github.com/Azure/azure-sdk-for-net/pull/29179 contains previous change. 
Would you please refresh RPs in the latest azure-sdk-for-net, so that we can check the modification affected by your only changes?

PR #2303, opened 2022-06-10
Fix serialization of common type

fengzhou-msft:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -160,6 +161,37 @@ private bool GetReadOnly(PropertyInfo property)
...
+                //ReferenceType for inheritance do not have public constructor, and currently there are ResourceData, TrackedResourceData, WritableResourceData.
+                //We treat Id, Name, ResourceType, Location in ResourceData/TrackedResourceData as required.
+                if (!property.DeclaringType!.Name.Equals("WritableResourceData", StringComparison.Ordinal))
+                {
+                    if (property.Name == "Id" ||
+                        property.Name == "Name" ||
+                        property.Name == "ResourceType" ||
+                        property.Name == "Location")
+                        return true;
+                }
```
fengzhou-msft, line 178: Although the swagger definition does not define id, name, type of ResourceData as required, but I believe in real cases they should be required and this avoids breaking changes in GAed libraries.

src/AutoRest.CSharp/Mgmt/Decorator/PropertyMatchDetection.cs
```diff
@@ -98,9 +98,14 @@ internal static bool DoesPropertyExistInParent(Type sourceType, MgmtObjectType t
...
-                if (childProperty.Declaration.Name.EndsWith("Type"))
+                // If exact property name match fails, try to match ResourceType with Type, match ManagedServiceIdentityType and SystemAssignedServiceIdentityType with Type or XXType.
+                if (childProperty.Declaration.Name.Equals("Type", StringComparison.Ordinal))
                 {
-                    parentProperty = parentDict.FirstOrDefault(p => p.Key.EndsWith("Type")).Value;
+                    parentProperty = parentDict.FirstOrDefault(p => p.Key.EndsWith("Type", StringComparison.Ordinal)).Value;
+                }
+                else if (childProperty.Declaration.Name.EndsWith("Type", StringComparison.Ordinal))
+                {
+                    parentProperty = parentDict.FirstOrDefault(p => p.Key.EndsWith("Type", StringComparison.Ordinal) && !p.Key.Equals("ResourceType", StringComparison.Ordinal)).Value;
```
ArcturusZhang, line 108: I have a few concerns about this piece of code (not from this PR)
This matching logic is kind of fragile, we are matching things if they ends with the word `Type`, would this matches something that are totally irrelevant? What if the class we are trying to compare has two different properties ending with the word `Type`?

Before this causes any issues, we might need to change this to compare the field name which this property serializes into

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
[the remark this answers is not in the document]
fengzhou-msft: The hard-code logic is actually for ResourceData/TrackedResourceData. Based on swagger definition, none of id, name, type is required and there is no public constructor which takes all required properties as parameters in the ReferenceTypes. Therefore, by default every property in ReferenceTypes is non-required. But since Azure.ResourceManager is GAed and in real life it seems reasonable to make id, name, type required in ResourceData/TrackedResourceData. So I added the hard-code logic here.

src/AutoRest.CSharp/Mgmt/Decorator/PropertyMatchDetection.cs
fengzhou-msft: The original code caused issue as I didn't expect the method was used separately in [superset match](https://github.com/Azure/autorest.csharp/blob/15c1e6cee6998c3c0d3088231804e007f5fa3797/src/AutoRest.CSharp/Mgmt/Decorator/InheritanceChooser.cs#L93https://github.com/Azure/autorest.csharp/blob/15c1e6cee6998c3c0d3088231804e007f5fa3797/src/AutoRest.CSharp/Mgmt/Decorator/InheritanceChooser.cs#L93). I thought it was only used in exact match and this fuzzy match relies on the checking of property count match and exact match of other properties. The change here removed the fuzzy match for superset match where the only property ending with Type is ResourceType. 

As for the field name match, we do not have that information for ReferenceTypes like ResourceData in the SDK code as we don't even have a serialization class for it. 

In general, I agree that the property match logic needs to be enhanced, it has too many patches now and could break with new corner cases. One potential improvement is that we should remove a property from parentDict after it's matched to avoid matching two different properties ending with the word Type with the same parent property as you mentioned.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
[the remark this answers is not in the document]
fengzhou-msft: I think you may have misunderstood the purpose of this piece of code, it's not for generating common types in ResourceManager (ResourceData and TrackedResourceData are all customized code). It's for building the SystemObjectType to replace the original type or use as base type in a model of some RP.  Here we build SystemObjectType from the common type models in ResourceManager, not from swagger. The challenge here is that we need to tell generator about whether the properties are required based on the ResourceData/TrackedResourceData classes themselves.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -160,6 +161,37 @@ private bool GetReadOnly(PropertyInfo property)
             return property.GetSetMethod() == null;
         }
 
+        private bool GetRequired(PropertyInfo property)
+        {
+            var publicCtor = property.DeclaringType?.GetConstructors().Where(c => c.IsPublic).FirstOrDefault();
```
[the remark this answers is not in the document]
fengzhou-msft: Updated

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -160,6 +161,37 @@ private bool GetReadOnly(PropertyInfo property)
...
+                    if (property.Name == "Id" ||
+                        property.Name == "Name" ||
+                        property.Name == "ResourceType" ||
+                        property.Name == "Location")
+                        return true;
+                }
+                return false;
+            }
+            return publicCtor.GetParameters().Any(param => param.Name?.Equals(property.Name, StringComparison.OrdinalIgnoreCase) == true && param.GetType() == property.GetType());
+        }
```
ArcturusZhang, line 182: I still do not think this makes sense - taking ResourceData and TrackedResourceData as an example: since they could match with an optional of SystemData, should this mean everything except systemData is required?
This could work for all its properties - except for `Tags`, but it is a collection, what would it change if it is required? I believe nothing, since it has an initializer and always has a value when serialized

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -160,6 +161,37 @@ private bool GetReadOnly(PropertyInfo property)
             return property.GetSetMethod() == null;
         }
 
+        private bool GetRequired(PropertyInfo property)
+        {
+            var publicCtor = property.DeclaringType?.GetConstructors().Where(c => c.IsPublic).OrderBy(c => c.GetParameters().Count()).FirstOrDefault();
```
Yao725, line 166: I believe we might have several ctors in the future as we might need to avoid the breaking changes. So is there any reason that makes us just use the ctor with the least number of parameters here?

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
fengzhou-msft: The serialization code for the derived classes of ResourceData and TrackedResourceData will change based on whether the property is required. For instance, if Tags is treated as required, the serializer will simply write the tag property and values without checking whether it exists in the requests. Services handle null values differently, so it might cause issue with the previous assumption that everything is required.

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
fengzhou-msft: Here it's for the constructors of common type models, and there is only a public constructor of all required parameters today. I'm not sure about the likelihood of adding new constructors for them as currently it does not take any optional parameters (unlike track 1). But if we add a new required parameter, then it's a breaking change in swagger itself and I guess we're not able to make compatible customized changes for that in the same model. We probably need to use a new model name like XXV2 with the new required parameter. If we will add optional parameters, then I believe all the required parameters need to be there, too, therefore it's the ctor with the least number of parameters that contains the list of all required parameters.

ArcturusZhang (approved):

PR #2330, opened 2022-06-20
Limit the scope of the uri type change

Yao725:
# Description
<i>Previously we will change the schema type to `Uri` if the property's name ends with `Uri`. This works well when the property is a string. However, there are some edge cases that a property uses Uri as the name suffix but is defined as `Object` type. Considering that all the properties marked as `Object` type will share the same schema, this change will result in that all the properties' type is changed to `Uri`, which is unexpected and wrong.</i>

# Checklist

# Ready to Land?


ArcturusZhang (approved):

Yao725:
In this [PR](https://github.com/Azure/autorest.csharp/pull/2318), we introduced a configuration `format-by-name-rule` to control the type change behavior, which means that the type of properties end with `Uri` suffix will not be changed to `Uri` type automatically, so just close this PR.

Issue #2326, opened 2022-06-18
Serialized names of properties are needed in the shared models in `Azure.ResourceManager` in order to unblock test generator


ArcturusZhang:
In our test generator, we will have the raw json of a real payload, and to generate the test cases, we need to build some expression of class initialization from that raw json. Since the json payload is the serialization result of one class, we need to have a way to get the serialized name of a property in order to get the value in the payload.

This works well for all the models generated in this SDK because the code model stores the serialized names of properties.. But this has issues if one class gets replaced by a shared model from `Azure.ResourceManager`.
We also need to construct those classes from the payload, but there is not a very simple way to get the serialized name of the properties. (Similar in [this issue](https://github.com/Azure/autorest.csharp/issues/2325))

A possible solution is that we could use attribute to mark the serialized names on the properties in `Azure.ResourceManager` models.

ArcturusZhang:
We decide to add a mapping of the serialized names inside mgmt target.
Anytime we add a new shared model to resourcemanager, we need to add the new mappings as well

PR #2346, opened 2022-06-21
add a mapping for the serialized names of system object types

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/2326

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -156,29 +166,21 @@ protected override IEnumerable<ObjectTypeConstructor> BuildConstructors()
...
-                if (property.Name == "Tags")
-                    return false;
-                return true;
-            }
-            return publicCtor.GetParameters().Any(param => param.Name?.Equals(property.Name, StringComparison.OrdinalIgnoreCase) == true && param.GetType() == property.GetType());
+            var dict = ReferenceClassFinder.GetPropertyMetadata(SystemType);
+            if (dict.TryGetValue(property.Name, out var metadata))
+                return metadata.Required;
+
+            return true;
```
fengzhou-msft, line 183: We still need to check publicCtor for PropertyReferenceType to decide whether the property is required although for PropertyReferenceType, the required attribute is never used in serialization of a certain RP as we just use JsonSerializer.Serialize/JsonSerializer.Deserialize to use the method from PropertyReferenceType itself.

src/AutoRest.CSharp/Mgmt/Decorator/ReferenceClassFinder.cs
```diff
@@ -23,6 +25,57 @@ public class ReferenceClassFinder
...
+                ["PrincipalId"] = new PropertyMetadata("principalId"),
+                ["TenantId"] = new PropertyMetadata("tenantId"),
+                ["ManagedServiceIdentityType"] = new PropertyMetadata("type"),
+                ["UserAssignedIdentities"] = new PropertyMetadata("userAssignedIdentities"),
+            },
+            [typeof(SystemAssignedServiceIdentity)] = new()
+            {
+                ["PrincipalId"] = new PropertyMetadata("principalId"),
+                ["TenantId"] = new PropertyMetadata("tenantId"),
+                ["SystemAssignedServiceIdentityType"] = new PropertyMetadata("type"),
```
fengzhou-msft, line 64: Only type is required. 

You may consider make default to be optional as most properties in a `PropertyReferenceType` model is optional? Then we need to set all `ReferenceType` proeprties explicitly since the default for them should be required and you've already done that. Don't forget that we have a `TypeReferenceType` `ReponseError` in Azure.Core.

ArcturusZhang: Since most of the things are just following the rules, and I am tired to add all of them here, (it is hard to maintain if we have a really long list), I am planning:
1. make the required default to false in PropertyMetadata, and assign it to true in some special cases.
2. if the mapping of a model is so normal (just propertyName: ToCamelCase(propertyName)), I will not include that in this list. I will add a function to construct this mapping from a type. Whether a property is required is fetched from its public constructor.

fengzhou-msft (approved):

ArcturusZhang:
The regeneration in azure-sdk-for-net shows no changes

PR #2392, opened 2022-06-29
Update the mgmt. readme

ArthurMa1978:


ArcturusZhang (commented):
We should also include some documentation about the formats available

ArcturusZhang (approved):

PR #2397, opened 2022-06-29
Fix #2323: [Bug] Api-Version not being attached during pagination

### LINKED ISSUE #2323: [Bug] Api-Version not being attached during pagination
Api-Version is not being added as a query parameter when building next page request

The request for first page works as expected (api-version is being added)

MicrosoftTeams-image (4)
.

The request for the next page is not adding api-version, instead it relies solely on the nextLink that is being sent back from the server.

MicrosoftTeams-image (5)

Resources:-
- Link to arch board decision: https://github.com/Azure/azure-sdk/issues/3817#issuecomment-1033226508
- Autorest configuration used: https://github.com/Azure/azure-sdk-for-net/blob/feature/jobrouter-repo-setup/sdk/communication/Azure.Communication.JobRouter/src/autorest.md
- Swagger file used to generate client: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/communication/data-plane/JobRouter/preview/2021-10-20-preview2/communicationservicejobrouter.json
- Operation pasted in the screenshot: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/communication/data-plane/JobRouter/preview/2021-10-20-preview2/communicationservicejobrouter.json#L1529-L1615

AlexanderSher:
This PR adds api-version parameter to the next page methods


archerzz:
LGTM. @m-nash @AlexanderSher I've created a preview PR on `azure-sdk-for-net`: https://github.com/Azure/azure-sdk-for-net/pull/29795
There are too many changes including both mgmt and data planes. I randomly checked some changes, and I think they are good.

m-nash (approved):
Lets double check the test cases will run in playback without the apiversion in the loop calls, otherwise looks good

PR #2405, opened 2022-07-01
Fix #2385: Refactor the HLC generator request error path to be consistent with DPG

### LINKED ISSUE #2385: Refactor the HLC generator request error path to be consistent with DPG
# Summary
In order to simplify the process by which the Azure.Core pipeline can choose to buffer response content we need a consistent mechanism to indicate when a response should throw `RequestFailedException`. DPG already has such a mechanism, wish is the ResponseClassifier and the resulting result of the `Response.IsError` property.

# More Details
This is related to the need to refactor customization of error messages for RequestFailedException such that it is now handled in the RequestFailedException constructor in cooperation with customizations made to the client.

See https://github.com/Azure/azure-sdk-for-net/pull/28466 for details of that change.

AlexanderSher:


AlexanderSher:
Changes in azure-sdk-for-net repo can be added here: https://github.com/Azure/azure-sdk-for-net/pull/29625

jsquire (approved):
Based on approvals for the linked PR in the .NET repository, approving here to unblock when ready.

PR #2408, opened 2022-07-04
fix mock test generation

changlong-liu:
1. fix generation on "duration" properties
2. fix approach of getting tenantResource in testcase


src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -225,19 +226,20 @@ private static CodeWriter AppendRawDictionary(this CodeWriter writer, Dictionary
             return writer;
         }
 
-        private static CodeWriter AppendRawValue(this CodeWriter writer, Type type, object? rawValue) => rawValue switch
+        private static CodeWriter AppendRawValue(this CodeWriter writer, Type type, object? rawValue, AllSchemaTypes? schemaType=null) => rawValue switch
```
ArcturusZhang, line 229: Please format this file

ArcturusZhang (approved):

PR #2409, opened 2022-07-05
Add tests to validate the generated authentication functions

### LINKED ISSUE #2272: Add tests to validate that generated authentication functions correctly
We have tests that validate that [this functionality](https://github.com/Azure/autorest.csharp/issues/2222) is generated correctly:

> We have test coverage for [this](https://github.com/Azure/autorest.csharp/tree/feature/v3/test/TestProjects/SecurityDefinition-LowLevel) sample [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServerLowLevel.Tests/SecurityDefinitionTests.cs) which has both "oauth2" and "apiKey" security schemes.

We need to add tests to validate that the generated code works as expected.

pshao25:
Fix #2272


test/TestServerProjectsLowLevel/security-aad/AutorestSecurityAadClient.cs
```diff
@@ -0,0 +1,87 @@
...
+            options ??= new AutorestSecurityAadClientOptions();
+
+            ClientDiagnostics = new ClientDiagnostics(options, true);
+            _tokenCredential = credential;
+            _pipeline = HttpPipelineBuilder.Build(options, Array.Empty<HttpPipelinePolicy>(), new HttpPipelinePolicy[] { new MockBearerTokenAuthenticationPolicy(_tokenCredential, AuthorizationScopes) }, new ResponseClassifier());
+            _endpoint = endpoint;
+        }
+
+        // Only for bypassing HTTPS check purpose
+        public class MockBearerTokenAuthenticationPolicy : BearerTokenAuthenticationPolicy
```
archerzz, line 36: Can we move this out of the generated codes. And inject this mock policy when initializing the client? Thus we don't need to maintain `[CodeGenSuppress]` which is easy to loose synchronization with latest autorest changes.

archerzz (approved):

PR #2410, opened 2022-07-05
make put operation also has required body parameters

ArcturusZhang:
Regeneration: https://github.com/Azure/azure-sdk-for-net/pull/29632

# Description

# Checklist

# Ready to Land?


samples/Azure.Management.Storage/Generated/RestOperations/BlobContainersRestOperations.cs
```diff
@@ -693,13 +693,10 @@ internal HttpMessage CreateCreateOrUpdateImmutabilityPolicyRequest(string subscr
...
             }
             request.Headers.Add("Accept", "application/json");
-            if (data != null)
-            {
-                request.Headers.Add("Content-Type", "application/json");
-                var content = new Utf8JsonRequestContent();
-                content.JsonWriter.WriteObjectValue(data);
-                request.Content = content;
-            }
+            request.Headers.Add("Content-Type", "application/json");
```
ArthurMa1978, line 696: throw exception if data is null?

src/AutoRest.CSharp/Mgmt/Decorator/Transformer/SchemaFormatByNameTransformer.cs
```diff
@@ -43,7 +43,7 @@ internal static void Update()
         private IEnumerable<Schema> allGeneralSchemas;
         private IEnumerable<OperationGroup> allOperationGroups;
         private IReadOnlyDictionary<string, string> allFormatByNameRules;
-        private Dictionary<Guid, string> schemaCache = new Dictionary<Guid, string>();
+        private Dictionary<Schema, string> schemaCache = new();
```
ArthurMa1978, line 46: Not sure here the `Schema` as key would drop down the perf

ArthurMa1978 (approved):

src/AutoRest.CSharp/Mgmt/Decorator/Transformer/SchemaFormatByNameTransformer.cs
ArcturusZhang: Absolutely no. We are doing that a lot (in MgmtOutputLibrary). These factors impact performance very minorly

samples/Azure.Management.Storage/Generated/RestOperations/BlobContainersRestOperations.cs
ArcturusZhang: We assert the data is not null in the corresponding client method. Therefore we no longer need to assert again

src/AutoRest.CSharp/Mgmt/Decorator/Transformer/SchemaFormatByNameTransformer.cs
Yao725, line 46: Just for curiosity, why is this change needed?

ArcturusZhang: 1. If we have to introduce an Id in schema, we should not do that in the CodeModel.cs file. We should put it in CodeModelPartial.
2. This has no point since the instance itself works as an unique Id of a schema object. We are doing that a lot in MgmtOutputLibrary, quite a few dictionaries are using the objects from code model as keys (such as Schema, Operation, etc)

PR #2411, opened 2022-07-05
Support interim status in lro

Yao725:
# Description
<i>This should be merged after https://github.com/Azure/azure-sdk-for-net/pull/29937
Fix https://github.com/Azure/azure-sdk-for-net/issues/29060
Regen PR https://github.com/Azure/azure-sdk-for-net/pull/29638
</i>

# Checklist

# Ready to Land?


src/assets/Generator.Shared/IArmOperationSource.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Azure.Core
+{
+    internal interface IArmOperationSource<T> : IOperationSource<T>
```
AlexanderSher, line 9: Only mgmt uses `IOperationSource`, so you can merge base interface into this one.

Yao725:
> @KrzysztofCwalina @tg-msft Do we have any archboard discussion about this type of LRO's?

This requirement  initially comes from storage team, you can see the details in this [issue](https://github.com/Azure/azure-sdk-for-net/issues/29060).

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/SampleArmOperationOfT.cs
```diff
@@ -43,7 +51,7 @@ internal SampleArmOperation(IOperationSource<T> source, ClientDiagnostics client
 #pragma warning restore CA1822
 
         /// <inheritdoc />
-        public override T Value => _operation.Value;
+        public override T Value => !HasValue && _source.SupportInterimStatus ? _interimValue ??= _source.CreateResult(_interimResponse, default) : _operation.Value;
```
m-nash, line 54: When I thought about this change I thought we might make it directly in the base classes.  The feature itself should default to false but if try affect how [`Value` works here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs#L157).

Off the top of my head if we add an optional param to `OperationInternal<T>` ctor and each `[RP]ArmOperation<T>` then each LRO method would know if this should be true and pass that in when constructing the Operation object otherwise the calling code would remain unchanged.

This would eliminate the need for the new interface `IArmOperationSource` and it allows everyone to take advantage of the functionality if they want to opt in.

m-nash (changes requested):

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/SampleArmOperationOfT.cs
Yao725: Added an optional parameter to `OperationInternal<T>` in https://github.com/Azure/azure-sdk-for-net/pull/29937

Yao725:
This way was discarded, will achieve this goal in another separate PR.

PR #2418, opened 2022-07-07
Fix internalizer for discriminator

Yao725:
# Description
<i>In the internalizer, we only use the deserialization method to check if the class is a base class with discriminator. And if it is, the accessibility of its derived classes will be changed to public. However, there are some base classes that are only used as input parameters, which means that they will not have a deserialization method and therefore we will incorrectly modify its derived classed to be internal. 

Regen PR: https://github.com/Azure/azure-sdk-for-net/pull/29683
</i>

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/Internalizer.cs
```diff
@@ -157,6 +157,18 @@ private static bool HasDiscriminator(BaseTypeDeclarationSyntax node, [MaybeNullW
...
+                else if (classDeclaration.HasLeadingTrivia)
+                {
+                    var syntaxTriviaList = classDeclaration.GetLeadingTrivia();
+                    if (BuilderHelpers.DiscriminatorDescFixedPart.All(syntaxTriviaList.ToFullString().Contains))
+                    {
+                        var identifierNodes = syntaxTriviaList.FirstOrDefault().GetStructure()?.DescendantNodes().OfType<XmlCrefAttributeSyntax>();
+                        identifiers = identifierNodes.Select(identifier => identifier.Cref.ToFullString()).ToImmutableHashSet();
+                        return true;
+                    }
+                }
```
ArcturusZhang, line 171: If we guaranteed that all the class with a discriminator would have the leading trivia with some "cref"s, we could just use this as a criteria and do not need to check the deserialization method. This will simplify the code in a great deal.
There are some other cref might be in the xml document, do you think we need to verify the content of that trivia?

Yao725: I think it is guaranteed that all the classes with discriminator will have this trivia. And yes, the verification of the content is also needed, I will extract the cref only from the specified line.

src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/Internalizer.cs
```diff
@@ -139,22 +139,18 @@ private static bool HasDiscriminator(BaseTypeDeclarationSyntax node, [MaybeNullW
...
-                            // get the identifiers in the switch statements, some of them are class names of the derived classes
-                            // but since we do not have the full semantic model here, we cannot directly get which of them is
-                            // therefore here we just return all of them, and see if the derived type name shows in this list
-                            var identifierNodes = switchNode.Sections.SelectMany(section => section.DescendantNodes().OfType<IdentifierNameSyntax>());
-                            identifiers = identifierNodes.Select(identifier => identifier.Identifier.ToFullString()).ToImmutableHashSet();
-                            return true;
-                        }
+                        var descendantNodes = filteredTriviaList.First().GetStructure()?.DescendantNodes().ToList();
+                        var filteredDescendantNodes = descendantNodes.Where((val, index) =>
+                            index >= descendantNodes?.FindLastIndex(node => node.ToFullString().Contains(BuilderHelpers.DiscriminatorDescFixedPart.Last())));
```
ArcturusZhang, line 150: This nested linq is really hard to understand, can we simplify this by adding some method for this?

Yao725: Good point, I added a separate method to replace this linq. And in the method also some comments are added.

ArcturusZhang (approved):

PR #2423, opened 2022-07-11
[sample generation] generate samples for apiexplorer

changlong-liu:
Samples in this PR is for incubation project apiexplorer


src/AutoRest.CSharp/MgmtTest/AutoRest/MgmtTestTarget.cs
```diff
@@ -23,25 +28,112 @@ public static void Execute(GeneratedCodeWorkspace project, CodeModel codeModel,
...
+            {
+                try
+                {
+                    var extensionWrapperTestWriter2 = new ExtensionWrapMockTestWriter(new CodeWriter(), extensionWrapperTest, library.ExtensionMockTests);
+                    extensionWrapperTestWriter2.WriteSample(testCase);
+                    var filename = GenExampleFileName(testCase);
+                    if (!generated.Contains(filename))
+                    {
+                        project.AddGeneratedFile(GenExampleFileName(testCase), extensionWrapperTestWriter2.ToString());
+                        generated.Add(filename);
```
ArcturusZhang, line 125: There here if we have the example generated with the same name, we just skip the current one. Will this make us to miss some examples?
How about we append some indices after the filename or something?

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -44,6 +48,75 @@ public virtual void Write()
...
+        public virtual void WriteSample(MockTestCase testCase)
+        {
+            generatingSample = true;
+            Regex pattern = new Regex(@".*(Azure.ResourceManager.\w+)/.*");
+            Match match = pattern.Match(Configuration.OutputFolder);
+            if (match.Success)
+            {
+                _writer.additionalComments.Add($"// Need to install {match.Groups[1].Value} v1.0.0 to execute below sample code");
+            }
+            _writer.addUsingRegion = true;
```
ArcturusZhang, line 60: If we need to change something in our code writer, please use a property instead. Changing a public field is always not recommended

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -44,6 +48,75 @@ public virtual void Write()
...
+            _writer.UseNamespace("Azure.ResourceManager");
+            _writer.UseNamespace("System");
+            _writer.UseNamespace("Azure.Identity");
+            _writer.UseNamespace("System.Threading.Tasks");
```
ArcturusZhang, line 64: Should not we import the usings automatically whenever they are used?
If a namespace is not used in our sample, do we need to use them and show them in the sample?

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -44,6 +48,75 @@ public virtual void Write()
...
+
+            _writer.Line($"#region Parameter decalarations");
```
ArcturusZhang, line 66: Maybe we could consider to add a new method in codeWriter to make sure the code writer to write regions (like CodeWriter.Scope). This could make sure we will never forget to append the `#endregion`

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -44,6 +48,75 @@ public virtual void Write()
...
+            _writer.Line($"// Below parameters are extracted from swagger example files, please use your own values to execute the API lively.");
+            _writer.Line($"ArmClient GetArmClient() => new ArmClient(new DefaultAzureCredential());");
```
ArcturusZhang, line 68:
```suggestion
            _writer.Line($"{typeof(ArmClient)} GetArmClient() => new {typeof(ArmClient)}(new DefaultAzureCredential());");
```

src/AutoRest.CSharp/MgmtTest/Models/MockTestCase.cs
```diff
@@ -79,7 +81,7 @@ private MgmtRestOperation GetRestOperationFromOperationId()
         /// <param name="requestPath"></param>
         /// <returns></returns>
         /// <exception cref="InvalidOperationException"></exception>
-        public IEnumerable<FormattableString> ComposeResourceIdentifierParameterValues(RequestPath requestPath)
+        public IEnumerable<FormattableString> ComposeResourceIdentifierParameterValues(RequestPath requestPath, bool generatingSample)
```
ArcturusZhang, line 84: I do not like the fact that we will have to check the literal values to determine whether this is a subscriptionId.
If we have to do something differently when it is a subscriptionId, how about (take either)
- instead of returning a FormattableString, we return a collection of a class when some more verbose information, we could add anything in that (for instance, `IsSubscriptionId`)
- if the extra action only is done to subscriptionId, since we already have that check inside this method, why do we do it here?

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -44,6 +48,75 @@ public virtual void Write()
...
+            //RequestPath requestPath = testCase.Carrier switch
+            //{
+            //    ResourceCollection parentCollection => parentCollection.RequestPath,
+            //    Resource parentResource => parentResource.RequestPath,
+            //    MgmtExtensions parentExtension => parentExtension.ContextualPath,
+            //    _ => throw new InvalidOperationException($"Unknown parent {testCase.Carrier.GetType()}"),
+            //};
+            RequestPath requestPath = testCase.RequestPath;
+            foreach (var fs in testCase.ComposeResourceIdentifierParameterValues(requestPath, true))
```
ArcturusZhang, line 77: Based on the changes in MockTestCase class, the extra flag `true` changes the meaning of this method, it is returning the serialized names instead of the "parameterValues" as the method names indicated, we need to
- rename the method to make it makes sense in both cases
- change the return value from FormattableString to a more verbose descriptive class or struct to contain both information of the parameter name and parameter value.

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -44,6 +48,75 @@ public virtual void Write()
...
+                    if (description is not null)
+                    {
+                        _writer.Line($";  // {description}");
+                    }
+                    else
+                    {
+                        _writer.Line($";");
+                    }
+                    parameterVariables.Add(s);
+                }
```
ArcturusZhang, line 108: The complicated calculation breaks the pattern that "we should prepare everything we want to use and use them directly in the writer". Therefore these complicated logic should be moved into the MockTestCase class

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -44,6 +48,75 @@ public virtual void Write()
...
+            }
+            _writer.Line($"#endregion");
+            _writer.Line($"");
+            _writer.Line($"#region API invocation");
+            _writer.Line($"// api-version: {testCase.RestOperation.Operation.ApiVersions.FirstOrDefault().Version}");
+            _writer.Line($"// x-ms-original-file: {testCase.Example.OriginalFile}");
+            WriteTestMethodBody(testCase);
+            _writer.Line($"#endregion");
+            generatingSample = false;
```
ArcturusZhang, line 117: This flag is really not a good desgin. We should have a separate class for writing the samples, and the class will automatically becomes the flag, therefore we do not need to maintain the state of this flag anymore

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -135,7 +208,7 @@ protected CodeWriterDeclaration WriteGetResource(Resource carrierResource, MockT
         {
             var idVar = new CodeWriterDeclaration($"{carrierResource.Type.Name}Id".ToVariableName());
             _writer.Append($"var {idVar:D} = {carrierResource.Type}.CreateResourceIdentifier(");
-            foreach (var value in testCase.ComposeResourceIdentifierParameterValues(carrierResource.RequestPath))
+            foreach (var value in testCase.ComposeResourceIdentifierParameterValues(carrierResource.RequestPath, generatingSample))
```
ArcturusZhang, line 211: Same here. If we are meant to reuse this method, this flag should be defined as a virtual property, in the writer that generates mock tests, it returns false, and in the writer that generates samples, it returns true

ArcturusZhang: And I did not get the point why when we are writing samples, this method has to return the serialized name of the parameters instead of values? Because the values are placeholders to let the customer fulfill?

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -211,27 +284,85 @@ protected void WriteTestOperation(CodeWriterDeclaration declaration, MockTestCas
...
+                _writer.Append($"{(generatingSample ? $"" : $"await ")}foreach (var _ in ");
                 WriteTestMethodInvocation(declaration, testCase);
                 using (_writer.Scope($")"))
-                { }
+                {
+                    if (generatingSample && !testCase.ClientOperation.ReturnType.IsFrameworkType && testCase.ClientOperation.ReturnType.Implementation is Resource)
+                    {
+                        _writer.UseNamespace("System.Text.Json");
+                        _writer.Line($"Console.WriteLine($\"ResourceId: {{_.Data.Id}}\");");
+                        _writer.Line($"Console.WriteLine($\"Full Resource content: {{JsonSerializer.Serialize(_.Data)}}\");");
```
ArcturusZhang, line 295: I do not think using the JsonSerializer to serialize the resource data here is a good idea, because our serializer is designed for the request body, therefore the readonly properties will not be printed out. This must introduce a big confusion to our customers if they just copy this sample somewhere else.
We could consider to write another helper method like `Print{ResourceName}` which writes the properties one by one from that resource data

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -211,27 +284,85 @@ protected void WriteTestOperation(CodeWriterDeclaration declaration, MockTestCas
...
+                        {
+                            _writer.Line($"Console.WriteLine($\"Succeed with return: {{{operation}.Value}}\");");
+                        }
+                        else if (testCase.ClientOperation.ReturnType.Arguments[0].Implementation is Resource)
+                        {
+                            _writer.UseNamespace("System.Text.Json");
+                            _writer.Line($"Console.WriteLine($\"Succeed on ResourceId: {{{operation}.Value.Data.Id}}\");");
+                            _writer.Line($"Console.WriteLine($\"Full Resource content: {{JsonSerializer.Serialize({operation}.Value.Data)}}\");");
+                        }
+                        else
```
ArcturusZhang, line 340: Once again we should not have this complicated calculation in the writer, and if this piece of code is specific to sample generation, we should not put it in the base writer.

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -211,27 +284,85 @@ protected void WriteTestOperation(CodeWriterDeclaration declaration, MockTestCas
...
+                        if (testCase.ClientOperation.ReturnType.Arguments.Length > 0 && testCase.ClientOperation.ReturnType.Arguments[0].Implementation is Resource)
+                        {
+                            _writer.Line($"Console.WriteLine($\"Succeed on ResourceId: {{{operation}.Value.Data.Id}}\");");
+                            _writer.Line($"Console.WriteLine($\"Full Resource content: {{JsonSerializer.Serialize({operation}.Value.Data)}}\");");
+                        }
+                        else
+                        {
+                            _writer.Line($"Console.WriteLine($\"Response: {{JsonSerializer.Serialize({operation}.Value)}}\");");
+                        }
+                    }
```
ArcturusZhang, line 327: We should have a flag on ClientOperation to tell you whether it is a LRO or not (including if this is a fake LRO). 
The idea is that if we have something high level, we will not depend on something low level (because low level things are complicated, processing them is usually a duplicated work and they are not as stable as the high level encapsulation)

src/AutoRest.CSharp/MgmtTest/Generation/Mock/MgmtMockTestBaseWriter.cs
```diff
@@ -211,27 +284,85 @@ protected void WriteTestOperation(CodeWriterDeclaration declaration, MockTestCas
...
+                }
             }
         }
 
         protected void WriteTestMethodInvocation(CodeWriterDeclaration declaration, MockTestCase testCase)
         {
             var clientOperation = testCase.ClientOperation;
-            var methodName = CreateMethodName(clientOperation.Name);
+            var methodName = CreateMethodName(clientOperation.Name, !generatingSample);
```
ArcturusZhang, line 357: Therefore in samples we do not use async method? May I know why?

src/AutoRest.CSharp/MgmtTest/Models/MockTestCase.cs
```diff
@@ -38,6 +38,8 @@ internal class MockTestCase
 
         public MgmtTypeProvider Carrier { get ; }
 
+        public ExampleModel Example => _example;
```
ArcturusZhang, line 41: It does not expose the ExampleModel on purpose. Therefore I do not think it is a good idea to expose this here.
If the writer needs something in this model, this MockTestCase should parse it properly and the writer could directly consume.

ArcturusZhang: On the other hand that this class is a "Mock Test" class, which is designed to store the data for generating a mock test.
If we need something similar for generating the samples, we should have a class either inherits from this, or holds an instance of MockTestCase, but in summary, we should not change something in this class to make it also has the ability to generate a sample

src/AutoRest.CSharp/readme.md
```diff
@@ -51,11 +51,14 @@ pipeline:
 
 modelerfour:
   emit-yaml-tags: true
+  include-x-ms-examples-original-file: true
+include-x-ms-examples-original-file: true
```
ArcturusZhang, line 55: Why do we have two copies of this configuration? Does modelerfour support this?

test/TestProjects/MgmtMockTest/Samples/Generated/DiskEncryptionSets_CreateOrUpdate$$Create a disk encryption set with key vault from a different subscription..cs
```diff
@@ -0,0 +1,46 @@
...
+// Below parameters are extracted from swagger example files, please use your own values to execute the API lively.
+ArmClient GetArmClient() => new ArmClient(new DefaultAzureCredential());
+var subscriptionId = Environment.GetEnvironmentVariable("SUBSCRIPTION_ID") ?? "00000000-0000-0000-0000-000000000000";
+var resourceGroupName = "myResourceGroup";  // Name of the resource group that contains the key vault.
+var diskEncryptionSetName = "myDiskEncryptionSet";  // The name of the disk encryption set that is being created. The name can't be changed after the disk encryption set is created. Supported characters for the name are a-z, A-Z, 0-9, _ and -. The maximum name length is 80 characters.
+#endregion
+
+#region API invocation
+// api-version: 2021-10-01
+// x-ms-original-file: file:///C:/ZZ/projects/codegen/autorest.csharp/test/TestProjects/MgmtMockTest/src/examples/diskEncryptionSetExamples/DiskEncryptionSet_Create_WithKeyVaultFromADifferentSubscription.json
```
ArcturusZhang, line 31: We should not use the absolute path here - this exposes something on your local, and makes the build analyze on this PR will never pass (since this value is different when regenerating in agents)

test/TestProjects/MgmtMockTest/Samples/Generated/DiskEncryptionSets_CreateOrUpdate$$Create a disk encryption set with key vault from a different subscription..cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#region Usings
```
ArcturusZhang, line 3: This line is not in the correct position

test/TestProjects/MgmtMockTest/Samples/Generated/DiskEncryptionSets_CreateOrUpdate$$Create a disk encryption set with key vault from a different subscription..cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ArcturusZhang, line 1: I am not sure if this file could compile since this file does not have a namespace definition nor a class definition.
But this clearly breaks the format and simplication functionalities in rolsyn -  some global module name is not properly trimmed off (on line 34.
If you are meant to write a file that is using the simple program functionality (I forgot the official name of it) that in a console application, when you want to write a Main function, you could omit the template code just starting from the first line of your program.
But here it might not work, because our test files are not treated as a standalone console application.

ArcturusZhang: And the filename contains an extra dot

test/TestProjects/MgmtMockTest/tests/Generated/CodeModel.yaml
```diff
@@ -7641,6 +7641,7 @@ operationGroups:
                 resourceGroupName: sample-resource-group
                 subscriptionId: 00000000-0000-0000-0000-000000000000
                 vaultName: sample-vault
+              x-ms-original-file: file:///C:/ZZ/projects/codegen/autorest.csharp/test/TestProjects/MgmtMockTest/src/examples/createVault.json
```
ArcturusZhang, line 7644: Something in the testmodeler might need change - we should not output an absolute path here (reason please see above comments)

test/TestProjects/MgmtMockTest/Samples/Generated/DiskEncryptionSets_CreateOrUpdate$$Create a disk encryption set with key vault from a different subscription..cs
```diff
@@ -0,0 +1,46 @@
...
+using Azure.ResourceManager.Models;
+using Azure.ResourceManager.Resources;
+using MgmtMockTest;
+using MgmtMockTest.Models;
+#endregion
+
+#region Parameter decalarations
+// Below parameters are extracted from swagger example files, please use your own values to execute the API lively.
+ArmClient GetArmClient() => new ArmClient(new DefaultAzureCredential());
+var subscriptionId = Environment.GetEnvironmentVariable("SUBSCRIPTION_ID") ?? "00000000-0000-0000-0000-000000000000";
```
ArcturusZhang, line 24: Since we have the comment "Below parameters are extracted from swagger example files, please use your own values to execute the API lively." we really do not need to put the extra GetEnvironmentVariable thing, we could just 
```suggestion
var subscriptionId = "00000000-0000-0000-0000-000000000000";
```
and expect our customer reads the comment above and change it to their own subscriptionId, either hard code here or getting it from environment variables.
This also simplify our generator a few

test/TestProjects/MgmtMockTest/Samples/Generated/DiskEncryptionSets_CreateOrUpdate$$Create a disk encryption set with key vault from a different subscription..cs
```diff
@@ -0,0 +1,46 @@
...
+using Azure.Identity;
+using Azure.ResourceManager;
+using Azure.ResourceManager.Models;
+using Azure.ResourceManager.Resources;
+using MgmtMockTest;
+using MgmtMockTest.Models;
+#endregion
+
+#region Parameter decalarations
+// Below parameters are extracted from swagger example files, please use your own values to execute the API lively.
```
ArcturusZhang, line 22: This line we might need to put it under the ArmClient

test/TestProjects/MgmtMockTest/Samples/Generated/DiskEncryptionSets_CreateOrUpdate$$Create a disk encryption set with key vault from a different subscription..cs
```diff
@@ -0,0 +1,46 @@
...
+ArmClient GetArmClient() => new ArmClient(new DefaultAzureCredential());
```
ArcturusZhang, line 23: Why we are defining a function here instead of a local variable?

test/TestProjects/MgmtMockTest/Samples/Generated/DiskEncryptionSets_CreateOrUpdate$$Create a disk encryption set with key vault from a different subscription..cs
```diff
@@ -0,0 +1,46 @@
...
+var subscriptionId = Environment.GetEnvironmentVariable("SUBSCRIPTION_ID") ?? "00000000-0000-0000-0000-000000000000";
+var resourceGroupName = "myResourceGroup";  // Name of the resource group that contains the key vault.
+var diskEncryptionSetName = "myDiskEncryptionSet";  // The name of the disk encryption set that is being created. The name can't be changed after the disk encryption set is created. Supported characters for the name are a-z, A-Z, 0-9, _ and -. The maximum name length is 80 characters.
+#endregion
+
+#region API invocation
+// api-version: 2021-10-01
+// x-ms-original-file: file:///C:/ZZ/projects/codegen/autorest.csharp/test/TestProjects/MgmtMockTest/src/examples/diskEncryptionSetExamples/DiskEncryptionSet_Create_WithKeyVaultFromADifferentSubscription.json
+// Example: Create a disk encryption set with key vault from a different subscription.
```
ArcturusZhang, line 32: We should not need this line.

test/TestProjects/MgmtMockTest/Samples/Generated/DiskEncryptionSets_CreateOrUpdate$$Create a disk encryption set with key vault from a different subscription..cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Text.Json;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Identity;
+using Azure.ResourceManager;
+using Azure.ResourceManager.Models;
+using Azure.ResourceManager.Resources;
+using MgmtMockTest;
+using MgmtMockTest.Models;
+#endregion
```
ArcturusZhang, line 19: Just out of curiosity, why do we need these regions? Are these related with something in the api explorer?

PR #2425, opened 2022-07-12
Put swagger summary into comment summary

### LINKED ISSUE #2388: Swagger "summary" is ignored during autorest generation
I have a rather long method description in "description" property. I wanted to use a shorter summary in "summary" property, but autorest ignores "summary" and only uses "description" property value.

Use "summary" property (when present) for "summary" source code documentation and use the "description" property value as "remark" for example. In situation where "summary" swagger element is missing, fall back to "description" property value.

pshao25:
Put swagger summary into comment summary, put swagger description into comment remarks.

SDK changes: https://github.com/Azure/azure-sdk-for-net/pull/29798

Fix #2388


src/AutoRest.CSharp/Common/Output/Models/Requests/RestClientMethod.cs
```diff
@@ -12,13 +12,14 @@ namespace AutoRest.CSharp.Output.Models.Requests
 {
     internal class RestClientMethod
     {
-        public RestClientMethod(string name, string? description, CSharpType? returnType, Request request, Parameter[] parameters, Response[] responses, DataPlaneResponseHeaderGroupType? headerModel, bool bufferResponse, string accessibility, Operation operation, RequestConditionHeaders conditionRequestFlag = RequestConditionHeaders.None)
+        public RestClientMethod(string name, string? description, string? summary, CSharpType? returnType, Request request, Parameter[] parameters, Response[] responses, DataPlaneResponseHeaderGroupType? headerModel, bool bufferResponse, string accessibility, Operation operation, RequestConditionHeaders conditionRequestFlag = RequestConditionHeaders.None)
```
archerzz, line 15: same sequence problem

src/AutoRest.CSharp/Common/Output/Models/ConstructorSignature.cs
```diff
@@ -6,5 +6,5 @@
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal record ConstructorSignature(string Name, string? Description, MethodSignatureModifiers Modifiers, IReadOnlyList<Parameter> Parameters, ConstructorInitializer? Initializer = null) : MethodSignatureBase(Name, Description, Modifiers, Parameters);
+    internal record ConstructorSignature(string Name, string? Description, string? Summary, MethodSignatureModifiers Modifiers, IReadOnlyList<Parameter> Parameters, ConstructorInitializer? Initializer = null) : MethodSignatureBase(Name, Description, Summary, Modifiers, Parameters);
```
archerzz, line 9: `Summary` should be before `Description` in the sequence. See the official document and example: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d312-remarks

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal record MethodSignature(string Name, string? Description, MethodSignatureModifiers Modifiers, CSharpType? ReturnType, FormattableString? ReturnDescription, IReadOnlyList<Parameter> Parameters) : MethodSignatureBase(Name, Description, Modifiers, Parameters)
+    internal record MethodSignature(string Name, string? Description, string? Summary, MethodSignatureModifiers Modifiers, CSharpType? ReturnType, FormattableString? ReturnDescription, IReadOnlyList<Parameter> Parameters) : MethodSignatureBase(Name, Description, Summary, Modifiers, Parameters)
```
archerzz, line 15: Same sequence problem

src/AutoRest.CSharp/Common/Output/Models/RestClient.cs
```diff
@@ -23,7 +23,7 @@ internal abstract class RestClient : TypeProvider
         internal OperationGroup OperationGroup { get; }
         public IReadOnlyList<Parameter> Parameters { get; }
         public RestClientMethod[] Methods => _allMethods ??= BuildAllMethods().ToArray();
-        public ConstructorSignature Constructor => _constructor ??= new ConstructorSignature(Declaration.Name, $"Initializes a new instance of {Declaration.Name}", MethodSignatureModifiers.Public, Parameters.ToArray());
+        public ConstructorSignature Constructor => _constructor ??= new ConstructorSignature(Declaration.Name, $"Initializes a new instance of {Declaration.Name}", null, MethodSignatureModifiers.Public, Parameters.ToArray());
```
archerzz, line 26: I think in this case, it makes more sense to set `Summary` of the constructor.

src/AutoRest.CSharp/DataPlane/Generation/DataPlaneClientWriter.cs
```diff
@@ -68,14 +69,15 @@ private void WriteClientMethod(CodeWriter writer, ClientMethod clientMethod, boo
             responseType = async ? new CSharpType(typeof(Task<>), responseType) : responseType;
 
             var parameters = clientMethod.RestClientMethod.Parameters;
-            writer.WriteXmlDocumentationSummary($"{clientMethod.Description}");
+            writer.WriteXmlDocumentationSummary($"{GetXmlDocumentationSummaryText(clientMethod.RestClientMethod)}");
```
archerzz, line 72: better to move this method to `clientMethod.DescriptionText`

src/AutoRest.CSharp/Common/Output/Models/MethodSignatureBase.cs
```diff
@@ -4,10 +4,15 @@
...
 using AutoRest.CSharp.Output.Models.Shared;
+using AutoRest.CSharp.Utilities;
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal abstract record MethodSignatureBase(string Name, string? Description, MethodSignatureModifiers Modifiers, IReadOnlyList<Parameter> Parameters);
+    internal abstract record MethodSignatureBase(string Name, string? Description, string? Summary, MethodSignatureModifiers Modifiers, IReadOnlyList<Parameter> Parameters)
+    {
+        public string? SummaryText => Summary.IsNullOrEmpty() ? Description : Summary;
+        public string? DescriptionText => Summary.IsNullOrEmpty() ? string.Empty : Description;
```
archerzz, line 14: I remember the decision is that we just follow the M4 output? So if M4 copies the `Summary` into `Description`, we will have duplicated texts. If it's empty, then we just don't print it.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -174,11 +174,18 @@ public static CodeWriter.CodeWriterScope WriteMethodDeclaration(this CodeWriter
 
         public static CodeWriter WriteMethodDocumentation(this CodeWriter writer, MethodSignatureBase methodBase)
         {
-            writer.WriteXmlDocumentationSummary($"{methodBase.Description}");
+            writer.WriteMethodDocumentationBase(methodBase);
+            writer.WriteXmlDocumentation("remarks", $"{methodBase.DescriptionText}");
```
archerzz, line 178: Why not move `writer.WriteXmlDocumentation("remarks", $"{methodBase.DescriptionText}");` into `WriteMethodDocumentationBase`?

src/AutoRest.CSharp/DataPlane/Generation/DataPlaneClientWriter.cs
```diff
@@ -419,6 +423,16 @@ private void WriteStartOperationOperation(CodeWriter writer, LongRunningOperatio
             writer.Line();
         }
 
+        private string? GetXmlDocumentationSummaryText(RestClientMethod restClientMethod)
+        {
+            return restClientMethod.Summary.IsNullOrEmpty() ? restClientMethod.Description : restClientMethod.Summary;
```
archerzz, line 428: maybe we just not print the `<Summary>` or `<Remarks>` if the corresponding property is empty?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -591,7 +596,12 @@ private static void WriteSchemaDocumentationRemarks(CodeWriter writer, LowLevelC
                         schemaDesription = "Below is the JSON schema for one item in the pageable response.";
                     }
                 }
-                writer.WriteXmlDocumentation("remarks", $"{schemaDesription}{Environment.NewLine}{docinfo}{schemas}");
+                string? descriptionInRemarks = (!methodSignature.DescriptionText.IsNullOrEmpty()) ? $"{methodSignature.DescriptionText}{Environment.NewLine}{Environment.NewLine}" : string.Empty;
+                writer.WriteXmlDocumentation("remarks", $"{descriptionInRemarks}{schemaDesription}{Environment.NewLine}{docinfo}{schemas}");
```
archerzz, line 600: The method is `WriteSchemaDocumentationRemarks`, why do we want to print `Description` here?

archerzz (changes requested):

ArthurMa1978:
Currently the M4 copies the `Summary` text to `Description` if it's empty, would like to ask M4 fix it

src/AutoRest.CSharp/Common/Output/Models/ConstructorSignature.cs
pshao25: This line is about variable sequence. Are you saying the output sequence?

src/AutoRest.CSharp/Common/Output/Models/MethodSignatureBase.cs
pshao25: I didn't get. I do follow this logic. If there is duplicate text, I will put them both into summary and description.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
pshao25: That is because the "description" should be in the <remarks>.

src/AutoRest.CSharp/DataPlane/Generation/DataPlaneClientWriter.cs
pshao25: Correct. This logic is handled in `writer.WriteXmlDocumentation(Summary)`. If there is no content for the label, the whole label will not be generated.

lirenhe:
> Currently the M4 copies the `Summary` text to `Description` if it's empty, would like to ask M4 fix it

Based on our discussion in Monday, the M4 fix would be breaking change for some SDK.  Do you see any problem with current approach in this PR?

test/TestProjects/MgmtExtensionCommonRestOperation/Generated/RestOperations/CommonRestOperations.cs
```diff
@@ -54,7 +54,7 @@ internal HttpMessage CreateListTypeOnesBySubscriptionRequest(string subscription
             return message;
         }
 
-        /// <summary> Validate information for a certificate order. </summary>
+        /// <summary> Description for Validate information for a certificate order. </summary>
```
archerzz, line 57: This really looks weird. Is it added by us?

archerzz (approved):
Personally, I think we should compare the `summary` with `description`. If they are the same, probably we should not print `description` into `<remarks>`.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
archerzz: wrong commit

PR #2519, opened 2022-07-31
Add serialization code generation for simple CADL input case + some bugfixes

AlexanderSher:


src/AutoRest.CSharp/LowLevel/Generation/LowLevelModelWriter.cs
```diff
@@ -2,40 +2,17 @@
...
 using AutoRest.CSharp.Output.Models.Types;
-using AutoRest.CSharp.Utilities;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Azure.Core;
 
 namespace AutoRest.CSharp.Generation.Writers
 {
     internal class LowLevelModelWriter
     {
-        public static void WriteModel(CodeWriter writer, TypeProvider model)
```
AlexanderSher, line 19: @archerzz I have removed the input type pattern matching because I don't expect we will need it. CADL has models and enums as separate collections (https://github.com/microsoft/cadl/blob/main/packages/compiler/core/types.ts#L275), and I think we should follow this pattern.

samples/AppConfiguration/readme.md
```diff
@@ -6,6 +6,6 @@
 title: AppConfiguration
 generation1-convenience-client: true
 require: $(this-folder)/../../readme.md
-input-file: https://github.com/Azure/azure-rest-api-specs/blob/e606243e5297312781dd7dbfd7ab76d2329cc088/specification/appconfiguration/data-plane/Microsoft.AppConfiguration/stable/1.0/appconfiguration.json
+input-file: https://github.com/Azure/azure-rest-api-specs/blob/a63ddbed578f026d4e140345d240eff90cdd67ce/specification/appconfiguration/data-plane/Microsoft.AppConfiguration/stable/1.0/appconfiguration.json
```
archerzz, line 9: just curious, why do we update the commit id of the sample spec? Is that because we want to sync up with the latest spec?

AlexanderSher: Yep. Just to be sure this PR doesn't break anything.

AlexanderSher: But you're right, it is not required for this PR. Reverting.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -15,44 +18,90 @@ namespace AutoRest.CSharp.Output.Models.Types
...
-            Fields = CreateFields(inputModel).ToArray();
-            PublicConstructor = BuildPublicConstructor(inputModel);
+
+            (_inputsToFields, _parametersToFields) = CreateParametersAndFieldsForRoundTripModel(inputModel, typeFactory);
+
+            Fields = _inputsToFields.Values.ToList();
+            PublicConstructor = BuildPublicConstructor(Declaration.Name, _parametersToFields.Keys.ToList());
+
+            // Since we consider all models roundtrip for now, use the same constructor for serialization
+            SerializationConstructor = PublicConstructor;
```
ArcturusZhang, line 48: Will this be overridden by derived classes of this type like in a specific type of model?
In mgmt plane, models are initialized using their internal constructor which contains all it properties as parameters

ArcturusZhang:
I have a question about the model generation in DPG:
will we have (or do we already have) some new types deriving from `ModelTypeProvider` to replace the things directly derived from `TypeProvider` to represent different type of models? Like we had `ObjectType`, `SchemaObjectType`, `EnumType` etc, and mgmt derived something from them - `MgmtObjectType`, `ResourceData` and `ReferenceType`. Would we have some equivalent classes in CADL?

src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonAdditionalPropertiesSerialization.cs
```diff
@@ -6,16 +6,13 @@
 
 namespace AutoRest.CSharp.Output.Models.Serialization.Json
 {
-    internal class JsonAdditionalPropertiesSerialization
+    internal class JsonAdditionalPropertiesSerialization : JsonPropertySerialization
```
archerzz, line 9: Just a minor issue: should this be renamed as `JsonAdditionalPropertySerialization`?

src/AutoRest.CSharp/Common/Input/CadlInputTypeConverter.cs
```diff
@@ -75,20 +74,32 @@ private InputType CreateObject(ref Utf8JsonReader reader, JsonSerializerOptions
             _                   => CadlInputModelTypeConverter.CreateModelType(ref reader, id, name, options, _referenceHandler.CurrentResolver)
         };
 
-        public static InputPrimitiveType ReadPrimitiveType(ref Utf8JsonReader reader)
+        public static InputPrimitiveType ReadPrimitiveType(ref Utf8JsonReader reader, string? id, JsonSerializerOptions options)
```
archerzz, line 77: `options` seems not used?

src/AutoRest.CSharp/Common/Input/CadlInputTypeConverter.cs
```diff
@@ -42,14 +43,12 @@ private InputType CreateObject(ref Utf8JsonReader reader, JsonSerializerOptions
         {
             string? id = null;
             string? name = null;
-            bool isNullable = false;
             InputType? result = null;
             var isFirstProperty = true;
             while (reader.TokenType != JsonTokenType.EndObject)
             {
                 var isIdOrNameOrNullable = reader.TryReadReferenceId(ref isFirstProperty, ref id)
```
archerzz, line 50: shall we change the variable to `isIdOrName`, since we're not reading `IsNullable` here.

archerzz (approved):
No blocker issue found. Feel free to resolve my comments later.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -387,5 +387,15 @@ public static bool IsCollectionType(CSharpType type)
...
         }
+
+        public static bool RequiresToList(CSharpType from, CSharpType to)
+        {
+            if (!to.IsFrameworkType || !from.IsFrameworkType || from.FrameworkType != typeof(IEnumerable<>))
+            {
+                return false;
+            }
+
+            return to.FrameworkType == typeof(IReadOnlyList<>) || to.FrameworkType == typeof(IList<>);
```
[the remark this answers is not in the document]
AlexanderSher: No, `IsList` won't work because it returns `true` for `IEnumerable<T>`. In this case, if both `from` and `to` are `IEnumerable<T>`, we don't need conversion.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
AlexanderSher: No, there is no plan to have types derived from `ModelTypeProvider` yet. `EnumTypeProvider` would be a separate type.

Issue #2430, opened 2022-07-14
Support generating abstract base class in polymorphism


Yao725:
Considering that the customers might feel confused when they need to take care of the polymorphism, we are going to generate the base class in mgmt. plane as abstract class, this work will include:
1. Add a configuration to decide if we need to enable this feature
2. Add an `abstract` modifier in the base class
3. Modify the deserialization method of base class

PR #2520, opened 2022-08-01
Ensure base class is absrtact in polymorphism

Yao725:
fix https://github.com/Azure/autorest.csharp/issues/2430

# Description
<i>Regen PR in sdk repo: https://github.com/Azure/azure-sdk-for-net/pull/30560</i>

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
```diff
@@ -201,7 +201,7 @@ private void WriteJsonDeserialize(CodeWriter writer, SchemaObjectType model, Jso
                     }
                 }
 
-                if (model.Declaration.IsAbstract)
+                if (model.Declaration.IsAbstract || model.IsAbstract)
```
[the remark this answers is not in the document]
ArcturusZhang: Now the property `IsAbstract` is defined in `SchemaObjectType`. But I believe it makes sense that we move it to `TypeProvider` since if a `TypeProvider` provides a type, it makes perfect sense that it is providing an abstract type (class).
And we could change the function `BuilderHelpers.CreateTypeAttributes` to consume a new parameter `isAbstract` so that these two calls will align with each other

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
```diff
@@ -166,6 +166,7 @@ internal static RenameRuleTarget Parse(string rawValue)
             JsonElement? resourceModelRequiresType = default,
             JsonElement? resourceModelRequiresName = default,
             JsonElement? singletonRequiresKeyword = default,
+            JsonElement? abstractBaseClassWithDiscriminator = default,
```
ArthurMa1978, line 169: add this function to the document

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
```diff
@@ -201,7 +201,7 @@ private void WriteJsonDeserialize(CodeWriter writer, SchemaObjectType model, Jso
...
                 {
                     writer.Line($"throw new {typeof(NotSupportedException)}(\"Deserialization of abstract type '{model.Type}' not supported.\");");
```
ArthurMa1978, line 206: How about:
Abstract class '{model.Type}' cannot be instantiated.

test/TestProjects/MgmtDiscriminator/Generated/Models/DeliveryRuleAction.cs
```diff
@@ -12,10 +12,10 @@ namespace MgmtDiscriminator.Models
     /// Please note <see cref="DeliveryRuleAction"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
     /// The available derived classes include <see cref="DeliveryRuleCacheExpirationAction"/>, <see cref="DeliveryRuleCacheKeyQueryStringAction"/>, <see cref="DeliveryRuleRequestHeaderAction"/>, <see cref="DeliveryRuleResponseHeaderAction"/>, <see cref="OriginGroupOverrideAction"/>, <see cref="DeliveryRuleRouteConfigurationOverrideAction"/>, <see cref="UrlRedirectAction"/>, <see cref="UrlRewriteAction"/> and <see cref="UrlSigningAction"/>.
     /// </summary>
-    public partial class DeliveryRuleAction
+    public abstract partial class DeliveryRuleAction
```
ArthurMa1978, line 15: Not enable `EnableAbstractBaseClassWithDiscriminator` but why this is changed to abstract?

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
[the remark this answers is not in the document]
AlexanderSher: Agreed. If both `TypeDeclarationOptions` and `SchemaObjectType` provides `IsAbstract` property, their values should be identical. Maybe we shouldn't even have one of these properties

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
Yao725: added

test/TestProjects/MgmtDiscriminator/Generated/Models/DeliveryRuleAction.cs
Yao725: Has changed the configuration to a list `suppress-abstract-base-class`

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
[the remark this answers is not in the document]
Yao725: Moved the `IsAbstract` property to the base type TypeProvider and ensured the value is identical with the `TypeDeclarationOptions`

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
Yao725: Data Plane also uses this message, so I prefer to leave it as it is.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -278,5 +279,9 @@ private ObjectTypeProperty UpdatePropertyDescription(ObjectTypeProperty property
                 property.ValueType,
                 property.OptionalViaNullability);
         }
+
+        private bool IsBaseClassWithDiscriminator => ObjectSchema.Discriminator?.All != null &&
+            ObjectSchema.Parents?.All.Count == 0 &&
+            !Configuration.MgmtConfiguration.SuppressAbstractBaseClass.Contains(ExistingType != null ? ExistingType.Name : DefaultName);
```
ArcturusZhang, line 285: This contains kind of complicated logic here, could we cache this value?
Is there a scenario that this value could change over time? If not, I suggest we cache this value or cache the `IsAbstract` property, so that this expression will only be invoked once.
And I have a bit concern that this configuration works on the generated type name instead of its "swagger name" - this introduces inconsistency with other configurations.
Please make sure this is well documented.

test/TestProjects/MgmtSafeFlatten/Generated/Models/LayerOneBaseType.Serialization.cs
```diff
@@ -30,16 +31,7 @@ internal static LayerOneBaseType DeserializeLayerOneBaseType(JsonElement element
...
-            foreach (var property in element.EnumerateObject())
-            {
-                if (property.NameEquals("name"))
-                {
-                    name = new LayerOneTypeName(property.Value.GetString());
-                    continue;
-                }
-            }
-            return new LayerOneBaseType(name);
+            throw new NotSupportedException("Deserialization of abstract type 'global::MgmtSafeFlatten.Models.LayerOneBaseType' not supported.");
```
ArcturusZhang, line 34: Would this be an issue when the service supports a new one for a customer using the old SDK? We should make sure there is a fallback class here

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
Yao725: I refactored this change, this part is reverted. And now the configuration will also use the swagger name.

test/TestProjects/MgmtSafeFlatten/Generated/Models/LayerOneBaseType.Serialization.cs
Yao725: Added the fallback internal class.

src/AutoRest.CSharp/Mgmt/Decorator/Transformer/AbstractTypeDetection.cs
```diff
@@ -0,0 +1,65 @@
...
+using AutoRest.CSharp.Mgmt.AutoRest;
+using AutoRest.CSharp.Mgmt.Output;
+using AutoRest.CSharp.Output.Models.Types;
+using Microsoft.CodeAnalysis;
+
+namespace AutoRest.CSharp.Mgmt.Decorator.Transformer
+{
+    internal static class AbstractTypeDetection
+    {
+        public static readonly ConcurrentDictionary<string, ObjectSchema> AbstractTypeToInternalDerivedClass = new();
```
Yao725, line 20: It isn't best to use the original schema name as the key to the dictionary, it's possible we couldn't find the corresponding value when the name is updated after this detection, need an update here.

test/TestProjects/MgmtSafeFlatten/Generated/Models/LayerOneBaseType.cs
```diff
@@ -28,5 +30,13 @@ internal LayerOneBaseType(LayerOneTypeName name)
 
         /// <summary> Gets or sets the name. </summary>
         internal LayerOneTypeName Name { get; set; }
+
+        /// <summary> Create an wrapped subclass instance of LayerOneBaseType by serializing the provided BinaryData. </summary>
+        /// <param name="data"> The BinaryData that will be used in serializing, it should be a valid json object. </param>
+        /// <returns> The wrapped subclass instance of LayerOneBaseType. </returns>
+        public static LayerOneBaseType FromBinaryData(BinaryData data)
```
m-nash, line 37: Lets keep this out of the first version.  I believe this was in relation to the round trip of unknown types?  For now we recognize that round trip of unknown types can result in an issue lets address that in a future change

test/TestProjects/MgmtSafeFlatten/Generated/Models/UnknownLayerOneBaseType.cs
```diff
@@ -0,0 +1,26 @@
...
+
+#nullable disable
+
+namespace MgmtSafeFlatten.Models
+{
+    /// <summary> The UnknownLayerOneBaseType. </summary>
+    internal partial class UnknownLayerOneBaseType : LayerOneBaseType
+    {
+        /// <summary> Initializes a new instance of UnknownLayerOneBaseType. </summary>
+        public UnknownLayerOneBaseType()
```
m-nash, line 14: this ctor should not be needed as it will never be called

m-nash:
We might need to make exceptions for [this ](https://github.com/Azure/azure-sdk-for-net/pull/30235/files#diff-396658de0319f8a4304752dd71c8bc76bf8a8f60b583086b1bd4345013a2c760R20) in GA'd types as its a breaking change.  Technically if someone is using that ctor today it wouldn't work so it might be an acceptable break I have asked @KrzysztofCwalina to weigh in before we move forward.

src/AutoRest.CSharp/Mgmt/Decorator/Transformer/AbstractTypeDetection.cs
Yao725: Remove this change.

test/TestProjects/MgmtSafeFlatten/Generated/Models/LayerOneBaseType.cs
Yao725: Removed

test/TestProjects/MgmtSafeFlatten/Generated/Models/UnknownLayerOneBaseType.cs
Yao725: Removed

ArthurMa1978 (approved):

Issue #2335, opened 2022-06-20
Generate convenience methods corresponding to protocol methods defined on a client, from Cadl


annelo-msft:
### Convenience methods for all operations

This issue tracks the work to generate convenience methods for all operations described in Cadl file.  Generating just a subset of methods is still in design https://github.com/Azure/autorest.csharp/issues/2443, and out of scope for this issue.

#### Input Cadl file
[petstore.cadl](https://github.com/microsoft/cadl/blob/main/packages/samples/petstore/petstore.cadl)

#### Expected output

Operations added to generated client:
- [ ] [ReadValue](https://github.com/Azure/azure-sdk-for-net/pull/30004/files#diff-0ec10628434582dfdc035202ee4a69f55d2b3d4ab4b6f90fdbbe725f63139367R17-R55)
- [ ] [Create](https://github.com/Azure/azure-sdk-for-net/pull/30004/files#diff-0ec10628434582dfdc035202ee4a69f55d2b3d4ab4b6f90fdbbe725f63139367R57-R79)
- [ ] [DeleteValue](https://github.com/Azure/azure-sdk-for-net/pull/30004/files#diff-0ec10628434582dfdc035202ee4a69f55d2b3d4ab4b6f90fdbbe725f63139367R81-R119)

Generated model files:
- [ ] [Pet.cs](https://github.com/Azure/azure-sdk-for-net/pull/30004/files#diff-d77e56a151ec6556711070374e5523122011c4282cca9ba50daf4f527dc3a2e3)
- [ ] [Pet.Serialization.cs](https://github.com/Azure/azure-sdk-for-net/pull/30004/files#diff-b316976d184a1d51b769e4a8bb40d2f64ec890c6d394bf7b75382629370c8c26), including
    - [ToRequestContent](https://github.com/Azure/azure-sdk-for-net/pull/30004/files#diff-b316976d184a1d51b769e4a8bb40d2f64ec890c6d394bf7b75382629370c8c26R56-R61)
    - [FromResponse](https://github.com/Azure/azure-sdk-for-net/pull/30004/files#diff-b316976d184a1d51b769e4a8bb40d2f64ec890c6d394bf7b75382629370c8c26R63-R67)

#### Requirements
- [ ] Documentation comments are provided from `@doc` and `@summary` decorators in the Cadl definition.  Json schema definitions in remarks and examples are not included.
[...]

PR #2521, opened 2022-08-01
[DPG] Add convenience methods 

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/2335


src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -49,11 +49,19 @@ internal class RestClientBuilder
...
+
+        private Dictionary<InputOperation, CSharpType> _returnTypes;
+        public Dictionary<InputOperation, CSharpType> ReturnTypes => _returnTypes;
+
         public RestClientBuilder(IEnumerable<InputParameter> clientParameters, TypeFactory typeFactory)
         {
             _serializationBuilder = new SerializationBuilder();
             _typeFactory = typeFactory;
             _parameters = clientParameters.ToDictionary(p => p.Name, BuildConstructorParameter);
+            _bodyParameters = new Dictionary<InputOperation, Parameter>();
```
archerzz, line 63: It would be better to initialize`_bodyParameters` in the constructor.

archerzz: Also, can we deduce `_bodyParameters` from `_parameters`?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -55,8 +55,9 @@ public void WriteClient(CodeWriter writer, LowLevelClient client)
                     WriteConstructors(writer, client);
 
                     var exampleComposer = new LowLevelExampleComposer(client);
-                    foreach (var clientMethod in client.ClientMethods)
+                    foreach (var convenienceMethod in client.ConvenienceMethods)
```
archerzz, line 58: We'll filter convenience methods through configuration. So here we cannot assume a 1:1 mapping.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -49,11 +49,19 @@ internal class RestClientBuilder
         private readonly TypeFactory _typeFactory;
         private readonly Dictionary<string, Parameter> _parameters;
 
+        private Dictionary<InputOperation, Parameter> _bodyParameters;
```
AlexanderSher, line 52: This is the wrong place for this change. We don't add any new `CreateXxxRequest` or protocol methods, only convenience methods will be added, so request methods shouldn't be affected.

pshao25: I got all the information for `ConvenienceMethods` from `ClientMethods`. But when we build `ClientMethods` by `RestClientBuilder`,  it throws away the important information like return type and body type. I just add its original information.

test/TestProjects/CadlPetStore/Generated/PetsClient.cs
```diff
@@ -6,14 +6,15 @@
...
 using System.Threading.Tasks;
 using Azure;
 using Azure.Core;
 using Azure.Core.Pipeline;
 
 namespace CadlPetStore
 {
-    /// <summary> The Pets service client. </summary>
+    /// <summary> Manage your pets. You can delete or get the Pet from pet store. </summary>
     public partial class PetsClient
```
annelo-msft, line 18: Curious: where is the input specified?  
I am looking at https://github.com/microsoft/cadl/blob/main/packages/samples/petstore/petstore.cadl, and I'd like to understand better how you are choosing the name for the client, client options, and namespace.  Thanks!

pshao25: The input for the codegen is https://github.com/Azure/autorest.csharp/blob/66570505a5724751905560bfc8357fbf1834c81e/test/TestProjects/CadlPetStore/Generated/cadl.json.
Currently it does not respect https://github.com/microsoft/cadl/blob/main/packages/samples/petstore/petstore.cadl because the emitter is still using the stable version of compiler (but I think Crystal will update later).

Here is how we choosing:
Name for the client: https://github.com/Azure/autorest.csharp/blob/66570505a5724751905560bfc8357fbf1834c81e/test/TestProjects/CadlPetStore/Generated/cadl.json#L65, append `Client`;

Name for the client options: 
If there is only one client, append `ClientOptions` to client name; if there are more than one clients, use `InputNamespace.Name` https://github.com/Azure/autorest.csharp/blob/66570505a5724751905560bfc8357fbf1834c81e/test/TestProjects/CadlPetStore/Generated/cadl.json#L3

Name for namespace:
If there is `Configuration.Namespace` https://github.com/Azure/autorest.csharp/blob/66570505a5724751905560bfc8357fbf1834c81e/test/TestProjects/CadlPetStore/Generated/Configuration.json#L3, then use this, else use `InputNamespace.Name` https://github.com/Azure/autorest.csharp/blob/66570505a5724751905560bfc8357fbf1834c81e/test/TestProjects/CadlPetStore/Generated/cadl.json#L3

src/AutoRest.CSharp/LowLevel/Generation/LowLevelConvenienceMethodWriter.cs
```diff
@@ -0,0 +1,104 @@
...
+using AutoRest.CSharp.Common.Generation.Writers;
+using AutoRest.CSharp.Output.Models;
+using AutoRest.CSharp.Output.Models.Shared;
+using Azure;
+
+namespace AutoRest.CSharp.Generation.Writers
+{
+    internal class LowLevelConvenienceMethodWriter : ClientWriter
+    {
+        internal static void WriteClientConvenienceMethod(CodeWriter writer, LowLevelConvenienceMethod convenienceMethod, string diagnosticsPropertyName, bool async)
```
archerzz, line 18: I suggest to make those method non-static. You can put some common parameters like `CodeWriter writer` as the properties of this class, and then initialize them in the constructor. See `MgmtClientBaseWriter`.
That can greatly improve the codes readability.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -217,7 +217,7 @@ private static void SetRequestToClient(ClientInfo clientInfo, InputOperation ope
             clientInfo.Requests.Add(operation);
         }
 
-        private IEnumerable<LowLevelClient> CreateClients(IEnumerable<ClientInfo> clientInfos, TypeFactory typeFactory, ClientOptionsTypeProvider clientOptions, LowLevelClient? parentClient)
+        private IEnumerable<LowLevelClient> CreateClients(IEnumerable<ClientInfo> clientInfos, TypeFactory typeFactory, ClientOptionsTypeProvider clientOptions, LowLevelClient? parentClient, bool cadlInput)
```
archerzz, line 220:
```suggestion
        private IEnumerable<LowLevelClient> CreateClients(IEnumerable<ClientInfo> clientInfos, TypeFactory typeFactory, ClientOptionsTypeProvider clientOptions, LowLevelClient? parentClient, bool isCadlInput)
```

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -46,7 +47,7 @@ internal class LowLevelClient : TypeProvider
         public bool IsSubClient { get; }
         public bool IsResourceClient { get; }
 
-        public LowLevelClient(string name, string ns, string description, string libraryName, LowLevelClient? parentClient, IEnumerable<InputOperation> operations, RestClientBuilder builder, InputAuth authorization, SourceInputModel? sourceInputModel, ClientOptionsTypeProvider clientOptions)
+        public LowLevelClient(string name, string ns, string description, string libraryName, LowLevelClient? parentClient, IEnumerable<InputOperation> operations, RestClientBuilder builder, InputAuth authorization, SourceInputModel? sourceInputModel, ClientOptionsTypeProvider clientOptions, bool cadlInput)
```
archerzz, line 50:
```suggestion
        public LowLevelClient(string name, string ns, string description, string libraryName, LowLevelClient? parentClient, IEnumerable<InputOperation> operations, RestClientBuilder builder, InputAuth authorization, SourceInputModel? sourceInputModel, ClientOptionsTypeProvider clientOptions, bool isCadlInput)
```

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -69,6 +70,10 @@ public LowLevelClient(string name, string ns, string description, string library
                 .OrderBy(m => m.LongRunning != null ? 2 : m.PagingInfo != null ? 1 : 0) // Temporary sorting to minimize amount of changed files. Will be removed when new LRO is implemented
                 .ToArray();
 
+            ConvenienceMethods = cadlInput
+                ? ClientMethods.Select(clientMethod => BuildConvenienceMethod(clientMethod, builder)).ToArray()
```
archerzz, line 74: Better to directly cast to `List`.
```suggestion
                ? ClientMethods.Select(clientMethod => BuildConvenienceMethod(clientMethod, builder)).ToList()
```

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -134,6 +139,45 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
+        {
+            List<Parameter> parameters = new List<Parameter>();
+            MethodSignature protocolSignature = clientMethod.Signature;
+            InputOperation operation = clientMethod.RequestMethod.Operation;
+            Parameter? bodyParameter = null;
+            foreach (var parameter in protocolSignature.Parameters)
+            {
+                if (parameter == KnownParameters.RequestContent || parameter == KnownParameters.RequestContentNullable)
+                {
+                    bodyParameter = builder.GetBodyParameter(operation, DefaultNamespace);
```
archerzz, line 152: See my comment at the top. Here you can change to:
```suggestion
                    bodyParameter = operation.GetBodyParameter(DefaultNamespace);
```

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -134,6 +139,45 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
+                    parameters.Add(KnownParameters.CancellationTokenParameter);
+                }
+                else
+                {
+                    parameters.Add(parameter);
+                }
+            }
+
+            CSharpType? responseType = builder.GetReturnType(operation, DefaultNamespace);
+            CSharpType? returnType = responseType == null ? typeof(Azure.Response) : new CSharpType(typeof(Azure.Response<>), responseType!);
```
archerzz, line 166: Can we fold this line into `GetReturnType()`?

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -86,6 +86,32 @@ private static string GetRequestParameterName(RequestParameter requestParameter)
             return language.SerializedName ?? language.Name;
         }
 
+        public CSharpType? GetReturnType(InputOperation operation, string defaultNamespace)
```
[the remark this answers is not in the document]
pshao25: I need `_typeFactory` which is a private field in `RestClientBuilder`.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: Same reason above. I cannot pass the `_typeFactory` to `InputOperation`.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: I'm afraid not. I need two things: `Response<responseType>` and `responseType`. `Response<responseType>` will pass into the `MethodSignature` to generate signature, and `responseType` will pass into `LowLevelConvenienceMethod` for model deserialization.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -86,6 +86,32 @@ private static string GetRequestParameterName(RequestParameter requestParameter)
...
+        {
+            var bodyType = operation.Responses.FirstOrDefault()?.BodyType;
+            if (bodyType != null && bodyType is InputModelType)
+            {
+                return new CSharpType(new ModelTypeProvider((bodyType as InputModelType)!, _typeFactory, defaultNamespace, null));
```
chunyu3, line 94: Do we current we only support json object as return value, primitive such as string not supported?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
archerzz: Then I suggest to rename this method to `GetResponseType`, since `RestClientBuilder` is low level.

archerzz (approved):

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -86,6 +86,32 @@ private static string GetRequestParameterName(RequestParameter requestParameter)
...
+
+        public Parameter? GetBodyParameter(InputOperation operation, string defaultNamespace)
+        {
+            var requestParameters = operation.Parameters
+                .Where(rp => !IsIgnoredHeaderParameter(rp));
+
+            var bodyParameters = requestParameters.Where(parameter => parameter.Location == RequestLocation.Body);
+            var requiredParameter = bodyParameters.Where(parameter => parameter.IsRequired).FirstOrDefault();
+
+            var bodyParameter = requiredParameter ?? bodyParameters.FirstOrDefault();
```
chunyu3, line 108: When there is no body parameter? it will return default. What is the default value? null?
if so, need to check in following code

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -86,6 +86,32 @@ private static string GetRequestParameterName(RequestParameter requestParameter)
...
+            var bodyParameterType = bodyParameter.Type is InputModelType
```
chunyu3, line 109: can bodyParameter be null?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -54,9 +54,40 @@ public void WriteClient(CodeWriter writer, LowLevelClient client)
                     WriteClientFields(writer, client);
                     WriteConstructors(writer, client);
 
+                    HashSet<LowLevelClientMethod> visitedMethods = new HashSet<LowLevelClientMethod>();
                     var exampleComposer = new LowLevelExampleComposer(client);
+                    foreach (var convenienceMethod in client.ConvenienceMethods)
```
chunyu3, line 59: Why not traverse the client.ClientMethods, and in each loop, check if it is need to generate convenience method, if so, just add one?
Write in this way may not cause too many duplicate code.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
pshao25: Correct. We don't support non-user defined type.

archerzz: Why not `_typeFactory.CreateType()`?

chunyu3 (approved):

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
pshao25: We can, but it is meaningless. We don't support this scenario and `_typeFactory.CreateType()` will never be hit.

lirenhe (approved):

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
AlexanderSher: This is a sister issue of the https://github.com/Azure/autorest.csharp/issues/2535, but we can address it only after model generation is fully implemented

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -54,6 +54,7 @@ public void WriteClient(CodeWriter writer, LowLevelClient client)
                     WriteClientFields(writer, client);
                     WriteConstructors(writer, client);
 
+                    HashSet<LowLevelClientMethod> visitedMethods = new HashSet<LowLevelClientMethod>();
```
AlexanderSher, line 57: Do you need this hashset?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -70,6 +71,12 @@ public void WriteClient(CodeWriter writer, LowLevelClient client)
                         }
                         else
                         {
+                            var convenienceMethod = client.ConvenienceMethods.FirstOrDefault(convenienceMethod => convenienceMethod.LowLevelClientMethod == clientMethod);
```
AlexanderSher, line 74: @annelo-msft , is it important that we write convenience methods right before its client method rather than writing them all at once?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -134,6 +139,45 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
+                {
+                    parameters.Add(KnownParameters.CancellationTokenParameter);
+                }
+                else
+                {
+                    parameters.Add(parameter);
+                }
+            }
+
+            CSharpType? responseType = builder.GetResponseType(operation, DefaultNamespace);
```
AlexanderSher, line 165: If you don't use `GetResponseType` anywhere else, I'd move it into `LowLevelClient`, cause it isn't used for REST operation logic

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -134,6 +139,45 @@ public static IEnumerable<LowLevelClientMethod> BuildMethods(RestClientBuilder b
...
+            foreach (var parameter in protocolSignature.Parameters)
+            {
+                if (parameter == KnownParameters.RequestContent || parameter == KnownParameters.RequestContentNullable)
+                {
+                    bodyParameter = builder.GetBodyParameter(operation, DefaultNamespace);
+                    parameters.Add(bodyParameter!);
+                }
+                else if (parameter == KnownParameters.RequestContext)
+                {
+                    parameters.Add(KnownParameters.CancellationTokenParameter);
```
AlexanderSher, line 157: Do we know that `RequestContext` is always the last parameter?

src/AutoRest.CSharp/Properties/launchSettings.json
```diff
@@ -152,10 +152,6 @@
       "commandName": "Project",
       "commandLineArgs": "--standalone $(SolutionDir)\\test\\TestProjects\\BodyAndPath-LowLevel\\Generated"
     },
-    "CadlFirstTest": {
```
AlexanderSher, line 155: Please don't kill this project yet, it is usable as a playground.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
pshao25: No, forget to remove. Will update.

src/AutoRest.CSharp/Properties/launchSettings.json
pshao25: @chunyu3 plans to remove it. Do we have a substitute one?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: No, but the problem is `LowLevelClient` does have a `TypeFactory`.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
pshao25: Currently, it is. https://github.com/Azure/autorest.csharp/blob/66570505a5724751905560bfc8357fbf1834c81e/src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs#L282

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
pshao25: I think the answer is yes. See https://github.com/Azure/autorest.csharp/issues/2335#issuecomment-1198164908

PR #2530, opened 2022-08-03
fix pageable item type issue

### LINKED ISSUE #2531: [P0] Generating GetAll function may return the wrong value if the return value has multiple sub-resources
When generating the GetAll method, if the returned resource has more than one child resource, the first-word resource will be used as the return value, which is incorrect.
For example, the return value of the `GetAllRecordSetsAsync` in [Dns](https://github.com/Azure/autorest.csharp/blob/2d3b347ee36a0b609e6c0553e68be502a102a9fe/test/TestProjects/MgmtExpandResourceTypes/Generated/ZoneResource.cs#L700) method should be `RecordSetData`, but the generated code is a subresource of `RecordSetData`: `RecordSetAResource`. This is inconsistent with the [swagger](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/dns/resource-manager/Microsoft.Network/stable/2018-05-01/dns.json#L698) definition

ArcturusZhang:
Fixes #2531 

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Models/MgmtRestOperation.cs
```diff
@@ -360,11 +360,15 @@ private CSharpType GetWrappedMgmtReturnType(CSharpType? originalType)
             if (Resource is not null && Resource.ResourceData.Type.Equals(originalType))
                 return Resource.Type;
 
-            var foundResource = MgmtContext.Library.ArmResources.FirstOrDefault(resource => resource.ResourceData.Type.Equals(originalType));
-            if (foundResource is not null)
-                return foundResource.Type;
+            var foundResources = MgmtContext.Library.ArmResources.Where(resource => resource.ResourceData.Type.Equals(originalType));
```
AlexanderSher, line 363: Please call `ToList` or `ToArray`, otherwise enumeration would happen twice.

ArcturusZhang: Will do

ArthurMa1978 (approved):

Issue #2535, opened 2022-08-04
Correctly use enums in protocol method generation


AlexanderSher:
Right now, RestClientBuilder generates methods with string and int parameters instead of enums because `TypeFactory.CreateType` incorrectly resolves type for enums:
https://github.com/Azure/autorest.csharp/blob/7da8185a8500150c26766b65f03fe2a2a030e932/src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs#L32-L37

When fixed, either `RestClientBuilder` or `LowLevelClientWriter` must be updated to generate correct signatures for protocol methods and CreateXxxMethods.

lirenhe:
@AlexanderSher,  seems this item is related to code refactoring.  So do you think this is required for August MVP? And could you provide the cost estimate for this item?  Thanks

PR #2536, opened 2022-08-04
Unified serialization + partial enum support

AlexanderSher:
This PR contains 2 changes:
- it unifies json serialization/deserialization codegen for DPG, MPG and HLC. Now all three use the same logic. However, generation of `JsonPropertySerialization` is flavor-specific, so one more PR will follow to resolve it
- `EnumType` now created for `InputEnumType` in addition to `ChoiceSchema`/`SealedChoiceSchema`. DpgOutputLibrary can resolve enum types. Known issue: https://github.com/Azure/autorest.csharp/issues/2535


src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -215,7 +216,8 @@ private string GetPropertySummary(MethodInfo? setter)
             return setter != null ? " or sets" : string.Empty;
         }
 
-        protected override ObjectTypeConstructor BuildSerializationConstructor() => BuildConstructor(GetCtor(_type, ReferenceClassFinder.SerializationCtorAttributeName));
+        protected override ObjectTypeConstructor BuildSerializationConstructor()
```
AlexanderSher, line 219: There was a bug when `SystemObjectType` has been using initialization constructor as a base constructor instead of serialization one.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -65,7 +75,8 @@ public SchemaObjectType(ObjectSchema objectSchema, BuildContext context)
         protected override TypeKind TypeKind => IsStruct ? TypeKind.Struct : TypeKind.Class;
         public bool IsStruct => ExistingType?.IsValueType == true;
 
-        public ObjectSerialization[] Serializations => _serializations ??= BuildSerializations();
+        public JsonObjectSerialization? JsonSerialization => _hasJsonSerialization ? _jsonSerialization ??= _serializationBuilder.BuildJsonObjectSerialization(ObjectSchema, this) : null;
```
AlexanderSher, line 78: Collection of serializations has been replaced with 2 explicit fields. This has made logic in SerializationWriter more straightforward.

test/AutoRest.TestServerLowLevel.Tests/LowLevel/Generation/ModelWriterTests.cs
```diff
@@ -190,7 +190,7 @@ public PrimitivePropertyModel(string requiredString,int requiredInt,long require
 
 namespace Cadl.TestServer.PrimitiveProperties
 {
-public partial class PrimitivePropertyModel : global::Azure.Core.IUtf8JsonSerializable
+public partial class PrimitivePropertyModel: global::Azure.Core.IUtf8JsonSerializable
 {
 void global::Azure.Core.IUtf8JsonSerializable.Write(global::System.Text.Json.Utf8JsonWriter writer)
```
ArcturusZhang, line 195: These are example snippets? Looks like our simplifier cannot simplify things in the xml document, would this be a problem?

src/AutoRest.CSharp/Common/Output/Models/Types/OutputLibrary.cs
```diff
@@ -10,5 +11,8 @@ internal abstract class OutputLibrary
     {
         public abstract CSharpType FindTypeForSchema(Schema schema);
         public abstract CSharpType? FindTypeByName(string originalName);
+
+        public abstract CSharpType ResolveEnum(InputEnumType enumType);
+        public abstract CSharpType ResolveModel(InputModelType model);
```
ArcturusZhang, line 16: Can we add some document to explain what these method are for?
And would these two methods be a temporary solution during the phase that we have both modeler model and cadl model? Or this is permenant?

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
```diff
@@ -3,36 +3,40 @@
...
-                // This is technically invalid behavior, we are hitting this in generating responses we throw away.
-                // https://github.com/Azure/autorest.csharp/issues/1108
-                // throw new InvalidOperationException($"FindTypeForSchema of invalid schema {schema.Name} in LowLevelOutputLibrary");
-                _ => new CSharpType(typeof(object))
-            };
+        public override CSharpType ResolveEnum(InputEnumType enumType) => _enums != null
+            ? _enums.TryGetValue(enumType with {IsNullable = false}, out var typeProvider) ? typeProvider.Type : _typeFactory.CreateType(enumType.EnumValueType)
+            : throw new InvalidOperationException($"{nameof(ResolveEnum)} is called before enums are generated.");
+        public override CSharpType ResolveModel(InputModelType model) => _models != null
+            ? _models.TryGetValue(model, out var typeProvider) ? typeProvider.Type : new CSharpType(typeof(object), model.IsNullable)
```
archerzz, line 36: Maybe use _typeFactory to be consistent?
```suggestion
            ? _models.TryGetValue(model, out var typeProvider) ? typeProvider.Type : _typeFactory.CreateType(model)
```

test/TestServerProjects/multiple-inheritance/Generated/Models/Cat.cs
```diff
@@ -28,14 +28,8 @@ public Cat(string name) : base(name)
         /// <param name="likesMilk"></param>
         /// <param name="meows"></param>
         /// <param name="hisses"></param>
-        /// <exception cref="ArgumentNullException"> <paramref name="name"/> is null. </exception>
         internal Cat(string name, bool? likesMilk, bool? meows, bool? hisses) : base(name)
         {
-            if (name == null)
```
annelo-msft, line 34: Just curious: why does this validation go away?

test/AutoRest.TestServerLowLevel.Tests/LowLevel/Generation/ModelWriterTests.cs
AlexanderSher: No, this hasn't gone through the simplifier, we do comparison straight out of the writer.

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
```diff
@@ -3,36 +3,40 @@
...
-            {
-                AllSchemaTypes.Choice => _typeFactory.CreateType(((ChoiceSchema)schema).ChoiceType, false),
-                AllSchemaTypes.SealedChoice => _typeFactory.CreateType(((SealedChoiceSchema)schema).ChoiceType, false),
-                // This is technically invalid behavior, we are hitting this in generating responses we throw away.
-                // https://github.com/Azure/autorest.csharp/issues/1108
-                // throw new InvalidOperationException($"FindTypeForSchema of invalid schema {schema.Name} in LowLevelOutputLibrary");
-                _ => new CSharpType(typeof(object))
-            };
+        public override CSharpType ResolveEnum(InputEnumType enumType) => _enums != null
+            ? _enums.TryGetValue(enumType with {IsNullable = false}, out var typeProvider) ? typeProvider.Type : _typeFactory.CreateType(enumType.EnumValueType)
```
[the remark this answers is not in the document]
AlexanderSher: yeah, that's not ideal. We probably should have equality comparer that ignores nullables (that would remove the requirement for custom collection as well). I'll adjust this.

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
```diff
@@ -50,34 +52,21 @@ private EnumType(Schema schema, BuildContext context, Schema baseType, IEnumerab
                         $" expected enum or struct got {ExistingType.TypeKind}")
                 };
 
-                _typeMapping = context.SourceInputModel?.CreateForModel(ExistingType);
+                _typeMapping = sourceInputModel?.CreateForModel(ExistingType);
             }
 
-            BaseType = context.TypeFactory.CreateType(baseType, false);
-            Description = schema.CreateDescription();
+            BaseType = typeFactory.CreateType(input.EnumValueType);
```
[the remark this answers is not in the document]
AlexanderSher: This is a bad name for the property. It is not a base type but rather the type of the value. I'll rename it.

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
AlexanderSher: That would create an infinite recursion. This check is for swagger-based DPG that doesn't have models yet. When HLC is switched on the DPG rails, I'll remove this check.

test/TestServerProjects/multiple-inheritance/Generated/Models/Cat.cs
AlexanderSher: It is validated in the base constructor. Previously HLC model has been using last constructor in collection, which may be initialization constructor if initialization and serialization constructors have the same signature. Serialization constructors explicitly skip serialization in HLC. I can revert this part to minimize the change.

annelo-msft: No, it's fine, I was just curious.  Thanks!

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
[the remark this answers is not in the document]
AlexanderSher: Fixed

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
```diff
@@ -50,34 +52,21 @@ private EnumType(Schema schema, BuildContext context, Schema baseType, IEnumerab
...
-
-            if (Configuration.ModelNamespace)
-            {
-                return $"{context.DefaultNamespace}.Models";
-            }
-
-            return context.DefaultNamespace;
-        }
+        private static string GetDefaultNamespace(string? ns, string defaultNamespace)
+            => ns ?? (Configuration.ModelNamespace ? $"{defaultNamespace}.Models" : defaultNamespace);
```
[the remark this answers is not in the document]
AlexanderSher: namespace must be set in the base class. Maybe this can be eliminated but I don't want to spend much time on this now.

test/TestProjects/CadlFirstTest/Generated/cadl.json
```diff
@@ -2,6 +2,32 @@
   "Name": "demo.helloworld",
   "Description": "",
   "ApiVersions": ["0.1.0"],
+  "Enums": [
+    {
+      "$id": "4",
+      "Name": "SimpleEnum",
+      "Description": "Simple enum",
+      "EnumValueType": "string",
```
AlexanderSher, line 10: Enum is properly generated only for the case when value type is a string. Since we're using existing generator code for enums, it looks like we don't have services that have different value type.

archerzz (approved):

PR #2538, opened 2022-08-04
Support `integer` type format

ArthurMa1978:


src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -217,6 +217,7 @@ internal static bool IsOperationOfPageable(CSharpType type)
 
         internal static Type? ToFrameworkType(Schema schema, RecordOfStringAndAny? extensions) => schema.Type switch
         {
+            AllSchemaTypes.Integer => typeof(int),
```
AlexanderSher, line 220: I need to remember to add this case to the `CodeModelConverter` :)

Yao725 (approved):

PR #2544, opened 2022-08-05
Update summary for BinaryData in mgmt plane

### LINKED ISSUE #2447: [P0][Bug] The serialization will fail if we change a string type to BinaryData
We changed the type of string to BinaryData at some times. Here is one example of [Cdn](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/cdn/Azure.ResourceManager.Cdn/src/Generated/Models/WafPolicySettings.cs#L44). Unfortunately, currently our serialization code will fail in such case because we will try to [parse](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/cdn/Azure.ResourceManager.Cdn/src/Generated/Models/WafPolicySettings.Serialization.cs#L51-L55) the string, but it isn't a valid json value.

Considering that this has impact on the packages with stable version, I think we should fix it ASAP.

Yao725:
# Description
Fixes #2447 

The original implementation of BinaryData was intended for `AnyObject` which can have any structure but must be a valid json.  It was expanded to include `Any` but when that happened we didn't adjust the serialization to handle the additional cases.

This PR is exploring options on how to make `Any` work in the serialization code.  The deserialization should remain untouched as no interpretation of the data happens during deserialization we simply store the raw bytes into the BinaryData and let the customer choose how to interpret the data.

For serialization we need to inspect the BinaryData and determine if it should be written as raw bytes (`json`) or as one of the primitive types `null`, `bool`, `number`, `string`.

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
@@ -288,13 +288,48 @@ namespace AutoRest.CSharp.Generation.Writers
 
         private FormattableString CreatePropertyDescription(ObjectTypeProperty property, string? overrideName = null)
         {
+            FormattableString binaryDataExtraDescription = Configuration.AzureArm &&
+                property.Declaration.Type.IsFrameworkType &&
+                property.Declaration.Type.FrameworkType == typeof(BinaryData) ?
+                CreateBinaryDataExtraDescription() : $"";
             if (!string.IsNullOrWhiteSpace(property.Description))
             {
-                return $"{property.Description}";
+                return $"{property.Description}{binaryDataExtraDescription}";
             }
-            return property.CreateDefaultPropertyDescription(overrideName);
+            return $"{property.CreateDefaultPropertyDescription(overrideName)}{binaryDataExtraDescription}";
         }
 
+        private FormattableString CreateBinaryDataExtraDescription()
+        {
+            return $@".
+<para>
+To assign an object to this property use <see cref=""{typeof(BinaryData)}.FromObjectAsJson{{T}}(T, System.Text.Json.JsonSerializerOptions?)""/>.
+</para>
+<para>
+To assign an already formated json string to this property use <see cref=""{typeof(BinaryData)}.FromString(string)""/>.
+</para>
+<para>
+Examples:
+<list type=""bullet"">
+<item>
+<term>BinaryData.FromObjectAsJson(""foo"")</term>
+<description>Creates a payload of ""foo"".</description>
+</item>
+<item>
+<term>BinaryData.FromString(""\""foo\"""")</term>
+<description>Creates a payload of ""foo"".</description>
+</item>
+<item>
+<term>BinaryData.FromObjectAsJson(new {{ key = ""value"" }})</term>
+<description>Creates a payload of {{ ""key"": ""value"" }}.</description>
+</item>
+<item>
+<term>BinaryData.FromString(""{{\""key\"": \""value\""}}"")</term>
+<description>Creates a payload of {{ ""key"": ""value"" }}.</description>
+</item>
+</list>
+</para>";
+        }
         private string GetAbstract(SchemaObjectType schema)
         {
             return schema.IsAbstract ? "abstract " : string.Empty;
--- a/src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
@@ -288,8 +288,7 @@ namespace AutoRest.CSharp.Generation.Writers
 
         private FormattableString CreatePropertyDescription(ObjectTypeProperty property, string? overrideName = null)
         {
-            FormattableString binaryDataExtraDescription = Configuration.AzureArm &&
-                property.Declaration.Type.IsFrameworkType &&
+            FormattableString binaryDataExtraDescription = property.Declaration.Type.IsFrameworkType &&
                 property.Declaration.Type.FrameworkType == typeof(BinaryData) ?
                 CreateBinaryDataExtraDescription() : $"";
             if (!string.IsNullOrWhiteSpace(property.Description))
--- a/samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineCaptureResult.cs
+++ b/samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineCaptureResult.cs
@@ -64,7 +64,34 @@ namespace Azure.ResourceManager.Sample.Models
         public string ContentVersion { get; }
         /// <summary>
         /// parameters of the captured virtual machine
-        /// Serialized Name: VirtualMachineCaptureResult.parameters
+        /// Serialized Name: VirtualMachineCaptureResult.parameters.
+        /// <para>
+        /// To assign an object to this property use <see cref="BinaryData.FromObjectAsJson{T}(T, System.Text.Json.JsonSerializerOptions?)"/>.
+        /// </para>
+        /// <para>
+        /// To assign an already formated json string to this property use <see cref="BinaryData.FromString(string)"/>.
+        /// </para>
+        /// <para>
+        /// Examples:
+        /// <list type="bullet">
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson("foo")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("\"foo\"")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson(new { key = "value" })</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// </list>
+        /// </para>
         /// </summary>
         public BinaryData Parameters { get; }
         /// <summary>
--- a/samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineExtensionUpdate.cs
+++ b/samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineExtensionUpdate.cs
@@ -52,12 +52,66 @@ namespace Azure.ResourceManager.Sample.Models
         public bool? EnableAutomaticUpgrade { get; set; }
         /// <summary>
         /// Json formatted public settings for the extension.
-        /// Serialized Name: VirtualMachineExtensionUpdate.properties.settings
+        /// Serialized Name: VirtualMachineExtensionUpdate.properties.settings.
+        /// <para>
+        /// To assign an object to this property use <see cref="BinaryData.FromObjectAsJson{T}(T, System.Text.Json.JsonSerializerOptions?)"/>.
+        /// </para>
+        /// <para>
+        /// To assign an already formated json string to this property use <see cref="BinaryData.FromString(string)"/>.
+        /// </para>
+        /// <para>
+        /// Examples:
+        /// <list type="bullet">
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson("foo")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("\"foo\"")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson(new { key = "value" })</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// </list>
+        /// </para>
         /// </summary>
         public BinaryData Settings { get; set; }
         /// <summary>
         /// The extension can contain either protectedSettings or protectedSettingsFromKeyVault or no protected settings at all.
-        /// Serialized Name: VirtualMachineExtensionUpdate.properties.protectedSettings
+        /// Serialized Name: VirtualMachineExtensionUpdate.properties.protectedSettings.
+        /// <para>
+        /// To assign an object to this property use <see cref="BinaryData.FromObjectAsJson{T}(T, System.Text.Json.JsonSerializerOptions?)"/>.
+        /// </para>
+        /// <para>
+        /// To assign an already formated json string to this property use <see cref="BinaryData.FromString(string)"/>.
+        /// </para>
+        /// <para>
+        /// Examples:
+        /// <list type="bullet">
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson("foo")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("\"foo\"")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson(new { key = "value" })</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// </list>
+        /// </para>
         /// </summary>
         public BinaryData ProtectedSettings { get; set; }
     }
--- a/samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineScaleSetExtensionPatch.cs
+++ b/samples/Azure.ResourceManager.Sample/Generated/Models/VirtualMachineScaleSetExtensionPatch.cs
@@ -115,12 +115,66 @@ namespace Azure.ResourceManager.Sample.Models
         public bool? EnableAutomaticUpgrade { get; set; }
         /// <summary>
         /// Json formatted public settings for the extension.
-        /// Serialized Name: VirtualMachineScaleSetExtensionUpdate.properties.settings
+        /// Serialized Name: VirtualMachineScaleSetExtensionUpdate.properties.settings.
+        /// <para>
+        /// To assign an object to this property use <see cref="BinaryData.FromObjectAsJson{T}(T, System.Text.Json.JsonSerializerOptions?)"/>.
+        /// </para>
+        /// <para>
+        /// To assign an already formated json string to this property use <see cref="BinaryData.FromString(string)"/>.
+        /// </para>
+        /// <para>
+        /// Examples:
+        /// <list type="bullet">
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson("foo")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("\"foo\"")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson(new { key = "value" })</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// </list>
+        /// </para>
         /// </summary>
         public BinaryData Settings { get; set; }
         /// <summary>
         /// The extension can contain either protectedSettings or protectedSettingsFromKeyVault or no protected settings at all.
-        /// Serialized Name: VirtualMachineScaleSetExtensionUpdate.properties.protectedSettings
+        /// Serialized Name: VirtualMachineScaleSetExtensionUpdate.properties.protectedSettings.
+        /// <para>
+        /// To assign an object to this property use <see cref="BinaryData.FromObjectAsJson{T}(T, System.Text.Json.JsonSerializerOptions?)"/>.
+        /// </para>
+        /// <para>
+        /// To assign an already formated json string to this property use <see cref="BinaryData.FromString(string)"/>.
+        /// </para>
+        /// <para>
+        /// Examples:
+        /// <list type="bullet">
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson("foo")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("\"foo\"")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson(new { key = "value" })</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// </list>
+        /// </para>
         /// </summary>
         public BinaryData ProtectedSettings { get; set; }
         /// <summary>
--- a/samples/Azure.ResourceManager.Sample/Generated/VirtualMachineExtensionData.cs
+++ b/samples/Azure.ResourceManager.Sample/Generated/VirtualMachineExtensionData.cs
@@ -115,12 +115,66 @@ namespace Azure.ResourceManager.Sample
         public bool? EnableAutomaticUpgrade { get; set; }
         /// <summary>
         /// Json formatted public settings for the extension.
-        /// Serialized Name: VirtualMachineExtension.properties.settings
+        /// Serialized Name: VirtualMachineExtension.properties.settings.
+        /// <para>
+        /// To assign an object to this property use <see cref="BinaryData.FromObjectAsJson{T}(T, System.Text.Json.JsonSerializerOptions?)"/>.
+        /// </para>
+        /// <para>
+        /// To assign an already formated json string to this property use <see cref="BinaryData.FromString(string)"/>.
+        /// </para>
+        /// <para>
+        /// Examples:
+        /// <list type="bullet">
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson("foo")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("\"foo\"")</term>
+        /// <description>Creates a payload of "foo".</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromObjectAsJson(new { key = "value" })</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// </list>
+        /// </para>
         /// </summary>
         public BinaryData Settings { get; set; }
         /// <summary>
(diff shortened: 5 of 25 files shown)


src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +140,26 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
                     throw new NotSupportedException("Not supported type " + value.GetType());
             }
         }
+
+        public static void WriteBinaryDataValue(this Utf8JsonWriter writer, BinaryData value)
+        {
+            string inputValue = value.ToString();
+            // First we check if it is possible to serialize the value as json value.
+            if (inputValue.StartsWith("{", StringComparison.InvariantCultureIgnoreCase) && (inputValue.EndsWith("}", StringComparison.InvariantCultureIgnoreCase) || inputValue.EndsWith($"}}{Environment.NewLine}", StringComparison.InvariantCultureIgnoreCase)) ||
+                inputValue.StartsWith("[", StringComparison.InvariantCultureIgnoreCase) && (inputValue.EndsWith("]", StringComparison.InvariantCultureIgnoreCase) || inputValue.EndsWith($"]{Environment.NewLine}", StringComparison.InvariantCultureIgnoreCase)) ||
```
ArthurMa1978, line 149: consider to have a perf test

Yao725: I tested the performance using implementation with try-catch and without try-catch respectively and below is the result.

- When the BinaryData is a json object

|         Method |           Job |       Runtime |       Mean |    Error |   StdDev |     Median |        Min |        Max |  Gen 0 | Allocated |
|--------------- |-------------- |-------------- |-----------:|---------:|---------:|-----------:|-----------:|-----------:|-------:|----------:|
| TestWithoutTry |      .NET 6.0 |      .NET 6.0 |   920.7 ns |  2.85 ns |  2.53 ns |   920.4 ns |   917.1 ns |   924.7 ns | 0.1574 |      1 KB |
|    TestWithTry |      .NET 6.0 |      .NET 6.0 |   952.4 ns | 10.76 ns | 10.07 ns |   954.7 ns |   937.2 ns |   968.7 ns | 0.1574 |      1 KB |
| TestWithoutTry | .NET Core 3.1 | .NET Core 3.1 | 1,576.0 ns | 23.19 ns | 21.69 ns | 1,566.8 ns | 1,551.4 ns | 1,621.7 ns | 0.1583 |      1 KB |
|    TestWithTry | .NET Core 3.1 | .NET Core 3.1 | 1,618.4 ns | 33.16 ns | 94.60 ns | 1,568.3 ns | 1,541.2 ns | 1,862.0 ns | 0.1583 |      1 KB |

- When the BinaryData is a string and we correctly treat it as string

|         Method |           Job |       Runtime |     Mean |   Error |  StdDev |      Min |      Max |  Gen 0 |  Gen 1 | Allocated |
|--------------- |-------------- |-------------- |---------:|--------:|--------:|---------:|---------:|-------:|-------:|----------:|
| TestWithoutTry |      .NET 6.0 |      .NET 6.0 | 201.0 ns | 0.62 ns | 0.55 ns | 200.1 ns | 202.1 ns | 0.1013 | 0.0002 |     848 B |
|    TestWithTry |      .NET 6.0 |      .NET 6.0 | 201.5 ns | 0.75 ns | 0.70 ns | 200.6 ns | 203.2 ns | 0.1013 | 0.0002 |     848 B |
| TestWithoutTry | .NET Core 3.1 | .NET Core 3.1 | 312.0 ns | 1.46 ns | 1.30 ns | 309.8 ns | 314.2 ns | 0.1020 |      - |     856 B |
|    TestWithTry | .NET Core 3.1 | .NET Core 3.1 | 333.8 ns | 3.62 ns | 3.21 ns | 328.7 ns | 338.1 ns | 0.1020 |      - |     856 B |

Based on above results, we can find that if we add a try-catch in this method, there will a little performance problem(however, sometimes the time are even very close to each other), but it can ensure that we could correctly handle the situation that a binary data is actually a string but we see it as a json. To be honest, I think this kind of case will be very rare, but we don't have the confidence to say it will never exist, therefore I think it's worth sacrificing a little performance here.

annelo-msft (commented):
Could you share what issue this is addressing?  I'm curious what motivates this addition.

Yao725:
> Could you share what issue this is addressing? I'm curious what motivates this addition.

Oh, my bad, I forgot to link the [issue](https://github.com/Azure/autorest.csharp/issues/2447) in the PR description. The context is that previously we changed the type from object to BinaryData and will parse it as we regard it as a valid json value. And in the process of releasing some management plane SDKs, we also have changed the type of some properties from string to BinaryData. At this time, the original serialization code will not work. Therefore, we want to have this change.

annelo-msft:
Great, thank you for the explanation!

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +140,33 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
                     throw new NotSupportedException("Not supported type " + value.GetType());
             }
         }
+
+        public static void WriteBinaryDataValue(this Utf8JsonWriter writer, BinaryData value)
```
AlexanderSher, line 144: Why not make it a case in `WriteObjectValue`?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
AlexanderSher: How big was the underlying byte array?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +140,33 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+            {
+                try
+                {
+#if NET6_0_OR_GREATER
+				    writer.WriteRawValue(value);
+#else
+                    JsonSerializer.Serialize(writer, JsonDocument.Parse(inputValue).RootElement);
+#endif
+                }
+                catch
```
AlexanderSher, line 161: Please specify the exception you're catching here.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +140,33 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
                     throw new NotSupportedException("Not supported type " + value.GetType());
             }
         }
+
+        public static void WriteBinaryDataValue(this Utf8JsonWriter writer, BinaryData value)
+        {
+            string inputValue = value.ToString();
```
AlexanderSher, line 146: I don't like that we need to create a copy of the `BinaryData` content. Can we cast `BinaryData` to `Span<byte>` and then use [UTF8Encoding.GetChars](https://docs.microsoft.com/en-us/dotnet/api/system.text.utf8encoding.getchars?view=net-6.0)?

AlexanderSher: Or maybe even use `Utf8JsonReader` to validate the span?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
ArthurMa1978: Object is Json Object, but here the BinaryData can ben raw string or raw binary

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: 1. Vaule: `{"foo":"bar","numbers":[1,2,3,4]}`, Length: 33
2. Vaule: `abcdef`, Length: 6

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: I think there are two reasons that we don't want to make it a case in `WriteObjectValue`:
1. In mgmt plane, we explicitly know that the input parameter is a BinaryData, so we don't need the type check in `WriteObjectValue`
2. `WriteObjectValue` already has a case about `BinaryData`, which writes the value as Base64 encoded json string, I don't think we want to break current algorithm.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: specified.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: I change it to ReadOnlySpan<char> using `GetChars`.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,34 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
                     throw new NotSupportedException("Not supported type " + value.GetType());
             }
         }
+
+        public static void WriteBinaryDataValue(this Utf8JsonWriter writer, BinaryData value)
+        {
+            ReadOnlySpan<char> inputValue = UTF8Encoding.UTF8.GetChars(value.ToArray()).AsSpan();
+            // Check if it is possible to serialize the value as json value.
+            // Possible match kinds: Json Object, Json Array, Json value false, Json value true, Json number.
+            if (inputValue.StartsWith("{".AsSpan(), StringComparison.InvariantCultureIgnoreCase) && (inputValue.EndsWith("}".AsSpan(), StringComparison.InvariantCultureIgnoreCase) || inputValue.EndsWith($"}}{Environment.NewLine}".AsSpan(), StringComparison.InvariantCultureIgnoreCase)) ||
```
m-nash, line 150: lets use index comparison vs startswith

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,34 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+                inputValue.StartsWith("[".AsSpan(), StringComparison.InvariantCultureIgnoreCase) && (inputValue.EndsWith("]".AsSpan(), StringComparison.InvariantCultureIgnoreCase) || inputValue.EndsWith($"]{Environment.NewLine}".AsSpan(), StringComparison.InvariantCultureIgnoreCase)) ||
```
m-nash, line 151: instead of endswith `}Newline` lets start at the end and move inwards until we no longer have whitespace then use that as the last character

test/AutoRest.Shared.Tests/WriterExtensionTests.cs
```diff
@@ -37,6 +38,20 @@ public class WriterExtensionTests
             },
         };
 
+        private static object[] BinaryDataCases =
+        {
+            new object[] { BinaryData.FromString("abcdefg"), "\"abcdefg\"" },
+            new object[] { BinaryData.FromString("123"), "123" },
```
m-nash, line 44: lets add test cases around .FromObject(123)

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,34 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+#endif
+                }
+                catch (JsonException)
+                {
+                    writer.WriteStringValue(inputValue);
+                }
+            }
+            else
+            {
+                writer.WriteStringValue(inputValue);
```
m-nash, line 170: If we are always writing string value do we ever want to be able to serialize `"foo": 51`?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: updated

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: I have added a trim logic and we don't need this NewLine check anymore.

test/AutoRest.Shared.Tests/WriterExtensionTests.cs
Yao725: Added

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: Our logic here will check if the first byte and last byte is between `0` and `9`, if it is, we will parse it as json number. There fore if you use an integer to instantiate the BinaryData, no matter you do it by `BinaryData.FromString("51")` or `BinaryData.FromObjectAsJson(51)`, the result will be `"foo": 51`

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,53 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
                     throw new NotSupportedException("Not supported type " + value.GetType());
             }
         }
+
+        public static void WriteBinaryDataValue(this Utf8JsonWriter writer, BinaryData value)
+        {
+            byte[] inputValue = value.ToArray();
+
+            // We will remove the trailing occurrences of the specified characters including: TAB, LF, space.
+            byte[] trimBytes = { 9, 10, 32 };
```
m-nash, line 150: can we use `char.IsWhiteSpace` here?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,53 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+                }
+                if (i == trimBytes.Length)
+                    break;
+            }
+
+            // Check if it is possible to serialize the value as json value.
+            // Possible match kinds: Json Object, Json Array, Json value false, Json value true, Json number.
+            // Our logic here is that, for example, we check if the first and the last bytes are '{' and '}', respectively.
+            if (end >= 0 && (
+                inputValue[0] == 123 && inputValue[end] == 125 ||
```
m-nash, line 168: can we add constants so we know what the ints stand for?

test/AutoRest.Shared.Tests/WriterExtensionTests.cs
```diff
@@ -6,6 +6,7 @@
 using System.IO;
 using System.Text.Json;
 using NUnit.Framework;
+using JsonObject = System.Collections.Generic.Dictionary<string, object>;
```
m-nash, line 9: Can we add test cases that test the boundaries where we start with '{' but don't end with '}' as well as the other combos we are looking for
Add test cases with various trailing whitespace
Cases with all whitespace

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,53 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
                     throw new NotSupportedException("Not supported type " + value.GetType());
             }
         }
+
+        public static void WriteBinaryDataValue(this Utf8JsonWriter writer, BinaryData value)
+        {
+            byte[] inputValue = value.ToArray();
```
m-nash, line 147: Does this create a copy?  If so is there a way to access the raw bytes without a copy?

test/AutoRest.Shared.Tests/WriterExtensionTests.cs
```diff
@@ -87,6 +103,17 @@ public static void WriteObjectValueNullIUtf8JsonSerializable ()
...
 
+        [Test, TestCaseSource("BinaryDataCases")]
+        public static void WriteBinaryDataValue(BinaryData value, string expectedValue)
+        {
+            using MemoryStream stream = new MemoryStream();
+            using Utf8JsonWriter writer = new Utf8JsonWriter(stream);
+            writer.WriteBinaryDataValue(value);
+
+            writer.Flush();
+            Assert.AreEqual(expectedValue, System.Text.Encoding.UTF8.GetString(stream.ToArray()));
```
m-nash, line 114: Can we add round trip tests to serialize, then deserialize and validate the original byte array is equal to the new byte array.

I don't know if it will make a difference but can we also validate these cases while they are a property of a larger object?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,53 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+                    break;
+            }
+
+            // Check if it is possible to serialize the value as json value.
+            // Possible match kinds: Json Object, Json Array, Json value false, Json value true, Json number.
+            // Our logic here is that, for example, we check if the first and the last bytes are '{' and '}', respectively.
+            if (end >= 0 && (
+                inputValue[0] == 123 && inputValue[end] == 125 ||
+                inputValue[0] == 91 && inputValue[end] == 93 ||
+                inputValue[0] == 116 && inputValue[end] == 101 ||
```
m-nash, line 170: we are saying that if the raw bytes = 't', ..., 'e' we will treat this as json?  Lets add a test case that sets the bytes equal to `terse`.

I am still not fully sold on the true, false, and number checks here.  If we send them to write as json 456 comes out as "456" according to the test case [here](https://github.com/Azure/autorest.csharp/pull/2544/files#diff-ede2da49b9fb105e853158da8dfe79723aa96156ae93c3577a7b0485fdf24c39R45)?  If we send it to write as string doesn't 456 also become "456"?

If we want to actually do anything here for true / false lets check length then validate the full word.

m-nash (changes requested):

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,53 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
                     throw new NotSupportedException("Not supported type " + value.GetType());
             }
         }
+
+        public static void WriteBinaryDataValue(this Utf8JsonWriter writer, BinaryData value)
```
annelo-msft, line 145: It would be good to get the Azure.Core v-team's eyes on this change per the change to Azure.Core shared source.  @jsquire, @JoshLove-msft, @christothes

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: changed

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: changed

test/AutoRest.Shared.Tests/WriterExtensionTests.cs
Yao725: added

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: Change to use ReadOnlySpan<byte> to avoid copying

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: I have added the test case. 

If we write 456 as string, the result will be "\"456\"", not "456". 

I understand the we can check the length and validate the full word for true/false, but what I want to say is, we are using the first and last byte to check if the binary data is likely to be a valid json and we should keep this pattern for the different json kinds. In 90% of the cases, we will pass this check and the json parse will succeed, that's what we are expecting. And we are clear that there are still some cases that the binary data can pass this check but it can't be parsed, therefore we will catch the exception and write it as string.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,52 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+            int end = inputValue.Length - 1;
+            while (end >= start && char.IsWhiteSpace(Convert.ToChar(inputValue[end])))
+            {
+                end--;
+            }
+            int len = end - start + 1;
+
+            // Check if it is possible to serialize the value as json value.
+            // Possible match kinds: Json Object, Json Array, Json string, Json value true, Json value false, Json number.
+            // Our logic here is that, for example, we check if the first and the last bytes are '{' and '}', respectively.
```
christothes, line 164: Is this heuristic enough? What about values like `{   this{ is { invalid { josn }` ?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,52 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+            if (len > 0 && (
```
jsquire, line 165: The allow-list approach makes me uncomfortable.   For example, this would consider `null` invalid and thus rule out valid patterns like:

```json
"someKey" : null
```

What problem are we trying to solve here?   Why do we need this level of validation rather than just writing out the bytes?   `Utf8JsonWriter` will validate that the JSON is structurally valid by default, unless you're setting the [SkipValidation](https://docs.microsoft.com/dotnet/api/system.text.json.jsonwriteroptions.skipvalidation?view=net-6.0#system-text-json-jsonwriteroptions-skipvalidation) option.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,52 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+            ReadOnlySpan<byte> inputValue = value;
+
+            // We will ignore the leading and trailing occurrences of the white space in the json check.
+            int start = 0;
+            while (start < inputValue.Length && char.IsWhiteSpace(Convert.ToChar(inputValue[start])))
+            {
+                start++;
+            }
+            int end = inputValue.Length - 1;
+            while (end >= start && char.IsWhiteSpace(Convert.ToChar(inputValue[end])))
```
jsquire, line 156: Why do we care to trim whitespace?  `Utf8JsonWriter` is going to collapse all whitespace and indentation by default unless we're explicitly setting the [Indented](https://docs.microsoft.com/dotnet/api/system.text.json.jsonwriteroptions.indented?view=net-6.0#system-text-json-jsonwriteroptions-indented) option.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
christothes: Taking a step back, is it even valid to store use `BinaryData` to store values that cannot be serialized to json?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
jsquire, line 156: `Convert.ToChar` just does a straight `(char)` cast; I'd recommend skipping the method call and just doing the cast.

_([src](https://source.dot.net/#System.Private.CoreLib/Convert.cs,05d59fb918111652))_

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
```diff
@@ -140,5 +141,52 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
...
+            int len = end - start + 1;
+
+            // Check if it is possible to serialize the value as json value.
+            // Possible match kinds: Json Object, Json Array, Json string, Json value true, Json value false, Json number.
+            // Our logic here is that, for example, we check if the first and the last bytes are '{' and '}', respectively.
+            if (len > 0 && (
+                inputValue[start] == (byte)'{' && inputValue[end] == (byte)'}' ||
+                inputValue[start] == (byte)'[' && inputValue[end] == (byte)']' ||
+                inputValue[start] == (byte)'"' && inputValue[end] == (byte)'"' ||
+                inputValue[start] == (byte)'t' && inputValue[end] == (byte)'e' ||
```
jsquire, line 169: What if my value is:

- telephone
- fortnite
- 9one1
- ["invalid]
- {"no]}

Unless I'm misreading, these would all pass this validation.

jsquire:
It would be quite helpful if this PR had some context in the description for what benefit this change is intending to provide.   The branch is called "fix-binarydata-serialization" ...  can you help us understand in what way it is broken?

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
christothes: oh -  I see now that we handle the edge case in the catch block

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
m-nash: We are trying to do a "fast" check to see if we should call WriteJson or WriteString.  If we incorrectly call WriteString for json it comes out as `"{"foo": "bar"}"` with extra quotes around it which is not correct.  We could just try to parse with JsonElement and catch always but this is very slow.  We could also use JsonReader which is bit faster.  Instead we believe we can be 90%+ accurate with this method and therefore only incur the slow parsing fail in a small percentage of cases.  Our worst case runtime is no worse than always doing a full parse, but our average case is significantly better.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
m-nash: Good call out on `null`.

Json and String are very simple to deal with, the challenge that we are attempting to handle here are the other primitives namely `null`, `number`, `bool`.

If we want to only use BinaryData for `AnyObject` we can safely just serialize as json which is the current behavior.  If we want to expand its use to `Any` as well we must deal with these other cases also.

Deserialization is easy as no interpretation happens we simply store the raw bytes in the binary data and let the customer choose how to interpret them.  Serialization is where we now need to know how to write out what was given to us.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
m-nash: The first three bullets are the same comment [here](https://github.com/Azure/autorest.csharp/pull/2544#discussion_r943715228).  The last two would be handled by the catch block and written as `"["invalid]"` and `"{"no]}"` respectively.  Again the pre-parse is not meant to be 100% accurate but its simply meant to lower our average case by eliminating things that we know for sure are NOT valid json.

test/AutoRest.Shared.Tests/WriterExtensionTests.cs
Yao725: I have added a test case in MgmtExtensionResourceTests to see if the result is expected when we first serialize the model, which contains a binary data property and then deserialize it.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: Yes, we think that this check will be right in most of the cases and use catch to handle the worst case, in which the `BinaryData` passed the check but isn't a valid json. We use this fast check to avoid trying to parse the json everytime and then catch the exception.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: Added `null` to the check list.

src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs
Yao725: changed, thank you!

m-nash:
@Yao725 based on our findings lets pivot this PR to address the following https://github.com/Azure/autorest.csharp/issues/2447#issuecomment-1217234786 and we will look into what else we can do for documentation on BinaryData itself in the future.

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
```diff
@@ -288,13 +288,48 @@ private void BuildHierarchy(ObjectTypeProperty property, Stack<ObjectTypePropert
 
         private FormattableString CreatePropertyDescription(ObjectTypeProperty property, string? overrideName = null)
         {
+            FormattableString binaryDataExtraDescription = Configuration.AzureArm &&
```
m-nash, line 291: remove the if arm check I believe this should apply to all libraries

m-nash (approved):

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
Yao725: removed

Issue #2490, opened 2022-07-26
Generate models with collections of model types from Cadl


annelo-msft:
### Round-trip model with collection type properties
This issue tracks the work to implement generation of round-trip models with that have properties that are collections of models.

Input and output models with properties that are collections of models should be generated as well as P1 (rather than P0), based on generalization of the rules described here and for input/output models with collection properties.

#### Input Cadl file
[Cadl testserver models/collections-models](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/models/collections-models/main.cadl#L36-L44)

#### Expected output

- [x] [ModelCollectionModel.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-test-collection-models/sdk/template/Azure.Template/src/Generated/Models/ModelCollectionModel.cs)
- [x] [ModelCollectionModel.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-test-collection-models/sdk/template/Azure.Template/src/Generated/Models/ModelCollectionModel.Serialization.cs)
- [x] Constructor takes initialization value for required `IEnumerable<SimpleModel>` property
- [x] Constructor initializes optional collection to `ChangeTrackingList<SimpleModel>`
- [x] Model collection properties are both of type `IList<SimpleModel>`
- [x] ModelCollectoinModel inherits from `IUtf8JsonSerializable`
- [x] IUtf8JsonSerializable.Write method is defined in serialization partial class
- [x] Write method calls `Utf8JsonWriterExtensions.WriteObjectValue` to serialize items in collection
- [x] Deserialize method is defined in serialization partial class
- [x] SimpleModel.DeserializeSimpleModel method is called to deserialize collection items

annelo-msft:
We make the assumption that models in the collection are serializable, which may not be true for user-defined schemas like we have in search, per @tg-msft, but these are out of scope for the August MVP.  We may want to revisit when we get to https://github.com/Azure/autorest.csharp/issues/2518.

lirenhe:
@archerzz, plesae help to verify whether our current collection generation could already cover models.  Thanks

archerzz:
I think we have a blocking issue: there is a circular dependency between `ModelTypeProvider => TypeFactory => OutputLibrary` for collections of model type and probably (I haven't test but pretty sure) nested model type.

[...]

Issue #2339, opened 2022-06-21
Generate models with optional properties from Cadl


annelo-msft:
### Input/Output/Round-trip models with optional properties
This issue tracks the work to implement generation of models with optional reference, value, and colleciton properties in the DPG generator.

If we need to triage this work, generation of Round-trip models is P0, and generation of Input and Output models is P1.

#### Input Cadl file
[Cadl testserver models/optional-properties](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/optional-properties)

#### Expected output
InputModel files:
- [x] [InputModel.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-models-optional-properties/sdk/template/Azure.Template/src/Generated/Models/InputModel.cs)
- [x] [InputModel.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-models-optional-properties/sdk/template/Azure.Template/src/Generated/Models/InputModel.Serialization.cs)
- [x] Input models have only one public constructor, and it takes no values (since there are no required values)
- [x] Optional primitive properties have both getter and setter
- [x] Optional list is initialized in constructor as ChangeTrackingList<T>
- [x] Optional list property is represented as IList<T>
- [x] Optional list property has only a getter
- [x] Write operation is defined in serialization partial class
- [x] Optional properties are written only if Optional.IsDefined(propertyName) check succeeds
- [x] Optional collection properties are written only if Optional.IsCollectionDefined(propertyName) check succeeds
- [x] Optional collection [is not serialized if not mutated](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/ModelShapeTests.cs#L269)
- [x] Optional collection [is serialized if mutated](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/ModelShapeTests.cs#L290)
- [x] Optional collection [is serialized if empty and mutated](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/ModelShapeTests.cs#L304)
- [x] Optional collection [is not serialized if reset](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/ModelShapeTests.cs#L320)

OutputModel files:
[...]

PR #2557, opened 2022-08-09
feat(dpg): support collection of model types

archerzz:
1. adopt lazy evaluation for fields related properties of `ModelTypeProvider`
    - the problem is that for nested types, we cannot resolve its type in the library
    - so if a field is a custom model, it cannot be initialized along with the parent type
2. update `JsonCodeWriterExentsion` to support serialization/deserialization codes of array element model types
3. add and refactor unit tests, move collection related tests into a separate test case

part of #2490

# Description

# Checklist

# Ready to Land?


archerzz:
Optional collection properties will be handled separately, probably along with the work of #2339

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -20,35 +20,108 @@ internal sealed class ModelTypeProvider : TypeProvider
...
+
+                _fieldsInitializer();
+                return _parameterNamesToFields!;
+            }
+        }
+
+        // lazy envaluation
+        private readonly Action _fieldsInitializer;
+        private IReadOnlyList<FieldDeclaration>? _fields;
+        public ConstructorSignature? _publicConstructor;
```
lirenhe, line 95: why are they public?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -20,35 +20,108 @@ internal sealed class ModelTypeProvider : TypeProvider
...
+                    return _parameterNamesToFields;
+                }
+
+                _fieldsInitializer();
+                return _parameterNamesToFields!;
+            }
+        }
+
+        // lazy envaluation
+        private readonly Action _fieldsInitializer;
```
lirenhe, line 93: is it better to name it _modelInitializer?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
archerzz: A mistake :)

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
archerzz: The `ModelTypeProvider` is partially useful after the constructor is invoked. What remains lazy evaluation is the fields related properties. So I name it after `_fieldsInitializer`.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -20,35 +20,108 @@ internal sealed class ModelTypeProvider : TypeProvider
...
         public ModelTypeProvider(InputModelType inputModel, TypeFactory typeFactory, string defaultNamespace, SourceInputModel? sourceInputModel)
             : base(inputModel.Namespace ?? defaultNamespace, sourceInputModel)
         {
             DefaultName = inputModel.Name;
             DefaultAccessibility = inputModel.Accessibility ?? "public";
 
-            (_fieldsToInputs, var serializationParameters, _parameterNamesToFields) = CreateParametersAndFieldsForRoundTripModel(inputModel, typeFactory);
+            _fieldsInitializer = delegate()
+            {
+                (_fieldsToInputs, var serializationParameters, _parameterNamesToFields) = CreateParametersAndFieldsForRoundTripModel(inputModel, typeFactory);
```
AlexanderSher, line 109: It this also because we don't have access to field names upfront?

archerzz: > It this also because we don't have access to field names upfront?

Sort of. Each field of a `ModelTypeProvider` is a `TypeProvider`. We resolve them in https://github.com/Azure/autorest.csharp/blob/c3e7e60ef50975a5dbc21c9ddd3eb484522f1949/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs#L84

The problem is that at that time the `_library` is not initialized. There are basically two options here:
- Use lazy evaluation, this is the approach I chose
- Need to traverse the type maps and get the dependency sequence. Then initialize from the types with least dependency first.

AlexanderSher: I've added simple model support in 2559: https://github.com/Azure/autorest.csharp/pull/2559/files#diff-d8015b998fd7c30224b9fff88beb8b080bcb5f7552c821d861c16c35cbdc83c8R31

Can you please try in after 2559 is merged, so we don't have to use lazy initialization here?

Full "model with model properties" support isn't there yet.

archerzz: > I've added simple model support in 2559: [#2559 (files)](https://github.com/Azure/autorest.csharp/pull/2559/files#diff-d8015b998fd7c30224b9fff88beb8b080bcb5f7552c821d861c16c35cbdc83c8R31)
> 
> Can you please try in after 2559 is merged, so we don't have to use lazy initialization here?
> 
> Full "model with model properties" support isn't there yet.

Verified in https://github.com/Azure/autorest.csharp/pull/2572 Thanks.

archerzz:
close this PR, since we've already supported collection of model types. The test cases are added in https://github.com/Azure/autorest.csharp/pull/2572

Issue #2556, opened 2022-08-08
Generate convenience methods with extensible enum parameters


annelo-msft:
### Convenience methods with extensible enum parameters

This issue tracks the work to generate convenience methods for operations described in a Cadl file that have extensible enum parameters.

#### Expected Input

The design is as follows:
- Protocol methods are generated with extensible enum parameter as string type
- Convenience methods are generated with extensible enum parameter as the extensible enum type

#### Expected Output

See the example in [ImportDevices method](https://gist.github.com/annelo-msft/8582310e0aab5c3dfa378de2c11849bb#devicemanagementclientimportdevices) where the protocol method takes `string action` and the convenience method takes `ImportAction action`.

Please create a test Cadl file in [cadl-ranch clients/parameters](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/clients) to illustrate this case.

PR #2559, opened 2022-08-10
Fix enum support

AlexanderSher:
- Fixes https://github.com/Azure/autorest.csharp/issues/2535
- Fixes https://github.com/Azure/autorest.csharp/issues/2556 (but tests in CADL-ranch are still required)
- Adds correct mapping between convenience and protocol method parameters
- Adds a feature to `CodeWriter` to enable pre-declaration of variables (`CodeWriter` will throw an exception if name is already available in any of the existing scopes)


src/AutoRest.CSharp/Common/Generation/Writers/CodeWriter.cs
```diff
@@ -36,12 +36,14 @@ public CodeWriter()
             _scopes.Push(new CodeWriterScope(this, "", false));
         }
 
-        public CodeWriterScope Scope(FormattableString line, string start = "{", string end = "}", bool newLine = true)
+        public CodeWriterScope Scope(FormattableString line, string start = "{", string end = "}", bool newLine = true, CodeWriterScopeDeclarations? scopeDeclarations = null)
```
AlexanderSher, line 39: @archerzz This is a part you can grab into the https://github.com/Azure/autorest.csharp/pull/2554 (or just wait until this one is merged).

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
```diff
@@ -98,6 +102,10 @@ private ClientFields(IEnumerable<Parameter> parameters, InputAuth? authorization
                     fields.Add(field);
                 }
                 parameterNamesToFields.Add(parameter.Name, field);
+                if (parameter.Name == "endpoint" && parameter.Type.Equals(typeof(Uri)))
```
AlexanderSher, line 105: Not a good option, perhaps we need to make `endpoint` a known parameter,

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
```diff
@@ -109,6 +117,7 @@ private ClientFields(IEnumerable<Parameter> parameters, InputAuth? authorization
             _fields = fields;
             _parameterNamesToFields = parameterNamesToFields;
             CredentialFields = credentialFields;
+            ScopeDeclarations = new CodeWriterScopeDeclarations(fields.Select(f => f.Declaration));
```
AlexanderSher, line 120: @archerzz This is where `ScopeDeclarations` is created to be used in scopes later

src/AutoRest.CSharp/Common/Output/Models/LongRunningOperation.cs
```diff
@@ -35,7 +35,7 @@ protected LongRunningOperation(InputOperation operation, BuildContext context, L
             if (finalResponseType != null)
             {
                 ResultType = TypeFactory.GetOutputType(context.TypeFactory.CreateType(finalResponseType with {IsNullable = false}));
-                ResultSerialization = new SerializationBuilder().Build(finalResponse.BodyMediaType, finalResponseType, ResultType);
+                ResultSerialization = SerializationBuilder.Build(finalResponse.BodyMediaType, finalResponseType, ResultType);
```
AlexanderSher, line 38: After all the clean-ups in serialization, `Build` can be declared static.

src/AutoRest.CSharp/Common/Output/Models/Requests/RestClientMethod.cs
```diff
@@ -2,18 +2,18 @@
...
 using AutoRest.CSharp.Output.Models.Shared;
 using AutoRest.CSharp.Utilities;
 
 namespace AutoRest.CSharp.Output.Models.Requests
 {
-    internal class RestClientMethod
+    internal record RestClientMethod
     {
-        public RestClientMethod(string name, string? summary, string? description, CSharpType? returnType, Request request, Parameter[] parameters, Response[] responses, DataPlaneResponseHeaderGroupType? headerModel, bool bufferResponse, string accessibility, InputOperation operation, RequestConditionHeaders conditionRequestFlag = RequestConditionHeaders.None)
+        public RestClientMethod(string name, string? summary, string? description, CSharpType? returnType, Request request, IReadOnlyList<Parameter> parameters, Response[] responses, DataPlaneResponseHeaderGroupType? headerModel, bool bufferResponse, string accessibility, InputOperation operation)
```
AlexanderSher, line 16: `RequestConditionHeaders conditionRequestFlag` isn't used in `RequestWriterHelpers`, so it is moved into `LowLevelClientMethod`

src/AutoRest.CSharp/Common/Output/Models/Requests/RestClientMethod.cs
```diff
@@ -2,18 +2,18 @@
...
-using AutoRest.CSharp.Common.Output.Models;
 using AutoRest.CSharp.Generation.Types;
 using AutoRest.CSharp.Output.Models.Responses;
 using AutoRest.CSharp.Output.Models.Shared;
 using AutoRest.CSharp.Utilities;
 
 namespace AutoRest.CSharp.Output.Models.Requests
 {
-    internal class RestClientMethod
+    internal record RestClientMethod
```
AlexanderSher, line 14: I probably don't need this to be a record, will revert

test/TestProjects/CadlFirstTest/Generated/Demo2Client.cs
```diff
@@ -375,7 +375,7 @@ internal HttpMessage CreateHelloDemo2Request(RequestContext context)
         private static RequestContext DefaultRequestContext = new RequestContext();
         internal static RequestContext FromCancellationToken(CancellationToken cancellationToken = default)
         {
-            if (cancellationToken == CancellationToken.None)
+            if (!cancellationToken.CanBeCanceled)
```
AlexanderSher, line 378: @annelo-msft IMO, it is better to use a property rather than compare struct to a default value. But I can revert it, up to you.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -136,6 +143,8 @@ public static CodeWriter.CodeWriterScope WriteMethodDeclaration(this CodeWriter
                 .Append($"{methodBase.Name}(")
                 .AppendRawIf("this ", methodBase.Modifiers.HasFlag(Extension));
 
+            var outerScope = writer.AmbientScope();
```
AlexanderSher, line 146: When method is declared, we need two scopes: inner for the body and outer that will include parameters.

test/TestProjects/CadlFirstTest/Generated/HelloWorldClient.cs
```diff
@@ -58,16 +59,20 @@ public HelloWorldClient(Uri endpoint, string apiVersion, DemoHelloworldClientOpt
         }
 
         /// <param name="action"> The String to use. </param>
+        /// <param name="exEnum"> The ExtensibleEnum to use. </param>
+        /// <param name="enum"> The SimpleEnum to use. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="action"/> is null. </exception>
-        public virtual async Task<Response<Thing>> TopActionValueAsync(string action, CancellationToken cancellationToken = default)
+        public virtual async Task<Response<Thing>> TopActionValueAsync(string action, ExtensibleEnum exEnum, SimpleEnum? @enum = null, CancellationToken cancellationToken = default)
```
AlexanderSher, line 66: Here are the enums, both required and optional

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriter.cs
archerzz: Thanks. I'd like to wait for this to merge in, #2554 is not a high priority.

test/TestProjects/CadlFirstTest/Generated/HelloWorldClient.cs
```diff
@@ -78,16 +83,20 @@ public virtual async Task<Response<Thing>> TopActionValueAsync(string action, Ca
...
 
         /// <param name="action"> The String to use. </param>
+        /// <param name="exEnum"> The ExtensibleEnum to use. </param>
+        /// <param name="enum"> The SimpleEnum to use. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="action"/> is null. </exception>
-        public virtual Response<Thing> TopActionValue(string action, CancellationToken cancellationToken = default)
+        public virtual Response<Thing> TopActionValue(string action, ExtensibleEnum exEnum, SimpleEnum? @enum = null, CancellationToken cancellationToken = default)
         {
+            Argument.AssertNotNull(action, nameof(action));
```
AlexanderSher, line 92: @annelo-msft We must validate all the parameters that are validated in protocol methods. Otherwise, in case of incorrect path parameter value, we will open the scope and serialize body parameter into `RequestContent` before path parameter is validated.

annelo-msft:
> (but tests in CADL-ranch are still required)

@AlexanderSher, are you able to add these tests?

test/TestProjects/CadlFirstTest/Generated/Demo2Client.cs
annelo-msft: I believe we had this discussion and decided CanBeCancelled has [too many nuances](https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken.canbecanceled?view=net-6.0#remarks) and what we want here is to compare it to CancellationToken.None instead of checking CanBeCancelled.  @JoshLove-msft, @tg-msft, do you remember this conversation, and can you help me confirm?

test/TestProjects/CadlFirstTest/Generated/HelloWorldClient.cs
annelo-msft: Parameters are validated outside diagnostic scopes being opened in protocol methods, but yes, in the case we need to open the scope in the convenience method you are right.  To make it a simple rule to follow let's go ahead and validate parameters in convenience methods as well, following the same rules for parameters that are validated in protocol methods.  I'll update relevant docs.  Thanks!

test/TestProjects/CadlFirstTest/Generated/Demo2Client.cs
AlexanderSher: For example, `CancellationTokenSource` uses `CanBeCancelled`: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs#L737

JoshLove-msft: I thought the recommendation was to use CanBeCancelled rather than comparing to None.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelConvenienceMethodWriter.cs
```diff
@@ -1,119 +0,0 @@
...
-            }
-            return scope;
-        }
-
-        private void WriteClientConvenienceMethodBody(bool async)
-        {
-            string contextVariableName = _convenienceMethod.Signature.Parameters.FirstOrDefault(parameter => parameter.Name == KnownParameters.RequestContext.Name) != null ? $"{KnownParameters.RequestContext.Name}1" : KnownParameters.RequestContext.Name;
-            _writer.Line($"{typeof(RequestContext)} {contextVariableName} = FromCancellationToken({KnownParameters.CancellationTokenParameter.Name});");
-
-            string responseVariableName = _convenienceMethod.Signature.Parameters.FirstOrDefault(parameter => parameter.Name == "response") != null ? "response1" : "response";
```
AlexanderSher, line 90: @pshao25 Variables should always be declared via `CodeWriterDeclaration`, so that `CodeWriter` can handle name conflicts

src/AutoRest.CSharp/LowLevel/Generation/LowLevelConvenienceMethodWriter.cs
```diff
@@ -1,119 +0,0 @@
...
-                _writer.Line($"return {responseVariableName};");
-            }
-            else
-            {
-                _writer.Line($"return Response.FromValue({_convenienceMethod.ResponseType}.FromResponse({responseVariableName}), {responseVariableName});");
-            }
-
-            string ParameterToItsName(Parameter parameter)
-            {
-                if (parameter.Name == KnownParameters.RequestContext.Name)
```
AlexanderSher, line 105: @pshao25 Ideally, we should compare references to parameters, i.e.: `parameter == KnownParameters.RequestContext`. If it isn't possible (e.g. in case of endpoint), we need to compare `Name` and `Type` (with or without nullability, depending on context).

AlexanderSher:
> @AlexanderSher, are you able to add these tests?

https://github.com/Azure/cadl-ranch/pull/42

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
```diff
@@ -12,29 +13,30 @@ namespace AutoRest.CSharp.Output.Models.Types
...
 
+        public TypeFactory TypeFactory { get; }
         public IEnumerable<EnumType> Enums => _enums.Values;
-        public IEnumerable<ModelTypeProvider> Models => _models.Values;
+        public IEnumerable<ModelTypeProvider> Models => _modelFactories.Values.Select(f => f());
         public IReadOnlyList<LowLevelClient> RestClients { get; }
         public ClientOptionsTypeProvider ClientOptions { get; }
 
-        public DpgOutputLibrary(Func<TypeFactory, IReadOnlyDictionary<InputEnumType, EnumType>> enumsFactory, Func<TypeFactory, IReadOnlyDictionary<InputModelType, ModelTypeProvider>> modelsFactory, Func<TypeFactory, IReadOnlyList<LowLevelClient>> restClientsFactory, ClientOptionsTypeProvider clientOptions)
+        public DpgOutputLibrary(IReadOnlyDictionary<InputEnumType, EnumType> enums, IReadOnlyDictionary<InputModelType, Func<ModelTypeProvider>> modelFactories, IReadOnlyList<LowLevelClient> restClients, ClientOptionsTypeProvider clientOptions)
```
archerzz, line 24: So here we break the circular dependency between `ModelTypeProvider`, `TypeFactory` and `DpgOutputLibrary`?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -49,7 +50,7 @@ public void WriteClient(CodeWriter writer, LowLevelClient client)
             using (writer.Namespace(clientType.Namespace))
             {
                 writer.WriteXmlDocumentationSummary($"{client.Description}");
-                using (writer.Scope($"{client.Declaration.Accessibility} partial class {clientType:D}"))
+                using (writer.Scope($"{client.Declaration.Accessibility} partial class {clientType:D}", scopeDeclarations: client.Fields.ScopeDeclarations))
                 {
                     WriteClientFields(writer, client);
```
archerzz, line 55: After the `scopeDeclarations`, we can invert the sequence of `WriteClientFields` and `WriteConstructors`?

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -40,24 +41,40 @@ public DpgOutputLibrary Build(bool isCadlInput)
...
+            var clients = new List<LowLevelClient>();
 
-            IReadOnlyDictionary<InputEnumType, EnumType> EnumFactory(TypeFactory typeFactory)
-                => isCadlInput ? _rootNamespace.Enums.ToDictionary(e => e, e => new EnumType(e, _defaultNamespace, "public", typeFactory, _sourceInputModel), InputEnumType.IgnoreNullabilityComparer) : new Dictionary<InputEnumType, EnumType>();
+            var library = new DpgOutputLibrary(enums, modelFactories, clients, clientOptions);
 
-            IReadOnlyDictionary<InputModelType, ModelTypeProvider> ModelsFactory(TypeFactory typeFactory)
-                => isCadlInput ? _rootNamespace.Models.ToDictionary(m => m, m => new ModelTypeProvider(m, typeFactory, _defaultNamespace, _sourceInputModel)) : new Dictionary<InputModelType, ModelTypeProvider>();
+            CreateEnums(enums, library.TypeFactory);
+            if (isCadlInput)
```
archerzz, line 55: For swagger input, where do we put the model?

archerzz (approved):
I left some comments, but no obvious blocking issue.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -0,0 +1,391 @@
...
+        private CSharpType? ChangeTypeForProtocolMethod(InputType type) => type switch
+        {
+            InputEnumType enumType => _typeFactory.CreateType(enumType.EnumValueType).WithNullable(enumType.IsNullable),
+            InputModelType modelType => new CSharpType(typeof(object), modelType.IsNullable),
+            _ => null
+        };
+
+        private record ReturnTypeChain(CSharpType Convenience, CSharpType Protocol, CSharpType? ConvenienceResponseType);
+
+        private record ParameterChain(Parameter? Convenience, Parameter? Protocol, Parameter? CreateMessage);
```
pshao25, line 389: How about we call it `ParameterSet`? `ParameterSuit`?

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
AlexanderSher: Yes. Scopes require some rework, but it would be a big change in the code that uses it, so it will be a separate work item.

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
AlexanderSher: Kind of. I hope that when we move MPG on the CADL rails, we can fix this completely.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
AlexanderSher: We don't support models for swagger input (at least for now).

src/AutoRest.CSharp/LowLevel/Generation/LowLevelConvenienceMethodWriter.cs
```diff
@@ -1,119 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
AlexanderSher: This one has several signs of a lazy class:
- only one code path (technically two, but that is the same method called with the same parameters, just with or without scope)
- leaky encapsulation: method _assumes_ that some other code has generated protocol method that matches what is generated. Changes in protocol method generations will leak here and visa versa
- no reusability: code is used only in one place

`LowLevelClientWriter` has indeed become huge, but if we want split it by the units of work, then IMO we need to split it into "create normal methods", "create lro methods", "create pageable methods", "create LRO pageable methods" and "utilities".

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
AlexanderSher: `ParameterGroup` will be misleading because we have grouped parameters in HLC. `ParameterSet` doesn't represent the semantics cause the idea that these parameters not just some random parameters brought together - they have a connection between them.

Issue #2341, opened 2022-06-21
Generate models with read-only properties from Cadl


annelo-msft:
### Input/Output/Round-trip models with read-only properties
This issue tracks the work to implement generation of models with read-only properties in the DPG generator.

Input models are not addressed, because if a model is input-only, it should not have any read-only properties.

If we need to triage this work, generation of Round-trip models is P0, and generation of Output models is P1.

#### Overview

Read-only properties are indicated via the `@visibility("read")` decorator in Cadl, see https://github.com/microsoft/cadl/blob/main/docs/tutorial.md#visibility-decorators.

#### Input Cadl file
[Cadl testserver models/readonly-properties](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/readonly-properties)

#### Expected output

OutputModel files:
- [x] [OutputModel.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-model-readonly-property/sdk/template/Azure.Template/src/Generated/Models/OutputModel.cs)
- [x] [OutputModel.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-model-readonly-property/sdk/template/Azure.Template/src/Generated/Models/OutputModel.Serialization.cs)
- [x] Output model has no initialization constructor, unless it contains collection properties. Collection properties are initialized in an internal initialization constructor. (See discussion below in https://github.com/Azure/autorest.csharp/issues/2341#issuecomment-1220387199)
- [x] Output model has serialization constructor that takes all property values
- [x] Readonly properties are get-only
- [x] Readonly model properties are get-only
- [x] Readonly collection properties are IReadOnlyLists
- [x] DeserializeOutputModel method is defined in serialization partial class
- [x] Readonly properties are deserialized

RoundTripModel files:
- [x] [RoundTripModel.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-model-readonly-property/sdk/template/Azure.Template/src/Generated/Models/RoundTripModel.cs)
- [x] [RoundTripModel.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-model-readonly-property/sdk/template/Azure.Template/src/Generated/Models/RoundTripModel.Serialization.cs)
- [x] Round-trip model initialization constructor does not take read-only property values
- [x] Round-trip model initialization constructor initialized read-only collection properties to ReadOnlyList
[...]

PR #2564, opened 2022-08-11
feat(dpg): generate readonly properties for round-trip model

archerzz:
1. update `TypeFactory.CreateType()` to support create `IReadOnlyList` and `IReadOnlyDictionary` for read-only properties
2. update constructor generation logic
    - split logic into public constructor and serialization constructor
    - generate initialization codes for `IReadOnlyList` fields (`IReadOnlyDictaionry` will be added later)
3. add test case

part of #2341

# Description

# Checklist

# Ready to Land?


archerzz:
Need to wait for the support of nested model type / collection of model type.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -29,9 +29,12 @@ public TypeFactory(OutputLibrary library)
             _library = library;
         }
 
-        public CSharpType CreateType(InputType inputType) => inputType switch
+        public CSharpType CreateType(InputType inputType, bool readOnly = false) => inputType switch
         {
-            InputListType listType             => new CSharpType(typeof(IList<>), listType.IsNullable, CreateType(listType.ElementType)),
+            InputListType listType when readOnly is true                => new CSharpType(typeof(IReadOnlyList<>), listType.IsNullable, CreateType(listType.ElementType)),
```
AlexanderSher, line 34: I'm not sure I like the idea of adding a `readOnly` parameter here because it is applicable only in a small subset of cases and requires a context that isn't available, e.g.:
- Read-only version of `IList<T>` is `IReadOnlyList<T>` for properties but `IEnumerable<T>` for method parameters
- Read-only version of  IList<IList<T>> is not well defined

Can you use `TypeFactory.GetOutputType` for your case?

archerzz: Agree. I also felt smell here, but I didn't know a better place to put the logic. Thanks for the information. I implemented a similar method in `ModelTypeProvider`, like `SchemaObjectType`: https://github.com/Azure/autorest.csharp/blob/f7ea672aeaa84049dfccc5c5e5ab6b93d3e7c86c/src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs#L472-L483

archerzz:
Update since last change:
- revert the change in `TypeFactory.CreateType`
- add `ModelTypeProvider.GetDefaultPropertyType` which is like `SchemaObject.GetDefaultPropertyType` which will handle read-only property types
- update `Parameter.FromModelProperty` which has duplicated property type calculation
- fix a bug in parameter list comparison
- fix unit tests
- update test project

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -80,29 +82,47 @@ private static (IReadOnlyDictionary<FieldDeclaration, InputModelProperty> Fields
...
-            return (fieldsToInputs, serializatoinParameters, parametersToFields);
+            return (fieldsToInputs, publicParameters, serializatoinParameters, parametersToFields);
         }
 
-        private static (ConstructorSignature PublicConstructor, ConstructorSignature SerializationConstructor) BuildConstructors(string name, IReadOnlyList<Parameter> serializationParameters)
+        private static CSharpType GetDefaultPropertyType(in InputModelProperty property, TypeFactory typeFactory)
+        {
+            var valueType = typeFactory.CreateType(property.Type);
+
+            if (//!_usage.HasFlag(SchemaTypeUsage.Input) ||
```
archerzz, line 108: I comment out the codes, but I think later we will handle output model case as well.

archerzz:
Optional readonly properties will be handled later.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
AlexanderSher, line 108: We can use `InputModelType.Usage` here

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -80,29 +82,47 @@ private static (IReadOnlyDictionary<FieldDeclaration, InputModelProperty> Fields
...
         {
             ConstructorSignature publicConstructor;
             ConstructorSignature serializationConstructor;
-            if (serializationParameters.Where(IsIList).Any())
+            if (!publicParameters.SequenceEqual(serializationParameters) || serializationParameters.Any(p => TypeFactory.IsList(p.Type)))
             {
                 serializationConstructor = new ConstructorSignature(name, $"Initializes a new instance of {name}", null, MethodSignatureModifiers.Internal, serializationParameters);
                 publicConstructor = new ConstructorSignature(name, $"Initializes a new instance of {name}", null, MethodSignatureModifiers.Public,
-                    serializationParameters.Select(p => IsIList(p) ? FromIListToIEnumerable(p) : p).ToList());
+                    publicParameters.Select(p => TypeFactory.IsReadWriteList(p.Type) ? FromIListToIEnumerable(p) : p).ToList());
```
AlexanderSher, line 125: A bit hard to read. Can we introduce `publicConstructorP

AlexanderSher, line 125: Can we extract `CreateConstructorParameter`:
```cs
private static Parameter CreateConstructorParameter(Parameter p)
    => TypeFactory.IsReadWriteList(p.Type) ? p with { Type = new CSharpType(typeof(IEnumerable<>), p.Type.IsNullable, p.Type.Arguments) } : p;
```

AlexanderSher:
Couple small comments, the rest is LGTM

AlexanderSher (approved):

PR #2579, opened 2022-08-13
Fix CanBeInitializedInline and GetParameterInitializer

AlexanderSher:
- In `CanBeInitializedInline`, `type` may be initialized inline (e.g. `string`), but `defaultValue`, though convertible to `type`, may not be used for inline initialization (e.g. extensible enum). So, if `type` is different than the type of `defaultValue`, we need to check if `defaultValue` can be assigned to its own type, then check for the provided type
- `GetParameterInitializer` should add conversion extension method if required

### DIFF
--- a/src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
+++ b/src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
@@ -14,6 +14,7 @@ namespace AutoRest.CSharp.Output.Models.Types
     {
         private readonly IReadOnlyDictionary<InputEnumType, EnumType> _enums;
         private readonly IReadOnlyDictionary<InputModelType, Func<ModelTypeProvider>> _modelFactories;
+        private readonly bool _isCadlInput;
 
         public TypeFactory TypeFactory { get; }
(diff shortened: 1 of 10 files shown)


src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
```diff
@@ -14,26 +14,38 @@ internal class DpgOutputLibrary : OutputLibrary
...
             RestClients = restClients;
             ClientOptions = clientOptions;
         }
 
         public override CSharpType ResolveEnum(InputEnumType enumType)
-            => _enums.TryGetValue(enumType, out var typeProvider)
-                ? typeProvider.Type
-                : throw new InvalidOperationException($"No {nameof(EnumType)} has been created for `{enumType.Name}` {nameof(InputEnumType)}.");
+        {
+            if (!_isCadlInput)
```
AlexanderSher, line 37: With this fix, swagger-based DPG projects may generate parameter initializer that references enum which is not generated.

AlexanderSher:
Fixes build issue in https://github.com/Azure/azure-sdk-for-net/pull/30472

archerzz (approved):

PR #2599, opened 2022-08-16
fix the simplification issue in test gen

ArcturusZhang:
# Description

<del>Our simplifier in <del>rolsyn</del> has some issues when simplifying expressions like this:</del>
```csharp
global::Azure.ResourceManager.Compute.AvailabilitySetData data = new ...;
```
<del>The best it can do is</del>
```csharp
Compute.AvailabilitySetData data = new ...;
```
Since we are already using the namespace `Azure.ResourceManager.Compute`, we are actually expecting
```csharp
AvailabilitySetData data = new ...;
```

<del>As a workaround, I borrowed the format identifier `:I` and passed it into the `AppendType` method of `CodeWriter`, to let it not write the full type name of this type. </del>

To fix this, this PR adds all the source file in the source code project into the workspace of the test project's `GeneratedCodeWorkspace` so that Roslyn knows how to simplify them, thanks to @AlexanderSher  to point this out.

# Checklist

# Ready to Land?


ArcturusZhang:
@AlexanderSher could you have a review on this PR?
Do you have insights about what else we could try to fix this? We would not want to see there are redundant namespaces in our generated samples.

AlexanderSher (changes requested):
Types in `autorest.csharp\test\TestProjects\MgmtMockAndSample\src\Generated` aren't included in the `GeneratedCodeWorkspace._project`, hence Roslyn has no idea how to resolve these types. You need to update either `MgmtTestTarget` so that types are known to the project.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -104,6 +104,38 @@ private static ImmutableHashSet<string> GetSuppressedTypeNames(Compilation compi
...
+
+        /// <summary>
+        /// Add the files in the directory to a project per a given predicate with the folders specified
+        /// </summary>
+        /// <param name="project"></param>
+        /// <param name="directory"></param>
+        /// <param name="skipPredicate"></param>
+        /// <param name="folders"></param>
+        /// <returns></returns>
+        internal static Project AddDirectory(Project project, string directory, Predicate<string>? skipPredicate = null, IEnumerable<string>? folders = null)
```
AlexanderSher, line 126: Let's use `Func<string, bool>` instead of `Predicate<string>`

ArcturusZhang: Fixed

AlexanderSher (approved):
LGTM

Issue #2562, opened 2022-08-10
[FEATURE] [Polymorphic Resources] Add client hierarchy to resource clients when the Data is shared


m-nash:
Original options https://gist.github.com/m-nash/0bc0e2a1871365708ad88d3052a236c6

Any time more than 1 resource shares a [DataPrefix]Data object we want to create a base Resource named [DataPrefix]Resource.  This class should be abstract and the data properties and initialization should move to this class.

```c#
    public abstract class ConfigResource : ArmResource
    {
        private readonly ConfigData _data;

        internal ConfigResource(ArmClient client, ConfigData data) : this(client, data.Id)
        {
            HasData = true;
            _data = data;
        }

        public virtual bool HasData { get; }

        public virtual ConfigData Data
        {
            get
            {
                if (!HasData)
                    throw new InvalidOperationException("The current instance does not have data, you must call Get first.");
                return _data;
            }
        }
    }
```

The derived classes will then inherit from this base and no longer deal with the data portion directly.

```c#
    public class WebConfigResource : ConfigResource
    {
        public static readonly ResourceType ResourceType = "Microsoft.Heirarchy/webConfigs";

        internal WebConfigResource(ArmClient client, ConfigData data) : base(client, data)
        {
        }
    }
```

Any common methods should also go on this new base resource class and only methods that are unique to one type should go on their derived classes.

If the method returns a resource we will follow this pattern to ensure mockability as well as getting back the right type based on which instance type the method is called from.

First we will make a protected virtual method with the same method name but `Core` appended to the end.  In the example of Get we would have the following for the core methods

```c#
        protected virtual Response<T> GetCore<T>(CancellationToken cancellationToken = default)
        {
            //implementation
        }
```

Then in the base class we will have a non virtual public method that calls this guy and returns the base resource type

```c#
        public Response<ConfigResource> Get(CancellationToken cancellationToken = default) => GetCore<ConfigResource>(cancellationToken);
```

In each of the derived classes they will use the `new` keyword to change the response type of this and still forward the call to the core method.  We must also keep the `virtual` keyword for back compat here in already GA libraries, but in new libraries this won't exist since the goal is to mock the GetCore to have consistent behavior across all reference types.

```c#
        public virtual new Response<WebConfigResource> Get(CancellationToken cancellationToken = default) => GetCore<WebConfigResource>(cancellationToken);
```

If the method doesn't return a resource then we can simply have a normal public virtual method on the base class that all derived classes will inherit.

[...]

PR #2611, opened 2022-08-18
Introduce polymorphic resources when resources are sharing the same resource data

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/2562

# Description

Check lists
- [x] structure of base resource and resource
- [x] static resource factory
- [x] static methods to dynamically check type of resources from resource id
- [x] base resource should collect all the "common" operations and create corresponding abstract XXXCore method
- [x] base resource should have the XXX method which is returning the base resource as its return type
- [x] each resource should override the XXXCore method and has their real implementation of XXX directing XXXCore
- [x] weird casting issues everywhere
- [x] unwrap the response returned by Core and cast it to its concrete type of resource and wrap it again
- [x] unwrap the lro returned by Core and cast it or rewrap it to its concrete type of resource
- [ ] extensible enum to describe which type this resource is (like a discriminator)
- [x] if a method returning the resource data and we cannot determine which resource to wrap into, we should return the base resource instead
- [ ] a configuration to determine whether we will add `virtual` keyword to those `new` methods

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -767,19 +808,39 @@ protected virtual void WriteNormalMethodBranch(MgmtRestOperation operation, IEnu
         }
         #endregion
 
+        protected bool NeedFactoryMethod(MgmtClientOperation clientOperation)
+        {
+            if (This is not Resource resource || resource.BaseResource == null)
+                return false;
+
+            return resource.BaseResource.CommonOperations.Any(operation => operation.Contains(clientOperation));
+        }
```
ArcturusZhang, line 817: Maybe instead of using this method, we could override the `WriteMethod` method in PolymorphicResourceWriter, to evaluate this bool there, and pass it through.
On those method, this new parameter could be optional and defaults to `false`.

ArcturusZhang: BTW we also need to find those places that are directly returning the concrete types and adding a cast there

ArcturusZhang: Or let them to call the corresponding Core method to keep things consistent (like our SetTags are calling Update, we should call UpdateCore if there is one, and if there is not UpdateCore, we should add a cast to keep it compile)

samples/Azure.ResourceManager.Sample/Generated/VirtualMachineExtensionResource.cs
```diff
@@ -120,7 +101,17 @@ public virtual async Task<Response<VirtualMachineExtensionResource>> GetAsync(st
         /// </summary>
         /// <param name="expand"> The expand expression to apply on the operation. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual Response<VirtualMachineExtensionResource> Get(string expand = null, CancellationToken cancellationToken = default)
+        [ForwardsClientCalls]
+        public new async Task<Response<VirtualMachineExtensionResource>> GetAsync(string expand = null, CancellationToken cancellationToken = default)
```
ArcturusZhang, line 105:
```suggestion
        public new virtual async Task<Response<VirtualMachineExtensionResource>> GetAsync(string expand = null, CancellationToken cancellationToken = default)
```

PR #2614, opened 2022-08-18
Fix -fast and -debug flags support

AlexanderSher:


eng/Generate.ps1
```diff
@@ -292,7 +292,9 @@ if ($reset -or $env:TF_BUILD)
 if (!$noBuild)
 {
     Invoke "dotnet build $autoRestPluginProject"
-    Invoke-CadlSetup
+    if (!$fast) {
+        Invoke-CadlSetup
```
AlexanderSher, line 296: Fast path doesn't invoke generation of `cadl.json`, so we don't need to setup CADL

AlexanderSher: In fact, I think we should make this an explicit command. Even at full rebuild we don't need it more than once.

chunyu3: !Fast path, we need to invoke generation of `cadl.json` and it need to re-build setup CADL, so I think we need to remain previous version
if (!$noBuild)
{
    Invoke "dotnet build $autoRestPluginProject"
    if (!$fast -and  $cadlCou]nt -gt 0) {
        Invoke-CadlSetup
   }
}

eng/Generate.ps1
```diff
@@ -310,6 +299,26 @@ if (![string]::IsNullOrWhiteSpace($filter))
...
+if ($reset -or $env:TF_BUILD)
+{
+    $cadlCount = ([string]::IsNullOrWhiteSpace($filter) ? $cadlDefinitions : $cadlDefinitions.Keys.Where({$_ -match $filter})).Count
+    $swaggerCount = $keys.Count - $cadlCount
+    if ($swaggerCount -gt 0) 
+    {
+        AutoRest-Reset;
+    }
+    
+    if ($cadlCount -gt 0)
```
chunyu3, line 311:
```suggestion
    move this into if (!$noBuild) block with the check $fast and $cadlCount
```

chunyu3: !Fast path, we need to invoke generation of cadl.json and it need to re-build setup CADL

AlexanderSher: > we need to invoke generation of cadl.json and it need to re-build setup CADL

Do we need to re-build CADL for every change in CADL input file?

chunyu3: In CI we need to use the emitter and autorest.csharp built by the source code to generate SDK from Cadl. So we need to build the emitter and autorest.csharp once every time run Generate.ps1 if `-noBuild` is not specified.

lirenhe:
@AlexanderSher, do you still need this PR? Seems all change are moved to https://github.com/Azure/autorest.csharp/pull/2631

AlexanderSher:
I'd rather merge this as a separate PR

Issue #2503, opened 2022-07-28
Expand partial resource functionality - scope resource should extend ArmClient and accept an Id as parameter instead of extending ArmResource


ArcturusZhang:
Initial purpose for a partial resource is to resolve the cross RP resources (see details in [this issue](https://github.com/Azure/azure-sdk-for-net/issues/28185)).
This is about the request paths like this:
`/subscriptions/{}/resourceGroups/{}/providers/Microsoft.Compute/virtualMachines/{}/publicIps` which extends a resource path (a virtual machine) in another RP (like this request is in network RP).

The other form of this kind of issue is in a form of scope resource, like the following path:
`/{resourceUri}/providers/Microsoft.Insights/vmInsightsOnboardingStatuses/default` in monitor RP.
It looks like a scope resource, but actually its scope only applies to a very limited list of resources from other RPs (given the name of the resource in this example, probably only virtual machine).
<del>We need to use partial resource to handle this "scope" resource.</del>

We will generate this kind of "scope" resource like in this PR: https://github.com/Azure/azure-sdk-for-net/pull/30295

PR #2627, opened 2022-08-19
changes on the scope resources

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/2503
Regenerate: https://github.com/Azure/azure-sdk-for-net/pull/30654

# Description

WIP
- [x] The scope resource now should by default write the code like:
```csharp
public static XXXCollection GetXXXs(this ArmClient client, ResourceIdentifier scope)
{
/* ... */
}
```
instead of extending `ArmResource`. The same changes have been made to the pair of methods that get an single instance of a resource, such like
```csharp
public static async Task<Response<XXXResource>> GetXXXAsync(this ArmClient client, ResourceIdentifier scope, string name)
{
/* ... */
}
public static async Response<XXXResource> GetXXX(this ArmClient client, ResourceIdentifier scope, string name)
{
/* ... */
}
```
- [x] Introduce a new configuration `generate-arm-resource-extensions` to restore the old behaviour to keep consistency.
- [ ] (maybe another PR) Support the case that in this RP we are adding a child resource to a resource in another RP, like 
`/subscriptions/{}/resourceGroups/{}/providers/Microsoft.Compute/virtualMachines/{}/providers/Microsoft.GuestConfiguration/guestConfigurations/{}` -> we need to extract the part `/subscriptions/{}/resourceGroups/{}/providers/Microsoft.Compute/virtualMachines/{}` as a scope which only accepts virtual machines.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Generation/ArmResourceExtensionsWriter.cs
```diff
@@ -12,45 +12,136 @@
...
 using System;
 using AutoRest.CSharp.Utilities;
+using System.Linq;
+using Azure.ResourceManager;
+using Azure.Core;
+using static AutoRest.CSharp.Output.Models.MethodSignatureModifiers;
 
 namespace AutoRest.CSharp.Mgmt.Generation
 {
     internal class ArmResourceExtensionsWriter : MgmtExtensionWriter
```
ArcturusZhang, line 22: This class previously has no reference in our code. This PR takes it back

ArthurMa1978 (approved):

Issue #2592, opened 2022-08-16
Cannot generate endpoint with server decorator


pshao25:
Add below decorator to the cadl file:
`@server("http://localhost:3000", "Testserver endpoint")`

Run the emitter, there is no parameter with `"IsEndpoint": true`

Therefore, no endpoint generated.

annelo-msft:
We will need to generate endpoint in order to generate any basic client.  Adding this to the August MVP requirements - it is covered as part of the parity requirement.

annelo-msft:
Related to https://github.com/Azure/cadl-azure/issues/1785, which is needed to indicate that the endpoint is a Uri.

Issue #2577, opened 2022-08-12
Generate clients with authentication specified by Cadl @useAuth decorator


annelo-msft:
Cadl support for the `@useAuth` decorator is available with https://github.com/microsoft/cadl/pull/795.

We will support generation of the same set of client constructors for authentication types we supported in DPG v1.0:
- AADToken (a.k.a OAuth2Flow): While Cadl has [several different models]( https://github.com/microsoft/cadl/blob/main/packages/rest/lib/auth.cadl#L86), DPG 1.0 creates one constructor that accepts `TokenCredential` and leaves instantiation of the specific `TokenCredential`-derived type to the user.
- AzureKey (a.k.a ApiKeyAuth): Cadl supports adding api key to [header, query or cookie](https://github.com/microsoft/cadl/blob/main/packages/rest/lib/auth.cadl#L36). DPG 1.0 uses `AzureKeyCredentialPolicy` that always adds key [to the header](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/AzureKeyCredentialPolicy.cs#L30).

Right now, at most one AADToken and one AzureKey are supported, both in HLC and DPG. HLC [merges identical security schemas](https://github.com/Azure/autorest.csharp/blob/f7ea672aeaa84049dfccc5c5e5ab6b93d3e7c86c/src/AutoRest.CSharp/DataPlane/Generation/DataPlaneClientWriter.cs#L426), DPG 1.0 doesn't have well-defined behavior (see https://github.com/Azure/autorest.csharp/issues/2590). 

Details of DPG v1.0 support can be found [here](https://github.com/Azure/azure-sdk-for-net/blob/main/doc/DataPlaneCodeGeneration/AzureSDKCodeGeneration_DataPlane_Quickstart.md#authentication-configuration-in-autorestmd).

This is needed for parity with DPG v1.0.

lirenhe:
@chunyu3   please help to make @useAuth  info could be extracted from emitter.

annelo-msft:
@schaabs, @tg-msft, @KrzysztofCwalina, please let us know if you have any concerns with following the DPG1.0 spec in generating client constructor authentication. I know historically much of this has been done in the convenience layer.  We have tracking the design work to support additional Cadl features.

AlexanderSher:
autorest.csharp part is done, emitter part remains

chunyu3:
I am working on emitter part for auth.

[...]

PR #2628, opened 2022-08-19
support default server and user authentication and Serialize

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2592
Fix https://github.com/Azure/autorest.csharp/issues/2577

- support server with default url string, no endpoint parameter
- emit user authentication and serialize

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
```diff
@@ -0,0 +1,118 @@
...
+            => throw new NotSupportedException("Writing not supported");
+
+        public static InputConstant CreateInputConstant(ref Utf8JsonReader reader, string? id, string? name, JsonSerializerOptions options, ReferenceResolver resolver)
+        {
+            var isFirstProperty = id == null;
+            Object? value = null;
+            InputType? type = null;
+            while (reader.TokenType != JsonTokenType.EndObject)
+            {
+                var isKnownProperty = reader.TryReadReferenceId(ref isFirstProperty, ref id)
```
AlexanderSher, line 37: Since there can be only two properties (Type and Value) and we require type to go after value, I think we can enforce strict order here:
```cs
reader.TryReadReferenceId(ref isFirstProperty, ref id);
if (!TryReadWithConverter(nameof(InputConstant.Type), options, ref type)) 
{
    throw 
}
ReadConstantValue(ref reader, nameof(InputConstant.Value), options, type, ref value);
```

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
```diff
@@ -0,0 +1,118 @@
...
+                        case InputTypeKind.Boolean:
+                            value = reader.GetBoolean();
+                            break;
+                        default:
+                            value = reader.GetString() ?? throw new JsonException();
+                            break;
+
+                    }
+                    break;
+                case InputModelType model:
```
AlexanderSher, line 102: I suggest we just throw in this case (except if value is `null`). We have no design for default values of the properties which type is a user-defined model.

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            var constant = new InputConstant(value, type);
+            if (id != null)
+            {
+                resolver.AddReference(id, constant);
+            }
+            return constant;
+        }
+
+        public static bool TryReadConstantValue(ref Utf8JsonReader reader, string propertyName, JsonSerializerOptions options, InputType? type, ref Object? value)
```
AlexanderSher, line 59: Let's make this `object ReadConstantValue`. `InputConstant` must have a value, and type is defined already.

src/AutoRest.CSharp/Common/Input/Inputs.cs
```diff
@@ -161,7 +161,9 @@ internal record InputDictionaryType(string Name, InputType KeyType, InputType Va
 
     internal record InputModelProperty(string Name, string? SerializedName, string Description, InputType Type, bool IsRequired, bool IsReadOnly, bool IsDiscriminator) { }
 
-    internal record InputConstant(object? Value, InputType Type);
+    internal record InputConstant(object? Value, InputType Type) {
+        public InputConstant() : this(null, InputPrimitiveType.String) { }
```
AlexanderSher, line 165: I don't think we need default constructor cause we have provided a custom converter.

src/AutoRest.CSharp/Common/Input/Inputs.cs
```diff
@@ -161,7 +161,9 @@ internal record InputDictionaryType(string Name, InputType KeyType, InputType Va
 
     internal record InputModelProperty(string Name, string? SerializedName, string Description, InputType Type, bool IsRequired, bool IsReadOnly, bool IsDiscriminator) { }
 
-    internal record InputConstant(object? Value, InputType Type);
+    internal record InputConstant(object? Value, InputType Type) {
```
AlexanderSher, line 164: Do we need `Value` to be nullable?

AlexanderSher:
Good work!

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
chunyu3: updated

src/AutoRest.CSharp/Common/Input/Inputs.cs
chunyu3: value should not be null as constant. I will update it.

src/AutoRest.CSharp/Common/Input/Inputs.cs
chunyu3: removed.

src/AutoRest.CSharp/Common/Input/CadlSerialization.cs
```diff
@@ -30,6 +30,9 @@ internal static class CadlSerialization
             options.Converters.Add(new CadlInputEnumTypeValueConverter(referenceHandler));
             options.Converters.Add(new CadlInputModelTypeConverter(referenceHandler));
             options.Converters.Add(new CadlInputModelPropertyConverter(referenceHandler));
+            options.Converters.Add(new CadlInputConstantConverter(referenceHandler));
+            options.Converters.Add(new CadlInputAuthConverter(referenceHandler));
```
AlexanderSher, line 34: Do we need custom converter for `InputAuth`? I'd expect such input type to be handled automatically.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -189,6 +193,41 @@ function createModel(program: Program): any {
...
+    for (const option of authentication.options) {
+        for (const schema of option.schemes) {
+            switch (schema.type) {
+                case "apiKey":
+                    auth.ApiKey = schema.name;
+                    break;
+                case "oauth2":
+                    let scopes: string[] = [];
+                    for (const flow of schema.flows) {
+                        switch (flow.type) {
```
AlexanderSher, line 209: I'd say we should just check if `flow` has `scopes`, `flow.type` is irrelevant.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -189,6 +193,41 @@ function createModel(program: Program): any {
...
+): InputAuth {
+    const auth = {} as InputAuth;
+    for (const option of authentication.options) {
+        for (const schema of option.schemes) {
+            switch (schema.type) {
+                case "apiKey":
+                    auth.ApiKey = schema.name;
+                    break;
+                case "oauth2":
+                    let scopes: string[] = [];
```
AlexanderSher, line 207: We probably need `let scopes = new Set<string>()` here to avoid repeated scopes.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -359,6 +398,16 @@ function loadOperation(
...
 }
 
+function isLROOperation(op: OperationDetails) {
+    return false;
+}
+function isPagingOperation(op: OperationDetails) {
+    return false;
+}
+function isSupportedOperation(op: OperationDetails) {
+    if (isLROOperation(op) || isPagingOperation(op)) return false;
```
AlexanderSher, line 408: `return !isLROOperation(op) && !isPagingOperation(op)`?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -359,6 +398,16 @@ function loadOperation(
     }
 }
 
+function isLROOperation(op: OperationDetails) {
```
AlexanderSher, line 401: nit: `isLroOperation`

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
```diff
@@ -0,0 +1,105 @@
...
+                        default:
+                            value = reader.GetString() ?? throw new JsonException();
+                            break;
+
+                    }
+                    break;
+                case InputModelType model:
+                    throw new JsonException("Not supported type.");
+                default:
+                    break;
```
AlexanderSher, line 99: Let's throw exception for `default` as well. That way we should be able to return `object` instead of `object?`

src/AutoRest.CSharp/Common/Input/CadlSerialization.cs
chunyu3: Yes, we need custom converter. We emit `InputAuth` as an object with id, and other two fields(ApiKey, OAuth2). Need customer converter to handle it.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: When the schema is 'oauth2' type, it will contain flows and each flow will have type to identify its related definition. if it is 'clientCredentials', it will contain scopes.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: Thanks @AlexanderSher Good suggestion. I will update.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: updated

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: Thanks. but We will add other supported scenarios. so we will not directly return at first statement.

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
chunyu3: Thanks. I will update it.

src/AutoRest.CSharp/Common/Input/CadlSerialization.cs
AlexanderSher: Ok, I'll check it myself

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
AlexanderSher: All currently defined flows have scopes:
- https://github.com/microsoft/cadl/blob/main/packages/rest/lib/auth.cadl#L96
- https://github.com/microsoft/cadl/blob/main/packages/rest/lib/auth.cadl#L106
- https://github.com/microsoft/cadl/blob/main/packages/rest/lib/auth.cadl#L116
- https://github.com/microsoft/cadl/blob/main/packages/rest/lib/auth.cadl#L126

AlexanderSher (approved):

Issue #2361, opened 2022-06-21
Generate round-trip polymorphic models from Cadl with extends and discriminator keywords


annelo-msft:
### Round trip polymorphic models

This issue tracks the work to implement generation of round-trip models:
- That use inheritance via the Cadl `extends` keyword
- That use [polymorphism with discriminators](https://github.com/microsoft/cadl/blob/main/docs/tutorial.md#polymorphism-with-discriminators) via the Cadl `@discriminator` decorator
- That have properties that are polymorphic types

#### Input Cadl file
[Cadl testserver models/inheritance](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/inheritance)

#### Expected output

Basic inheritance using `extends` keyword:
- [ ] [DerivedFromBaseClassA.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-test-polymorphic-basic/sdk/template/Azure.Template/src/Generated/Models/DerivedFromBaseClassA.cs)
- [ ] [DerivedFromBaseClassA.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-test-polymorphic-basic/sdk/template/Azure.Template/src/Generated/Models/DerivedFromBaseClassA.Serialization.cs)
- [ ] Type derives from BaseClass
- [ ] Constructor calls BaseClass constructor to initialize baseClassProperty
- [ ] Constructor only initializes DerivedClassAProperty
- [ ] Serialization Write method writes both BaseClassProperty and DerivedClassAProperty
- [ ] Serialization DeserializeDerivedFromBaseClassA method reads both baseClassProperty and derivedClassAProperty

Polymorphism with discriminators:
- [ ] [BaseClassWithDiscriminator.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-test-polymorphic-basic/sdk/template/Azure.Template/src/Generated/Models/BaseClassWithDiscriminator.cs)
- [ ] [BaseClassWithDiscriminator.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-test-polymorphic-basic/sdk/template/Azure.Template/src/Generated/Models/BaseClassWithDiscriminator.Serialization.cs)
- [ ] Discriminator is not set in initialization constructor
- [ ] Discriminator is set in serialization constructor
- [ ] DiscriminatorProperty is an internal property
[...]

PR #2631, opened 2022-08-22
Add nested model support

AlexanderSher:
PR adds support for nested models and adds `discriminatorPropertyName` and `discriminatorValue` generation to emitter


src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -8,6 +8,7 @@ import {
     getServiceTitle,
     getServiceVersion,
     getSummary,
+    ModelType,
```
archerzz, line 11: Is this import required? It seems it's not used.

archerzz (approved):

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -407,27 +349,121 @@ export function getInputType(
...
+            const baseModel = getInputModelBaseType(m.baseModel);
+            const properties: InputModelProperty[] = [];
+
+            model = {
+                Name: name,
+                Namespace: m.namespace?.name,
+                Description: getDoc(program, m),
+                IsNullable: false,
+                DiscriminatorPropertyName: getDiscriminator(program, m)?.propertyName,
+                DiscriminatorValue: getDiscriminatorValue(baseModel),
```
chunyu3, line 383: DiscriminatorValue is defined in derived model, not the base Model.

AlexanderSher: Yes, it picks up model from the outer scope, but I'll change the signature to make it easier to read

AlexanderSher: Fixed

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
AlexanderSher: Fixed. Cleaned up other unused imports

eng/Generate.ps1
```diff
@@ -310,6 +299,26 @@ if (![string]::IsNullOrWhiteSpace($filter))
...
+    if ($cadlCount -gt 0) 
+    {
+        Invoke-CadlSetup
+    }
+}
+
+if (!$noBuild)
+{
+    Invoke "dotnet build $autoRestPluginProject"
+}
```
ArcturusZhang, line 320: Is there a flag which makes us to "include the build of the generator, but skip the CadlSetup"?
I believe in some scenarios, when we do not touch anything in the emitter, it does not have to rebuild everytime? Especially when we are changing something in the generator

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -407,27 +349,121 @@ export function getInputType(
...
+            } as InputModelType;
+
+            models.set(name, model);
+
+            // Resolve properties after model is added to the map to resolve possible circular dependencies
+            addModelProperties(m.properties, properties, baseModel?.DiscriminatorPropertyName);
+
+            // Temporary part. Derived types may not be referenced directly by any operation
+            // We should be able to remove it when https://github.com/Azure/cadl-azure/issues/1733 is closed 
+            if (model.DiscriminatorPropertyName && m.derivedModels) {
```
chunyu3, line 395: shall we also emit derived models as fields in the base model? in case, in the late codegen, the operation will set base model as input and we can easy to get its derived models to generate the sdk.

src/CADL.Extension/Emitter.Csharp/src/type/InputType.ts
```diff
@@ -20,8 +20,8 @@ export interface InputModelType extends InputType {
     Description: string;
     Properties: InputModelProperty[];
     BaseModel?: InputModelType;
-    DerivedModels?: InputModelType[];
-    DiscriminatorValue?: string[];
+    DiscriminatorPropertyName?: string;
+    DiscriminatorValue?: string;
```
chunyu3, line 24: shall we also emit derived Models also?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -407,27 +349,121 @@ export function getInputType(
...
-}
+
+    function addModelProperties(
+        inputProperties: Map<string, ModelTypeProperty>,
+        outputProperties: InputModelProperty[],
+        discriminatorPropertyName?: string
+    ) : void {
+        inputProperties.forEach(
+            (value: ModelTypeProperty, key: string) => {
+                if (value.name !== discriminatorPropertyName) {
```
chunyu3, line 425: discriminator proeprty will not be added into the model. When use this function to resolve properties for base model, it seems that the discriminator property will be missing.

eng/Generate.ps1
AlexanderSher: Sorry, I don't understand the question.
`Invoke-CadlSetup` builds emitter. By default (without any flags), cadl will be regenerated (similar to codemodel.yaml), but emitter itself won't be rebuilt.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
AlexanderSher: No, derived models be restored in autorest.csharp without any additional input (because of possible circular references, it is actually easier).

src/CADL.Extension/Emitter.Csharp/src/type/InputType.ts
AlexanderSher: No, see response above.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
AlexanderSher: Yes, that is intentional. Discriminator property is not really a property (essentially, it is an internal constant), existing code in HLC has to deal with it, not always successfully. I want to try to avoid it by not making it a property at all.

eng/Generate.ps1
AlexanderSher: There is a separate PR for this change (https://github.com/Azure/autorest.csharp/pull/2614), I'm just working on top of it cause my machine is too slow, and I can't work without this fix.

ArcturusZhang: Thank you! I am actually asking if we have a flag that could build the generator and regenerate the test project but skip the cadl setup because it takes a long time and mgmt generator never needs it.
I suppose when working on DPG generator, not every time when you change something in the generator, you will have to involve `cadl setup` in the generate script

AlexanderSher: > I am actually asking if we have a flag that could build the generator and regenerate the test project but skip the cadl setup

With this change, that would be a default behavior. To rebuild emitter you'll have to add `-reset` flag.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -407,27 +349,121 @@ export function getInputType(
...
+                    const vis = getVisibility(program, value);
+                    let isReadOnly: boolean = false;
+                    if (vis && vis.includes("read") && vis.length == 1) {
```
lirenhe, line 428: use ===

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -407,27 +349,121 @@ export function getInputType(
...
+                        SerializedName: value.name,
+                        Description: "",
+                        Type: getInputType(
+                            program,
+                            value.type,
+                            models,
+                            enums
+                        ),
+                        IsRequired: !value.optional,
+                        IsReadOnly: isReadOnly, //TODO: get the require and readonly value from cadl.
```
lirenhe, line 442: Can we remove this TODO now?

test/TestProjects/Models-Cadl/Generated/BaseModelWithDiscriminator.cs
```diff
@@ -0,0 +1,18 @@
...
+#nullable disable
+
+namespace ModelsInCadl
+{
+    /// <summary> Base model with discriminator property. </summary>
+    public partial class BaseModelWithDiscriminator
+    {
+        /// <summary> Initializes a new instance of BaseModelWithDiscriminator. </summary>
+        public BaseModelWithDiscriminator()
+        {
```
lirenhe, line 15: Is this the final output or just a placeholder for the test project?

AlexanderSher: This isn't a final output cause discriminator isn't handled in autorest.csharp yet

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: Discriminator property will be an class Field of the base model. When we generate the model, codegen will add discriminator as a field together with other properties of the base model. OK. thanks @AlexanderSher

test/TestProjects/Models-Cadl/Generated/BaseModelWithDiscriminator.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License.
+
+// <auto-generated/>
+
+#nullable disable
+
+namespace ModelsInCadl
+{
+    /// <summary> Base model with discriminator property. </summary>
+    public partial class BaseModelWithDiscriminator
```
chunyu3, line 11: The 'discriminatorProperty' is missing in the base model. The discriminator is an field of the base model.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: removed in another PR

test/TestProjects/Models-Cadl/Models-Cadl.cadl
```diff
@@ -0,0 +1,101 @@
...
+
+@doc("Derived model")
+model DerivedModel extends BaseModel {
+  @doc("Required collection")
+  requiredCollection: CollectionItem[];
+}
+
+@discriminator("discriminatorProperty")
+@doc("Base model with discriminator property.")
+model BaseModelWithDiscriminator extends BaseModel {}
```
chunyu3, line 24: discriminator is an property of the BaseModelWithDiscriminator, why is it not defined in the model definition?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: Hello @AlexanderSher The discriminator property is missing in base model (see the generated test project). And in the test project, the discriminator even does not defined as a property in the base model in cadl file. Does cadl support the polymorphism that way which is different with API spec definition?

test/TestProjects/Models-Cadl/Generated/BaseModelWithDiscriminator.cs
AlexanderSher: discriminator support is not implemented in autorest.csharp in this PR. That would be a separate PR after we close several open issues:
- should base class with discriminator be abstract?
- do we need discriminator property or constant is enough?
- what should we do if discriminator property value is different from anything we handle?

test/TestProjects/Models-Cadl/Models-Cadl.cadl
AlexanderSher: That is the way it is implemented in CADL: https://github.com/microsoft/cadl/blob/0a528079a54ab74de2614cbdd59eb4a907d58e05/docs/tutorial.md#polymorphism-with-discriminators

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
AlexanderSher: We don't have finalized design for polymorphic models yet. https://github.com/Azure/autorest.csharp/issues/2361 still has `needs-design` tag:
image

test/TestProjects/Models-Cadl/Generated/BaseModelWithDiscriminator.cs
ArcturusZhang: FYI mgmt plane decides the base class to be abstract: https://github.com/Azure/autorest.csharp/issues/2430
Should we align?

lirenhe: discriminator is not in the scope of for August MVP so can we leave as it is for now?  @annelo-msft 
This part is likely to be redesigned later.

chunyu3 (approved):

PR #2632, opened 2022-08-22
Add autorest.csharp part of  useAuth support

AlexanderSher:
Since autorest.csharp supports at most one ApiKey and one OAuth2, they are made explicit fields in the `InputAuth`. We can change the schema in case more authentication types is supported


src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -404,5 +404,93 @@ public static IReadOnlyList<string> GetApiVersions(CodeModel codeModel)
...
+
+            foreach (var scheme in securitySchemes.Distinct(SecuritySchemesComparer.Instance))
+            {
+                switch (scheme)
+                {
+                    case AzureKeySecurityScheme:
+                        throw new NotSupportedException($"{typeof(AzureKeySecurityScheme)} is not supported. Use {typeof(KeySecurityScheme)} instead");
+                    case AADTokenSecurityScheme:
+                        throw new NotSupportedException($"{typeof(AADTokenSecurityScheme)} is not supported. Use {typeof(OAuth2SecurityScheme)} instead");
+                    case KeySecurityScheme when apiKey is not null:
```
archerzz, line 421: This might not be necessary, since in `SecuritySchemesComparer` it's ensured that only one `.Type` can exist.

archerzz (approved):

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
AlexanderSher: True. Which means that it is implemented incorrectly. Nice catch!

AlexanderSher:
And... we have found an example of the service that defines more than one ApiKey schema with different keys:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/TranslatorText/stable/v3.0/TranslatorText.json#L14

In azure-sdk-for-net, we use only first value: https://github.com/Azure/azure-sdk-for-net/blob/e7f6846812af0756d2278df605e06ebeb74c7934/sdk/translation/Azure.AI.Translation.Document/src/Constants.cs#L8 I don't know if this is correct.

We can either relax the checking condition (allow multiple schemas and pick the first one) or define some behavior for this case

CC: @KrzysztofCwalina , @tg-msft , @annelo-msft , @lmazuel

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -404,5 +404,93 @@ public static IReadOnlyList<string> GetApiVersions(CodeModel codeModel)
...
+
+                if (x.GetType() != y.GetType())
+                {
+                    return false;
+                }
+
+                return (x, y) switch
+                {
+                    (KeySecurityScheme apyKeyX, KeySecurityScheme apyKeyY)
+                        => apyKeyX.Type == apyKeyY.Type && apyKeyX.Name == apyKeyY.Name,
```
lirenhe, line 468: Is it apiKeyX and apiKeyY?

PR #2633, opened 2022-08-22
Add a patch in GetResourceFromResourceType

Yao725:
This PR includes two parts:

- Add a patch in GetResourceFromResourceType

> Previous we use resource type to get the related Resources in MgmtRestClient, it works well in most of the scenarios, but will introduce some unexpected behaviors in some rare cases.
> For example, we have following three resources in GuestConfigurations:
> ```text
> /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}/providers/Microsoft.GuestConfiguration/guestConfigurationAssignments/{guestConfigurationAssignmentName}
> /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.HybridCompute/machines/{machineName}/providers/Microsoft.GuestConfiguration/guestConfigurationAssignments/{guestConfigurationAssignmentName}
> /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachineScaleSets/{vmssName}/providers/Microsoft.GuestConfiguration/guestConfigurationAssignments/{name}
> ```
> All of these three resources have the same resource type, aka Microsoft.GuestConfiguration/guestConfigurationAssignments, but they expand from different resource. If we only compare the resource type, we will miss the extension information and provide three candidates for every related method of these three resources. It will make us fail in the construction of MgmtRestOperation. Therefore I put more restrictions in the method `GetResourceFromResourceType` . When all the segments of the operation's resource type are constant and more than one candidate resources are found, we will do a further check in request path to select the right candidate.

- Introduce a new configuration to prevent wrapping return type

> A list method should return a data instead of the resource when the data is shared by multiple resources. However, our current implementation can't do that when the return type is exactly the data of the rest operation's resource, so I added a configuration to make the code generate as expected...

Regen PR in sdk repo: https://github.com/Azure/azure-sdk-for-net/pull/30655
# Description
<i></i>

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/OperationExtensions.cs
```diff
@@ -229,6 +229,13 @@ internal static IEnumerable<Resource> GetResourceFromResourceType(this Operation
             var resourceType = operation.GetRequestPath().GetResourceType();
             var candidates = MgmtContext.Library.ArmResources.Where(resource => resource.ResourceType.DoesMatch(resourceType));
 
+            // We should not find more than one candidate if all the segments in ResourceType is constant.
+            // If this happen, We determine the right candidate by doing a further check on the request path.
+            if (candidates.Count() > 1 && !resourceType.Any(segment => segment.IsReference))
```
ArcturusZhang, line 234: We could use `resourceType.IsConstant` instead of this linq expression here
```suggestion
            if (candidates.Count() > 1 && resourceType.IsConstant)
```

src/AutoRest.CSharp/Mgmt/Decorator/OperationExtensions.cs
```diff
@@ -229,6 +229,13 @@ internal static IEnumerable<Resource> GetResourceFromResourceType(this Operation
...
+            {
+                return candidates.Where(resource => resource.RequestPath.Equals(operation.GetRequestPath()));
```
ArcturusZhang, line 236: Can we introduce a variable to hold the result of `operation.GetRequestPath()` above and use it directly here?

src/AutoRest.CSharp/Mgmt/Decorator/OperationExtensions.cs
Yao725: fixed

src/AutoRest.CSharp/Mgmt/Decorator/OperationExtensions.cs
Yao725: fixed

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
```diff
@@ -421,6 +427,9 @@ internal static MgmtConfiguration LoadConfiguration(JsonElement root)
             var prependRPPrefixList = prependRPPrefix.ValueKind == JsonValueKind.Array
                 ? prependRPPrefix.EnumerateArray().Select(t => t.ToString()).ToArray()
                 : Array.Empty<string>();
+            var disableResourceReturnList = disableResourceReturn.ValueKind == JsonValueKind.Array
+                ? prependRPPrefix.EnumerateArray().Select(t => t.ToString()).ToArray()
```
ArcturusZhang, line 431: This is incorrect

src/AutoRest.CSharp/Mgmt/Models/MgmtRestOperation.cs
```diff
@@ -385,14 +385,17 @@ private CSharpType GetWrappedMgmtReturnType(CSharpType? originalType)
...
                 return Resource.Type;
 
             var foundResources = MgmtContext.Library.ArmResources.Where(resource => resource.ResourceData.Type.Equals(originalType)).ToList();
             return foundResources.Count switch
             {
                 0 => throw new InvalidOperationException($"No resource corresponding to {originalType?.Name} is found"),
-                1 => foundResources.Single().Type,
+                1 => shouldReturnData ?
+                    throw new InvalidOperationException($"When resource data is only used by one resource, we don't support directly returning the data.") :
+                    foundResources.Single().Type,
```
ArcturusZhang, line 398: What is the usage if this configuration? I only see it is used here, and if it is configured, we will throw an exception?

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
Yao725: My bad, I just cope-paste and forgot to change it ==

src/AutoRest.CSharp/Mgmt/Models/MgmtRestOperation.cs
Yao725: When I fix the issue in GuestConfiguration, I noticed that when the data model is the same as the data in the resource, this [check](https://github.com/Azure/autorest.csharp/blob/10b43d4cc1bc01f5c114501fe89a050683b69d1d/src/AutoRest.CSharp/Mgmt/Models/MgmtRestOperation.cs#L389) will be true, which means that we will return the resource even though the data model is shared. And I believe this is not expected behavior.

ArcturusZhang (approved):
LGTM
BTW is this new configuration a workaround or a permanent configuration?
If this is a workaround, could you please add some comment in the MgmtConfiguration class to emphasize that this is a workaround and will be removed in the future?
If this is a permanent configuration, please also update the document in `doc/mgmt` to tell everyone what this confgiuration is for and how to use it (maybe also when to use it)

ArcturusZhang (approved):

PR #2640, opened 2022-08-24
Fix field rewriting

AlexanderSher:
FieldDeclarationSyntax may contain more than one field declaration, so `SemanticModel.GetDeclaredSymbol` has been returning `null` for it. 

For testing, added customization to the `SchemaMappingClientOptions`.


heaths (approved):
What about if 1 or the other of `LatestVersion` or `ServiceVersion` is defined? Is that even supported? If not, shouldn't there be some actionable error message? I'd be fine with it if not supported, but an actionable error message would be very helpful to others. If it is supported, maybe 2 additional test cases for each isolated case.

AlexanderSher:
If more than one property is defined, that would generate normal compiler error that class already contains definition of said field or constant. Resulting code won't compile.

It maybe a good idea to generate warnings for every duplicated type or member, but that is a separate work item.

heaths:
No, I mean, as a client dev, can I also only define either `LatestVersion` or `ServiceVersion` and let the generator generate the other? Perhaps an edge case, but still worth testing if supported to make sure it doesn't get regressed from whatever is currently supported.

PR #2641, opened 2022-08-24
Fix #2391: DPG replaces existing ServiceVersion values in ClientOptions

### LINKED ISSUE #2391: DPG replaces existing ServiceVersion values in ClientOptions
The DPG-generated `*ClientOptions.ServiceVersion` enum only ever has 1 version defined as value 1. This is a problem for CLU since we just shipped a 1.0 GA with `V2022_05_01 = 1` and now need to ship a CLU 1.1 beta with `V2022_05_15_Preview = 2` (note also that the casing is wrong by default; see #1524).

After a lot of trial and error as well as some offline advice, I found I could work around this for now using `[CodeGenSuppressType("ConversationsClientOptions")]` and just author the whole thing myself, but since SDKs will often have to support new versions this is something we'll want to handle better, I hope.

AlexanderSher:
PR introduces `CodeGenOverrideServiceVersionsAttribute` that replaces service versions defined in swagger, propagating these versions into ClientOptions type.


archerzz (approved):
LGTM. Just one open question: we assume the SDK developers should add the current swagger API version into the override attribute? How about joining them? I feel that could be easier to use. 

Anyway, current implementation is sufficient to support multiple api versions.

Issue #2550, opened 2022-08-05
Confirm DPG v1.0 and DPG v2.0 parity with regard to resiliency


annelo-msft:
### DPG v1.0 and DPG v2.0 parity

This issue tracks the work to show that the DPG v2.0 generator can generate the same client output from a Cadl file as the DPG v1.0 generator can from a swagger file.

#### Cadl Inputs

- [dev-driven](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/resiliency/dev-driven)
- [srv-driven-1](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/resiliency/srv-driven-1)
- [srv-driven-2](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/resiliency/srv-driven-2)

#### Expected Outputs

The generated clients should match those generated from the corresponding autorest.testserver swagger files:

- [dpg-customization.json](https://github.com/Azure/autorest.testserver/blob/main/swagger/dpg-customization.json)
- [dpg-initial.json](https://github.com/Azure/autorest.testserver/blob/main/swagger/dpg-initial.json)
- [dpg-update1.json](https://github.com/Azure/autorest.testserver/blob/main/swagger/dpg-update1.json)

#### Test Coverage

The tests written for the DPG v1.0 clients should work against the DPG v2.0 clients as well.

lirenhe:
I think this could be part  of the effort of Test server support. 
@annelo-msft, what is the priority of this item?

pshao25:
@annelo-msft looks like we only need to do the comparison, there is no coding effort, correct?

I would still put these three test scenarios into #2467

pshao25:
Current gap:
1. Fixed enum is not in the August scope: https://github.com/Azure/cadl-ranch/blob/fa1f20e21bf8dfa6316f7b2025cb7670ffbff068/packages/cadl-ranch-specs/http/resiliency/dev-driven/main.cadl#L26
2. paging and lro cannot be filtered now. Tracked by https://github.com/Azure/autorest.csharp/issues/2565
3.

lirenhe:
Will move to 2.0 beta as srv-driven2 is not a correct CADL file.

lirenhe:
@pshao25, seems only 9&10 left and they are not the scope for beta2,   I plan to move this item to 2.0GA.
Is there anything else left for this item?

[...]

PR #2652, opened 2022-08-29
Fix #2650: Support Customization with CADL input in DPG

### LINKED ISSUE #2650: Support Customization with CADL input in DPG
Need to verify whether CADL+DPG could still support the customization on models:
Detailed description could be found in https://github.com/Azure/autorest.csharp#customizing-the-generated-code

**Pre-generation customizations:**
- [ ] Make a model internal (only model itself becomes internal - all usages of the models must be updated manually)
- [ ] Rename a model class
- [ ] Change a model or client namespace
- [ ] Make model property internal
- [ ] Rename a model property
- [ ] Change a model property type
- [ ] Preserve raw Json value of a property
- [ ] Convert property to field
- [ ] Renaming an enum
- [ ] Renaming an enum member
- [ ] Changing an enum to an extensible enum
- [ ] Make a client internal
- [ ] Rename a client

**Post-generation customizations:**
- [ ] Replace any generated member
- [ ] Customize serialization/deserialization methods (replaces any )
- [ ] Changing member doc comment (doc comment is changed with generated member)
- [ ] Remove any generated member
- [ ] Remove any type
- [ ] Extending a model with additional constructors (simply means that models are generated as partial classes)

**Configuration settings (not really a customization):**
- [ ] Exclude models from namespace

**Swagger-specific customizations (out-of-scope)**
- Change model namespace or accessibility in bulk
- Change operation accessibility in bulk

With model generation , we should decide what kinds of customization we should support in CADL+DPG and verify/implement it.

cc @AlexanderSher  @annelo-msft

AlexanderSher:


src/CADL.Extension/Emitter.Csharp/src/type/InputTypeKind.ts
```diff
@@ -7,6 +7,7 @@ export enum InputTypeKind {
     Boolean = "Boolean",
     DateTime = "DateTime",
     Dictionary = "Dictionary",
+    Duration = "DurationISO8601",
```
AlexanderSher, line 10: Small fix to correctly support duration

test/TestProjects/Customizations-Cadl/Customizations/RootModel.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace CustomizationsInCadl
```
lirenhe, line 4: I didn't see this case in the doc, do we need to support this?

lirenhe:
@annelo-msft, i would like to confirm with you whether all those customization cases should be supported in CADL+DPG

test/TestProjects/Customizations-Cadl/Customizations/RootModel.cs
AlexanderSher: It is not a customization. This model is needed to reference all other models so that they are correctly referenced. Since `ModelToMakeInternal` customization changes accessibility of the model but not of its usages, we need to update it manually.

AlexanderSher:
Sidecar has to provide language-specific customizations, and I'm very skeptical about having a good way of customizing entities that exist only in C# code (multiple constructors, type conversions, etc.)

annelo-msft:
> @annelo-msft, i would like to confirm with you whether all those customization cases should be supported in CADL+DPG.

@lirenhe, we have not closed on the customization story for DPG+CADL yet.  Closing on this story is not required for August MVP.  It is up to you whether you want to merge this PR or not, but if it's required for anything in the August MVP, then we are doing something incorrectly.

AlexanderSher:
> if it's required for anything in the August MVP, then we are doing something incorrectly.

It is required to workaround https://github.com/Azure/cadl-ranch/pull/49, which is a blocker for https://github.com/Azure/autorest.csharp/issues/2550.

annelo-msft:
> It is required to workaround https://github.com/Azure/cadl-ranch/pull/49, which is a blocker for https://github.com/Azure/autorest.csharp/issues/2550.

That issue is for parity with DPG 1.0.  DPG 1.0 did not generate models, therefore we shouldn't need model customization to achieve parity with DPG 1.0.  So, something is incorrect.

AlexanderSher:
> That issue is for parity with DPG 1.0

Well, https://github.com/Azure/autorest.csharp/issues/2550 is not about parity in terms of API surface, it is about parity in terms of resiliency, meaning that "DPG 2.0 handles client generation from CADL when upgraded from v1 to v2 as good as DPG 1.0 handles client generation from swagger when upgraded from v1 to v2".

annelo-msft:
I disagree, but at the end of the day it doesn't matter what I think - please work with @lmazuel to clarify his intention for this test case and what the requirements for DGP 1.0 parity are.  Thanks!

annelo-msft:
Independent of whether or not this is needed for DPG 1.0 parity, I just chatted with @KrzysztofCwalina, and he would like to keep these customizations available for use in .NET SDKs.  We can slowly shift the customization balance to Cadl over time with the shift left effort, but this effort should not block DPG 2.0.  @lmazuel, @tg-msft, FYI.

lirenhe (approved):

tg-msft:
This is high interest technical debt.  If we're going to keep it, I'd propose we never let anything GA using it (possibly adding analyzers to warn about it) and only allow it for short term Preview workarounds until the Cadl customization story adds any missing features.

KrzysztofCwalina:
I would be fine with having a system in place where all such customizations need to be approved explicitly. I don't agree with banning such customizations in GA libraries altogether. The main mission of our team is to ship great libraries and we need to have tools that allow arbitrary customizations.

Issue #2643, opened 2022-08-24
No headers in request method


pshao25:
**Symptom:**
In request method, there is no header set:
https://github.com/pshao25/autorest.csharp/blob/4dc1fd395672d24b5f3f5219d8ef1b5456a65a75/test/CadlRanchProjects/dev-driven/Generated/TestserverClient.cs#L494-L507
Post content cannot be successfully sent. (at least cadl ranch cannot get this value.)

**Expected:**
https://github.com/pshao25/autorest.csharp/blob/4dc1fd395672d24b5f3f5219d8ef1b5456a65a75/test/TestServerProjectsLowLevel/dpg-customization/Generated/DPGClient.cs#L518-L531

We should have header as
https://github.com/pshao25/autorest.csharp/blob/4dc1fd395672d24b5f3f5219d8ef1b5456a65a75/test/TestServerProjectsLowLevel/dpg-customization/Generated/DPGClient.cs#L528-L529

**Root cause:**
We should have clientMethod.Request.Headers to write headers.
https://github.com/pshao25/autorest.csharp/blob/4dc1fd395672d24b5f3f5219d8ef1b5456a65a75/src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs#L221

We build header parameters by `InputParameter.Location`. However, there are two parts don't have it.

- Output of emitter don't have it
- We initialize InputParameter hard coded as `RequestLocation.None`
https://github.com/pshao25/autorest.csharp/blob/4dc1fd395672d24b5f3f5219d8ef1b5456a65a75/src/AutoRest.CSharp/Common/Input/Inputs.cs#L93

chunyu3:
Normal head parameter defined in operation will be emit and mark the location 'head'. but the top-level head such as 'content-type' is set to each operation.

AlexanderSher:
M4 has generated two parameters:
- contentType: https://github.com/Azure/autorest.csharp/blob/b20e5ee38dce9d743b5c3390118433e2499ff4c1/test/TestServerProjectsLowLevel/dpg-customization/Generated/CodeModel.yaml#L343-L355
- accept: https://github.com/Azure/autorest.csharp/blob/b20e5ee38dce9d743b5c3390118433e2499ff4c1/test/TestServerProjectsLowLevel/dpg-customization/Generated/CodeModel.yaml#L368-L380

Both have types declared as constants.

AlexanderSher:
[...]

PR #2656, opened 2022-08-29
support content-type and accept parameter

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2643

- support content-type in CADL
- support Accept parameter

# Checklist

# Ready to Land?


src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -129,6 +137,38 @@ function createModel(program: Program): any {
...
+            } as InputPrimitiveType,
+            Location: RequestLocation.Header,
+            IsApiVersion: false,
+            IsResourceParameter: false,
+            IsContentType: true,
+            IsRequired: true,
+            IsEndpoint: false,
+            SkipUrlEncoding: false,
+            Explode: false,
+            Kind: InputOperationParameterKind.Method,
```
AlexanderSher, line 160: This one should be `InputOperationParameterKind.Constant`

chunyu3: Hello @AlexanderSher Those parameters are used in method signature, they should be InputOperationParameterKind.Method. What is 'InputOperationParameterKind.Constant' for?

AlexanderSher: This is for pseudo-parameters that are not part of the signature, but instead we use their default value in request:
https://github.com/Azure/autorest.csharp/blob/8f0859eebff16a9f28cc6c6b5ea926dbb1153e03/src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs#L384

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -276,6 +330,8 @@ function loadOperation(
     const summary = getSummary(program, op);
     const externalDocs = getExternalDocs(program, op);
 
+    const consumes = getConsumes(program, op);
+    const produces = getProduces(program, op);
```
lirenhe, line 334: I didn't see this 'produces' used in the code, where do we plan to use it?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -129,6 +137,38 @@ function createModel(program: Program): any {
...
+            SkipUrlEncoding: false,
+            Explode: false,
+            Kind: InputOperationParameterKind.Method,
+            DefaultValue: {
+                Type: {
+                    Name: "String",
+                    Kind: InputTypeKind.String,
+                    IsNullable: false
+                } as InputPrimitiveType,
+                Value: consumes[0]
```
lirenhe, line 167: what if Comsumes contains multple values?

test/TestProjects/FirstTest-Cadl/Generated/Demo2Client.cs
```diff
@@ -84,39 +84,54 @@ public Demo2Client(TokenCredential credential, Uri endpoint, CadlfirsttestClient
...
         /// <summary> Return hi again. </summary>
+        /// <param name="p2"> The String to use. </param>
+        /// <param name="p1"> The String to use. </param>
         /// <param name="action"> The RoundTripModel to use. </param>
+        /// <param name="contentType"> The String to use. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="action"/> is null. </exception>
-        public virtual async Task<Response<Thing>> HelloAgainAsync(RoundTripModel action, CancellationToken cancellationToken = default)
+        /// <exception cref="ArgumentNullException"> <paramref name="p2"/>, <paramref name="p1"/>, <paramref name="action"/> or <paramref name="contentType"/> is null. </exception>
+        public virtual async Task<Response<Thing>> HelloAgainAsync(string p2, string p1, RoundTripModel action, string contentType, CancellationToken cancellationToken = default)
```
pshao25, line 93: If there is only one value for content type, we should hard code it instead of exposing a parameter. See example https://github.com/pshao25/autorest.csharp/blob/0ea9774167b6e88bfeb539cbfc1cfa0a794721e3/test/TestServerProjects/media_types/Generated/MediaTypesClient.cs#L193

test/TestProjects/FirstTest-Cadl/Generated/Demo2Client.cs
```diff
@@ -167,15 +182,17 @@ public virtual Response<Thing> HelloAgain(RoundTripModel action, CancellationTok
         /// </code>
         /// 
         /// </remarks>
-        public virtual async Task<Response> HelloAgainAsync(RequestContent content, RequestContext context = null)
+        public virtual async Task<Response> HelloAgainAsync(string p2, string p1, RequestContent content, ContentType contentType, RequestContext context = null)
```
pshao25, line 185: Where do we define `ContentType`? In Azure.Core or Models?

pshao25 (dismissed):
This is not what I expect. I want there is hard code header like:
https://github.com/pshao25/autorest.csharp/blob/0ea9774167b6e88bfeb539cbfc1cfa0a794721e3/test/TestServerProjectsLowLevel/dpg-customization/Generated/DPGClient.cs#L529

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -174,6 +214,20 @@ function createModel(program: Program): any {
                 modelMap,
                 enumMap
             );
+            if (
+                contentTypeParameter &&
+                op.Parameters.some(
+                    (value) => value.Location === RequestLocation.Body
+                ) &&
+                !op.Parameters.some(
```
archerzz, line 222: So looks like we won't add the content type parameter if it's already defined in the operation parameters? It seems a defensive fix just in case service team forget to define content type parameter?

AlexanderSher (changes requested):
`getConsumes` and `getProduces` are going to be deprecated: https://github.com/microsoft/cadl/issues/937

test/TestProjects/FirstTest-Cadl/Generated/Demo2Client.cs
chunyu3: In Azure.Core

AlexanderSher:
We can use this logic to extract content types: https://github.com/microsoft/cadl/blob/872a3897bb6e91269fe5b974719e6cf266e8507e/packages/openapi3/src/openapi.ts#L624-L629
To get values for accept header values, we can use `HttpOperationResponse.responses[0].body.contentTypes`, assuming that we support only one response content per status code.

test/TestProjects/FirstTest-Cadl/Generated/Demo2Client.cs
pshao25: I think we need to discuss it because HLC is using the generated `ContentType` model.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: Thanks @AlexanderSher I updated it.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: updated. only when one value, it will set as default value for later usage in codegen. If multiple values, don't set default value.

test/TestProjects/FirstTest-Cadl/Generated/Demo2Client.cs
chunyu3: I think if the content-type is one value and it is 'application/json' we may hard code in Create***Request; if multiple-value, will add contentType parameter

test/TestProjects/FirstTest-Cadl/Generated/Demo2Client.cs
chunyu3: In protocol method, it use Azure.Core.ContentType since there is no model class generated in DPG.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: global consumes/produces parameter will apply to all operations. if one operation defines its own content-type, it will override global setting.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: we will not support @consumes @produces as operation decorator. remove them.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -303,11 +309,16 @@ export function getInputType(
...
+    function getInputTypeForEnum(
+        e: EnumType,
+        addToCollection: boolean = true
+    ): InputEnumType {
         let enumType = enums.get(e.name);
         if (!enumType) {
             if (e.members.length == 0) {
-                throw new Error(`Enum type '${e.name}' doesn't define any values.`);
+                throw new Error(
+                    `Enum type '${e.name}' doesn't define any values.`
```
AlexanderSher, line 320: Why this formatting changes? Do we have any rules about where to insert new lines?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -326,19 +427,32 @@ function loadOperation(
...
-            InputOperationParameterKind.Method;
+        const isContentType: boolean =
+            requestLocation === RequestLocation.Header &&
+            name.toLowerCase() === "content-type";
+        const kind: InputOperationParameterKind = isContentType
+            ? InputOperationParameterKind.Constant
+            : InputOperationParameterKind.Method;
         return {
-            Name: name,
+            Name: param.name,
```
AlexanderSher, line 446: What's the difference between `name` and `param.name`?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: It's prettier rule. There is an constrain of the length of each line.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: `param.name` will remain the same as the parameter defined. but cadl compile will change the `name` according to name convention. e.g. `@header contentType`, cadl compile will change the `name` to 'content-type', and remain the `param.name` as defined

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
lirenhe: Let's skip this case for now

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -388,25 +400,34 @@ export function getInputType(
...
+                properties,
+                baseModel?.DiscriminatorPropertyName
+            );
 
             // Temporary part. Derived types may not be referenced directly by any operation
-            // We should be able to remove it when https://github.com/Azure/cadl-azure/issues/1733 is closed 
+            // We should be able to remove it when https://github.com/Azure/cadl-azure/issues/1733 is closed
             if (model.DiscriminatorPropertyName && m.derivedModels) {
-                m.derivedModels.forEach(dm => {
+                m.derivedModels.forEach((dm) => {
```
AlexanderSher, line 412: Since you're changing this place, can we use `for .. of` here?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -130,6 +138,72 @@ function createModel(program: Program): any {
...
+            IsApiVersion: false,
+            IsResourceParameter: false,
+            IsContentType: true,
+            IsRequired: true,
+            IsEndpoint: false,
+            SkipUrlEncoding: false,
+            Explode: false,
+            Kind: InputOperationParameterKind.Constant,
+            DefaultValue:
+                consumes.length === 1
```
AlexanderSher, line 162: Can we extract this into a helper function? It looks like logic is the same for `produces` and `consumes`

src/AutoRest.CSharp/Common/Input/CadlInputTypeConverter.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using AutoRest.CSharp.Mgmt.AutoRest.PostProcess;
```
lirenhe, line 7: Do we need to use this class?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: updated.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: Thanks @AlexanderSher . Good Suggestion. I wrapped them into a function.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -207,6 +253,41 @@ function createModel(program: Program): any {
...
+        IsApiVersion: false,
+        IsResourceParameter: false,
+        IsContentType: isContentType,
+        IsRequired: true,
+        IsEndpoint: false,
+        SkipUrlEncoding: false,
+        Explode: false,
+        Kind: InputOperationParameterKind.Constant,
+        DefaultValue:
+            mediaTypes.length === 1
```
lirenhe, line 282: Shall we still use     mediaTypes.length >= 1  
So if there are multiple values, we pick the first one.

lirenhe (approved):
Left one comment, other parts look good.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: No. when it has multiple types, no default value will be set.

AlexanderSher (approved):

PR #2658, opened 2022-08-29
Some auth code cleanup

AlexanderSher:
- ApiKey presented as object instead of string
- OAuth2 collects scopes from all flows instead of just the last one
- All OAuth2 flows that have `scopes` are included
- Custom converters for `InputAuth` and `InputOAuth2Auth` are removed in favor of default constructors


chunyu3:
Hello @AlexanderSher I saw the CaldInputAutherConverter and CaldInputOAuth2OAuthConverter have been deleted. Can Utf8Json Serialize support `id` and `ref` and `nested object` by default, and don't need to write converter to handle?

archerzz (approved):

AlexanderSher:
> Can Utf8Json Serialize support id and ref and nested object by default, and don't need to write converter to handle?

Yes, it can handle reference resolution, but it can't handle type inheritance and reference resolution for the same object (at least version of System.Text.Json that is currently in use). This is why we have converters for `InputType` and its derived types.

PR #2660, opened 2022-08-30
Support string union parameter

pshao25:
We need to support this
https://github.com/Azure/cadl-ranch/blob/7435655e82b732b3573379f76e0e613423d244e9/packages/cadl-ranch-specs/http/resiliency/srv-driven-2/main.cadl#L109-L115


src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -215,7 +216,8 @@ export function getInputType(
     program: Program,
     type: Type,
     models: Map<string, InputModelType>,
-    enums: Map<string, InputEnumType>
+    enums: Map<string, InputEnumType>,
+    name: string = ""
```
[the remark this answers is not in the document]
pshao25: Take this line as an example: `@path contentTypePath: "json" | "jpeg"`

`name` is the parameter name `contentTypePath`
`type` is the union type `"json" | "jpeg"`, it is actually an anonymous type so there is no name for type.

AlexanderSher, line 220: This is a potential error prone:
- We call `getInputType` from other `getXxxType` methods, but in these methods name will always be default.
- Two parameters may have the same name but different set of stings in the enum.

AlexanderSher:
I think the right solution for the unions would be to add `InputUnionType` to emitter and autorest.csharp and pass it directly into converter, with their own collection in `InputNamespace`, separate from enums:
```cs
internal record InputNamespace(
    string Name,
    string Description,
    IReadOnlyList<string> ApiVersions,
    IReadOnlyList<InputEnumType> Enums,
    IReadOnlyList<InputUnionType> Unions,
    IReadOnlyList<InputModelType> Models,
    IReadOnlyList<InputClient> Clients,
    InputAuth Auth)
```

Then, resolve types while creating type providers. 

However, I think that for August MVP it will be enough to treat union of strings simply as string.

pshao25:
> I think the right solution for the unions would be to add `InputUnionType` to emitter and autorest.csharp and pass it directly into converter, with their own collection in `InputNamespace`, separate from enums:

The equivalent in swagger is Enum, see [here](https://cadlplayground.z22.web.core.windows.net/?c=aW1wb3J0ICJAY2FkbC1sYW5nL3Jlc3QiOw0K0xt2ZXJzaW9uaW5nxCENCnVzaW5nIENhZGwuVskbxRpAxytlZCjHGnMpDQpAc2VydmljZVRpdGxlKCJEUEdDbGllbnQiyxzHLygiMS4xLjDJGmVyKCJodHRwOi8vbG9jYWxob3N0OjMwMDAiLCAiVGVzdMYlIGVuZHBvacdTZG9jxWYgU3dhZ2dlciwgdGhpcyBpcyB0aGUgaW5pdGlhbCBzxh0gYSDnAIMgY291bGQgZG%2FFRXJvdXRlKCIvxx1Ecml2ZW4yxBtuYW1lc3BhY2UgRFBHU80fO%2B0BHUh0dHDlARdlbnVtIOgBESB7DQogIHYxMDogIjEuMC4wIizGETHFEeQA%2FiwNCn3EOG1vZGVsIFBvc3RJbnB1dMY6dXJsOiBzdHLmAXTLKE1lc3NhZ2XGJm3GDVByb3BlcnR50DLmASRBbnN3ZXIgZnJvbegBBeQA4c1PUmVzcG9uc8dXQHN0YXR1c0NvZGUgygs6IDIwMDvFIGJvZHkgxAU6yEHJc%2FUBUsRwaW50ZXJm5AFRcGFyYW3nASLJMMUVZXRlcnMve2NvbnRlbnRUeXBlUGF0aH3EPyAg5gDOUE9TVCBhIEpTT04gb%2BQB2EpQRUfHIXBvc3TECcQIUMlQKMQTxB5hdGggz1s6ICJqc29uIiB8ICJqcGVn5gGmICBA6QHbLmhlYWRlciAixzotdHlwZSI6ICJhcHBsaWNhdGlvbi%2FFROUBG%2BgAp0kgYW0gYeUBJOYA8OQAiy4gTXkgb25seSB2YWxpZOYAxWVudHJ55AK3e%2BYB9VzoAwtleGFtcGxlLm9yZy9teWltYWdlLuQApFwiIOcBEucBisYLyWY66gJMxBop6QGk6AHl5gGs). And we will treat it as an enum in HLC as well. What is the benefit if we introduce an `InputUnionType`.

> However, I think that for August MVP it will be enough to treat union of strings simply as string.

What do you mean by "treat union of strings simply as string"? Do you mean we should not introduce an enum?

AlexanderSher:
> No matter modelAsString set to true or false, the results are the same.

No, the difference is that when `modelAsString: false`, M4 produces `SealedChoiceSchema`, and when `modelAsString: true`, M4 produces `ChoiceSchema`:

image

image

But this is swagger-to-codemodel logic, which is irrelevant for our case.

With CADL, we can have access to types directly, and I don't want to split type creation between emitter and autorest.csharp, considering all the known issues with name resolution, accessibility and usage scope.  Even in this case, if we change `@path` to `@header`, we should generate protocol method with `ContentType` parameter, and two separate convenience methods. 

Since @lmazuel has confirmed that we can skip the test in question, I would prefer not to spend time on it now.

lirenhe:
As discussed in today's meeting, let's hold this PR.

test/TestProjects/FirstTest-Cadl/FirstTest-Cadl.cadl
```diff
@@ -63,7 +63,7 @@ model RoundTripModel {
 namespace Hello.Demo {
     @doc("Return hi")
     @get
-    op sayHi(): Thing;
+    op sayHi(@path pathParameter: "json" | "jpeg"): Thing;
```
ArcturusZhang, line 66: May I comprehend this type expression as a sealed choice that contains two values `json` and `jpeg`?

PR #2688, opened 2022-09-08
Fix convenience method generating/naming issue

pshao25:
This issue is to fix two things:
1. When there is no input model and output model, we should not generate convenience method. E.g., `op delete(...PetId): Response<200> | Error;`
2. When extensible enum is an input parameter, we don't need to change the convenience method by appending a `value` suffix, because the signature of protocol method and convenience method has already been different. E.g., `@get op getPetByKind(@path kind: PetKind): Pet | Error;`


src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -143,7 +166,7 @@ private ReturnTypeChain BuildReturnTypes()
 
         private ConvenienceMethod BuildConvenienceMethod(ReturnTypeChain returnTypeChain)
         {
-            string name = _protocolBodyParameter == null
+            string name = !_orderedParameters.Where(parameter => parameter.Convenience != KnownParameters.CancellationTokenParameter).ToList().Exists(parameter => !IsParameterTypeSame(parameter.Convenience, parameter.Protocol))
```
lirenhe, line 169: Better create a function instead of duplicate the code.

lirenhe (approved):

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
ArcturusZhang, line 169: since we are using linq, better to use `Any` instead of `ToList().Exists(predicate)`, they should have the same result.
And as a complicated condition like this, please create a method to make this more readable
```suggestion
            string name = !_orderedParameters.Where(parameter => parameter.Convenience != KnownParameters.CancellationTokenParameter).Any(parameter => !IsParameterTypeSame(parameter.Convenience, parameter.Protocol))
```
And BTW, semantically, `!array.Any(predicate)` is equivalent to `array.All(predicate)`. Avoiding to use double negative could make our code cleaner

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -143,7 +166,7 @@ private ReturnTypeChain BuildReturnTypes()
...
                 ? _restClientMethod.Name.IsLastWordSingular()
                     ? $"{_restClientMethod.Name}Value"
                     : $"{_restClientMethod.Name.LastWordToSingular()}Values"
```
ArcturusZhang, line 172: And BTW this name has two layer of ternary condition operator which is not very readable, please also consider convert this to if else statement in a method and  reduce one layer of the ternary condition

test/TestProjects/PetStore-Cadl/Generated/PetsClient.cs
```diff
@@ -488,6 +446,126 @@ public virtual Response Create(RequestContent content, RequestContext context =
             }
         }
 
+        /// <param name="kind"> The PetKind to use. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<Response<Pet>> GetPetByKindAsync(PetKind kind, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
pshao25: Yes

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
pshao25: Actually in this class, there is a lot of such kind of inline operations. If we always to create a function for these operations, I'm worried about it would be hard to read. So I'd prefer Dapeng's suggestion.

PR #2695, opened 2022-09-09
Add description of CADL Pipeline high level design

AlexanderSher:
First version of CADL support design. It is not where current implementation is today, but rather what we want to have at the time of release.

@annelo-msft, feel free to fix any styling issues you see in the document


docs/CADL.md
```diff
@@ -0,0 +1,93 @@
...
+## Basic Terms
+- CADL pipeline: the process and components of C# code generation from cloud services API specification written in [CADL](https://github.com/microsoft/cadl).
+- Swagger pipeline: the process and components of C# code generation from REST API specification written in [OpenAPI](https://swagger.io/).
+- generator: autorest.csharp per se, last part of both pipelines that is directly responsible for code generation.
+- generator input: file or stream with input for autorest.csharp, produced by previous part in the pipeline.
+- CADL C# Emitter: part of the CADL pipeline. Emitters are used by CADL compiler to output result of its work.
+- CodeModel.yaml: generator input in Swagger pipeline.
+- CodeModel: deserialized object representation of CodeModel.yaml in autorest.csharp.
+- cadl.json: generator input in CADL pipeline produced by CADL emitter.
+- InputModel: deserialized object representation of cadl.json in autorest.csharp.
```
lirenhe, line 12: Here, the name inputModel & outputModel conflict with the model name used in ModelGeneration ( Input/Output/RoundTrip models).  is there any plan to do refactor/rename to avoid confusion?

docs/CADL.md
```diff
@@ -0,0 +1,93 @@
...
+
+## CADL C# Emitter
+Emitter generates cadl.json from subset of CADL compiler output object graph. Resulting output is a standard JSON file with circular references support. It doesn't have explicit specification. Instead, it is assumed that InputModel types defined in [TS](https://github.com/Azure/autorest.csharp/tree/feature/v3/src/CADL.Extension/Emitter.Csharp/src/type) and in [C#](https://github.com/Azure/autorest.csharp/tree/feature/v3/src/AutoRest.CSharp/Common/Input) match each other. When new features in generator require InputModel types to be modified or extended, there is no backwards compatibility requirement. Hence, releases of autorest.csharp and CADL C# Emitter must be synchronized so that autorest.csharp can deserialize `cadl.json` produced by emitter without errors.
+
+Emitter can apply C#-specific transformations and generation rules to the spec before or in process of cadl.json generation (language-agnostic transformations and rules are expected to be applied to spec before it becomes available to the emitter).
```
lirenhe, line 22: For C#-specific transformations, is there any guideline on what should be implemented in emitter or codegen?

docs/CADL.md
```diff
@@ -0,0 +1,93 @@
...
+
+InputModel is a slim, unvalidated representation of the cadl.json in memory. OutputModel is a rich, self-consistent set of objects that has all the data code writers need to generate valid C# code.
+
+With the exception of enums, no object should belong to InputModel and OutputModel at the same time.
+
+#### InputModel requirements
+- Minimalistic design. Only nessesary data is available, data duplication should be avoided (e.g. `InputModelType.DiscriminatorPropertyName` makes `InputModelProperty.IsDiscriminator` unnessesary). 
+- All InputModel objects are init-immutable (instance can't be modified after becoming accessible to external code).
+- InputModel is final. All decorator-based and configuration-based transformations and rules must be applied in emitter.
+- Lazy initializations must be avoided.
```
lirenhe, line 36: why do we highlight this? Is there any problem caused by lazy initialization?

docs/CADL.md
```diff
@@ -0,0 +1,93 @@
...
+## Unification of CADL and Swagger pipelines
+
+In Swagger pipeline, generator input is defined by [CodeModel specification](https://github.com/Azure/autorest/blob/main/packages/libs/codemodel/.resources/all-in-one/json/code-model.json). CodeModel.yaml itself is written in YAML. CodeModel types are over-complicated yet they aren't good enough to represent some schemas that can be defined in CADL. To share generation logic and types between Swagger pipeline and CADL pipeline, CodeModel object graph is transformed into InputModel object graph by `CodeModelConverter`. 
+
+#### Unification requirements
+
+- InputModel is sufficient. Its objects should have no properties or methods that return CodeModel objects. CodeModel-only parts of InputModel must have meaningful default values for CADL pipeline.  
+- All CodeModel rules and transformations must be applied before `CodeModelConverter` is invoked.
+
+## CADL + Swagger pipeline diagram
```
chunyu3, line 53: what is the process line `sidecar->codewriter->codegen target`? Is it the process to generate `configuration.json`?

docs/CADL.md
chunyu3, line 22: what c#-specific transformations or greneration rules the emitter will apply? The customization (something like as directive in swagger)?

docs/CADL.md
AlexanderSher: Yeah, this name conflict exists for years. We have model layers in generator itself and we generate models. I'm open to suggestions.

docs/CADL.md
AlexanderSher: Not yet. I hope this will be clarified when we start putting MPG on CADL rails, cause MPG is a heavy user of such transformations.

docs/CADL.md
AlexanderSher: Good point, I'll add `configuration.json` as a block

docs/CADL.md
AlexanderSher: Something like [RearrangeParameterOrder](https://github.com/Azure/autorest.csharp/blob/8f0859eebff16a9f28cc6c6b5ea926dbb1153e03/src/AutoRest.CSharp/Mgmt/Decorator/Transformer/RearrangeParameterOrder.cs#L12). 
Again, it is a subject of further discussion should we put this transformations into emitter or into generator (to share them between swagger and CADL paths).

docs/CADL.md
AlexanderSher: Yes. 

Lazy initialization pattern should be used when we don't want to evaluate expression until its value is needed and want to share the result (the other name of the pattern is [call-by-need](https://en.wikipedia.org/wiki/Lazy_evaluation)).

In autorest.csharp, however, it is used for completely different purpose. We know upfront that expression will be called, but we can't call it immediately because it doesn't have all the variables defined. So we declare it lazy, **believing** that we can delay access to the expression value up to the point when all its variables are defined already or can be defined  using similar logic. This is an anti-pattern. In MPG, it has already resulted in code that creates [all types twice](https://github.com/Azure/autorest.csharp/blob/110a403f43e6e6364bba97ced59f9e754d047774/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs#L748). That is not the worst thing that may happen though. In [Python extension for Visual Studio Code](https://github.com/Microsoft/vscode-python), where lazy initialization has been used for similar purposes (type resolution for IntelliSense and debugging), we got a case of 3 files with 20 lines of code combined that required analyzer more than an hour to process. After several unsuccessful attempts to fix it granularly, I had to rewrite the whole type resolution code. That work took several months.

So I want get rid of lazy initialization misuse while can do it bit by bit, starting by not adding more lazy initializations.

docs/CADL.md
AlexanderSher: Fixed

docs/CADL.md
lirenhe: I created https://github.com/Azure/autorest.csharp/issues/2712 and we could address that in MQ.

lirenhe (approved):

chunyu3 (approved):

PR #2705, opened 2022-09-14
Move npm modules from emitter folder to outside folder

pshao25:
1. Currently, we have two node_modules. One is in the root folder and the other is in the emitter folder, which is confusing. This change merges them together into the root folder.

2. Currently we need to run `.\eng\Generate.ps1 cadl-project -reset` to apply the latest change in the emitter, which costs. This change links the `@azure-tools/cadl-csharp` to the source code.


src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -36,14 +36,14 @@
         "dist/**"
     ],
     "devDependencies": {
+        "@azure-tools/cadl-azure-core": "0.6.0",
```
chunyu3, line 39: It is better to keep the order as before, just to update the version. In that way, the diff may more clear about what is changed.

eng/Generation.psm1
```diff
@@ -93,42 +93,17 @@ function Invoke-CadlSetup()
...
-            $package = $packages;
-            if ($packages.count -gt 1)
-            {
-                $package = $packages[0]
-            }
-            npm install $package
-        }
-        git checkout $repoRoot/package.json
-        git checkout $repoRoot/package-lock.json
+        npm run build
```
chunyu3, line 99: Will it use the emitter version build locally?

chunyu3 (commented):
Why delete package-lock.json under src/CADL.Extension/Emitter.Csharp?

eng/Generation.psm1
pshao25: You mean in CI? It will use the build in the build agent.

pshao25:
> Why delete package-lock.json under src/CADL.Extension/Emitter.Csharp?

There is no place we need that. Now we just need to run `npm install` from the root folder and all the dependencies are installed in the root folder.

src/CADL.Extension/Emitter.Csharp/package.json
pshao25: I'm trying to make it in alphabetical order.

chunyu3 (approved):

PR #2713, opened 2022-09-16
Apply default value "application/json" to contentype and accept

### LINKED ISSUE #2702: Set default/global content-type and accept header
As these decorators are [deprecated](https://github.com/Azure/cadl-ranch/pull/91#discussion_r968738512), we need to figure out a way to set default content-type and accept.
```
@Cadl.Rest.produces("application/json")
@Cadl.Rest.consumes("application/json")
```

Solutions:
1. We hard code the default value of content-type/accept header as "application/json".
2. We force user to define the content-type/accept header at operation level.
3. We configure default value in sidecar since this issue only exists in .NET.

pshao25:
Fix #2702, hard cord default value as "application/json". See details at: https://github.com/Azure/autorest.csharp/issues/2702#issuecomment-1249109692


lirenhe:
@pshao25, could you solve the conflict from this PR? thanks

archerzz (approved):

PR #2735, opened 2022-09-23
Ignore non-schema property on models

pshao25:
CADL:
```
@doc("Check whether client is authenticated.")
@get
@route("/invalid")
op invalid(): NoContentResponse | InvalidAuth;

@error
model InvalidAuth {
  @statusCode _: 403;
  error: string;
}
```

Expected:
```C#
public partial class InvalidAuth
{
    /// <summary> Initializes a new instance of InvalidAuth. </summary>
    /// <param name="error"></param>
    /// <exception cref="ArgumentNullException"> <paramref name="error"/> is null. </exception>
    public InvalidAuth(string error)
    {
        Argument.AssertNotNull(error, nameof(error));

        Error = error;
    }

    public string Error { get; set; }
}
```

Actual: Broken code with a property called `_`.
```C#
public partial class InvalidAuth
{
    /// <summary> Initializes a new instance of InvalidAuth. </summary>
    /// <param name=""></param>
    /// <param name="error"></param>
    /// <exception cref="ArgumentNullException"> <paramref name="error"/> is null. </exception>
    public InvalidAuth(int , string error)
    {
        Argument.AssertNotNull(error, nameof(error));

        _ = ;
        Error = error;
    }

    public int _ { get; set; }

    public string Error { get; set; }
}
```


src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -182,21 +182,25 @@ export function getEffectiveSchemaType(program: Program, type: Type): Type {
...
     }
 }
 
+/**
+ * A "schema property" here is a property that is emitted to OpenAPI schema.
+ *
+ * Headers, parameters, status codes are not schema properties even they are
+ * represented as properties in Cadl.
+ */
+function isSchemaPropertyOnProgram(program: Program, property: ModelProperty) {
```
chunyu3, line 196:
```suggestion
function isSchemaProperty(program: Program, property: ModelProperty) {
```

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -182,21 +182,25 @@ export function getEffectiveSchemaType(program: Program, type: Type): Type {
 
     return target;
 
-    /**
-     * A "schema property" here is a property that is emitted to OpenAPI schema.
-     *
-     * Headers, parameters, status codes are not schema properties even they are
-     * represented as properties in Cadl.
-     */
     function isSchemaProperty(property: ModelProperty) {
```
chunyu3, line 185:
```suggestion
    function isSchemaPropertyInternal(property: ModelProperty) {
```

eng/CadlGenerate.ps1
```diff
@@ -1,5 +1,7 @@
 #Requires -Version 7.0
 param($filter)
+
+$ErrorActionPreference = 'Stop'
```
chunyu3, line 4: It stops without error string. It is better to  check after command execution and give out error if fail.

pshao25: It will have error message. See example here: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1870922&view=logs&j=b5cdfb1e-c180-561f-918e-a169e451719e&t=aff748e3-476c-56d1-ae31-4029fde28653

Conversely, without this line the CI pipeline will always pass instead of stopping with error.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: It is better to change the function name.

chunyu3 (approved):

PR #2736, opened 2022-09-23
feat(dpg): generate lro protocol methods

archerzz:
- add lro running metadata for operation models in emitter
- upate generator input models to support deserialization of JSON input
- add test project

part of #2698

# Description

# Checklist

# Ready to Land?


src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -583,20 +598,45 @@ function loadOperation(
...
+
+        return {
+            FinalStateVia: OperationFinalStateVia.Location, // data plane only supports `location`
+            FinalResponse: finalResponse,
+        } as OperationLongRunning;
+    }
+
+    function loadLongRunningFinalResponse(program: Program, op: OperationDetails, resourceOperation?: ResourceOperation): OperationResponse | undefined {
+        for (const response of op.responses) {
+            if (response.statusCode == "200" || response.statusCode == "204")
```
chunyu3, line 616: If there are  both `200` and `204` responses, what is expected behavior? throw error? or choose one?

chunyu3: what is final response? how do we use it? Does every LRO have final response?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -197,7 +200,7 @@ function createModel(program: Program): any {
 
         for (const operation of routes) {
             console.log(JSON.stringify(operation.path));
-            if (!isSupportedOperation(operation)) continue;
+            if (!isSupportedOperation(operation, getAllLroMonitorOperations(routes, program))) continue;
```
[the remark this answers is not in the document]
archerzz: I saw the change in paging support. I think we still need to filter out the LRO Monitor operations. I don't think we should generate them. I talked with Java team who is going to do the similar thing.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
archerzz: > If there are both `200` and `204` responses, what is expected behavior? throw error? or choose one?

I think we prefer `200` which has some return content. `204` normally should not co-exist with `200`. If they both exist, `200` is better because it returns something. Actually here I copied the logic of converting autorest LRO: https://github.com/Azure/autorest.csharp/blob/bdc076cfe6dcfb1003416c9926c08039a5a05005/src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs#L74-L92

archerzz: > what is final response? how do we use it? Must every LRO have final response?

@chunyu3 This should be the return value of a success LRO. It doesn't necessarily return a value. We support HTTP `204` which returns no content. In the test project, I defined two operations: one has return value, another doesn't.

chunyu3: If we choose `200` when both `200` and `204` exist. But I am afraid that we cannot make sure that `200` response is always before `204` response in cadl (there is no rule to force the response order). so it is possible that the `204` response will be return as final response if `204` response is before `200` in `Response` array.

archerzz:
This PR is blocked by the excessive mode generation issue which should be fixed in #2747

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
archerzz: I'll update the logic to find all supported return responses, and return the best match.

archerzz:
> This PR is blocked by the excessive mode generation issue which should be fixed in #2747

Blocking issue is resolved.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -644,10 +664,46 @@ function loadOperation(
...
+        return {
+            FinalStateVia: OperationFinalStateVia.Location, // data plane only supports `location`
+            FinalResponse: finalResponse,
+        } as OperationLongRunning;
+    }
+
+    function loadLongRunningFinalResponse(program: Program, op: OperationDetails, resourceOperation?: ResourceOperation): OperationResponse | undefined {
+        let finalResponse: any | undefined;
+        for (const response of op.responses) {
+            if (response.statusCode == "200") {
```
chunyu3, line 683: use `===`

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -644,10 +664,46 @@ function loadOperation(
...
+                finalResponse = response;
+                break;
+            }
+            if (response.statusCode == "204") {
```
chunyu3, line 687: use "==="

chunyu3 (commented):
run `npm run prettier:fix` to prettier the .ts files. Thanks

chunyu3 (approved):

PR #2738, opened 2022-09-23
set name for anonymous body parameter

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2733

- when there is anonymous body parameter, set the name for it.

# Checklist

# Ready to Land?


src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -431,6 +431,12 @@ function loadOperation(
                 cadlParameters.bodyType
             );
             if (effectiveBodyType.kind === "Model") {
+                /* handle anonymous body model. */
+                if (effectiveBodyType.name === "") {
+                    let name = `${op.name}Properties`;
```
pshao25, line 436: Why we append the name with `Properties`? 

I think it should go to arch board? Even more, I think we should confirm whether such kind of cadl is valid first.

pshao25 (changes requested):
I think this need to be discussed.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -533,8 +539,8 @@ function loadOperation(
         const kind: InputOperationParameterKind =
             InputOperationParameterKind.Method;
         return {
-            Name: body.name,
-            NameInRequest: body.name,
+            Name: body.name !== "" ? body.name : "body",
```
archerzz, line 542: Can we consolidate the same logic into a method?

PR #2747, opened 2022-09-27
fix(dpg): intrinsic type is not correctly deduced

archerzz:
When a type has an intrinsic mode type, we should try to return the exact type instantly. The `indexer` property is not null for intrinsic model type. So sometimes the deduced type for intrinsic model type is not correct.

# Description

# Checklist

# Ready to Land?


src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -255,10 +257,19 @@ export function getInputType(
...
+        if (intrinsicName) {
+            switch (intrinsicName) {
+                case "string":
+                    const values = getKnownValues(program, m);
+                    if (values) {
+                        return getInputModelForExtensibleEnum(m, values);
+                    };
+                default:
+                    return {
+                        Name: m.name,
```
archerzz, line 269: I don't use `mapCadlIntrinsicModelToCsharpModel` here, because it will erase the original model's name. For example, `ResourceLocation` will be changed into `string`.
The cost is that in many existing `cadl.json` there will be changes of property names. But that won't change the generated codes.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -255,10 +257,19 @@ export function getInputType(
 
     function getInputModelType(m: Model): InputType {
         const intrinsicName = getIntrinsicModelName(program, m);
-        if (intrinsicName && intrinsicName === "string") {
-            const values = getKnownValues(program, m);
-            if (values) {
-                return getInputModelForExtensibleEnum(m, values);
+        if (intrinsicName) {
+            switch (intrinsicName) {
```
chunyu3, line 261: Shall we  wrap all the model type who has intrinsicName as intrinsic type? need to discuss with arch board.

archerzz: I think we should unbox all types, just like a compiler does.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -255,10 +257,23 @@ export function getInputType(
...
+                case "string":
+                    const values = getKnownValues(program, m);
+                    if (values) {
+                        return getInputModelForExtensibleEnum(m, values);
+                    };
```
AlexanderSher, line 266: What happens here in `else` case?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: If we wrap an model to primitive model, the model name should change too. The origin model is not emitted anymore.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
archerzz: @AlexanderSher I didn't put a `break;` here. So this case will fall back to `default:` which will just return a string type.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
archerzz: The primitive model is already represented by the `Kind`. I suggest we should keep the name for debug purpose.

test/TestProjects/Models-Cadl/Generated/cadl.json
```diff
@@ -649,7 +649,7 @@
      "Description": "",
      "Type": {
       "$id": "82",
-      "Name": "TimeSpan",
+      "Name": "duration",
```
lirenhe, line 652: why is this change?

lirenhe: There are some other changes like:
Datetime=>zonedDateTime,
bool=>boolean
Int32=>int32

Are they expected?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -279,19 +294,7 @@ export function getInputType(
             }
         }
 
-        if (m.name === intrinsicName) {
-            // if the model is one of the Cadl Intrinsic type.
-            // it's a base Cadl "primitive" that corresponds directly to an c# data type.
-            // In such cases, we don't want to emit a ref and instead just
-            // emit the base type directly.
-            return {
-                Name: mapCadlIntrinsicModelToCsharpModel(program, m) ?? m.name,
```
lirenhe, line 288: What is the reason behind to remove 'mapCadlIntrinsicModelToCsharpModel'

test/TestProjects/Models-Cadl/Generated/cadl.json
archerzz: This is exactly what Crystal and I discussed above: https://github.com/Azure/autorest.csharp/pull/2747#discussion_r984156640

If a property is of type `ResourceLocation` which is actually a `string`. Previously we used `string` as the value for both `Kind` and `Name`. I think we should use `string` as `Kind`, but keep the original name for `Name`.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
archerzz: I think this code will not be hit. So I deleted it.

chunyu3 (approved):

PR #2752, opened 2022-10-06
Fix #2726: DPG LROs for PATCH method do not return final response from original URI

### LINKED ISSUE #2726: DPG LROs for `PATCH` method do not return final response from original  URI

AlexanderSher:


heaths (approved):

AlexanderSher:
Non-live tests have passed

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -233,7 +233,7 @@ internal static bool TryGetApiVersion(Uri startRequestUri, out ReadOnlySpan<char
             // 1.Original request is a put method;
             // 2.Original request is a management plane patch method. For data plane patch method, the final uri will be determinned by the original response header and "_finalStateVia";
             // 3.Original response header contains "Location" and FinalStateVia is configured to OriginalUri.
-            if (_requestMethod == RequestMethod.Put || _requestMethod == RequestMethod.Patch && _startRequestUri.Scheme == "https" && _startRequestUri.Host == "management.azure.com" || _originalResponseHasLocation && _finalStateVia == OperationFinalStateVia.OriginalUri)
```
Yao725, line 236: Hi @AlexanderSher, just FYI, the check condition you took off here is used to avoid regression in the data plane package `Azure.Communication.PhoneNumbers`. According to the [swagger](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/communication/data-plane/PhoneNumbers/stable/2021-03-07/phonenumbers.json#L276-L351), the final result of the operation `PhoneNumbers_UpdateCapabilities` in this RP will be retrieved by the URL in the Location header. Therefore I think you might need to double check this change when trying to update the autorest version in sdk repo.

PR #2758, opened 2022-10-10
Revert fix for #2726: DPG LROs for PATCH method do not return final response from original URI

AlexanderSher:
Reverting due to failure in Azure.Communication.PhoneNumbers.Tests (see comment from @Yao725: https://github.com/Azure/autorest.csharp/pull/2752#discussion_r990727037)


jsquire:
@AlexanderSher :  Before we revert, would you please try re-recording the PhoneNumbers tests and see if the failure is due to stale recordings or there's something bigger going on?

archerzz (approved):

PR #2764, opened 2022-10-12
Add LocationOverride value to OperationFinalStateVia

AlexanderSher:
This PR adds a LocationOverride value to OperationFinalStateVia which is never generated, but can be used in partial classes to override behavior when `x-ms-long-running-operation-options.final-state-via` is set explicitly, e.g.: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/communication/Azure.Communication.PhoneNumbers/src/UpdatePhoneNumberCapabilitiesOperation.cs#L14

### DIFF
--- a/src/assets/Generator.Shared/OperationFinalStateVia.cs
+++ b/src/assets/Generator.Shared/OperationFinalStateVia.cs
@@ -7,8 +7,10 @@ namespace Azure.Core
 {
     internal enum OperationFinalStateVia
     {
+        LocationOverride,
         AzureAsyncOperation,
         Location,
-        OriginalUri
+        OriginalUri,
+        OperationLocation
     }
 }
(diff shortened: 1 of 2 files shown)


src/assets/Generator.Shared/OperationFinalStateVia.cs
```diff
@@ -7,8 +7,10 @@ namespace Azure.Core
 {
     internal enum OperationFinalStateVia
     {
+        LocationOverride,
         AzureAsyncOperation,
         Location,
-        OriginalUri
+        OriginalUri,
+        OperationLocation
```
heaths, line 14: Tip: I recommend adding a terminal comma so that future edits are only new lines. Makes for a cleaner diff.

heaths (approved):

PR #2767, opened 2022-10-13
Make Autorest.CSharp a dependency of Emitter.Csharp

pshao25:
Make Autorest.CSharp a dependency of Emitter.Csharp

### DIFF
--- a/eng/pipelines/cadl-ci.yml
+++ b/eng/pipelines/cadl-ci.yml
@@ -81,8 +81,8 @@ stages:
           displayName: 'Set Emitter version'
           workingDirectory: $(Build.SourcesDirectory)/src/CADL.Extension/Emitter.Csharp
         - script: |
-            depedencyVersion = "$(node -p -e "require('./package.json').version")-beta.$(Build.BuildNumber)"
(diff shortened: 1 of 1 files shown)


eng/pipelines/cadl-ci.yml
```diff
@@ -70,15 +70,27 @@ stages:
...
+          workingDirectory: $(Build.SourcesDirectory)
+        - pwsh: |
+            $currentVersion = node -p -e "require('./package.json').version"
+            $devVersion = "$currentVersion-beta.$(Build.BuildNumber)"
+            Write-Host "Setting version to $devVersion"
+            npm version --no-git-tag-version $devVersion | Out-Null;
+          displayName: 'Set Emitter version'
+          workingDirectory: $(Build.SourcesDirectory)/src/CADL.Extension/Emitter.Csharp
+        - script: |
+            dependencyVersion="3.0.0-beta.$(Build.BuildNumber))"
```
chunyu3, line 84: autorest.csharp minus version may not the same as the emitter minus version

chunyu3: And it is better to set the autorest.charp dependency in the package.json with the special version.

pshao25: How could we know the version of autorest.charp

Issue #2336, opened 2022-06-20
Generate a subset of models based on convenience methods specified in a Cadl file


annelo-msft:
This tracks the work to generate input and output models for operations listed in autorest.md.  It tracks creation of the mechanism to specify the operation in the autorest.md file and from it determine the list of models, not to generate those models.  Generation of the models themselves will be tracked in a separate issue.

The input describing the API is Cadl.

annelo-msft:
Pending https://github.com/Azure/cadl-azure/issues/1839

lirenhe:
based on the discussion with @lmazuel, let's use 

@OpenAPI.extension("x-ms-convenient-api", true)
op foo(): {};

As a workaround to unblock this work.  Please note, the JSON/YAML that comes out of the emitter must be designed correctly from the start,

lirenhe:
We shall also filter the models for convinience methods instead of generating a full list of models.

chunyu3:
depend on https://github.com/microsoft/cadl/issues/1068,
https://github.com/microsoft/cadl/issues/1046

chunyu3:
Wait for the cadl compiler and rest new version to include the resolve of issue 1068 and 1046

PR #2768, opened 2022-10-13
Generate models of selected operations

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2336

- upgrade cadl libraries and cadl-azure libraries.
- set None/Input/Output/RoundTrip for each models/enums according to the operations selected.
- Only generate input/output/roundtrip models and their serialize or deserialize function accordingly.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -379,6 +379,7 @@ private static InputType CreateType(Schema schema, string? format, IReadOnlyDict
             Namespace: schema.Extensions?.Namespace,
             Accessibility: schema.Extensions?.Accessibility,
             Description: schema.CreateDescription(),
+            Usage: InputModelTypeUsage.RoundTrip,
```
lirenhe, line 382: For the swagger path, we won't generate models, so will this value be InputModelTypeUsage.None?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -519,3 +525,71 @@ export function getInputType(
...
+            let value = usagesMap.get(name);
+            if (!value) value = UsageFlags.None;
+            if (usages.isUsedAs(type, UsageFlags.Input))
+                value = value | UsageFlags.Input;
+            if (usages.isUsedAs(type, UsageFlags.Output))
+                value = value | UsageFlags.Output;
+            usagesMap.set(name, value);
+        }
+    }
+    /* handle resource operation. */
```
lirenhe, line 572: Could you add some comment on why there is some additional handling here?

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
chunyu3: No. We will support HLC from swagger. We will convert to the right Usage which is calculated by autorest and recorded in schema. And there is a flag in ModelGenerator which filter to generate model from swagger.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: For ResourceOperation, we will not use the finally model as input or response return type, but use the resource type which act as template argument in ResouceOperation definition. And those are not included in models returned by cadl compile, and as a result they are not calculated in resolveUsage, we need to handle those separately.

pshao25:
We could add a namespace level or service level decorator so that we don't need to add the convenience decorator for every single operation.

lirenhe:
> We could add a namespace level or service level decorator so that we don't need to add the convenience decorator for every single operation.

@pshao25, can you discuss this with CADL team? This may not be a priority now, we should think about it aftere 2.0beta2.

eng/Generate.ps1
```diff
@@ -57,7 +58,7 @@ function Add-CadlRanch-Cadl([string]$testName, [string]$projectPrefix, [string]$
     if ($cadlFolders) {
         $cadlFolder = $cadlFolders[0]
         $cadlMain = Join-Path $cadlFolder "main.cadl"
-        Add-Cadl "$projectPrefix$testName" $projectDirectory $cadlMain
+        Add-Cadl "$projectPrefix$testName" $projectDirectory $cadlMain "--option @azure-tools/cadl-csharp.generateConvenienceAPI=true"
```
lirenhe, line 61: I think this is controlled inside cadl file. why do we need this flag?

chunyu3: It's a workaround for caldRanch projects. Current Convenience decorator has not be applied into cadlRanch project, so we set the flag as now, when we add Convenience decorators, we will remove this workaround.

lirenhe: Better to add a 'TODO' here as this is a temp solution.

pshao25:
> > We could add a namespace level or service level decorator so that we don't need to add the convenience decorator for every single operation.
> 
> @pshao25, can you discuss this with CADL team? This may not be a priority now, we should think about it aftere 2.0beta2.

Why we need to discuss with CADL team? We own cadl-dpg.

lirenhe:
> > > We could add a namespace level or service level decorator so that we don't need to add the convenience decorator for every single operation.
> > 
> > 
> > @pshao25, can you discuss this with CADL team? This may not be a priority now, we should think about it aftere 2.0beta2.
> 
> Why we need to discuss with CADL team? We own cadl-dpg.

Better to confirm with them before we add new features.

eng/Generation.psm1
```diff
@@ -50,7 +50,7 @@ function AutoRest-Reset()
     Invoke "$script:autoRestBinary --reset"
 }
 
-function Invoke-Cadl($baseOutput, $projectName, $mainFile, $sharedSource="", $fast="", $debug="")
+function Invoke-Cadl($baseOutput, $projectName, $mainFile, $arguments="", $sharedSource="", $fast="", $debug="")
```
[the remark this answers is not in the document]
chunyu3: @archerzz It may be many parameters which will be used in cadl comple command, something like the command parameter to call autorest. it is not the parameter for this function. So I think we just provide such function parameter '$arguments' for custom to provide options/argument to call an command.

src/AutoRest.CSharp/Common/Input/InputModelType.cs
```diff
@@ -33,11 +33,13 @@ private static IEnumerable<InputModelType> EnumerateBase(InputModelType? model)
         }
     }
 
+    /*
```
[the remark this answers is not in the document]
chunyu3: deleted

src/AutoRest.CSharp/Common/Input/CadlInputModelTypeConverter.cs
```diff
@@ -44,7 +44,7 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
                     || reader.TryReadString(nameof(InputModelType.Namespace), ref ns)
                     || reader.TryReadString(nameof(InputModelType.Accessibility), ref accessibility)
                     || reader.TryReadString(nameof(InputModelType.Description), ref description)
-                    || reader.TryReadWithConverter(nameof(InputModelType.Usage), options, ref usage)
+                    || reader.TryReadString(nameof(InputEnumType.Usage), ref usageString)
```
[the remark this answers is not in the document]
chunyu3: In cadl.json, it will use the string as Usage, not number as InputModelType.Usage, there is no existing convert to convert string to number. And it is simple one, and it does not need to write an json Convert for InputModeType.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -69,15 +69,21 @@ private void CreateEnums(IDictionary<InputEnumType, EnumType> dictionary, TypeFa
         {
             foreach (var inputEnum in _rootNamespace.Enums)
             {
-                dictionary.Add(inputEnum, new EnumType(inputEnum, _defaultNamespace, "public", typeFactory, _sourceInputModel));
+                if (inputEnum.Usage != InputModelTypeUsage.None)
```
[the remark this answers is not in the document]
chunyu3: @archerzz in our design, we will not emit an array for input/output/rountrip models/enum in cadl.json, we just add Usage flag for each model/enum.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -519,3 +525,71 @@ export function getInputType(
...
+    for (const op of ops) {
+        const resourceOperation = getResourceOperation(program, op.operation);
+        if (resourceOperation) {
+            if (!op.parameters.bodyParameter && op.parameters.bodyType) {
+                const resourceName = resourceOperation.resourceType.name;
```
lirenhe, line 577: Will this be changed if we enable discriminator?

chunyu3: No. it is not related with discriminator.

eng/Generation.psm1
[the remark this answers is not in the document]
archerzz: > @archerzz It may be many parameters which will be used in cadl comple command, something like the command parameter to call autorest. it is not the parameter for this function. So I think we just provide such function parameter '$arguments' for custom to provide options/argument to call an command.

I see. Then I suggest to move this to the end.

src/AutoRest.CSharp/Common/Input/CadlInputModelTypeConverter.cs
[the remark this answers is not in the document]
archerzz: > @archerzz In cadl.json, it will use the string as Usage, not number as InputModelType.Usage, there is no existing convert to convert string to number. And it is simple one, and it does not need to write an json Convert for InputModeType.

@chunyu3 Since it's for setting `InputMdelTypeUsge`, I think we should use `nameof(InputModelType.Usage)` for clarity.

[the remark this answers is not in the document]
chunyu3: resolved. Thanks

archerzz (approved):

PR #2769, opened 2022-10-13
fix(dpg): protocol/convenience method parameters should order by default value

### LINKED ISSUE #2766: AutoRest C# generator generating DPG code with compilation errors
We are trying to generate DPG code for the swaggers listed [here](https://github.com/Azure/azure-sdk-for-net/blob/2ed2c1d5aa3576f556216a9da95600fa8b52a71f/sdk/devcenter/Azure.Developer.DevCenter/src/autorest.md) and we are getting [these](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1914087&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=a880e989-7d1a-5c96-a41f-d540b383cc43&l=69) errors for most of the methods:

> Error CS1737: Optional parameters must appear after all required parameters

For more details please see the failures in https://github.com/Azure/azure-sdk-for-net/pull/31636 PR

archerzz:
- fix the sequence of parameters not sorted by default value (e.g. optional parameters should be at the end)
- fix a method access privilege
- regen samples & tests

fix #2766

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -80,7 +80,7 @@ public void BuildNextPageMethod(IReadOnlyDictionary<InputOperation, OperationMet
         public LowLevelClientMethod BuildOperationMethodChain()
         {
             var returnTypeChain = BuildReturnTypes();
-            var protocolMethodParameters = _orderedParameters.Select(p => p.Protocol).WhereNotNull().ToArray();
+            var protocolMethodParameters = _orderedParameters.Select(p => p.Protocol).WhereNotNull().OrderBy(p => p.DefaultValue != null).ToArray();
```
lirenhe, line 83: Can we have one test case for this fix?

archerzz: @lirenhe added test cases

ShivangiReja (approved):
LGTM

lirenhe (approved):

AlexanderSher:
`Sort` doesn't guarantee that order of parameters that are already sorted (that's why the field is called `_orderedParameters`) is preserved.

The order itself is defined here: https://github.com/azure/autorest.csharp/issues/1771#issue-1077102837
Test that covers this scenario is this one: https://github.com/Azure/autorest.csharp/pull/1781/files#diff-2129f5efd7dc8c8b923b5972cdbcf7c6afde0fa26aa04fe7228c9d40f4cdff24

One last comment is about the test project. Every separate test project increases build time and makes it harder to find correct test. If there is existing test project that handles similar cases, please add new case there instead of creating new test.

jsquire:
@m-nash:  Would you please take a look at the feedback from Alex and offer your thoughts?

m-nash:
Let me revive this PR which meant to codify this https://github.com/Azure/azure-sdk/pull/4385.  We have had agreement on this ordering in the past but its very loosely documented in many different places

archerzz:
> The bug happens in `BuildParameters` where all path parameters are added as required.
> 
> Probably we should add another parameter to the `AddUriOrPathParameters`, e.g.:
> 
> ```cs
> private void AddUriOrPathParameters(string uriPart, IReadOnlyDictionary<string, InputParameter> requestParameters, bool addRequiredParameters)
> {
>     foreach ((ReadOnlySpan<char> span, bool isLiteral) in StringExtensions.GetPathParts(uriPart))
>     {
>         if (isLiteral)
>         {
>             continue;
>         }
> 
>         var text = span.ToString();
>         if (requestParameters.TryGetValue(text, out var requestParameter))
>         {
>             var isRequired = requestParameter.IsRequired && requestParameter.DefaultValue == null;
>             if (addRequiredParameters == isRequired)
>             {
>                 AddParameter(text, requestParameter);
>             }
>         }
>         else
>         {
>             ErrorHelpers.ThrowError($"\n\nError while processing request '{uriPart}'\n\n  '{text}' in URI is missing a matching definition in the path parameters collection{ErrorHelpers.UpdateSwaggerOrFile}");
>         }
>     }
> }
> ```
> 
> and then used in `BuildParameters`:
> 
> ```cs
> AddWaitForCompletion();
> AddUriOrPathParameters(Operation.Uri, pathParameters, true);
> AddUriOrPathParameters(Operation.Path, pathParameters, true);
> AddQueryOrHeaderParameters(requiredRequestParameters);
> AddBody(bodyParameter, contentTypeRequestParameter);
> AddUriOrPathParameters(Operation.Uri, pathParameters, false);
> AddUriOrPathParameters(Operation.Path, pathParameters, false);
> AddQueryOrHeaderParameters(optionalRequestParameters);
> AddRequestConditionHeaders(requestConditionHeaders, requestConditionRequestParameter, requestConditionSerializationFormat);
> AddRequestContext();
> ```
> 
> `Sort` uses introspective sort, which is unstable sorting algorithm, so it doesn't guarantee that order of parameters that are already in correct order (that's why the field is called `_orderedParameters`) is preserved.
> 
> The order itself is defined here: [#1765 (comment)](https://github.com/Azure/autorest.csharp/issues/1765#issue-1076126970) Test that covers this scenario is this one: [#1781 (files)](https://github.com/Azure/autorest.csharp/pull/1781/files#diff-2129f5efd7dc8c8b923b5972cdbcf7c6afde0fa26aa04fe7228c9d40f4cdff24)
> 
> One last comment is about the test project. Every separate test project increases build time and makes it harder to find correct test. If there is existing test project that handles similar cases, please add new case there instead of creating new test.

@AlexanderSher Thank you very much! I've composed another PR to resolve this issue, according to your suggestion: https://github.com/Azure/autorest.csharp/pull/2779

PR #2770, opened 2022-10-13
Add tests for property-optional

pshao25:
Add tests for property-optional


test/CadlRanchProjects/property-optional/Generated/BytesClient.cs
```diff
@@ -0,0 +1,519 @@
...
+
+#nullable disable
+
+using System;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace property_optional
```
lirenhe, line 14: same question with this namespace

test/CadlRanchProjects.Tests/property-optional.cs
```diff
@@ -0,0 +1,214 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Threading.Tasks;
+using System.Xml;
+using AutoRest.TestServer.Tests.Infrastructure;
+using Azure;
+using Models.Property.Optional;
+using NUnit.Framework;
+using property_optional;
```
lirenhe, line 11: Is it possible to make the namespace and class name different?

test/CadlRanchProjects.Tests/property-optional.cs
```diff
@@ -0,0 +1,214 @@
...
+
+namespace CadlRanchProjects.Tests
+{
+    public class property_optional : CadlRanchTestBase
```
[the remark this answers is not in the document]
pshao25: This is the class name of test class. I don't think we need to follow naming pattern.

test/CadlRanchProjects.Tests/property-optional.cs
pshao25: Yes, we could change the namespace of generated SDK to Property.Optional

test/CadlRanchProjects.Tests/property-optional.cs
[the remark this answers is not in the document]
lirenhe: Seems this is manual written so it is better to make naming consistent, shall we use PropertyOptional instead

lirenhe:
@pshao25, could you resolve the build failure?

lirenhe (approved):

PR #2773, opened 2022-10-14
First step in merging mode generation between the generator types

m-nash:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -559,5 +562,38 @@ private static string GetConversion(CodeWriter writer, CSharpType from, CSharpTy
             var propertyName = property.PropertyName;
             return writer.Scope($"if ({propertyName} != null)");
         }
+
+        public static IDisposable WriteCommonMethodWithoutValidation(this CodeWriter writer, MethodSignature signature, FormattableString? returnDescription, bool isAsync, bool isPublicType, bool enableAttributes = false, IEnumerable<Attribute>? attributes = default)
```
ArcturusZhang, line 566: I believe we do not need this `enableAttributes` parameter. We could either use null value or empty value of the array `attributes` to state that we do not want to write attributes

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
```diff
@@ -351,13 +343,13 @@ private FormattableString ConstructBinaryDataDescription(string typeSpecificDesc
 </list>
 </para>";
         }
-        private string GetAbstract(SchemaObjectType schema)
+        private string GetAbstract(ObjectType schema)
         {
             // Limit this change to management plane to avoid data plane affected
             return schema.Declaration.IsAbstract && Configuration.AzureArm ? "abstract " : string.Empty;
         }
```
ArcturusZhang, line 350: Could we let this to return a boolean (or potentially move this to ObjectType to return a boolean), and in the corresponding place of the writer, we could use
```
writer.AppendRawIf("abstract ", IsAbstract(obj));
```

src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs
```diff
@@ -7,14 +7,15 @@
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal record FieldDeclaration(FormattableString? Description, FieldModifiers Modifiers, CSharpType Type, CodeWriterDeclaration Declaration, FormattableString? DefaultValue, bool WriteAsProperty = false)
+    internal record FieldDeclaration(FormattableString? Description, FieldModifiers Modifiers, CSharpType Type, CodeWriterDeclaration Declaration, FormattableString? DefaultValue, bool IsRequired, bool IsField = false, bool WriteAsProperty = false)
```
ArcturusZhang, line 10: what does `IsField` stand for? This is a "Field"Declaration, what else could it be? A property? If so I think maybe this is a bit redundant with `WriteAsProperty`

src/AutoRest.CSharp/Common/Output/Models/MethodSignatureBase.cs
```diff
@@ -25,6 +25,8 @@ internal enum MethodSignatureModifiers
         Static = 16,
         Extension = 32,
         Virtual = 64,
-        Async = 128
+        Async = 128,
+        New = 256,
+        Override = 512
```
ArcturusZhang, line 30: My PR of the polymorphic resource will introduce another value of `abstract` here. FYI

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -4,109 +4,347 @@
...
+        }
+
+        protected override ObjectTypeConstructor BuildSerializationConstructor()
+        {
+            ObjectTypeConstructor? baseCtor = GetBaseObjectType()?.SerializationConstructor;
+
+            return new ObjectTypeConstructor(SerializationConstructorSignature, GetPropertyInitializers(Fields.SerializationParameters), baseCtor);
+        }
+
+        private ObjectPropertyInitializer[] GetPropertyInitializers(IReadOnlyList<Parameter> parameters)
```
ArcturusZhang, line 216: Should this be `GetParameterInitializers`?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -4,109 +4,347 @@
...
+                //if (property == Discriminator?.Property)
+                //{
+                //    continue;
+                //}
+
+                ReferenceOrConstant? initializationValue;
+                Constant? defaultInitializationValue = null;
+
+                var propertyType = property.Declaration.Type;
+                if (property.SchemaProperty?.Schema is ConstantSchema constantSchema)
```
ArcturusZhang, line 237: Should we be using the models in the `Input` namespace here? Because I believe `ConstantSchema` is a class from modeler v4's model. Or this is a placeholder for future improvements?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -4,109 +4,347 @@
...
+                    defaultCtorInitializers.Add(new ObjectPropertyInitializer(property, initializationValue.Value, defaultInitializationValue));
+                }
+            }
+            return defaultCtorInitializers.ToArray();
+        }
+
+        protected override CSharpType? CreateInheritedType()
+        {
+            if (_inputModel.BaseModel is not null)
+                return new CSharpType(new ModelTypeProvider(_inputModel.BaseModel!, DefaultNamespace, null, _typeFactory));
```
ArcturusZhang, line 294: Should we get an existing model from the library here instead of creating a new model?
I believe it benefits if we could keep "for a specific model, we always use the same instance" if possible.
I think that this method is not invoked when the models are building, therefore it should be safe to find a model from the full list of models.

src/AutoRest.CSharp/Common/Output/Models/Types/SerializableObjectType.cs
```diff
@@ -0,0 +1,50 @@
...
+        public JsonObjectSerialization? JsonSerialization => HasJsonSerialization ? _jsonSerialization ??= EnsureJsonSerialization() : null;
+
+        private XmlObjectSerialization? _xmlSerialization;
+        public XmlObjectSerialization? XmlSerialization => HasXmlSerialization ? _xmlSerialization ??= EnsureXmlSerialization() : null;
+
+        private bool? _hasJsonSerialization;
+        private bool HasJsonSerialization => _hasJsonSerialization ??= EnsureHasJsonSerialization();
+
+        private bool? _hasXmlSerialization;
+        private bool HasXmlSerialization => _hasXmlSerialization ??= EnsureHasXmlSerialization();
```
ArcturusZhang, line 41: This is a bit weird.
What is the advantage of this pattern vs the following pattern?
```
protected abstract HasJsonSerialization {get;}
```
This is simpler

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -15,16 +19,24 @@ internal abstract class ObjectType : TypeProvider
         private ObjectTypeConstructor? _serializationConstructor;
         private ObjectTypeConstructor? _initializationConstructor;
         private string? _description;
+        private IEnumerable<ModelMethodDefinition>? _methods;
+        private ObjectTypeDiscriminator? _discriminator;
 
         protected ObjectType(BuildContext context) : base(context)
         {
         }
+        protected ObjectType(string defaultNamespace, SourceInputModel? sourceInputModel):base(defaultNamespace, sourceInputModel)
```
ArcturusZhang, line 28: nit: here we need some formatting

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -15,16 +19,24 @@ internal abstract class ObjectType : TypeProvider
...
+        {
+        }
 
+        public bool IsStruct => ExistingType?.IsValueType == true;
```
ArcturusZhang, line 32: I personally prefer this:
```suggestion
        public bool IsStruct => ExistingType?.IsValueType ?? false;
```
which is emphasizing that this defaults to false.

src/AutoRest.CSharp/Common/Output/Models/Types/SerializableObjectType.cs
ArcturusZhang: And if the concern is that we need to ensure that this value will only be calculated once, I believe all the derived class could do this by themselves.

archerzz: This property could be referred multiple times, so even for children we should avoid duplicated calculation.

src/AutoRest.CSharp/Mgmt/Generation/ArmResourceExtensionsWriter.cs
```diff
@@ -107,7 +107,7 @@ protected override void WriteResourceCollectionGetMethod(Resource resource)
                 resourceCollection.Type,
                 $"An object representing collection of {resource.Type.Name.LastWordToPlural()} and their operations over a {resource.Type.Name}.",
                 GetParametersForCollectionEntry(resourceCollection));
-            using (WriteCommonMethod(signature, null, false))
+            using (_writer.WriteCommonMethod(signature, null, false, This.Accessibility == "public"))
```
archerzz, line 110: shall we extract a string enum for `Accessiblity`?

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
m-nash: This is actually shared with mgmt plane that does have references to it.  If we want to refactor that later we can but trying to keep the changes to other libraries at a minimum in this pr

src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs
m-nash: whether or not the destination on the enclosing type is a property or an field.  I don't know if this is actually a valid case but it was one of the examples in the cadl projects so I needed to add support for it.

src/AutoRest.CSharp/Common/Output/Models/MethodSignatureBase.cs
m-nash: was thinking of doing it that way for this comment as well https://github.com/Azure/autorest.csharp/pull/2773#discussion_r995471586

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
m-nash: Based on the type name I would leave it as is.  I also think property initializer is more correct.  We are using the parameters to initialize the properties, not sure what "parameter initialization" would mean.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
m-nash: I will add a TODO here, this code is most likely a copy / paste and is no-op in its current state.

src/AutoRest.CSharp/Common/Output/Models/Types/SerializableObjectType.cs
m-nash: The advantage to the method is it allows for complex logic to be inserted if the answer is not "simple".

src/AutoRest.CSharp/Mgmt/Generation/ArmResourceExtensionsWriter.cs
m-nash: Not sure I follow.  I didn't want to change how mgmt plane determines if a method is "public" today.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
ArcturusZhang: Therefore the general principal is that we should never use Schema types from modeler v4 anymore once we are out of the constructor? We have a lot to refactor in mgmt generator - they coupled so deeply currently

src/AutoRest.CSharp/Mgmt/Generation/ArmResourceExtensionsWriter.cs
ArcturusZhang: I believe Mingzhe means if we could introduce a new enum with Flags for the `Accessibility` property like MethodSignatureModifiers is currently doing, so that we do not have to maintain it in strings.

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
m-nash: I think we can postpone this clean up for future PRs as this isn't introducing this behavior simply re-using it

src/AutoRest.CSharp/Mgmt/Generation/ArmResourceExtensionsWriter.cs
m-nash: I think we can do that, but I would like to leave this refactoring to a future PR as this PR doesn't introduce this and its just the first step, a big clean up PR will follow

archerzz (approved):

PR #2777, opened 2022-10-17
Add parameter skipSDKGeneration

pshao25:


src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -71,11 +71,13 @@ import { exec } from "child_process";
 export interface NetEmitterOptions {
     outputFile: string;
     logFile: string;
+    runCodegen: 'true' | 'false';
```
chunyu3, line 74: Can it be boolean?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -71,11 +71,13 @@ import { exec } from "child_process";
...
 }
 
 const defaultOptions = {
     outputFile: "cadl.json",
-    logFile: "log.json"
+    logFile: "log.json",
+    runCodegen: "true"
```
lirenhe, line 80: Can we find a better name to describe it, 'GenerateSDK'?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
pshao25: No. Compiler takes it as string even you declare it as boolean.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
ArcturusZhang: I personally think that a boolean option with a `true` as default is not very intuitive. Could we flip this switch to something like `skipCodegen` or `skipSDKGeneration`?

lirenhe (approved):

PR #2782, opened 2022-10-18
Reduce string allocations in RawRequestUriBuilder

AlexanderSher:
`RawRequestUriBuilder` creates enormous amount of unnecessary small strings that often get promoted to gen 1. This is a first set of fixes. Further improvements will be possible after https://github.com/Azure/azure-sdk-for-net/pull/31859


AlexanderSher:
Before/after results in conjunction with https://github.com/Azure/azure-sdk-for-net/pull/31859
image

christothes (approved):

PR #2785, opened 2022-10-19
Add IsErrorResponse parameter to response

### LINKED ISSUE #2744: Design of error response in convenience method
Cadl:
```
@get
@route("/invalid")
op invalid(): NoContentResponse | InvalidAuth;

@error
model InvalidAuth {
  @statusCode _: 403;
  error: string;
}
```
This is an operation with no-body successful response and with-model error response. See swagger [here](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAY2FkbC1sYW5nL3Jlc3QiOw0KDQp1c2luZyBDYWRsLkh0dHA7DQpAc2VydmljZVZlcnNpb24NCkBkb2MoIklsbHVzdHJhdGVzIGNsaWVudHMgZ2VuZcQSZCB3aXRoIEFwaUtleSBhdXRoZW50aWNhdGlvbi4iKQ0KQHVzZUF1dGgoxiPECzzGC0xvxydoZWFkZXIsICJ4LW1zLWFwaS1rZXkiPikNCm5hbWVzcGFjZSDEOMtaxkLlAMfmAKRDaGVjayB3aGV0aGVy5wCmIGlz7ACSZWTmAJFnZXQNCkByb3V0ZSgiL2ludmFsaWTEGm9wIMcOKCk6IE5vQ29udGVudFJlc3BvbnNlIHwgScYf5ACNxnxlcnJvcg0KbW9kZWzMHiB7DQogIEBzdGF0dXNDb2RlIF86IDQwMzvEF8U1OiBzdHJpbmc7DQp9DQo%3D). Current behavior [here](https://github.com/Azure/autorest.csharp/blob/bdc076cfe6dcfb1003416c9926c08039a5a05005/test/CadlRanchProjects/api-key/Generated/ApiKeyClient.cs).

There are three issues:
1. Generated convenience method should not return model
```C#
public virtual Response<InvalidAuth> InvalidValue(CancellationToken cancellationToken = default)
{
    using var scope = ClientDiagnostics.CreateScope("ApiKeyClient.InvalidValue");
    scope.Start();
    try
    {
        Response response = Invalid(context);
        return Response.FromValue(InvalidAuth.FromResponse(response), response);
    }
    catch (Exception e)
    {
        scope.Failed(e);
        throw;
    }
}
```
2. ResponseClassifier should not contain 403
```C#
private static ResponseClassifier _responseClassifier204403;
private static ResponseClassifier ResponseClassifier204403 => _responseClassifier204403 ??= new StatusCodeClassifier(stackalloc ushort[] { 204, 403 });
```
3. When there is error happening, we should throw a model type error `InvalidAuth` instead of an internal error.

For 1&2, what I think it add an isError flag to OperationResponse which is the output of emitter.

pshao25:
Fix #2744


src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -116,12 +116,15 @@ private bool IsParameterTypeSame(Parameter? first, Parameter? second)
...
+            if (operationBodyTypes != null && operationBodyTypes.Length != 0)
             {
-                0 => null,
-                1 => _typeFactory.CreateType(operationBodyTypes[0]),
-                _ => new CSharpType(typeof(object))
+                var firstBodyType = operationBodyTypes[0];
+                if (firstBodyType != null)
+                {
+                    responseType = _typeFactory.CreateType(firstBodyType);
+                }
```
ArcturusZhang, line 127: This is changing the previous logic.
In previous implementation, if there are multiple body types, we will use `typeof(object)` as the return type.
Now we will always use the first as return type.
Please double confirm this is expected. At least I believe when there are multiple in the result, and we are taking the first, we should print a warning somewhere

pshao25: It is expected. We cannot handle more than one return types.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -116,12 +116,15 @@ private bool IsParameterTypeSame(Parameter? first, Parameter? second)
...
-            CSharpType? responseType = operationBodyTypes.Length switch
+            var operationBodyTypes = Operation.Responses.Where(r => !r.IsErrorResponse).Select(r => r.BodyType).Distinct().ToArray();
+            CSharpType? responseType = null;
+            if (operationBodyTypes != null && operationBodyTypes.Length != 0)
             {
-                0 => null,
-                1 => _typeFactory.CreateType(operationBodyTypes[0]),
-                _ => new CSharpType(typeof(object))
+                var firstBodyType = operationBodyTypes[0];
+                if (firstBodyType != null)
```
chunyu3, line 124: what if it is null? No return body? shall we need to retrieve next one?

pshao25: Then responseType keeps null, which is expected.

chunyu3 (approved):

PR #2788, opened 2022-10-19
Fix bug in RawRequestUriBuilder

AlexanderSher:
Also add net461 target to `AutoRest.Shared.Tests.csproj` to cover assemblies built for netstandard2.0

### DIFF
--- a/test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj
+++ b/test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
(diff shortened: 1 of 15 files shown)


test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>net461;netcoreapp3.1</TargetFrameworks>
```
m-nash, line 4: lets add net6 as well to match azure-sdk-for-net since there are many shared files between the two repos.

m-nash (approved):

test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj
AlexanderSher: Done

PR #2807, opened 2022-10-25
dpg: Replace azure core error types

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/2450

# Description

This is a POC of replacing the models that match the definition in `Azure.Core` library. Currently the mapping defining which type to be replaced is hard-coded in the code. We will dynamically build that dictionary in the future if necessary.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -140,21 +141,24 @@ public override string ToString()
             return new CodeWriter().Append($"{this}").ToString(false);
         }
 
-        internal static CSharpType FromSystemType(BuildContext context, Type type)
+        internal static CSharpType FromSystemType(Type type, string defaultNamesapce, SourceInputModel? sourceInputModel)
```
lirenhe, line 144: typo,  defaultNamespace

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
```diff
@@ -13,17 +13,17 @@ namespace AutoRest.CSharp.Output.Models.Types
...
     {
         private readonly IReadOnlyDictionary<InputEnumType, EnumType> _enums;
-        private readonly IReadOnlyDictionary<InputModelType, ModelTypeProvider> _models;
+        private readonly IReadOnlyDictionary<InputModelType, ObjectType> _models;
         private readonly bool _isCadlInput;
 
         public TypeFactory TypeFactory { get; }
         public IEnumerable<EnumType> Enums => _enums.Values;
-        public IEnumerable<ModelTypeProvider> Models => _models.Values;
+        public IEnumerable<ObjectType> Models => _models.Values;
```
lirenhe, line 21: why is this change?

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -28,12 +28,19 @@ internal class DpgOutputLibraryBuilder
...
         public DpgOutputLibraryBuilder(InputNamespace rootNamespace, SourceInputModel? sourceInputModel)
         {
             _rootNamespace = rootNamespace;
             _sourceInputModel = sourceInputModel;
             _defaultNamespace = Configuration.Namespace ?? rootNamespace.Name;
             _libraryName = Configuration.LibraryName ?? rootNamespace.Name;
+            _azureCoreTypeMapping = new()
+            {
+                [("Azure.Core.Foundations", "Error")] = typeof(Azure.ResponseError),
+                [("Azure.Core.Foundations", "InnerError")] = typeof(Azure.ResponseError).Assembly.GetType("Azure.ResponseInnerError", true)!
```
lirenhe, line 42: Can we have a config file for this, so if we found new types, we don't need to release a new codegen.
And can we use typeof(Azure.ResponseInnerError)) directly?

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
ArcturusZhang: Here is the inheritance relationship among them: 
```
SystemObjectType : ObjectType
ModelTypeProvider : SerializableObjectType
SerializableObjectType : ObjectType
```
And we are not actually using the elements in this array as `ModelTypeProvider` anywhere (and we should not)

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
ArcturusZhang: We cannot use `typeof(Azure.ResponseInnerError))` because it is an internal class, I have to get it in this way.

In the future this map should be automatically built from the dependencies (maybe we could read it from the attributes on classes defined in `Azure.Core`.

ArcturusZhang: We no longer  need this therefore I already removed this part.

ArcturusZhang:
The replaced types (`Error` and `InnerError` in this case) will still be generated, but they are not used anywhere.
When we introduce the internalizer and remover to DPG, those unused classes will be automatically removed. This part will not be covered by this PR

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
ArcturusZhang: This change is not required by this PR right now. But I believe we should do this level's abstraction, therefore I still preserved this change on this PR

chunyu3: What actually Type will the Models list item be?  ModelTypeProvider only or it can be ObjectType. I am afraid it will affect the Model Generation. In Model Generation, it will check the type of the model, and then choose the method to generate.

test/TestProjects/Authoring-Cadl/Generated/InnerError.Serialization.cs
```diff
@@ -0,0 +1,49 @@
...
+
+#nullable disable
+
+using System.Text.Json;
+using Azure;
+using Azure.Core;
+
+namespace Azure.Core.Foundations
+{
+    public partial class InnerError
```
lirenhe, line 14: Can we git rid of this as well?

test/TestProjects/Authoring-Cadl/Generated/Error.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Azure;
+using Azure.Core;
+
+namespace Azure.Core.Foundations
+{
+    /// <summary> The error object. </summary>
+    public partial class Error
```
lirenhe, line 17: I expect this would be removed so why it is still shown in the test?

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -29,11 +29,15 @@ public TypeFactory(OutputLibrary library)
...
 
+        private Type AzureResponseErrorType => typeof(Azure.ResponseError);
+
         public CSharpType CreateType(InputType inputType) => inputType switch
         {
             InputListType listType             => new CSharpType(typeof(IList<>), listType.IsNullable, CreateType(listType.ElementType)),
             InputDictionaryType dictionaryType => new CSharpType(typeof(IDictionary<,>), inputType.IsNullable, typeof(string), CreateType(dictionaryType.ValueType)),
             InputEnumType enumType             => _library.ResolveEnum(enumType).WithNullable(inputType.IsNullable),
+            // TODO -- this is a temporary solution until we refactored the type replacement to use input types instead of code model schemas
+            InputModelType { Namespace: "Azure.Core.Foundations", Name: "Error" } => SystemObjectType.Create(AzureResponseErrorType, AzureResponseErrorType.Namespace!, null).Type,
```
lirenhe, line 40: shall we also add innerError their?

test/TestProjects/Authoring-Cadl/Generated/InnerError.Serialization.cs
ArcturusZhang: I explained this here: https://github.com/Azure/autorest.csharp/pull/2807#issuecomment-1306702826

test/TestProjects/Authoring-Cadl/Generated/Error.cs
ArcturusZhang: Same as above

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
ArcturusZhang: No. `Azure.ResponseInnerError` in `Azure.Core` is internal, and it does not have a `JsonConverter` or a public constructor or a deserialize method to deserialize. Therefore even if we have it here, the generator cannot know how to deserialize it.

And the type is internal, there should not be any property using this type, the only exception is the `innerError` property in the `Error` model. Therefore we only need to replace the `Error` with `Azure.ResponseError`, then the `InnerError` classes are already replaced.

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
ArcturusZhang: > 

That is how polymorphism in OO works. Everything is calculated when it is constructed no matter what type of variable the object is storing into.

ArcturusZhang: This change will never introduce any break - because this is what polymorphism is for.
But I change my mind and already reverted all these changes. I could always introduce it back when it is necessary

test/TestProjects/Authoring-Cadl/Generated/InnerError.Serialization.cs
ArcturusZhang: BTW I will introduce internalizer and remover to DPG in this [PR](https://github.com/Azure/autorest.csharp/pull/2842)

lirenhe (approved):

chunyu3 (approved):

PR #2816, opened 2022-10-27
fix some bugs found in polymorphic resources and introduce some refactor

ArcturusZhang:
# Description

This PR is applying some fixes and refactors introduced in the [polymorphic resource PR](https://github.com/Azure/autorest.csharp/pull/2739)

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -399,15 +393,7 @@ private MgmtExtensions EnsureExtensions(Type armCoreType, RequestPath contextual
...
-        private ArmClientExtensions? _armClientExtensions;
-        private ArmClientExtensions EnsureArmClientExtensions()
-        {
-            if (_armClientExtensions != null)
-                return _armClientExtensions;
-
-            _armClientExtensions = new ArmClientExtensions(GetChildOperations(RequestPath.Tenant));
-            return _armClientExtensions;
-        }
+        private ArmClientExtensions EnsureArmClientExtensions() => new ArmClientExtensions(GetChildOperations(RequestPath.Tenant));
```
ArthurMa1978, line 396: The previous code is more like a singleton, but new code is not, is this the expected change?

ArthurMa1978 (approved):

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
ArcturusZhang: All these instances work like a singleton because they only initialize once. Previously we are mixing two patterns:
First:
```csharp
private Something _property;
public Something Property => EnsureProperty();

private Something EnsureProperty(){
if (_property != null)
return _property;

_property = new Something();
return _property;
}
```
Second:
```csharp
private Something _property;
public Something Property => _property ??= EnsureProperty();

private Something EnsureProperty(){
return new Something();
}
```
These two are equivalent therefore I am unifying our code to use the simpler one (the second)

Issue #2823, opened 2022-10-31
Enabling polymorphic pattern for models in data plane


joseharriaga:
Using the Text Analytics client library as an example, REST API version 2022-10-01-preview introduced a new, read-only, polymorphic type. The base class is called `BaseResolution`, and it has multiple derived classes, such as `AgeResolution`, `DateTimeResolution`, `LengthResolution`, etc.  Today, the generated model has a few issues:
- `BaseResolution` should be generated as an _abstract_ class.
- `BaseResolution` should list its derived classes as part of its summary.
- The public properties of the derived classes should only have getters and no setters.

This pattern seems to work correctly for the management plane. For example:
- [AlertProcessingRuleAction.cs (base class)](https://github.com/Azure/azure-sdk-for-net/blob/84ec01fbfe07cefd19b0a1501a98d80dbe53b8bb/sdk/alertsmanagement/Azure.ResourceManager.AlertsManagement/src/Generated/Models/AlertProcessingRuleAction.cs)
- [AlertProcessingRuleAddGroupsAction.cs (derived class)](https://github.com/Azure/azure-sdk-for-net/blob/84ec01fbfe07cefd19b0a1501a98d80dbe53b8bb/sdk/alertsmanagement/Azure.ResourceManager.AlertsManagement/src/Generated/Models/AlertProcessingRuleAddGroupsAction.cs)

Other references:
- [autorest.md for Text Analytics](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/autorest.md)
- [swagger spec for Text Analytics](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/Language/preview/2022-10-01-preview/analyzetext.json)

For now, the workaround is to hand-write these models.

PR #2837, opened 2022-11-07
Remove some of the low level writers and add discriminator support

m-nash:
# Description
This normalized polymorphic models across generators in order to share common logic.  It resolves this issue here https://github.com/Azure/autorest.csharp/issues/2823

Regen PR in azure-sdk-for-net https://github.com/Azure/azure-sdk-for-net/pull/32335

# Checklist

# Ready to Land?


test/TestProjects/Models-Cadl/Generated/BaseModelWithDiscriminator.cs
```diff
@@ -5,25 +5,23 @@
...
+    /// The available derived classes include <see cref="DerivedModelWithDiscriminatorA"/> and <see cref="DerivedModelWithDiscriminatorB"/>.
+    /// </summary>
+    public abstract partial class BaseModelWithDiscriminator
     {
         /// <summary> Initializes a new instance of BaseModelWithDiscriminator. </summary>
-        /// <param name="discriminatorProperty"></param>
-        /// <exception cref="ArgumentNullException"> <paramref name="discriminatorProperty"/> is null. </exception>
-        public BaseModelWithDiscriminator(string discriminatorProperty)
+        /// <param name="discriminatorProperty"> Discriminator. </param>
+        protected BaseModelWithDiscriminator(string discriminatorProperty)
```
chunyu3, line 19: In HLC, base model will have an public constructor which has no discriminator as parameter. And in Mgmt, the constructor will be protected, and has no discriminator property as parameter also.

So will we change the base model constructor?

m-nash: The class is abstract so a public ctor doesn't make sense.  The only time we will keep a public ctor is IF the class was already GAd as non abstract we won't change it since that is a breaking change.  All new poly sets should work this way.  There is a way to override the existing poly sets so that autorest doesn't convert them, but it was easier to just leave hlc as non abstract for now until I get the "last code analyzer" finished.

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -238,8 +238,8 @@ private InputModelType GetOrCreateModel(ObjectSchema schema, SchemaUsageProvider
                     ? GetOrCreateModel(parent, schemaUsages)
                     : null,
                 DerivedModels: derived,
-                DiscriminatorValue: schema.DiscriminatorValue
-            );
+                DiscriminatorValue: schema.DiscriminatorValue,
```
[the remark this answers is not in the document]
m-nash: not sure I follow the question

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -631,5 +645,21 @@ protected override XmlObjectSerialization EnsureXmlSerialization()
...
+
+            _hasCalculatedDefaultDerivedType = true;
+            if (_context.BaseLibrary is null)
+                return null;
+
+            var defaultDerivedSchema = ObjectSchema.GetDefaultDerivedSchema();
+            if (defaultDerivedSchema is null)
+                return null;
+
+            return _context.BaseLibrary.FindTypeProviderForSchema(defaultDerivedSchema) as ObjectType;
```
ArcturusZhang, line 662: Looks like this is the only place to use this as a TypeProvider. We actually do not need to introduce this new abstract method, we could use the [`TryCast<ObjectType>` method](https://github.com/Azure/autorest.csharp/blob/943b721046b5a5cf94ee80e2c8e3bc3b2176ae9e/src/AutoRest.CSharp/Mgmt/Decorator/TypeExtensions.cs#L65) on the CSharpType returned by `FindTypeForSchema`.
We maybe we should throw when the type is not a `ObjectType` (well in the current implementation it is always a ObjectType)

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -138,6 +138,7 @@ internal static string SaveConfiguration()
                     WriteIfNotDefault(writer, Configuration.Options.SingleTopLevelClient, Configuration.SingleTopLevelClient);
                     WriteIfNotDefault(writer, Configuration.Options.ProjectFolder, Configuration.RelativeProjectFolder);
                     Utf8JsonWriterExtensions.WriteNonEmptyArray(writer, nameof(Configuration.ProtocolMethodList), Configuration.ProtocolMethodList);
+                    Utf8JsonWriterExtensions.WriteNonEmptyArray(writer, nameof(Configuration.SuppressAbstractBaseClass), Configuration.SuppressAbstractBaseClass);
```
lirenhe, line 141: This seems to be an array of classes. Can we rename it to SuppressAbstractBaseClassList?

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
m-nash: I actually thing I would want to deprecate the CSharpType version of this abstract method since TypeProvider is a superset of the info (even though CSharpType does have a two way reference).

For now I will leave this as there are other areas of consolidation that will be coming in the future.

JoshLove-msft (approved):

PR #2838, opened 2022-11-07
Fix #2672: Support for operation-location/resourceLocation polling

### LINKED ISSUE #2672: Support for operation-location/resourceLocation polling

AlexanderSher:
Change in Generator.Shared type, so no updates in generated code.


src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -241,6 +240,30 @@ internal static bool TryGetApiVersion(Uri startRequestUri, out ReadOnlySpan<char
                     return _startRequestUri.AbsoluteUri;
             }
 
+            // If body contains resourceLocation, use it: https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#target-resource-location
+            var contentStream = response.ContentStream;
+            if (contentStream is { CanSeek: true, Length: > 0 })
+            {
+                try
+                {
+                    using JsonDocument document = JsonDocument.Parse(contentStream);
```
m-nash, line 249: We are parsing in IsFinalState and in this method now.  Can we avoid the double parse?

AlexanderSher: Fixed

src/assets/Generator.Shared/NextLinkOperationImplementation.cs
```diff
@@ -288,17 +289,19 @@ private HttpMessage CreateRequest(string uri)
...
+
             if (headerSource == HeaderSource.Location)
             {
                 return response.Status != 202;
             }
 
             if (response.Status is >= 200 and <= 204)
             {
-                if (response.ContentStream?.Length > 0)
+                if (response.ContentStream is {CanSeek: true, Length: > 0})
```
archerzz, line 304: what if `CanSeek: false`? If `204` is returned, will the `ContentStrea.CanSeek == false`?

AlexanderSher: If `CanSeek` is false, we won't be able to call `response.ContentStream.Position = 0;` 
It's a good question if we should throw here if `CanSeek` is false or do something else.

m-nash: Is there ever a scenario where we don't own constructing this?  It seems we should never hit this case and maybe adding the check only adds confusion?

In the past the seek would have thrown an exception it might be safest to simply let that continue to happen instead of changing the behavior?

AlexanderSher: Resolved

m-nash (approved):

Issue #2841, opened 2022-11-08
[DPG] Introduce the `Internalizer` and `Remover` from mgmt plane to DPG to minimize the public API


ArcturusZhang:
The internalizer will make the models not using by any public API `internal`, which hides the class from the public API.
The remover will remove the unused models.

PR #2842, opened 2022-11-08
Combine `internalizer` and `remover` as `postProcessor` and introduce it to DPG

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/2841

# Description

This PR refactors the two components in mgmt plane, `internalizer` and `remover` to a new component `postProcessor` and introduces this new component to all the targets (especially DPG). This change might introduce breaking changes therefore this is not enabled for HLC.

1. The `PostProcessor` has a virtual method `IsRootDocument(Document)` allowing each target to define the logic which document is the "root document" which is the starting point of the accessibility traversal. The default behaviour will identify customized document and the client documents as root documents.
2. The `PostProcessor` has two different procedure - `Internalize` and `Remove`. By following a process in the post processor, it will have a list of type definitions, and `Internalize` procedure will mark their accessibility to `internal`, and the `Remove` procedure will remove all those definitions.

How does this `postProcessor` work?
1. The `PostProcessor` will build a reference map according to which procedure it is invoking.
The `Internalize` procedure will build a "public reference map" in which only public references on the signature will be included.
The `Remove` procedure will build a "all reference map" in which all references will be included (invocations inside method bodies for instance will also be included).
2. The `PostProcessor` will make a BFS starting from the types defined in the "root documents". All the types that are not visited during the BFS are the candidates to internalize or to remove.
3. The `PostProcessor` will internalize or remove all the types it finds in step 2.

To control the behaviour of post processor, this PR adds a new configuration `unreferenced-types-handling` which takes an enumeration: `removeOrInternalize` (default), `internalize` or `keepAll`.
In the default behaviour, the post processor will call Internalize first which will change all the types that are not publicly used to `internal`, then it call remove which will remove all the types that are not referenced.
If the `runreferenced-types-handling` is set to `internalize`, the post processor will only internalize types without removing them.
If the `unreferenced-types-handling` is set to `keepAll`, the post processor will do nothing.

Unfortunately HLC now has some issues therefore this new component is not enabled for HLC yet. We will cover that soon

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,259 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Threading.Tasks;
+using AutoRest.CSharp.AutoRest.Plugins;
+using AutoRest.CSharp.Mgmt.Output;
```
lirenhe, line 10: I expect this is a common part, so why it has the dependency on mgmt output?

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    public async Task<Project> InternalizeAsync()
+    {
+        var compilation = await _project.GetCompilationAsync();
+        if (compilation == null)
+            return _project;
+
+        // first get all the declared symbols
+        var definitions = await GetModelSymbolsAsync(compilation, true);
```
lirenhe, line 70: Is it better to use the real type instead of var here?

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,259 @@
...
+    protected Project _project;
+
+    public PostProcessor(Project project)
+    {
+        _project = project;
+    }
+
+    protected record ModelSymbols(HashSet<INamedTypeSymbol> DeclaredSymbols, Dictionary<INamedTypeSymbol, HashSet<BaseTypeDeclarationSyntax>> DeclaredNodesCache, Dictionary<Document, HashSet<INamedTypeSymbol>> DocumentCache);
+
+    protected virtual async Task<ModelSymbols> GetModelSymbolsAsync(Compilation compilation, bool publicOnly = true)
```
lirenhe, line 31: what is the meaning of 'publicOnly', could you add some description?

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            nodesToInternalize.AddRange(definitions.DeclaredNodesCache[symbol]);
+        }
+
+        foreach (var model in nodesToInternalize)
+        {
+            MarkInternal(model);
+        }
+
+        return _project;
```
lirenhe, line 91: As this class holds a ref of project , do we need to return project in InternalizeAsync() and RemoveAsync()?
It would make the calling place of those 2 APIs a bit strange.

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,259 @@
...
+using AutoRest.CSharp.Output.Models.Types;
+using AutoRest.CSharp.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Simplification;
+
+namespace AutoRest.CSharp.Common.Output.PostProcessing;
+
+internal abstract class PostProcessor
```
lirenhe, line 20: I think the public methods of PostProcessor can be tested in unit tests.
Could you add some tests to verify the functionality?

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
ArcturusZhang: Good catch. I will do a clean

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
ArcturusZhang: We are using `var` all over the place. I prefer not changing this.

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
ArcturusZhang: it is quite a common pattern that we return project whenever it gets changed because the project is immutable, we have to manually maintain its reference.
It is OK to make these two methods no longer return project and expose another property for `project`, but I do not like it because these two methods are changing it, and we need it to be reflected on the project instance right away.
For instance, if you call Internalize and you would like to change the project by yourself and then call remove, you will have to use `postProcessor.Project` first to get the reference. On the other hand, if you are not changing the project in the middle, just ignore the return value of it, it harms nothing.

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
ArcturusZhang: It is testable, but first I would like to focus on the feature

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,291 @@
...
+{
+    public PostProcessor()
+    {
+    }
+
+    protected record TypeSymbols(HashSet<INamedTypeSymbol> DeclaredSymbols, Dictionary<INamedTypeSymbol, HashSet<BaseTypeDeclarationSyntax>> DeclaredNodesCache, Dictionary<Document, HashSet<INamedTypeSymbol>> DocumentsCache);
```
AlexanderSher, line 25: Please make all the properties here of immutable types or read-only types

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,291 @@
...
+        {
+            if (ShouldIncludeDocument(document))
+            {
+                var root = await document.GetSyntaxRootAsync();
+                if (root == null)
+                    continue;
+
+                var semanticModel = compilation.GetSemanticModel(root.SyntaxTree);
+
+                foreach (var typeDeclaration in root.DescendantNodes().OfType<BaseTypeDeclarationSyntax>())
```
AlexanderSher, line 50: Do we need types defined within other types?

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,291 @@
...
+                {
+                    var symbol = semanticModel.GetDeclaredSymbol(typeDeclaration);
+                    if (symbol == null)
+                        continue;
+                    if (publicOnly && symbol.DeclaredAccessibility != Accessibility.Public)
+                        continue;
+                    result.Add(symbol);
+                    declarationCache.AddInList(symbol, typeDeclaration);
+                    documentCache.AddInList(document, symbol, () => new HashSet<INamedTypeSymbol>(SymbolEqualityComparer.Default));
```
AlexanderSher, line 59: Correct me if I'm wrong:
- `result` contains exactly the same symbols as `declarationCache.Keys`
- `documentCache` may contain only one type symbol per document, otherwise it is ignored in `ProcessSymbolAsync`
- It is assumed that we can't have more than one top level type in the file

src/AutoRest.CSharp/Common/Output/PostProcessing/ReferenceMapBuilder.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// The <paramref name="node"/> here should come from the result of <see cref="SymbolFinder"/>, therefore a result is guaranteed
+        /// </summary>
+        /// <param name="root"></param>
+        /// <param name="node"></param>
+        /// <returns></returns>
+        private static BaseTypeDeclarationSyntax GetOwner(SyntaxNode root, SyntaxNode node)
+        {
+            var candidates = root.DescendantNodes().OfType<BaseTypeDeclarationSyntax>();
+            var result = candidates.First(candidate => candidate.DescendantNodesAndSelf().Contains(node));
+            return result;
```
AlexanderSher, line 214: It should be much faster to go from the node up to the root.

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
lirenhe: Please consider this work in the MQ

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,291 @@
...
+
+    protected record TypeSymbols(HashSet<INamedTypeSymbol> DeclaredSymbols, Dictionary<INamedTypeSymbol, HashSet<BaseTypeDeclarationSyntax>> DeclaredNodesCache, Dictionary<Document, HashSet<INamedTypeSymbol>> DocumentsCache);
+
+    /// <summary>
+    /// This method reads the project, returns the types defined in it and build symbol caches to acceralate the calculation
+    /// By default, the types defined in shared documents are not included. Please override <see cref="ShouldIncludeDocument(Document)"/> to tweak this behaviour
+    /// </summary>
+    /// <param name="compilation">The <see cref="Compilation"/> of the <paramref name="project"/> </param>
+    /// <param name="project">The project to extract type symbols from</param>
+    /// <param name="publicOnly">If <paramref name="publicOnly"/> is true, only public types will be included. If <paramref name="project"/> is false, all types will be included </param>
```
lirenhe, line 33: If <paramref name="publicOnly"/> is false

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
ArcturusZhang: 1. Correct
2. No. In `ProcessSymbolAsync` when we are finding references of one symbol A, we can only get the text span and document of where it is used. But in the map we need the type symbol that is referencing symbol A. I did not find a handy API in roslyn to do that. Therefore considering the reference information contains the "document", I keep a cache of "what symbols a document could hold". When we want to get a symbol from the document, and the document only contains one, we just use that symbol without any calculation. If the document contains multiple, we still have to determine which symbol in the document is referencing symbol A (this case will be very rare in our generated code)
3. No, we did not make that assumption and it is also the reason why I iterate over all the type declaration nodes under one root

src/AutoRest.CSharp/Common/Output/PostProcessing/ReferenceMapBuilder.cs
ArcturusZhang: I did not get the point. And per my explanation in the above comment about the "documentCache", this piece of code will be reached very rarely.

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -136,6 +136,7 @@ internal static string SaveConfiguration()
                     WriteIfNotDefault(writer, Configuration.Options.SkipCSProjPackageReference, Configuration.SkipCSProjPackageReference);
                     WriteIfNotDefault(writer, Configuration.Options.Generation1ConvenienceClient, Configuration.Generation1ConvenienceClient);
                     WriteIfNotDefault(writer, Configuration.Options.SingleTopLevelClient, Configuration.SingleTopLevelClient);
+                    WriteIfNotDefault(writer, Configuration.Options.KeepUnusedModels, Configuration.KeepUnusedModels);
```
m-nash, line 139: Can we have this config support 3 values?

1. Don't change accessor
2. Make accessor internal
3. Delete file from library

Default being 3.  There are some debug use cases where we might want to keep the files internal and not remove them.  By default we do want to remove them to optimize package sizes.

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
```diff
@@ -41,6 +44,21 @@ public static void Execute(GeneratedCodeWorkspace project, InputNamespace inputN
...
+            await project.PostProcess(PostProcess);
+        }
+
+        private static async Task<Project> PostProcess(Project project)
+        {
+            if (Configuration.KeepUnusedModels)
+                return project;
+
+            var postProcessor = new DpgPostProcessor();
+            project = await postProcessor.InternalizeAsync(project);
```
m-nash, line 57: I assume this two step process here works like the following:

If we have 3 types that get internalized
Foo.cs
Bar.cs
BangListResult.cs

step one will mark all 3 as internal

step two will delete Foo and Bar, but since BangListResult has internal references it will be left behind?

ArcturusZhang: Yes. This is how it works.
Now that since I have combined maybe I could let the remover to go first and by default the internalizer will have less classes to look at which might slightly accelerate the process. What do you think?

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
ArcturusZhang: Good idea. Therefore we are making something like the level we would like to keep the models?
I would like to reverse this order to make the default value to be `0`, like:
0 - Delete unused files from library (default)
1 - Make accessor internal
2 - Do nothing
What do you think? Or should we use strings instead of integer on this?

ArcturusZhang: Well, I refactored this configuration and made it an enumeration of three values:
1. RemoveAll (default, which will remove all unused models)
2. Internalize (only make unused models internal, but keep them)
3. KeepAll (do nothing)

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -34,6 +34,14 @@ public static class Options
             public const string SkipSerializationFormatXml = "skip-serialization-format-xml";
             public const string DisablePaginationTopRenaming = "disable-pagination-top-renaming";
             public const string SuppressAbstractBaseClasses = "suppress-abstract-base-class";
+            public const string RemoveUnusedTypes = "remove-unused-types";
+        }
+
+        public enum UnusedTypeRemovalLevel
+        {
+            RemoveAll = 0,
```
lirenhe, line 42: Please make the name to be consistent with the configuration

ArcturusZhang: Renamed accordingly

archerzz:
We could rename the configuration property from `remove-unused-types` to `process-unused-types` with 3 options:
- `none`: do nothing, e.g. keep them
- `internalizeOnly`
- `internalizeAndRemove`

What do you think?

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -83,9 +83,18 @@ await foreach (var file in workspace.GetGeneratedFilesAsync())
             }
         }
 
-        private static void WriteIfNotDefault(Utf8JsonWriter writer, string option, bool value)
+        private static void WriteIfEnumNotDefault<T>(Utf8JsonWriter writer, string option, T enumValue) where T : struct, Enum
```
archerzz, line 86: Since template is used, we don't need `Enum` in the name, e.g. original name is fine. And for bool value, we can use `WriteIfNotDefault(Utf8JsonWriter writer, string option, bool value)` as an overload.

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,294 @@
...
+        var compilation = await project.GetCompilationAsync();
+        if (compilation == null)
+            return project;
+
+        // find all the declarations, including non-public declared
+        var definitions = await GetTypeSymbolsAsync(compilation, project, false);
+        // build reference map
+        var referenceMap = await new ReferenceMapBuilder(compilation, project, HasDiscriminator).BuildAllReferenceMapAsync(definitions.DeclaredSymbols, definitions.DocumentsCache);
+        // get root nodes
+        var rootNodes = GetRootSymbols(project, definitions);
```
archerzz, line 134: To be consistent with the codes above.
```suggestion
        var rootSymbols = GetRootSymbols(project, definitions);
```

src/AutoRest.CSharp/Common/Output/PostProcessing/ReferenceMapBuilder.cs
ArcturusZhang: I get it now. Now we find its owner from the node up to the root and we actually do not need the root here.

PR #2847, opened 2022-11-09
Add check enforcer

benbp:
Adding check enforcer v2 action to repository. See https://github.com/Azure/azure-sdk-tools/pull/3287 for details.


weshaggard (changes requested):
Lets avoid check-enforcer here and instead add the branch protection rule for the pipelines in the repo.

Issue #2832, opened 2022-11-03
We should not generate enums by default if there is no convenience API


lirenhe:
From https://apiview.dev/Assemblies/Review/5b7d805ce109492094e0f8dc193b11f4?revisionId=035edfa2ebf24e81bbcf5bc3a14fa226&doc=False,  even there is no convenience API, we still see a couple of Enum generated.

There enum are not used so we should not generate code from them.

PR #2849, opened 2022-11-10
correct enum usage and decide enum generation according to usage

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2832

- correct usage of enum
- Resolve enum type according to the usage

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -380,7 +380,7 @@ private static InputType CreateType(Schema schema, string? format, IReadOnlyDict
             Namespace: schema.Extensions?.Namespace,
             Accessibility: schema.Extensions?.Accessibility,
             Description: schema.CreateDescription(),
-            Usage: InputModelTypeUsage.RoundTrip,
+            Usage: InputModelTypeUsage.None,
```
archerzz, line 383: why do we change to `None`?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -450,11 +450,7 @@ function setUsage(
         } else if (usages.roundTrips.includes(name)) {
             m.Usage = Usage.RoundTrip;
         } else {
-            if ((m as InputEnumType).IsExtensible) {
-                m.Usage = Usage.RoundTrip;
-            } else {
-                m.Usage = Usage.None;
-            }
+            m.Usage = Usage.None;
```
archerzz, line 453: I think we need to report to CADL team to have a usage calculation here.

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
chunyu3: It is from swagger, and there is no need to generate enum, so the Usage is None

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: No. cadl compiler correctly resolve the usage for enum. We used to hard-code extensible enum to RoundTrip as workround. Now we will resolve the enum issue, and this workround need to be removed.

lirenhe (commented):
I didn't see any tests updated, so we don't have any coverage for this part?
Could you add tests to verify the behaviour?

archerzz (approved):

chunyu3:
> I didn't see any tests updated, so we don't have any coverage for this part? Could you add tests to verify the behaviour?

Add extensible enum as response in FirstTest-Cadl project, and this method is not chosen to generate convenience method, so the enum will not be generated.

lirenhe (approved):

Issue #2817, opened 2022-10-28
Add convenience API of paging methods


pshao25:
Example of protocol method:
```C#
public virtual AsyncPageable<BinaryData> GetLedgerEntriesAsync(RequestContext context = null) {}
public virtual Pageable<BinaryData> GetLedgerEntries(RequestContext context = null) {}
```

Corresponding convenience method:
```C#
public virtual AsyncPageable<LedgerEntry> GetLegerEntryValuesAsync(CancellationToken cancellationToken = default)
{
    var context = new RequestContext { CancellationToken = cancellationToken };

    AsyncPageable<BinaryData> pageableBindaryData = GetLedgerEntriesImplementationAsync("PaginationClient.GetLegerEntryValues", context);
    return PageableHelpers.Select(pageableBindaryData, response => CustomPage.FromResponse(response).Value);
}

public virtual Pageable<LedgerEntry> GetLegerEntryValues(CancellationToken cancellationToken = default)
{
    var context = new RequestContext { CancellationToken = cancellationToken };

    Pageable<BinaryData> pageableBindaryData = GetLedgerEntriesImplementation("PaginationClient.GetLegerEntryValues", context);
    return PageableHelpers.Select(pageableBindaryData, response => CustomPage.FromResponse(response).Value);
}
```

lirenhe:
@pshao25,  please confirm about the right approach of implementing this.

chunyu3:
@pshao25 will help to do bench mark performance test to verify which implementation has the better performance, calling protocol method and serialize results or call rest api and serialize result page by page as mgmt-plan does.

pshao25:
Performance result:
![image](https://user-images.githubusercontent.com/97225342/199448107-06087f62-49cb-4e74-958f-b7f62da30c9d.png)
- Data: one page with 1000 items `{ received: "model" }`
- networkLatency: call the local test server on the same machine, test server returns this page with 1000 items
- protocolMethod: call protocol method first and then desiarialize the 1000 items one by one.
- hlc: call generated hlc code, which will desiarialize the 1000 items together.

Conclusion:
Maybe the data is not large enough, however, compared with the network latency, I don't think we would get the performance improvement by changing the desiarialization method. 2.792 ms is just local call. When it comes to real internet, it would be longer.

m-nash:
[...]

PR #2855, opened 2022-11-11
Add convenience method for paging

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/2817

No changes in SDK repo.


src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
```diff
@@ -84,7 +88,7 @@ private static void WriteClientCtor(CodeWriter writer, DataPlaneRestClient restC
             writer.Line();
         }
 
-        private static void WriteOperation(CodeWriter writer, DataPlaneRestClient restClient, RestClientMethod operation, bool async)
+        public static void WriteOperation(CodeWriter writer, RestClientMethod operation, string pipelineName, bool async, WritePageFuncBody writePageFuncBody, string methodNameSuffix = "", WriteStatusCodeImplementation? writeStatusCodeImplementation = null, FieldDeclaration? fieldDeclaration = null)
```
lirenhe, line 91: Why do we need 'methodNameSuffix'?  It looks the value is always empty.

lirenhe (commented):
Do we have a cadl-ranch project that could verify the correctness of the generated code?

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
```diff
@@ -17,6 +18,9 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal class RestClientWriter
     {
+        internal delegate void WritePageFuncBody(CodeWriter writer, CodeWriterDeclaration messageVariable, RestClientMethod operation, string pipelineName, bool async);
+        internal delegate void WriteStatusCodeImplementation(CodeWriter writer, string messageVariable, RestClientMethod operation, bool async, FieldDeclaration fieldDeclaration);
```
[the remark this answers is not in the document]
archerzz: Also I think if it's overridden in sub-classes, you should consider make those as abstract methods, let sub-classes implement them and remove `static`. We have too many static methods in writers.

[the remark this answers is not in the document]
pshao25: @archerzz which static method you mean?

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
pshao25: In HLC, the paging method and its firstpage method have the same name. They are in different classes (one in client and the other in restclient), so there is no issue. 

However, in DPG, they are in the same class, so we need to rename the firstpage method.

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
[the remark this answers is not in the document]
archerzz: @pshao25 The implementation of the delegate. I think since they are implemented by sub-classes, those two should be declared as virtual. delegate is more for external interaction.

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
```diff
@@ -17,6 +19,9 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal class RestClientWriter
     {
+        internal delegate void WriteFuncBody(CodeWriter writer, CodeWriterDeclaration messageVariable, RestClientMethod operation, string pipelineName, bool async);
+        internal delegate void WriteStatusCodeImplementation(CodeWriter writer, CodeWriterDeclaration messageVariable, RestClientMethod operation, bool async, FieldDeclaration fieldDeclaration);
```
ArcturusZhang, line 23: Well, this is not what I meant - we could change this messageVariable to `FormattableString` which is far more flexible.
```suggestion
        internal delegate void WriteFuncBody(CodeWriter writer, Formattable messageVariable, RestClientMethod operation, string pipelineName, bool async);
        internal delegate void WriteStatusCodeImplementation(CodeWriter writer, Formattable messageVariable, RestClientMethod operation, bool async, FieldDeclaration fieldDeclaration);
```

ArcturusZhang: Also this `fieldDeclaration` is quite confusing, which field is this?
Based on what I read below, this should always be the clientDiagnostics, is this true? and could you rename this parameter to make it clear?

ArcturusZhang (commented):
Well, this PR is not quite following the principal of "reuse as much as we could".
Instead of making a static method that accepting a lot of parameters, we should use inheritance, making the important prat of write method "virtual" and override them in our corresponding writers (like in `LowLevelClientWriter` and `DataPlaneClientWriter`)

Another topic about the `delegate`s is that the implementation of that delegate should be define somewhere else, out of this class. Imagine this, why I use a delegate in the same class, letting a method to call it through a parameter, instead of that I just directly call it? This is an issue introduced by the fact that this method is static. If it is not, it could just call that virtual method and if LLC writer want to do something else.

But if we are urgent to merge this PR, I am OK to refactor all these later

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
pshao25: I know you mean `FormattableString`, but I prefer `CodeWriterDeclaration`. Any concern?

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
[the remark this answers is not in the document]
pshao25: The problem is: `LowLevelClientWriter` is not inheriting `RestClientWriter`. I'll move `WriteFuncBodyWithProcessMessage` to `LowLevelClientWriter`.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -111,6 +125,22 @@ public void WriteClient()
             }
         }
 
+        private void WriteFuncBodyWithProcessMessage(CodeWriter writer, CodeWriterDeclaration messageVariable, RestClientMethod operation, string pipelineName, bool async)
```
chunyu3, line 128: It seems that this the function is the same as the one in RestClientWriter. Can we use one?

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
ArcturusZhang: OK, here are my thoughts:
1. if this declaration is designed to be declared outside this method (write it with `:D`), we should pass in a `FormattableString` for flexibility. Because in some circumstances, we might need to pass in an expression (like `result.Value`). This is a delegate, and it should be general
2. if this declaration is designed to be declared inside this method, we might never need to have this parameter on the delegate. The method could define its own local variable (and the code writer will ensure its name never gets collision with others)
And if this variable is also needed by other parts that followed by invocation of this method, I believe it is not very great choice to pass things around.
We will need to make some refactor after this change

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
pshao25: Which one you mean?

src/AutoRest.CSharp/Common/Output/Builders/ClientBuilder.cs
```diff
@@ -103,15 +103,25 @@ public static IEnumerable<PagingMethod> BuildPagingMethods(InputClient inputClie
...
-                    method.Name,
-                    new Diagnostic($"{declaration.Name}.{method.Name}"),
-                    new PagingResponseInfo(paging.NextLinkName, paging.ItemName, objectResponseBody.Type));
+                yield return BuildPagingMethod(method.Name, paging.NextLinkName, paging.ItemName, method, nextPageMethod, declaration);
             }
         }
 
+        public static PagingMethod BuildPagingMethod(string methodName, string? nextLinkName, string? itemName, RestClientMethod method, RestClientMethod? nextPageMethod, TypeDeclarationOptions declaration)
+        {
+            if (!(method.Responses.SingleOrDefault(r => r.ResponseBody != null)?.ResponseBody is ObjectResponseBody objectResponseBody))
```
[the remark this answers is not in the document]
pshao25: Are you saying `BuildPagingMethod` is also called in `BuildPagingMethods`? This is a feature removed after the DPG method improved to be more efficient. So I want to keep this method isolated.

src/AutoRest.CSharp/Common/Generation/Writers/RestClientWriter.cs
pshao25: Fair enough.

src/AutoRest.CSharp/Common/Output/Builders/ClientBuilder.cs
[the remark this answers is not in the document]
chunyu3:
```
if (!(method.Responses.SingleOrDefault(r => r.ResponseBody != null)?.ResponseBody is ObjectResponseBody objectResponseBody))
            {
                throw new InvalidOperationException($"Method {method.Name} has to have a return value");
            }
```

I mean above check has been trigged twice.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
chunyu3: OK. You removed `WriteFuncBodyWithProcessMessage` function from RestClientWriter in latest commit. The issue was resolved.

src/AutoRest.CSharp/Common/Output/Builders/ClientBuilder.cs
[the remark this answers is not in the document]
pshao25: I see. Removed.

chunyu3 (approved):

PR #2859, opened 2022-11-14
Bump Azure.Core to 1.26.0 + Update RawRequestUriBuilder

AlexanderSher:
RawRequestUriBuilder is updated to use new properties and methods available in `RequestUriBuilder` in 1.26 to minimize string allocations.


m-nash (commented):
I would like to postpone this until we have the bandwidth to address this change completely.  Right now our top priority is to be able to generate the upcoming anomaly detector library which has quite a few issues with it.

Issue #2156, opened 2022-04-12
Implement property bag for optional parameters


archerzz:
See https://github.com/Azure/autorest.csharp/issues/1630#issuecomment-1018265506

PR #2862, opened 2022-11-16
Implement property bag in management plane

Yao725:
Fix: https://github.com/azure/autorest.csharp/issues/2156

Design: https://github.com/Azure/autorest.csharp/issues/2951

SDK regen: https://github.com/Azure/azure-sdk-for-net/pull/32605

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Decorator/PropertyBag.cs
```diff
@@ -0,0 +1,217 @@
...
+using AutoRest.CSharp.Output.Models.Shared;
+using AutoRest.CSharp.Output.Models.Types;
+using AutoRest.CSharp.Utilities;
+using Azure.ResourceManager;
+
+namespace AutoRest.CSharp.Mgmt.Decorator
+{
+    internal static class PropertyBag
+    {
+        public static RestClientMethod UpdateMgmtRestClientMethod(this RestClientMethod method, string? resourceName, string methodName, string operationId)
```
m-nash, line 34: I would have expected to see something along the lines of adding one additional method while we are building `AllMethods` in the Resource / ResourceCollection classes.

We cannot get rid of the old one if it existed, but if it never existed in a GA library we don't want the extra overload so I also would have expected some logic around that to determine when to build this.  Lets review this in person when I get back.

src/AutoRest.CSharp/Common/Output/Models/Requests/RestClientMethod.cs
```diff
@@ -13,7 +13,7 @@ namespace AutoRest.CSharp.Output.Models.Requests
 {
     internal class RestClientMethod
     {
-        public RestClientMethod(string name, string? summary, string? description, CSharpType? returnType, Request request, IReadOnlyList<Parameter> parameters, Response[] responses, DataPlaneResponseHeaderGroupType? headerModel, bool bufferResponse, string accessibility, InputOperation operation)
+        public RestClientMethod(string name, string? summary, string? description, CSharpType? returnType, Request request, IReadOnlyList<Parameter> parameters, Response[] responses, DataPlaneResponseHeaderGroupType? headerModel, bool bufferResponse, string accessibility, InputOperation operation, IReadOnlyList<Parameter>? originalParameters = null)
```
m-nash, line 16: Can we use one input for parameters and internally decide if we use the property bag?  We would decide if we want the bag, if so we would make a copy of parameters and modify it and expose it via the `Parameters` property, and expose the original as `OriginalParameters`.

This would make it easier to use across generator types.

We could then have a flag which defaults to true for mpg and dpg, but false for hlc.  Any single library could override that default.

If its easier for first release to have dpg default to false as well that could work

ArcturusZhang:
A high level comment: should we do this in our input classes instead of doing it in mgmt plane code?

src/AutoRest.CSharp/Common/Output/Models/Requests/RestClientMethod.cs
Yao725: fixed, and for now I just enabled property bag for mgmt plane.

samples/Azure.Management.Storage/Generated/Models/BlobContainerCollectionGetAllOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <summary> The BlobContainerCollectionGetAllOptions. </summary>
+    public partial class BlobContainerCollectionGetAllOptions
+    {
+        /// <summary> Initializes a new instance of BlobContainerCollectionGetAllOptions. </summary>
+        public BlobContainerCollectionGetAllOptions()
+        {
+        }
+
+        /// <summary> Optional. Specified maximum number of containers that can be included in the list. </summary>
+        public int? Maxpagesize { get; set; }
```
m-nash, line 19: do we have any cases where the original default on the method was not the type default?  Meaning do we ever have `int foo = 5`?

If so should we create the options bag such that it explicitly honors all the defaults?

`public int? Maxpagesize {get; set;} = null;`

This way we are covered when the default case is used?

samples/Azure.ResourceManager.Sample/Generated/Extensions/SubscriptionResourceExtensionClient.cs
```diff
@@ -490,24 +490,18 @@ public virtual Response<VirtualMachineImage> GetVirtualMachineImage(AzureLocatio
...
-        /// <param name="skus"> A valid image SKU. </param>
-        /// <param name="expand"> The expand expression to apply on the operation. </param>
-        /// <param name="top"> The Integer to use. </param>
-        /// <param name="orderby"> The String to use. </param>
+        /// <param name="options"> A property bag which contains all the parameters of this method except the LRO qualifier and request context parameter. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         /// <returns> An async collection of <see cref="VirtualMachineImageResource" /> that may take multiple service requests to iterate over. </returns>
-        public virtual AsyncPageable<VirtualMachineImageResource> GetVirtualMachineImagesAsync(AzureLocation location, string publisherName, string offer, string skus, string expand = null, int? top = null, string orderby = null, CancellationToken cancellationToken = default)
+        public virtual AsyncPageable<VirtualMachineImageResource> GetVirtualMachineImagesAsync(SampleExtensionsGetVirtualMachineImagesOptions options, CancellationToken cancellationToken = default)
```
m-nash, line 496: It appears in other classes like [here](https://github.com/Azure/autorest.csharp/pull/2862/files#diff-f5f2f87eae78ca10a21df07fa50fcb686542ca8652442cece0ee4fe39ffed17dR197) we are defaulting the options bag if its null since everything is optional, and then [here](https://github.com/Azure/autorest.csharp/pull/2862/files#diff-5bb05231d693a0ca35041f241a2ad7978d20ae6610dcae4f8a08d4930ddb211cR249) we are asserting not null since some of the parameters are required.

Why are we not doing either of those for this method?

src/AutoRest.CSharp/Common/Input/Inputs.cs
```diff
@@ -124,6 +124,14 @@ internal record InputParameter(
             ArraySerializationDelimiter: null,
             HeaderCollectionPrefix: null)
         { }
+
+        private readonly Schema? _schema;
+
+        public Schema Schema
```
m-nash, line 130: I don't think we want the concept of a Schema in the Input classes.  In the end we will be deleting all of the CodeModel classes and only using the Input classes so any mixing like this makes that final task much more difficult.  There should be a 1 way conversion from CodeModel classes to Input classes.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -19,6 +19,14 @@ internal record Parameter(string Name, string? Description, CSharpType Type, Con
         public CSharpAttribute[] Attributes { get; init; } = Array.Empty<CSharpAttribute>();
         public bool IsOptionalInSignature => DefaultValue != null;
 
+        private readonly bool? _isPropertyBag;
+
+        public bool IsPropertyBag
+        {
+            get => _isPropertyBag ?? false;
+            init => _isPropertyBag = value;
```
m-nash, line 27: This record was immutable before is there a reason we want to introduce the ability for it to change after construction?

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -848,7 +863,7 @@ protected virtual void WriteLROResponse(string diagnosticsVariableName, string p
         }
         #endregion
 
-        protected void WriteArguments(CodeWriter writer, IEnumerable<ParameterMapping> mapping, bool passNullForOptionalParameters = false)
+        protected void WriteArguments(CodeWriter writer, IEnumerable<ParameterMapping> mapping, bool passNullForOptionalParameters = false, bool isPropertyBagMethod = false)
```
m-nash, line 866: Can we remove this parameter and use [parameter.Parameter.IsPropertyBag](https://github.com/Azure/autorest.csharp/pull/2862/files#diff-1d8080793123f1f633e4fcb80852199dacf38a51b2fd96f50e256756507f0429R24) instead?

m-nash (changes requested):

samples/Azure.Management.Storage/Generated/Models/BlobContainerCollectionGetAllOptions.cs
Yao725: Yes, I agree that we should explicitly honor the default value, now all the optional parameters will be generated like [this](https://github.com/Azure/autorest.csharp/blob/d76abeb8c49264eef9de1c8c6dd05f01b9c25f2b/test/TestProjects/MgmtPropertyBag/Generated/Models/FooCollectionGetAllOptions.cs#L21-L24).

samples/Azure.ResourceManager.Sample/Generated/Extensions/SubscriptionResourceExtensionClient.cs
Yao725: Actually, we did, this asserting happened in the [public extension](https://github.com/Azure/autorest.csharp/blob/d76abeb8c49264eef9de1c8c6dd05f01b9c25f2b/samples/Azure.ResourceManager.Sample/Generated/Extensions/SampleExtensions.cs#L249).

src/AutoRest.CSharp/Common/Input/Inputs.cs
Yao725: I have removed the Schema here and now I'm using `InputModelType`, thanks for the suggestion.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
Yao725: fixed

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
Yao725: Removed

samples/Azure.ResourceManager.Sample/Generated/Extensions/SubscriptionResourceExtensionClient.cs
m-nash: Ah didn't realize I was looking at the internal extension client and the validation goes on the public interface.

src/AutoRest.CSharp/Common/Output/Models/Requests/RestClientMethod.cs
```diff
@@ -34,6 +36,15 @@ public RestClientMethod(string name, string? summary, string? description, CShar
                 .Distinct()
                 .OrderBy(c => c.Code ?? c.Family * 100);
             ResponseClassifierType = new ResponseClassifierType(statusCodes);
+
+            PropertyBag = null;
+            // By default, we enable property bag feature in management plane and the real behavior will be determined later.
+            if (Configuration.AzureArm)
```
m-nash, line 42: @lirenhe Can we make a follow up issue to enable this in DPG?

lirenhe: Created https://github.com/Azure/autorest.csharp/issues/3004 and @archerzz  will follow up

m-nash (approved):

Issue #2848, opened 2022-11-10
[Track] Use getApiVersion and getApiVersionParameter API to emit apiversion parameter


chunyu3:
Update to use `getApiVersion` and `getApiVersionParameter` in azure-dpg to identify the apiVersion parameter and its values.

Issue #2704, opened 2022-09-14
[emitter]Upgrade cadl compiler and rest version


chunyu3:
upgrade cadl compiler to 0.35.0 and rest to 0.17.0 in csharp emitter.

chunyu3:
duplicate with https://github.com/Azure/autorest.csharp/issues/2703

lirenhe:
@chunyu3 , please upgrade cadl related libraries for the Oct release.

Issue #2514, opened 2022-07-29
Implement default namespace for models, and enable Cadl override


annelo-msft:
We will either put models in a `.Models` namespace by default and allow configuration to override this, or put them in the main namespace by default and allow configuration to specify putting them in a `.Models` namespace.  @KrzysztofCwalina, @tg-msft, FYI.

See discussion https://github.com/Azure/azure-sdk-for-net/pull/30176/files#r932491403

This issue tracks the work to implement the default namespace for models and handle the override as specified from Cadl configuratoin.

lirenhe:
@archerzz, could you help to check whether there are any engineer effort in this issue?  thanks

archerzz:
> @archerzz, could you help to check whether there are any engineer effort in this issue? thanks

Right now we don't use `.Models` namespace for models. So, model classes are generated under the same namespace of client classes.

If we decided to use `.Models` namespace by default and provide options to override, then there will be engineering efforts that:
- change default namespace of model classes to `XXX.Models`
- support configuration to override namespace

archerzz:
Probably pending on the discuss in https://github.com/Azure/cadl-azure/issues/1787

lirenhe:
Updated based on the discussion. 


- [ ]  Have the configuration of 'namespace' in cadl-project.yaml
- [ ]  model-namespace in cadl-project.yaml, default 'true' to put models in a `.Models` namespace by default. Else, put models in the root namespace.

As this would impact all SDKs, please prioritize this work.

PR #2873, opened 2022-11-22
upgrade cadl and use single namespace for models

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2848
Fix https://github.com/Azure/autorest.csharp/issues/2514
Fix https://github.com/Azure/autorest.csharp/issues/2704

- upgrade cadl to latest version, compiler: 0.37.0, rest: 0.19.0
- use isApiVersion API in azure-dpg to identify the apiVersion parameter
- implement model-namespace option, and generate all models in one namespace

# Checklist

# Ready to Land?


pshao25:
I suppose it should include another API called `getDefaultApiVersion`

chunyu3:
> getDefaultApiVersion

Thanks @pshao25 getDefaultApiVersion is for you define apiversion via @cadl-lang/versioning, and it only get the default one( which is the last one), not all the apiversion. So now, I don't use getDefaultApiVersion, but get all the apiVersions defined in the cadl.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -64,6 +64,8 @@ public ModelTypeProvider(InputModelType inputModel, string defaultNamespace, Sou
             _defaultDerivedType = defaultDerivedType ?? (inputModel.IsDefaultDiscriminator ? this : null);
         }
 
+        private static string GetDefaultNamespace(string defaultNamespace)
```
archerzz, line 67: Can we consolidate this into one method?

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
```diff
@@ -20,12 +21,13 @@ public static void Execute(GeneratedCodeWorkspace project, InputNamespace inputN
                 var codeWriter = new CodeWriter();
                 var modelWriter = new ModelWriter();
                 modelWriter.WriteModel(codeWriter, model);
-                project.AddGeneratedFile($"{model.Type.Name}.cs", codeWriter.ToString());
+                var prefix = Configuration.ModelNamespace ? "Models/" : "";
```
archerzz, line 24:
```suggestion
                var path = Configuration.ModelNamespace ?? "Models/";
```

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -99,6 +101,7 @@ export interface NetEmitterOptions {
     csharpGeneratorPath: string;
     "clear-output-folder"?: boolean;
     "save-inputs"?: boolean;
+    "model-namespace"?: boolean;
```
archerzz, line 104: This configuration property name sounds more like an input of the model name space. Maybe the following is better.
```suggestion
    "separate-model-namespace"?: boolean;
```

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
chunyu3: Yes, I want to. But I cannot find an common place to provide this.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
lirenhe: It is from Michael's doc, we may update that after further discussing with all the names.  For now, I suggest leave it as it is.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
pshao25: Why this cannot be in the emitter side?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -127,7 +130,8 @@ const NetEmitterOptionsSchema: JSONSchemaType<NetEmitterOptions> = {
         "new-project": { type: "boolean", nullable: true },
         csharpGeneratorPath: { type: "string", nullable: true },
         "clear-output-folder": { type: "boolean", nullable: true },
-        "save-inputs": { type: "boolean", nullable: true }
+        "save-inputs": { type: "boolean", nullable: true },
+        "model-namespace": { type: "boolean", nullable: true }
```
lirenhe, line 134: I think we also need the 'namespace' from cadl-project.yaml,   do we already get that value?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: Yes. it is better to change to `enable-model-namespace` or `separate-model-namespace`. but now this flag has already used in mgmt, and we need to reuse the existing one now. Maybe we will change it to more clear name when we refact the whole project both mgmt and data-plane.

src/CADL.Extension/Emitter.Csharp/src/lib/cadlServer.ts
```diff
@@ -71,7 +71,9 @@ export function resolveServers(
                     IsNullable: false
                 } as InputPrimitiveType,
                 Location: RequestLocation.Uri,
-                IsApiVersion: false,
+                IsApiVersion:
+                    name.toLowerCase() === "apiversion" ||
```
lirenhe, line 75: can we use the similar API here:   isApiVersion(program, parameter);

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -53,7 +53,7 @@ internal sealed class ModelTypeProvider : SerializableObjectType
         public override ObjectTypeProperty? AdditionalPropertiesProperty => throw new NotImplementedException();
 
         public ModelTypeProvider(InputModelType inputModel, string defaultNamespace, SourceInputModel? sourceInputModel, TypeFactory? typeFactory = null, InputModelType[]? derivedTypes = null, ObjectType? defaultDerivedType = null)
-            : base(inputModel.Namespace ?? defaultNamespace, sourceInputModel)
+            : base(GetDefaultNamespace(defaultNamespace), sourceInputModel)
```
pshao25, line 56: Think about this scenario: model-namespace is false, and user defines namespace in cadl. Is there any issue?

Should it be: 
`GetDefaultNamespace(inputModel.Namespace ?? defaultNamespace)` ?

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
chunyu3: resolved. Change to 'folderPath'

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
chunyu3: No. emitter will not do such calculation/compute, it will emitter what is defined in cadl or configuration

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: yes

src/CADL.Extension/Emitter.Csharp/src/lib/cadlServer.ts
chunyu3: No. current isApiVerison only work on HttpOperationParamter. it is not work on ModelType/ModelProperty. I will ask azure-dpg to support this. But now, we only can calculate ourself. When the api is ready, we will update.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
chunyu3: No. now we will ignore model's namespace defined in cadl, we will use common namespace which is `<serviceNamespace>/<serviceNamespace>.Models`

pshao25 (approved):
Will you release a new version?

chunyu3:
> Will you release a new version?

yes.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -284,19 +291,30 @@ function createModel(
...
+        apiVersions.add(version);
+    }
+
+    const versions = getVersions(
+        program,
+        serviceNamespaceType
+    )[1]?.getVersions();
+    if (versions) {
+        for (const ver of versions) {
+            apiVersions.add(ver.value);
```
lirenhe, line 306: Do we need to remove duplicate for versions?

lirenhe (approved):
Left one comment on a minor issue, other LGTM

PR #2876, opened 2022-11-23
Change intial name to friendly name when resolving usages

pshao25:
Fix this scenario:

Cadl in ADP:
```
@Azure.DPG.convenienceAPI
list is CustomResourceList<ClassificationSchema>;

@doc("List {name} resources", TResource)
op CustomResourceList<
  TResource extends object,
  TCustom extends Foundations.CustomizationFields = {}
> is Foundations.ResourceList<
  TResource,
  Foundations.CustomParameters<TCustom>,
  Page<TResource>
>;
```
`Page<TResource>` has a friendly name `PagedTResource`, so we will generate a model called `PagedTResource`, however, resolveUsages doesn't check friendly name, so its `Usage` in cadl.json would be None.


ArcturusZhang (commented):
Could we add something in an existing test project or add a new test project about the changes in this PR to let us could what it does? I did not see anything about `friendlyname` in the test project changed in this PR

BTW if the changes in the test project in this PR is irrelevant to the changes here, it is a distraction, please try to do that in a different PR

chunyu3 (approved):

pshao25:
> Could we add something in an existing test project or add a new test project about the changes in this PR to let us could what it does? I did not see anything about `friendlyname` in the test project changed in this PR
> 
> BTW if the changes in the test project in this PR is irrelevant to the changes here, it is a distraction, please try to do that in a different PR

`friendlyName` is defined at cadl-azure side on the template `Page<TResource>`. It is not defined at our test side.

The test added is to show the fix in this PR.

ArcturusZhang:
> 

Can we add friendly name to a model in our cadl as well to show whether it works on that?

pshao25:
> > 
> 
> Can we add friendly name to a model in our cadl as well to show whether it works on that?

Added.

ArcturusZhang (approved):

PR #2878, opened 2022-11-24
feat(common): unify abstract base class

### LINKED ISSUE #2857: [Feature Req] Enable abstract base class for dataplane by default
Right now dataplane does not default to abstract for base classes of polymorphic sets.  We should use logic similar to [this](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs#L46).  As soon as we do this all types that are already GAd will become abstract which is a breaking change so we will have to add the [suppress-abstract-base-class](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs#L36) to each of those cases to ensure no breaking changes.

After that all new polymorphic sets will adopt the abstract keyword which is what we want.

archerzz:
- move logic of checking whether a class is abstract by object discriminator from `MgmtObjectType` to `SchemaObjectType`, so that data plane model can benefit the same behavior
- move existing logic of checking whether a class is abstract in `SchemaObjectType` to `MgmtReferenceType`, because the existing logic is for mgmt reference types only
- regen samples & tests

resolve #2857

# Description

# Checklist

# Ready to Land?


archerzz:
Corresponding PR on `azure-sdk-for-net`: https://github.com/Azure/azure-sdk-for-net/pull/32661

m-nash (approved):
Overall this a good change and the direction we want to move in.  Lets be careful in the regen PR to find all to make sure we don't introduce any breaking changes.  There is a configuration for maintain a single class as non-abstract so that we can grandfather in some previously GAd classes.

samples/CognitiveSearch/readme.md
```diff
@@ -12,6 +12,18 @@ input-file:
...
+- Analyzer
+- CharFilter
+- CognitiveServicesAccount
+- DataChangeDetectionPolicy
+- DataDeletionDetectionPolicy
+- ScoringFunction
+- Similarity
+- Skill
+- TokenFilter
+- Tokenizer
```
ArcturusZhang, line 25: How does this configuration work with the rename configurations?
We should fill in its generated name or its original name in the swagger?
Could we emphasize this part in its corresponding documentation?
Since this is not a mgmt-only configuration, do we have a place for its documentation?

test/TestProjects/Inheritance/Generated/Models/BaseClassWithEnumDiscriminator.cs
```diff
@@ -12,10 +12,10 @@ namespace Inheritance.Models
     /// Please note <see cref="BaseClassWithEnumDiscriminator"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
     /// The available derived classes include <see cref="DerivedClassWithEnumDiscriminator"/>.
     /// </summary>
-    internal partial class BaseClassWithEnumDiscriminator
+    internal abstract partial class BaseClassWithEnumDiscriminator
```
ArcturusZhang, line 15: This might be irrelevant to the changes in this PR, but why this class is internal?

ArcturusZhang: OK - it is not used in its swagger. Actually I am expecting that this class should be removed.
I will make some follow ups on this

samples/CognitiveSearch/readme.md
archerzz: The original name, unfortunately...

test/TestProjects/Inheritance/Generated/Models/BaseClassWithEnumDiscriminator.cs
archerzz: @ArcturusZhang  Because `BaseClassWithEnumDiscriminator` doesn't inherit [BaseClass](https://github.com/Azure/autorest.csharp/blob/1ca6314ddf96ae913b1993c7284b6991f10e26a9/test/TestProjects/Inheritance/Inheritance.json#L55) which is used in operations. So the usage analysis of this class is none, thus making it internal.

samples/CognitiveSearch/readme.md
archerzz: @ArcturusZhang I've updated the `polish.md` to emphasize the exact name. See the latest commit.

PR #2881, opened 2022-11-24
fix(dpg): required property in input model not serialized

### LINKED ISSUE #2877: Required field in non-roundtrip model becomes readonly
Cadl:
```
model TimeSeriesPoint {
  @doc("Optional argument, timestamp of a data point (ISO8601 format).")
  timestamp?: zonedDateTime;

  @doc("The measurement of that point, should be float.")
  value: float32;
}
```
Expected: both of these two fields could be set.
Actual: `value` becomes readonly so cannot be set.

Root cause: [here](https://github.com/Azure/autorest.csharp/blob/787427f570ab913fd4c0eb642f417fba892e2d55/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs#L93-L95) if a field is required in input/output model, it is set as readonly.

archerzz:
- fix the logic to determine whether a property should be serialized
- fix the test cases which were not correctly generated
- regen samples/tests

fix #2877

# Description

# Checklist

# Ready to Land?


lirenhe (approved):
@archerzz, please create an issue for the investigation of the usage value.  As discussed, this behaviour is a bit strange.

archerzz:
Filed issue https://github.com/Azure/autorest.csharp/issues/2882

Issue #2888, opened 2022-11-29
Determine why apiVersion shows up in the constructor parameters here instead of the Options class


m-nash:
As stated in this comment here https://apiview.dev/Assemblies/Review/0eacd79245e14c6fa88d5991aac2e02f#fa96b04973f74ca38b2082968ac3e7a6 we should not have the api version on the ctor it should be passable via the options ctor which it already is [here](https://apiview.dev/Assemblies/Review/0eacd79245e14c6fa88d5991aac2e02f#Azure.AI.AnomalyDetector.AnomalyDetectorClientOptions).

Issue #2800, opened 2022-10-21
CADL: The mixed use of template & non-template for operations would case codegen error


lirenhe:
Repro:
1. Use the attached package
2. Execute npm install & cadl compile . --emit @azure-tools/cadl-csharp

expect:
Code are generated successfully.

Actual:
Unhandled exception. System.ArgumentException: An item with the same key has already been added. Key: apiVersion

If I removed the line of  `op delete is ResourceDelete<Pet>;`, compile succeed again.

[petstore.zip](https://github.com/Azure/autorest.csharp/files/9837039/petstore.zip)

ArcturusZhang:
Root cause: when we are mixing the template and non-template operations, the generated api-version parameter is different in the emitter output.
In the `read` or `create` operation (non-template), the api-version parameter has the name `apiVersion`:
```
{
       "$id": "41",
       "Name": "apiVersion",
       "NameInRequest": "api-version",
       "Description": "",
       "Type": {
        "$id": "42",
        "Name": "String",
        "Kind": "String",
        "IsNullable": false
       },
```
In the `delete` operation (template), the api-version parameter has the name `api-version`:
```
{
       "$id": "27",
       "Name": "api-version",
       "NameInRequest": "api-version",
       "Description": "The API version to use for this operation.",
       "Type": {
        "$id": "28",
        "Name": "string",
        "Kind": "String",
        "IsNullable": false
       },
```

lirenhe:
No real service is impacted by this, move to GA

PR #2899, opened 2022-11-30
remove apiversion parameter from client constructor

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2888
Fix https://github.com/Azure/autorest.csharp/issues/2800

- remove apiVersion parameter from client constructor, put it in ClientOptions
- support InputConstant for enum type
- fix issue: exception occur when mix use of template & non  template operation

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
```diff
@@ -93,6 +93,22 @@ public static object ReadConstantValue(ref Utf8JsonReader reader, string propert
 
                     }
                     break;
+                case InputEnumType enumType:
+                    switch (enumType.EnumValueType.Kind)
+                    {
+                        case InputTypeKind.String:
+                            value = reader.GetString() ?? throw new JsonException();
```
lirenhe, line 100: Please include the error message for easy trouble shooting.

src/AutoRest.CSharp/Properties/launchSettings.json
```diff
@@ -420,6 +420,10 @@
       "commandName": "Project",
       "commandLineArgs": "--standalone $(SolutionDir)\\test\\TestProjects\\MgmtSubscriptionNameParameter\\Generated"
     },
+    "MixAPIVersion-Cadl": {
```
archerzz, line 423: Can we test in the existing test project like [APIVersoin](https://github.com/Azure/autorest.csharp/tree/feature/v3/test/TestProjects/ApiVersion)

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
```diff
@@ -93,6 +93,22 @@ public static object ReadConstantValue(ref Utf8JsonReader reader, string propert
...
+                            break;
+                        case InputTypeKind.Int32:
+                            value = reader.GetInt32();
+                            break;
```
lirenhe, line 104: Just curious, this is not related to the APIversion, right?

test/TestProjects/MixAPIVersion-Cadl/Generated/MixApiVersionClient.cs
```diff
@@ -0,0 +1,338 @@
...
+
+        internal HttpMessage CreateCreateRequest(RequestContent content, RequestContext context)
+        {
+            var message = _pipeline.CreateMessage(context, ResponseClassifier200);
+            var request = message.Request;
+            request.Method = RequestMethod.Post;
+            var uri = new RawRequestUriBuilder();
+            uri.Reset(_endpoint);
+            uri.AppendPath("/pets", false);
+            uri.AppendQuery("apiVersion", "2022-07-09", true);
```
archerzz, line 296: But this `apiVersion` parameter defined in cadl won't be exposed to external?

chunyu3: It is a special case, the operation is bind to a special apiversion, so it will define in the operation as @query apiVersion: "2022-07-09". So it is not the same as the one in clientOption, we must pass the exact value in Create***Request method.

src/AutoRest.CSharp/Properties/launchSettings.json
chunyu3: No. I want to define an special case for this mix scenario, And ApiVersion-Cadl is to test an apiversion as path paramter.

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
chunyu3: Yes, It is not for apiVersion. this is to support  InputConstant for enum type.

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
chunyu3: The occurs in json reader and we cannot know the detail, so we just return default exception.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -396,33 +401,21 @@ function createModel(
...
-                            apiVersionInOperation.DefaultValue?.Value
-                        )
-                    ) {
-                        apiVersions.add(
-                            apiVersionInOperation.DefaultValue.Value
-                        );
-                    } else {
-                        if (
-                            apiVersionInOperation.Location ===
+                    if ((!apiVersionInOperation.DefaultValue?.Value ||
```
lirenhe, line 404: This logic is complex, could you add some comments?

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
lirenhe: Can we just use the message like:  "The value is null for ......, expect 'string'."

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: added

archerzz (approved):

lirenhe (approved):
I am still no sure we should do 'throw new JsonException()', but i won't block the merge of this PR.

src/AutoRest.CSharp/Common/Input/CadlInputConstantConverter.cs
chunyu3: It may not because of the value is null or something. The message is also meaningless. Now we just use the default exception as we did before.

PR #2901, opened 2022-11-30
Fix #2880: When Client parameter Type is enum or model, codegen throw exception

### LINKED ISSUE #2880: When Client parameter Type is enum or model, codegen throw exception
When the type of client parameter is enum or model, authorest.csharp throw exception.

e.g. 
When define apiVersion as an enum APIVersion
```
@versioned(Versions)
@service({title: "Anomaly Detector"})
@useAuth(ApiKeyAuth<ApiKeyLocation.header, "Ocp-Apim-Subscription-Key">)
@server(
  "{Endpoint}/anomalydetector/{ApiVersion}",
  "The Anomaly Detector API detects anomalies automatically in time series data. It supports two kinds of mode, one is for stateless using, another is for stateful using. In stateless mode, there are three functionalities. Entire Detect is for detecting the whole series with model trained by the time series, Last Detect is detecting last point with model trained by points before. ChangePoint Detect is for detecting trend changes in time series. In stateful mode, user can store time series, the stored time series will be used for detection anomalies. Under this mode, user can still use the above three functionalities by only giving a time range without preparing time series in client side. Besides the above three functionalities, stateful model also provide group based detection and labeling service. By leveraging labeling service user can provide labels for each detection result, these labels will be used for retuning or regenerating detection models. Inconsistency detection is a kind of group based detection, this detection will find inconsistency ones in a set of time series. By using anomaly detector service, business customers can discover incidents and establish a logic flow for root cause analysis.",
  {
    @doc("""
Supported Cognitive Services endpoints (protocol and hostname, for example:
https://westus2.api.cognitive.microsoft.com).
""")
    Endpoint: string,
    @doc("Api Version")
    @path
    ApiVersion: APIVersion,
  }
)
@doc("CADL project to test api versions.")
namespace ApiVersionInCadl;

enum Versions {
  v1_1: "v1.1",
}

@knownValues(Versions)
model APIVersion is string {}

```

exception occur:
image

AlexanderSher:
The root cause of the problem was that `OperationMethodChainBuilder.BuildParameters` has been consolidating multiple types of parameters and then try to resolve them, resolving cases that are either impossible or not supported, e.g.:
- `ContentType` parameters can only be method parameters and only when there is more than one supported content type
- Parameter type for request conditions is always known upfront
- Constant parameter should always have a `DefaultValue` specified
- Generation of protocol or convenience method parameters for client fields is not needed

And so on. Handling of all these cases is now split, and for some conditions exceptions are thrown to avoid generation of incorrect code.

Fix doesn't contain the test because `cadl compile` for the example from https://github.com/Azure/autorest.csharp/issues/2880 shows the following error:
```
C:/GitHub/autorest.csharp/test/TestProjects/Parameters-Cadl/Parameters-Cadl.cadl:10:2 - error unknown-identifier: Unknown identifier versioned
> 10 | @versioned(Versions)
     |  ^^^^^^^^^
C:/GitHub/autorest.csharp/test/TestProjects/Parameters-Cadl/Parameters-Cadl.cadl:10:1 - error unknown-decorator: Unknown decorator
> 10 | @versioned(Versions)
     | ^^^^^^^^^^^^^^^^^^^^
```

No changes in existing generated code

### DIFF
--- a/src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
+++ b/src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
@@ -285,7 +285,7 @@ namespace AutoRest.CSharp.Output.Models
                 ? KnownParameters.RequestContent
                 : KnownParameters.RequestContentNullable;
 
-            var convenienceBodyParameter = Parameter.FromInputParameter(bodyParameter, _typeFactory.CreateType(bodyParameter.Type), _typeFactory);
+            var convenienceBodyParameter = Parameter.FromInputParameter(bodyParameter, _typeFactory);
             _orderedParameters.Add(new ParameterChain(convenienceBodyParameter, _protocolBodyParameter, _protocolBodyParameter));
 
             if (contentTypeRequestParameter != null && Operation.RequestMediaTypes?.Count > 0)
@@ -356,33 +356,29 @@ namespace AutoRest.CSharp.Output.Models
         private void AddParameter(string nameInRequest, InputParameter inputParameter)
         {
             var serializationFormat = SerializationBuilder.GetSerializationFormat(inputParameter.Type);
-
-            if (inputParameter.Kind == InputOperationParameterKind.Client)
+            switch (inputParameter.Kind)
             {
-                AddRequestPartFromClient(nameInRequest, inputParameter, serializationFormat);
-                return;
+                case InputOperationParameterKind.Client:
+                    AddRequestPartFromClient(nameInRequest, inputParameter, serializationFormat);
+                    return;
+                case InputOperationParameterKind.Constant:
+                    AddRequestPartFromConstant(nameInRequest, inputParameter, serializationFormat);
+                    return;
+                case InputOperationParameterKind.Grouped:
+                    AddGroupedParameter(nameInRequest, inputParameter, serializationFormat);
+                    return;
             }
 
-            if (inputParameter.Kind == InputOperationParameterKind.Constant)
+            if (inputParameter.Location == RequestLocation.None)
             {
-                AddRequestPartFromConstant(nameInRequest, inputParameter, serializationFormat);
+                AddConvenienceMethodOnlyParameter(inputParameter);
                 return;
             }
 
             var protocolMethodParameter = Parameter.FromInputParameter(inputParameter, ChangeTypeForProtocolMethod(inputParameter.Type), _typeFactory);
+            var convenienceMethodParameter = Parameter.FromInputParameter(inputParameter, _typeFactory);
+            var parameterChain = new ParameterChain(convenienceMethodParameter, protocolMethodParameter, protocolMethodParameter);
             _requestParts.Add(new RequestPartSource(nameInRequest, inputParameter, protocolMethodParameter, serializationFormat));
-
-            if (inputParameter.Kind == InputOperationParameterKind.Grouped)
-            {
-                _orderedParameters.Add(new ParameterChain(null, protocolMethodParameter, protocolMethodParameter));
-                return;
-            }
-
-            var convenienceMethodParameter = Parameter.FromInputParameter(inputParameter, _typeFactory.CreateType(inputParameter.Type), _typeFactory);
-            var parameterChain = inputParameter.Location == RequestLocation.None
-                ? new ParameterChain(convenienceMethodParameter, null, null)
-                : new ParameterChain(convenienceMethodParameter, protocolMethodParameter, protocolMethodParameter);
-
             _orderedParameters.Add(parameterChain);
         }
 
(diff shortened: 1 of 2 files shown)


lirenhe (commented):
Looks this change is big, can we have some test case for it?

AlexanderSher:
> Looks this change is big, can we have some test case for it?

`OperationMethodChainBuilder.BuildOperationMethodChain` is executed for every DPG and CADL project, so this code has pretty good coverage. Since there is no difference in the generated output for these projects, I would consider the change correct.

Unfortunately, I can't write a test for #2880 because `cadl compile` fails for the input. But I think this change fixes the issue cause we won't create parameters for client fields anymore.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -98,34 +96,20 @@ private bool ShouldConvenienceMethodGenerated(ReturnTypeChain returnTypeChain)
...
-            }
-
-            if (first == null && second == null)
-            {
-                return true;
-            }
-
-            return first!.Type.Equals(second!.Type);
-        }
+        private static bool IsParameterTypeSame(Parameter? first, Parameter? second) => first == null || second == null || first.Type.Equals(second.Type);
```
archerzz, line 99: So here if `first == null` and `second != null`, the two parameters are the same?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -319,106 +303,129 @@ public void AddRequestConditionHeaders(RequestConditionHeaders conditionHeaderFl
...
                 default:
-                    _orderedParameters.Add(new ParameterChain(KnownParameters.RequestConditionsParameter, KnownParameters.RequestConditionsParameter, KnownParameters.RequestConditionsParameter));
-                    AddReference(KnownParameters.RequestConditionsParameter.Name, null, KnownParameters.RequestConditionsParameter, serializationFormat);
+                    inputParameter = null;
+                    parameter = KnownParameters.RequestConditionsParameter;
+                    nameInRequest = parameter.Name;
                     break;
             }
+
+            _orderedParameters.Add(new ParameterChain(parameter, parameter, parameter));
```
archerzz, line 328: Can we have another constructor of `ParameterChain(parameter)`? so that:
```suggestion
            _orderedParameters.Add(new ParameterChain(parameter));
```

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -295,18 +284,13 @@ private void AddBody(InputParameter? bodyParameter, InputParameter? contentTypeR
             _protocolBodyParameter = bodyParameter.IsRequired
                 ? KnownParameters.RequestContent
                 : KnownParameters.RequestContentNullable;
-            _orderedParameters.Add(new ParameterChain(BuildParameter(bodyParameter), _protocolBodyParameter, _protocolBodyParameter));
 
-            if (contentTypeRequestParameter != null)
+            var convenienceBodyParameter = Parameter.FromInputParameter(bodyParameter, _typeFactory);
```
chunyu3, line 288: `BuildParameter` already support to build an convenience parameter. Why do we define another series of functions?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -295,18 +284,13 @@ private void AddBody(InputParameter? bodyParameter, InputParameter? contentTypeR
...
             {
-                if (Operation.RequestMediaTypes?.Count > 1)
-                {
-                    AddContentTypeRequestParameter(contentTypeRequestParameter, Operation.RequestMediaTypes);
-                }
-                else
-                {
-                    AddParameter(contentTypeRequestParameter, typeof(ContentType));
-                }
+                AddContentTypeParameter(contentTypeRequestParameter, Operation.RequestMediaTypes);
```
chunyu3, line 293: When contentTypeRequestParameter is not null, but it does not define RequestMediaTypes, will we not add this contentTypeRequestParameter? 
In that way, will we not set `content-type` or we will set default 'application/json' when create***Request?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
AlexanderSher: That was the old behavior. Probably the name of the method isn't the good one.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
AlexanderSher: I would prefer it to be explicit. Constructor with less parameters often considered to be default one, while in our case it is not: models and generated enums can't be parameter types in protocol methods.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
AlexanderSher: This is not possible in swagger (generated M4 should contain at least one media type). Is it possible in CADL?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
AlexanderSher: Sorry, I don't fully understand the question. `BuildParameter` has been relying on implicit check rather than explicit condition, which is the only option if at compile time we don't know if we have to change parameter type or not. However, it turned out that at the moment when parameter is created we do know explicitly if we need to change parameter type. E.g. (pseudo-code):
```cs
if (some_conditions) {
    Parameter.FromInputParameter(inputParameter, CreateTypeForParameter(...), _typeFactory)
} else {
    Parameter.FromInputParameter(inputParameter, _typeFactory)
}
```
instead of
```
var parameterType = some_conditions ? CreateTypeForParameter(...) : null;
BuildParameter(inputParameter, parameterType);
```

PR #2918, opened 2022-12-06
Refactor around safe flatten properties

ArcturusZhang:
Regen: https://github.com/Azure/azure-sdk-for-net/pull/33024

The changes in this regen are all from some unapplied previous changes in the generator.

# Description

We introduced the "safe flattened property" when we are writing the model, and the name for that flattened property is calculated in the model writer. This violates our principal that "the writer should be dumb and just simply writes whatever the type provider tells it.

It also introduces some difficulties when we want to use the PUBLIC properties, because the flattened property actually hides a public property, such as the model factory and sample generator.

Therefore I introduce this PR to refactor the logic around that. In this PR, I introduced a concrete property to represent the flattened property, and the property to flattened is determined to be internal during the process of building the properties.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
```diff
@@ -84,198 +84,145 @@ protected virtual void WriteProperties(CodeWriter writer, ObjectType schema)
...
-                Stack<ObjectTypeProperty> hierarchyStack = property.GetHeirarchyStack();
-                if (Configuration.AzureArm && hierarchyStack.Count > 1)
+                WriteProperty(writer, property);
+
+                if (property.FlattenedProperty != null)
+                    WriteProperty(writer, property.FlattenedProperty);
+            }
+        }
+
+        private static bool AllCtorsContainMyProperty(ObjectType enclosingType, ObjectTypeProperty property)
```
Yao725, line 94: Do we need this method?

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
```
Yao725, line 6: Can we remove this?

Yao725 (approved):

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
ArcturusZhang: I just moved these methods around in the same file. If no one is using it, I will remove it

ArthurMa1978 (approved):

PR #2934, opened 2022-12-12
Use overloads when both protocol method and convenience method are new

### LINKED ISSUE #2887: Use overloads when both protocol and convenience are new at the same time
If we are creating the protocol and convenience methods at the same time they should both use the same name and the overload that takes RequestContent should not have optional parameters.

As an example with this comment https://apiview.dev/Assemblies/Review/0eacd79245e14c6fa88d5991aac2e02f#5fdf7c9f2e154394a459e56ea4a166ce the two method would be
```c#
public virtual Pageable<Model> GetMultivariateModels(int? skip = null, int? maxCount = null, CancellationToken cancellationToken = default);
public virtual Pageable<BinaryData> GetMultivariateModels(int? skip, int? maxCount, RequestContext context);
```

pshao25:
SDK change: https://github.com/Azure/azure-sdk-for-net/pull/35724
Fix https://github.com/Azure/autorest.csharp/issues/2887
To resolve #2887, there need three steps:
1. We should have the ability to know the latest GA tag. https://github.com/Azure/azure-sdk-for-net/pull/35001. This is .net specific.
2. We should have a common script to download the latest code with the latest GA tag. https://github.com/Azure/azure-sdk-tools/pull/5587
3. And the .net generator for the specifc usage in this PR.


src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -177,6 +177,20 @@ public static async Task<GeneratedCodeWorkspace> Create(string projectDirectory,
             return new GeneratedCodeWorkspace(generatedCodeProject);
         }
 
+        public static GeneratedCodeWorkspace CreateExistingCodeProject(string outputDirectory, Func<string, bool>? skipPredicate = null)
```
ArcturusZhang, line 180: Comparing this method with the one above, here `outputDirectory` actually refers to the same thing as the `projectDirectory` above. Could we align?
Therefore I doubt if we have to add a new configuration `existingProjectPath`. If we have the ability to read all the customized documents, we should be able to read all the documents, not to mention actually previously we are skipping the generated document explicitly.
Also will it be a problem if we do not add the shared code in this project?

src/AutoRest.CSharp/Common/Input/Source/CompilationInput.cs
```diff
@@ -0,0 +1,13 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+using Microsoft.CodeAnalysis;
+
+namespace AutoRest.CSharp.Input.Source
+{
+    public record CompilationInput(Compilation Compilation, Func<INamedTypeSymbol, bool>? typeFilter = null, Func<IMethodSymbol, bool>? methodFilter = null);
```
ArcturusZhang, line 12: nit
```suggestion
    public record CompilationInput(Compilation Compilation, Func<INamedTypeSymbol, bool>? TypeFilter = null, Func<IMethodSymbol, bool>? MethodFilter = null);
```
Also I suggest we add two methods on this record so that we do not have to do the null check when we check the symbol in other classes, something like
```
public bool CheckType(INamedTypeSymbol type) => typeFilter?.Invoke(type) ?? false;

public bool CheckMethod(IMethodSymbol method) => methodFilter?.Invoke(method) ?? false;
```

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -91,11 +95,26 @@ public LowLevelClientMethod BuildOperationMethodChain()
             return new LowLevelClientMethod(protocolMethodSignature, convenienceMethod, _restClientMethod, requestBodyType, responseBodyType, diagnostic, _protocolMethodPaging, Operation.LongRunning, _conditionHeaderFlag);
         }
 
-        private bool ShouldConvenienceMethodGenerated(ReturnTypeChain returnTypeChain)
+        private bool ShouldConvenienceMethodGenerated()
```
ArcturusZhang, line 98: This method reads weird. Should it be `ShouldGenerateConvenienceMethod`?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -91,11 +95,26 @@ public LowLevelClientMethod BuildOperationMethodChain()
...
+                || !_returnType.Convenience.Equals(_returnType.Protocol));
+        }
+
+        private bool ShouldOverloadConvenienceMethod()
+        {
+            if (!ShouldConvenienceMethodGenerated() || _sourceInputModel == null)
+            {
+                return false;
+            }
+            var existingMethod = _sourceInputModel.FindForMethod(Operation.Name.ToCleanName());
```
ArcturusZhang, line 111: Do we have a concrete way to get the name of this method? We do not need to calculate that again here

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -91,11 +95,26 @@ public LowLevelClientMethod BuildOperationMethodChain()
...
+            if (existingMethod == null || existingMethod.Parameters.Length < 1 || existingMethod.Parameters.Last().IsOptional == false)
```
ArcturusZhang, line 112: Please add some comment to describe the predicates here for future references

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
pshao25: `CreateExistingCodeProject` and `Create` are differenct things. `Create` is for the final project we will generate. It contains the shared folders, metadata references, etc, which are not needed for `CreateExistingCodeProject`.

`existingProjectPath` is for test purpose. If we don't expose this parameter, we cannot test the scenario "if there is a protocol method exists, then next time when adding the convenience method, it should not be an overload", because it will nver pass the up-to-date code check.

Not adding the shared code in project is totally OK.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
pshao25: You mean cache the clean name in operation, or?

ArcturusZhang: maybe, having a centralized place for the name which this generated method is meant to have.
For instance, since this class is a builder for an operation, we should have a place that we return the operation with a proper name. We can extract that name and put it into at least a variable

src/AutoRest.CSharp/Common/Input/Source/SourceInputModel.cs
```diff
@@ -56,6 +84,46 @@ public ModelTypeMapping CreateForModel(INamedTypeSymbol? symbol)
...
+                    {
+                        continue;
+                    }
+
+                    var existingParameters = method.Parameters.ToArray();
+                    int index = 0;
+                    foreach (var parameter in parameters)
+                    {
+                        // It would be better to compare the types.
+                        if (parameter != existingParameters[index++].Name)
```
Yao725, line 112: If only the last one of `parameters` has the different name when compared with the penultimate of `existingParameters`, the increment operator will also be executed and index will be equal to the length of the parameters, is this expected?

pshao25: Not expected. Fixed.

src/AutoRest.CSharp/Common/Input/Source/CompilationInput.cs
```diff
@@ -9,9 +9,21 @@
...
+        public CompilationInput(Compilation compilation, Func<INamedTypeSymbol, bool>? typeFilter = null, Func<IMethodSymbol, bool>? methodFilter = null)
+        {
+            _compilation= compilation;
+            _typeFilter= typeFilter;
+            _methodFilter= methodFilter;
+        }
+
+        public Compilation Compilation => _compilation;
+        public bool FilterType(INamedTypeSymbol type) => _typeFilter == null || _typeFilter(type);
+        public bool FilterMethod(IMethodSymbol method) => _methodFilter == null || _methodFilter(method);
```
ArcturusZhang, line 27: If we make this abstract, we do not need these delegates, just use abstract methods.
```suggestion
public CompilationInput(Compilation compilation)
        {
            _compilation= compilation;
        }

        public Compilation Compilation => _compilation;
        public abstract bool FilterType(INamedTypeSymbol type);
        public abstract bool FilterMethod(IMethodSymbol method);
```

src/AutoRest.CSharp/Common/Input/Source/SourceInputModel.cs
```diff
@@ -56,6 +78,48 @@ public ModelTypeMapping CreateForModel(INamedTypeSymbol? symbol)
...
+        public IMethodSymbol? FindForMethod(string name, IEnumerable<string> parameters)
+        {
+            var methods = _methodSet.Where(m => m.Name == name).ToArray();
+            if (methods.Length == 0)
+            {
+                return null;
+            }
+            else if (methods.Length == 1)
+            {
+                return methods.First();
```
chunyu3, line 90: Why immediately return when there is only one method with the same method name? the method may don't have the same parameters.

pshao25: Because this is enough. We cannot define two methods with same name. What's more, we cannot compare the types technically.

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.CodeAnalysis;
+
+namespace AutoRest.CSharp.Input.Source
+{
+    public class ProtocolCompilationInput: CompilationInput
+    {
+        private readonly List<IMethodSymbol> _methodSet = new List<IMethodSymbol>();
+
+        public ProtocolCompilationInput(Compilation compilation)
+            : base(compilation) { }//, type => type.Name.EndsWith("Client"), method => method.Parameters.Length > 0 && method.Parameters.Last().Name == "context") { }
```
ArcturusZhang, line 18: the comment should be removed here

src/AutoRest.CSharp/Common/Input/Source/SourceInputModel.cs
```diff
@@ -56,6 +64,15 @@ public ModelTypeMapping CreateForModel(INamedTypeSymbol? symbol)
...
         }
 
+        public IMethodSymbol? FindProtocolMethod(string namespaceName, string clientName, string methodName, IEnumerable<string> parameters)
+        {
+            if (_existingCompilation is ProtocolCompilationInput protocolCompilation)
+            {
+                return protocolCompilation.FindSymbol(namespaceName, clientName, methodName, parameters);
+            }
+            return null;
+        }
```
ArcturusZhang, line 74: Because `SourceInputModel` is used in multiple target, a method that only finds "protocol" methods that only exist in DPG does not make sense.
We should generalize this method as `FindMethod`. From the parameter list of this method, there is no clue that shows these parameters can only map to a protocol method. If I pass in namespace, client name, method name and parameters of an existing other type of method, I should also get it.

ArcturusZhang: Also please add some comments on this method to explain the purpose of each parameters.
And I believe we should not use `clientName`, instead we should use `typeName`, to indicate that this method is "finding a method on a type with the specific signature".

ArcturusZhang: What does the `IEnumerable<string> parameters` mean? Is this a list of parameter names or parameter types?
Ideally, this should be a `IEnumerable<IParameterSymbol> parameters` or `IEnumerable<ITypeSymbol>` or `IEnumerable<CSharpType>`. If this means a list of parameter type names (which is a string array), it should also work fine. But I believe they should be fully qualified names (namespace included to disambiguate as much as possible)

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
```diff
@@ -0,0 +1,88 @@
...
+                            {
+                                _methodSet.Add(methodSymbol);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        public IMethodSymbol? FindSymbol(string namespaceName, string clientName, string methodName, IEnumerable<string> parameters)
```
ArcturusZhang, line 40: First, this method is finding a method symbol, therefore I believe a better name of this should be `FindMethodSymbol`.
This method is general enough to be the real abstract method.
The idea of abstraction and polymorphism is to extract common behaviors from all possible derived types.
Finding a method symbol on an existing compilation is kind of behavior that all possible derived type of `CompilationInput` should possess, therefore on the base type, we should make this method abstract, and override it here.

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
```diff
@@ -0,0 +1,88 @@
...
+                        {
+                            if (member is IMethodSymbol methodSymbol && isProtocolMethod(methodSymbol))
+                            {
+                                _methodSet.Add(methodSymbol);
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
ArcturusZhang, line 38: This method is building a cache. We should not bring this into the abstraction. True they all should have this step, but this is not a behavior a `CompilationInput` should have, external members actually should not invoke this method, because this method changes the internal status of this instance.
This cache should be maintained by this instance itself.
Please also refer the below comment as well.

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
ArcturusZhang: Also we should do the same on this method as the comments I put in SourceInputModel.cs
(adding descriptions, change the parameter name (clientName => className), and possibly change the type or specify the meaning of parameters)

src/AutoRest.CSharp/Common/Input/Source/SourceInputHelper.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.CodeAnalysis;
+using System.Collections.Generic;
+
+namespace AutoRest.CSharp.Input.Source
+{
+    public class SourceInputHelper
```
ArcturusZhang, line 9: Maybe we should make the class static as well?

src/AutoRest.CSharp/Common/Input/Source/SourceInputModel.cs
```diff
@@ -32,14 +35,19 @@ public SourceInputModel(Compilation compilation)
...
                     {
                         _nameMap.Add(schemaName, namedTypeSymbol);
                     }
                 }
             }
+
+            if (existingCompilation != null)
+            {
+                existingCompilation.FilterSymbols();
+            }
```
ArcturusZhang, line 50: I believe this method on this instance should only be invoke once, therefore it never should be public and we should not invoke it here. It should be invoked internally in this instance (for instance when first time we call the FindSymbol on it)

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -151,6 +154,7 @@ public static void Initialize(IPluginCommunication autoRest)
                 disablePaginationTopRenaming: GetOptionBoolValue(autoRest, Options.DisablePaginationTopRenaming),
                 unreferencedTypesHandling: GetOptionEnumValue<UnreferencedTypesHandlingOption>(autoRest, Options.UnreferencedTypesHandling),
                 projectFolder: autoRest.GetValue<string?>(Options.ProjectFolder).GetAwaiter().GetResult(),
+                existingProjectFolder: null,
```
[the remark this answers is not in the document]
pshao25: Currently, this feature is for convenience method, so only DPG is impacted.

src/AutoRest.CSharp/Common/Input/Source/SourceInputModel.cs
pshao25: We cannot do this. Different method would have different arguments.

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
pshao25: Why do you say we change `CompilationInput` to interface?

src/AutoRest.CSharp/Common/Input/Source/SourceInputModel.cs
pshao25: Synced to move this to base abstract class.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -178,7 +178,7 @@ public CSharpType GetNonNullable()
             if (!IsNullable)
                 return this;
 
-            return new CSharpType(Implementation, Namespace, Name, IsValueType, IsEnum, false, Arguments);
+            return IsFrameworkType ? new CSharpType(_type!, false, Arguments) : new CSharpType(Implementation, Namespace, Name, IsValueType, IsEnum, false, Arguments);
```
ArcturusZhang, line 181: this might be better? 
```suggestion
            return IsFrameworkType ? new CSharpType(FrameworkType, false, Arguments) : new CSharpType(Implementation, Namespace, Name, IsValueType, IsEnum, false, Arguments);
```

src/AutoRest.CSharp/Common/Input/Source/CompilationInput.cs
```diff
@@ -0,0 +1,22 @@
...
+    public abstract class CompilationInput
+    {
+        protected Compilation _compilation;
+
+        public CompilationInput(Compilation compilation)
+        {
+            _compilation= compilation;
+        }
+
+        private protected abstract void FilterSymbols();
```
ArcturusZhang, line 19: we could actually remove this method and let each derived class to handle this by themselves

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
```diff
@@ -0,0 +1,99 @@
...
+                        {
+                            if (member is IMethodSymbol methodSymbol && IsProtocolMethod(methodSymbol))
+                            {
+                                _methodSet.Add(methodSymbol);
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
ArcturusZhang, line 45: we could use our common pattern to write this:
```suggestion
        private List<IMethodSymbol>? _methodSet;
        private List<IMethodSymbol> MethodSet => _methodSet ??= EnsureMethodSet();
        private List<IMethodSymbol> void FilterSymbols()
        {
            var result = new List<IMethodSymbol>();
            foreach (IModuleSymbol module in _compilation.Assembly.Modules)
            {
                foreach (var type in SourceInputHelper.GetSymbols(module.GlobalNamespace))
                {
                    if (type is INamedTypeSymbol typeSymbol && IsClient(typeSymbol))
                    {
                        foreach (var member in typeSymbol.GetMembers())
                        {
                            if (member is IMethodSymbol methodSymbol && IsProtocolMethod(methodSymbol))
                            {
                                result.Add(methodSymbol);
                            }
                        }
                    }
                }
            }
            return result;
        }
```

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        internal override IMethodSymbol? FindMethod(string namespaceName, string typeName, string methodName, IEnumerable<CSharpType> parameters)
+        {
+            if (_methodSet == null)
+            {
+                FilterSymbols();
+            }
```
ArcturusZhang, line 52: we could directly use the property `MethodSet` instead in this method, not to worry about "whether this property is initialized or not" because that is covered by the property.

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
```diff
@@ -0,0 +1,99 @@
...
+                    {
+                        return method;
+                    }
+                }
+                return null;
+            }
+        }
+
+        private bool IsClient(INamedTypeSymbol type) => type.Name.EndsWith("Client");
+        private bool IsProtocolMethod(IMethodSymbol method) => method.Parameters.Length > 0 && method.Parameters.Last().Name == "context";
```
ArcturusZhang, line 97: Maybe we should be precise and check if the last parameter is `RequestContext context`?

src/AutoRest.CSharp/Common/Input/Source/SourceInputModel.cs
```diff
@@ -56,6 +59,11 @@ public ModelTypeMapping CreateForModel(INamedTypeSymbol? symbol)
             return new ModelTypeMapping(_modelAttribute, _schemaMemberNameAttribute, symbol);
         }
 
+        internal IMethodSymbol? FindMethod(string namespaceName, string typeName, string methodName, IEnumerable<CSharpType> parameters)
+        {
+            return _existingCompilation != null ? _existingCompilation.FindMethod(namespaceName, typeName, methodName, parameters) : null;
```
ArcturusZhang, line 64: how about just 
```suggestion
            return _existingCompilation?.FindMethod(namespaceName, typeName, methodName, parameters);
```

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
pshao25: You mean we add a property `MethodSet` with backing field `_methodSet`? What is the benefit?

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
pshao25: The problem is: `RequestContext` is in `Azure.Core`, when we load it as a symbol, we don't know its namespace name.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -205,6 +211,61 @@ public static void Initialize(IPluginCommunication autoRest)
...
+                rootFolder = rootFolder.TrimEnd(Environment.NewLine.ToCharArray());
+                var scriptPath = Path.Join(rootFolder, "eng", "common", "scripts", "Download-Latest-Contract.ps1");
+                // var scriptPath = "D:\\GIT\\sdk-for-pr\\eng\\common\\scripts\\Download-Latest-Contract.ps1";
+                if (File.Exists(scriptPath))
+                {
+                    if (projectFolder.EndsWith("src"))
+                    {
+                        projectFolder = Path.Join(projectFolder, "..");
+                    }
+                    projectFolder = "D:\\GIT\\sdk-for-pr\\sdk\\analysisservices\\Azure.ResourceManager.Analysis";
```
chunyu3, line 240: Why hardcode a sdk project folder?

chunyu3 (commented):
Can we add a test to download exsisting GA version?

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -205,6 +211,61 @@ public static void Initialize(IPluginCommunication autoRest)
             );
         }
 
+        private static string DownloadLatestContract(string projectFolder)
```
pshao25, line 214: To explain the change in this file:
1. When the `existingProjectFolder` is not empty, it means we are in the test.
2. When the `existingProjectFolder` is empty, it means we are in the product so we try to call `Download-Latest-Contract.ps1` implemented in https://github.com/Azure/azure-sdk-tools/pull/5587 and https://github.com/Azure/azure-sdk-for-net/pull/35001

In the function of `DownloadLatestContract`, we will start two processes sequentially:
1. Call `git rev-parse --show-toplevel` to get the root folder of current repo, so that we could know the script path for the next step. (Maybe this could be changed to just C# code, but this is not the biggest challenge I want to raise)
2. Call `Download-Latest-Contract.ps1` to download the existing contract. What's more, the script will print the project folder which contains the existing contract to stdout, so that I redirect the stdout to the pipe between the generator process and pwsh process and get the project folder from the pipe.

Everything goes well in standalone until I found it hangs as an autorest plugin. Not sure if it is because the generator and autorest has connection based on stdin and stdout, but the process will hang at `process.StandardOutput.ReadToEnd();`

Current workaround I could think of is:
1. In the `Download-Latest-Contract.ps1`, we print the project folder (containing existing contract) to a temp file.
2. In the generator, we don't redirect stdout. We just call the script `Download-Latest-Contract.ps1` and read data from the temp file generated above. After that, we delete temp file.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
pshao25: This will be deleted before merge.

eng/Generate.ps1
```diff
@@ -60,6 +63,21 @@ function Add-CadlRanch-Typespec([string]$testName, [string]$projectPrefix, [stri
...
+    if (Test-Path $clientPath) {
+        return $clientPath
+    }
+
+    $mainPath = Join-Path $directory "main.tsp"
+    if (Test-Path $mainPath) {
+        return $mainPath
+    }
+
+    $projectNamePath = Join-Path $directory "$($directory.Name).tsp"
```
chunyu3, line 77: please check the existence of this file also, and throw error if it does not exist.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -132,8 +145,52 @@ public enum UnreferencedTypesHandlingOption
...
         }
 
+        private static string? DownloadLatestContract(string projectFolder)
+        {
+            if (AzureArm || Generation1ConvenienceClient)
+            {
+                return null;
+            }
+
+            // projectFolder = "D:\\GIT\\azure-sdk-for-net\\sdk\\analysisservices\\Azure.ResourceManager.Analysis\\src";
```
chunyu3, line 155: remove this.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -132,8 +145,52 @@ public enum UnreferencedTypesHandlingOption
...
+            int sdkFolderIndex = projectFolder.LastIndexOf("sdk", StringComparison.InvariantCultureIgnoreCase);
```
chunyu3, line 156:
```suggestion
            int sdkFolderIndex = projectFolder.LastIndexOf("\\sdk\\", StringComparison.InvariantCultureIgnoreCase);
```

chunyu3: Better to add `\` before and after to avoid searching out the work include `sdk` in it.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -132,8 +145,52 @@ public enum UnreferencedTypesHandlingOption
...
+                return null;
+            }
+
+            string rootFolder = projectFolder.Substring(0, sdkFolderIndex);
+            var scriptPath = Path.Join(rootFolder, "eng", "common", "scripts", "Download-Latest-Contract.ps1");
+            // var scriptPath = "D:\\GIT\\sdk-for-pr\\eng\\common\\scripts\\Download-Latest-Contract.ps1";
+            if (File.Exists(scriptPath))
+            {
+                string projectDirectory = projectFolder.EndsWith("src") ? Path.GetFullPath(Path.Join(projectFolder, "..")) : projectFolder;
+                // projectDirectory = "D:\\GIT\\sdk-for-pr\\sdk\\analysisservices\\Azure.ResourceManager.Analysis";
```
chunyu3, line 168: remove

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -31,11 +32,13 @@ public static class Options
             public const string SingleTopLevelClient = "single-top-level-client";
             public const string AttachDebuggerFormat = "{0}.attach";
             public const string ProjectFolder = "project-folder";
+            public const string ExistingProjectfolder = "existing-project-folder";
             public const string ProtocolMethodList = "protocol-method-list";
             public const string SkipSerializationFormatXml = "skip-serialization-format-xml";
             public const string DisablePaginationTopRenaming = "disable-pagination-top-renaming";
             public const string SuppressAbstractBaseClasses = "suppress-abstract-base-class";
             public const string UnreferencedTypesHandling = "unreferenced-types-handling";
+            public const string AmbiguityHandling = "ambiguity-handling";
```
chunyu3, line 41: Is `ambiguity-handling` not a command line option? If so, how autorest configure this?

chunyu3:
hello @pshao25  would you please also add a test which use `AppendValue` ambiguity-handling-option, make sure the code and the diagnostics Scope are correct in that option?

pshao25:
> hello @pshao25 would you please also add a test which use `AppendValue` ambiguity-handling-option, make sure the code and the diagnostics Scope are correct in that option?

Yes, cadl-ranch "hello" is for that purpose.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
pshao25: I don't think autorest has such requirement. Others don't have generated convenience method. At the very least, autorest configures it in the `autorest.md`.

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
pshao25: Currently there is no solution because we don't know the namespace of RequestContext

chunyu3 (approved):

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -132,8 +145,49 @@ public enum UnreferencedTypesHandlingOption
...
+                    return Path.GetFullPath(Path.Join(rootFolder, "..", "sparse-spec", "sdk", projectName, relativeProject));
+                }
+            }
+
+            return null;
+        }
+
         private static string? _outputFolder;
         public static string OutputFolder => _outputFolder ?? throw new InvalidOperationException("Configuration has not been initialized");
+        public static string? ExistingProjectFolder { get; internal set; }
```
ArcturusZhang, line 190: should this be `private set`? We should not change any configuration outside this class.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -132,8 +145,49 @@ public enum UnreferencedTypesHandlingOption
             _intrinsicTypesToTreatEmptyStringAsNull.UnionWith(additionalIntrinsicTypesToTreatEmptyStringAsNull);
         }
 
+        private static string? DownloadLatestContract(string projectFolder)
```
ArcturusZhang, line 148: Is this the conclusion that we download the contract here?
It is weird that a configuration class is doing something like this.
We could always relying some script to download the contract and passing in the path and if we get null we could assume there is nothing downloaded.

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
ArcturusZhang: I still did not get why we have a blocker on checking the type here.

src/AutoRest.CSharp/Common/Input/Source/SourceInputHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+                }
+            }
+
+            foreach (INamedTypeSymbol symbol in namespaceSymbol.GetTypeMembers())
+            {
+                yield return symbol;
+            }
+        }
+
+        internal static bool IsEqualType(INamedTypeSymbol symbol, CSharpType type)
```
ArcturusZhang, line 29: I made some changes to this method [here](https://github.com/Azure/autorest.csharp/blob/6677daafde53094b1de56bba430c7a76f88e0b17/src/AutoRest.CSharp/Common/Utilities/NamedTypeSymbolExtensions.cs) (because I need it in another PR)
Could you take a look and apply those modifications here?
Also I think this method is better on the NamedTypeSymbolExtensions class as an extension class of `INamedTypeSymbol` therefore I move it there in my PR

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -94,12 +102,41 @@ public LowLevelClientMethod BuildOperationMethodChain()
...
+
+            if (HasAmbiguityBetweenProtocolAndConvenience() && Configuration.AmbiguityHandling == Configuration.AmbiguityHandlingOption.UseOverload)
+            {
+                return false;
+            }
+
+            return true;
         }
 
         private bool IsParameterTypeSame(Parameter? first, Parameter? second)
```
ArcturusZhang, line 142: is this method correct?
when first is null and second is NOT null, we return true, meaning that they have the same type?
This does not look correct to me.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -94,12 +102,41 @@ public LowLevelClientMethod BuildOperationMethodChain()
             return new LowLevelClientMethod(protocolMethodSignature, convenienceMethod, _restClientMethod, requestBodyType, responseBodyType, diagnostic, _protocolMethodPaging, Operation.LongRunning, _conditionHeaderFlag);
         }
 
-        private bool ShouldConvenienceMethodGenerated(ReturnTypeChain returnTypeChain)
+        private bool ShouldGenerateConvenienceMethod()
```
ArcturusZhang, line 105: Could we add some comments to explain the conditions we are using here?

test/AutoRest.TestServer.Tests/Common/Input/Source/SourceInputHelperTests.cs
ArcturusZhang, line 1: I also added some more test cases here: https://github.com/Azure/autorest.csharp/blob/6677daafde53094b1de56bba430c7a76f88e0b17/test/AutoRest.TestServer.Tests/Common/Utilities/NamedTypeSymbolExtensionsTests.cs
Could you please move the additional ones here?

test/AutoRest.TestServer.Tests/Common/Input/Source/SourceInputHelperTests.cs
```diff
@@ -0,0 +1,124 @@
...
+                                break;
+                            case "metadataModel":
+                                _modelSymbol = (INamedTypeSymbol)parameter.Type;
+                                break;
+                        }
+                    }
+                }
+            }
+
+            Configuration.Initialize("Generated", "", "", new string[] { }, false, false, true, false, false, false, false, false, false, false, false, new List<string> { }, Configuration.UnreferencedTypesHandlingOption.RemoveOrInternalize, Configuration.AmbiguityHandlingOption.UseOverload, ".", null, new string[] { }, new List<string>(), new List<string>(), new List<string>(), null, null);
```
ArcturusZhang, line 72: Please use the positional arguments here so that it could be easier to update next time we introduce a new parameter in this method.

test/TestProjects/PetStore-Typespec/Generated/PetStoreClient.cs
```diff
@@ -103,51 +103,34 @@ public virtual Response Delete(int petId, RequestContext context = null)
...
-                return Response.FromValue(Pet.FromResponse(response), response);
-            }
-            catch (Exception e)
-            {
-                scope.Failed(e);
-                throw;
-            }
+            RequestContext context = FromCancellationToken(cancellationToken);
+            Response response = Read(petId, context);
+            return Response.FromValue(Pet.FromResponse(response), response);
```
ArcturusZhang, line 120: Why this changes?
Previously because the method name is different therefore we have to create another scope and now they are the same therefore we do not need the extra scope and just using the same scope in the protocol method?

test/TestProjects/PetStore-Typespec/Generated/PetStoreClient.cs
```diff
@@ -326,49 +312,32 @@ public virtual Response GetPetByKind(string kind, RequestContext context = null)
...
         /// <param name="start"> The Int32 to use. </param>
         /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="context"/> is null. </exception>
         /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
         /// <returns> The response returned from the service. Details of the response body schema are in the Remarks section below. </returns>
         /// <include file="Docs/PetStoreClient.xml" path="doc/members/member[@name='GetFirstPetAsync(Int32,RequestContext)']/*" />
-        public virtual async Task<Response> GetFirstPetAsync(int? start = null, RequestContext context = null)
+        public virtual async Task<Response> GetFirstPetAsync(int? start, RequestContext context)
         {
+            Argument.AssertNotNull(context, nameof(context));
```
ArcturusZhang, line 339: I think we do not need to validate context is not null - it is working perfectly fine when we are not validating it.
We are making it required just because the compiler will not accept the overload, but null value is totally acceptable.

test/TestProjects/PetStore-Typespec/Generated/PetStoreClient.cs
pshao25: Yes, by design.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
pshao25: It is self-explaint.

ArcturusZhang (approved):

PR #2946, opened 2022-12-15
Pagination performance optimizations and refactoring

AlexanderSher:
- Two existing pagination implementations are replaced by one, more performant and with `WithCancellation` support
- BinaryData case is handled via internal hook, so generated code for protocol methods and convenience methods looks similar
- Generation code is moved to the `PageableMethodsWriterExtensions`. It used by DPG, MPG and legacy data plane (except for pageable LRO where pageable part is generated in the operation type, not in client type)

Here is the result of running autorest with this PR in [azure-sdk-for-net repo](https://github.com/Azure/azure-sdk-for-net/compare/main...AlexanderSher:azure-sdk-for-net:GeneratedForAutorest2946)


archerzz (approved):

archerzz (approved):

chunyu3 (approved):

pshao25:
Seems it is for mproving the performance of HLC convenience methods? How about the performance of DPG convenience method? Our goal is to call protocol method inside DPG convenience method.

AlexanderSher:
> Our goal is to call protocol method inside DPG convenience method.

Unfortunately, we can't do this until `JsonElement` type has access to the underlying UTF8 span. Without it, conversion from `BinaryData` to specific model would be very slow. We will have to convert UTF8 into string, then back into UTF8, then parse JSON to create a model.

m-nash (approved):
I like the consolidation of the implementations here.  Please make sure we run some of the recorded pageable tests against your PR example above before merging to validate there are no differences.

Issue #2112, opened 2022-03-28
Provide model factory which can support model mocking for unit-test


ArthurMa1978:

m-nash:
https://github.com/Azure/autorest.csharp/pull/2107

PR #2947, opened 2022-12-16
Introduce model factory to mgmt plane

ArcturusZhang:
Regen: https://github.com/Azure/azure-sdk-for-net/pull/32867

Fix https://github.com/Azure/autorest.csharp/issues/2112

# Description

This PR introduces model factory to mgmt libraries.

The model factory takes a list of models you would like to generate model factory method entries. Therefore it is easy to apply to DPG when it is ready.

Previously the existing model factory in HLC only generates entries of models with at least one readonly property defined in itself. This PR changes the criteria of introducing a model factory entry for a model:
1. The model at least has one public property and not abstract
2. If the model is abstract but it is a base type with discriminator

# Checklist

# Ready to Land?


ArthurMa1978 (approved):

samples/Azure.Management.Storage/Generated/StorageModelFactory.cs
```diff
@@ -0,0 +1,958 @@
...
+        /// <param name="enable"> When set to true last access time based tracking is enabled. </param>
+        /// <param name="name"> Name of the policy. The valid value is AccessTimeTracking. This field is currently read only. </param>
+        /// <param name="trackingGranularityInDays"> The field specifies blob object tracking granularity in days, typically how often the blob object should be tracked.This field is currently read only with value as 1. </param>
+        /// <param name="blobType"> An array of predefined supported blob types. Only blockBlob is the supported value. This field is currently read only. </param>
+        /// <returns> A new <see cref="Models.LastAccessTimeTrackingPolicy"/> instance for mocking. </returns>
+        public static LastAccessTimeTrackingPolicy LastAccessTimeTrackingPolicy(bool enable = default, Name? name = null, int? trackingGranularityInDays = null, IEnumerable<string> blobType = null)
+        {
+            blobType ??= new List<string>();
+
+            return new LastAccessTimeTrackingPolicy(enable, name, trackingGranularityInDays, blobType?.ToList());
```
m-nash, line 84: nit:  I don't think we need the blotType?. here since we are guaranteed its not null with [line 82](https://github.com/Azure/autorest.csharp/pull/2947/files#diff-12b4e2c93b7f384212c026136870359882fb7d7f7ac027b7d49bc2146a5f897cR82).

test/TestProjects/MgmtDiscriminator/Generated/MgmtDiscriminatorModelFactory.cs
```diff
@@ -0,0 +1,317 @@
...
+            return new DeliveryRuleAction(name, foo);
+        }
+
+        /// <summary> Initializes a new instance of DeliveryRuleRemoteAddressCondition. </summary>
+        /// <param name="foo"> For test. </param>
+        /// <param name="parameters"> Defines the parameters for the condition. </param>
+        /// <returns> A new <see cref="Models.DeliveryRuleRemoteAddressCondition"/> instance for mocking. </returns>
+        public static DeliveryRuleRemoteAddressCondition DeliveryRuleRemoteAddressCondition(string foo = null, RemoteAddressMatchConditionParameters parameters = null)
+        {
+            return new DeliveryRuleRemoteAddressCondition("RemoteAddress", foo, parameters);
```
m-nash, line 83: any reason we don't just reference the enum value directly here?

test/TestProjects/MgmtDiscriminator/Generated/MgmtDiscriminatorModelFactory.cs
```diff
@@ -0,0 +1,317 @@
...
+            return new DeliveryRuleProperties(order, conditions, actions?.ToList(), extraMappingInfo);
+        }
+
+        /// <summary> Initializes a new instance of DeliveryRuleCondition. </summary>
+        /// <param name="name"> The name of the condition for the delivery rule. </param>
+        /// <param name="foo"> For test. </param>
+        /// <returns> A new <see cref="Models.DeliveryRuleCondition"/> instance for mocking. </returns>
+        public static DeliveryRuleCondition DeliveryRuleCondition(string name = null, string foo = null)
+        {
+            return new UnknownDeliveryRuleCondition(name, foo);
```
m-nash, line 65: The discriminator here should default to "Unknown" if its an extensible enum or default(EnumType) if its a fixed enum.

Do we have a case for the fixed enum discriminator?

samples/Azure.Management.Storage/Generated/StorageModelFactory.cs
ArcturusZhang: well this comes from our common code writer extension methods - it checks if the instance is nullable, if so it will append `?.ToList` otherwise `.ToList`. I think this is kind of redundant because a thing that supports `ToList` is always nullable.
But I think to change this might involve a lot of unexpected things, I would like to leave it as is because it works.

test/TestProjects/MgmtDiscriminator/Generated/MgmtDiscriminatorModelFactory.cs
ArcturusZhang: Well no particular reason just because it is simple.
I will have some investigation to see if we could use the enum instance here instead of the raw value.

test/TestProjects/MgmtDiscriminator/Generated/MgmtDiscriminatorModelFactory.cs
ArcturusZhang: I wonder if we have a case that the discriminator is a fixed enum because fixed enums mean it could never expand, and if it never expands, what is the point of "Unknown"?

As for the default value, let me investigate if we could change it to Unknown.
But we cannot use Unknown enum value because we do not have one, here we can only use the Unknown literal.

test/TestProjects/MgmtDiscriminator/Generated/MgmtDiscriminatorModelFactory.cs
ArcturusZhang: This is fixed now

test/TestProjects/MgmtDiscriminator/Generated/MgmtDiscriminatorModelFactory.cs
ArcturusZhang: The default value for discriminators is `Unknown` if it appears in the method signature and it only appears in method signature when this is the base class

ArcturusZhang:
We also need to update the code to make sure the types removed by attribute `CodeGenSuppressType` will not be included in the model factory (otherwise we get compilation errors)

src/AutoRest.CSharp/AutoRest.CSharp.csproj
```diff
@@ -21,7 +21,7 @@
     <PackageReference Include="Azure.Core" Version="1.26.0" />
     <PackageReference Include="Azure.Core.Expressions.DataFactory" Version="1.0.0-alpha.20221110.1" />
     <PackageReference Include="Azure.Core.Experimental" Version="0.1.0-preview.18" />
-    <PackageReference Include="Azure.ResourceManager" Version="1.3.1" />
+    <PackageReference Include="Azure.ResourceManager" Version="1.4.0-alpha.20221218.1" />
```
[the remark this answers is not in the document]
ArcturusZhang: We will update to a new stable version once we have that.
We need the new version because we need the model factory in resource manager for some manual-written models (for instance SubResource)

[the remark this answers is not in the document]
ArcturusZhang: 1.4.0 just releases. updated to 1.4.0

m-nash (approved):

Issue #2297, opened 2022-06-08
Handle WaitForCompletionResponse scope in the Grow-up story


pshao25:
**Backgroud:**
To resolve https://github.com/Azure/autorest.csharp/issues/2236, we agree to add the scope `WaitForCompletionResponse` to `OperationInternal`/`OperationInternalOfT`, since DPG, Mgmt, HLC, grow-up from DPG all need this.

**Current behavior:** 
Only grow-up from DPG scnario has `WaitForCompletionResponse` scope because we implemented this wrapper at https://github.com/Azure/autorest.csharp/blob/12cb9bee234b546de6795a95328f1da7dd80acf6/src/assets/Generator.Shared/LowLevelOperationHelpers.cs#L143

**Solution to #2236:** 
Will add the `WaitForCompletionResponse` scope in https://github.com/Azure/azure-sdk-for-net/pull/29033

**Follow-up question: Do we need to remove the [scope wrapper](https://github.com/Azure/autorest.csharp/blob/12cb9bee234b546de6795a95328f1da7dd80acf6/src/assets/Generator.Shared/LowLevelOperationHelpers.cs#L143) in grow-up scenario?**
It is straightforward to remove this, but the problem is:
In grow-up LRO we created a new scope [DPGClient.LroValue](https://github.com/Azure/autorest.csharp/blob/12cb9bee234b546de6795a95328f1da7dd80acf6/test/TestServerProjectsLowLevel/dpg-customization/Customization/DPGClient.cs#L99) expecting we will have a scope named `DPGClient.LroValue.WaitForCompletionResponse`. But we call DPG method inside grow-up method [here](https://github.com/Azure/autorest.csharp/blob/12cb9bee234b546de6795a95328f1da7dd80acf6/test/TestServerProjectsLowLevel/dpg-customization/Customization/DPGClient.cs#L103). During the call to DPG method, we open a scope named `DPGClient.Lro` and pass this name to the procedure of creating the `LowLevelFuncOperation`,
https://github.com/Azure/autorest.csharp/blob/12cb9bee234b546de6795a95328f1da7dd80acf6/test/TestServerProjectsLowLevel/dpg-customization/Generated/DPGClient.cs#L301
which will pass to the constructor of `OperationInternal`
https://github.com/Azure/autorest.csharp/blob/12cb9bee234b546de6795a95328f1da7dd80acf6/src/assets/Generator.Shared/LowLevelFuncOperation.cs#L19-L24
[...]

Issue #2136, opened 2022-04-01
Design distributed tracing scopes for LROs


annelo-msft:
### Summary

In our HLC libraries, distributed tracing scopes are currently opened around every new "Start Operation" and "Get Operation Status"  network call.  For a call to Start Recognize Receipts in our HLC Form Recognizer library (below), you can see that multiple sibling scopes are opened. 

For LROs, we would like to have a single logical outer scope containing all network calls related to the operation.

### Design considerations

We have recently [changed the design of LRO service methods](https://github.com/Azure/azure-sdk-for-net/issues/25716) so that they take a parameter stating whether to return after the operation has started, or wait to return until after the operation has completed.  We will need to determine whether scopes should be different based on which option the caller selects.

### Success criteria
- [ ] We have agreement that this is the pattern we want and don't plan to change it after GA.

### Application Insights view of distributed tracing spans in HLC client
![image](https://user-images.githubusercontent.com/21269347/161311445-b168331a-e527-4ff2-9ab6-cfee9322490e.png)

annelo-msft:
@lmolkova, we discussed that this may be something we could postpone solving until after DPG GA.  Could you confirm?  If it is, I will remove the blocking-release tag.

annelo-msft:
**Proposal:** We will use the same design as HLCs in DPG clients, i.e. one span for each page request.

annelo-msft:
Approved by architects.

Issue #2233, opened 2022-05-12
What is expected behavior wrt Diagnostic Scopes of Operation?


annelo-msft:
We have:

1. LowLevelFuncOperation (which has OperationInternalBase inside with only _updateStatusScopeName) for raw LRO.

2. ConvertOperation (which has both _waitForCompletionScopeName and _updateStatusScopeName) for grow-up LRO.

That leads to the inconsistent behavior between raw lro and grow-up lro. After we call WaitForCompletionAsync, raw lro has scope DPGClient.Lro.UpdateStatus while grow-up lro has scope DPGClient.LroValue.WaitForCompletion.

Could you check out this and let me know what is the expected behavior? Then we could see if it could be fixed.

annelo-msft:
@pshao25, the guidance can be found here:
https://github.com/Azure/autorest.csharp/issues/2136#issuecomment-1092257699

In general, distributed tracing spans should have the same name as the public API call the customer made.  That means if someone does:

```csharp
 Operation<BinaryData> result = await new DPGClient(Key, host, null).LroAsync(WaitUntil.Started, "raw"); 
 await lro.WaitForCompletionAsync(); 
```

The spans created should look like:
- DPGClient.Lro
  - HTTP Call
- Operation.WaitForCompletion
 - HTTP Call
 - ...
 - HTTP Call

This should be true regardless of whether it is a protocol method or a grow-up method.

If someone instead does:
```csharp
Operation<BinaryData> result = await new DPGClient(Key, host, null).LroAsync(WaitUntil.Started, "raw"); 
while (!operation.HasCompleted)
{
    await operation.UpdateStatusAsync();
}
```

The spans created should look like:
- DPGClient.Lro
  - HTTP Call
- Operation.UpdateStatus
 - HTTP Call
- Operation.UpdateStatus
 - HTTP Call
- ...

This should also be true regardless of whether the customer calls a protocol method or a grow-up method.

annelo-msft:
@pshao25, could you file any follow-up issues needed to address getting this implemented consistent with the spec?

pshao25:
@annelo-msft Thank you! Then unfortunately, I have to file https://github.com/Azure/autorest.csharp/issues/2236

PR #2961, opened 2022-12-21
Add LRO Distributed Tracing Tests

m-redding:
# Description
The changes here add test cases for each of the following scenarios:
- LRO called with WaitUntil.Started and WaitForCompletionAsync
- LRO called with WaitUntil.Started and WaitForCompletionResponseAsync
- LRO called with WaitUntil.Started and manual polling with UpdateStatus
- LRO called with WaitUntil.Completed

Each scenario is tested on these cases:
- Operation and Operation\<T\>
- Handwritten and raw LRO's
- Nested span suppression enabled/disabled

The ability to disable nested span suppression is not inherently provided by generated DPG clients. An additional constructor was added to the `DPGClient` type within this testing project in order to test disabled nested span suppression.

The expected behavior being tested is defined by:
- [Nested span suppression enabled](https://github.com/Azure/autorest.csharp/issues/2136#issuecomment-1092257699)
- [Nested span suppression disabled](https://github.com/Azure/autorest.csharp/issues/2297#issuecomment-1152739475)

These tests are only being added to .NET right now, so `ignoreScenario` is being set to true for each new case.

Resolves: https://github.com/Azure/azure-sdk-for-net/issues/29215

# Checklist

# Ready to Land?
- [x] Build is completely green


test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -93,68 +93,344 @@ await foreach (Page<Product> page in allPages.AsPages())
...
 
-            Operation<BinaryData> result = await new DPGClient(Key, host, null).LroAsync(WaitUntil.Started, "raw");
+            Operation<BinaryData> result = await new DPGClient(Key, host, null, nestedScopeSuppression).LroAsync(WaitUntil.Started, "raw");
             diagnosticListener.AssertAndRemoveScope("DPGClient.Lro");
             CollectionAssert.IsEmpty(diagnosticListener.Scopes);
 
             await result.WaitForCompletionAsync();
             diagnosticListener.AssertAndRemoveScope("DPGClient.Lro.WaitForCompletion");
-            CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+            if (nestedScopeSuppression)
```
pshao25, line 109: I think no matter `nestedScopeSuppression` is true or not, we should both check `CollectionAssert.IsEmpty(diagnosticListener.Scopes);` ?

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -93,68 +93,344 @@ await foreach (Page<Product> page in allPages.AsPages())
...
+        {
+            using var diagnosticListener = new ClientDiagnosticListener("dpg_customization_LowLevel", asyncLocal: true);
+            CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+
+            Operation<BinaryData> result = await new DPGClient(Key, host, null, nestedScopeSuppression).LroAsync(WaitUntil.Started, "raw");
+            diagnosticListener.AssertAndRemoveScope("DPGClient.Lro");
+            CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+
+            await result.WaitForCompletionResponseAsync();
+            diagnosticListener.AssertAndRemoveScope("DPGClient.Lro.WaitForCompletion");
```
pshao25, line 131: Could you confirm with arch it should be `DPGClient.Lro.WaitForCompletion` or `DPGClient.Lro.WaitForCompletionResponse`?

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -93,68 +93,344 @@ await foreach (Page<Product> page in allPages.AsPages())
         });
 
         [Test]
-        public Task RawLRO() => Test(async (host) =>
+        [TestCase(true)]
+        [TestCase(false)]
+        public Task RawLRO(bool nestedScopeSuppression) => Test(async (host) =>
```
annelo-msft, line 98: nit: same comment as below regarding parameter naming.

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -93,68 +93,344 @@ await foreach (Page<Product> page in allPages.AsPages())
...
+            await result.WaitForCompletionResponseAsync();
+            diagnosticListener.AssertAndRemoveScope("DPGClient.Lro.WaitForCompletion");
+            if (nestedScopeSuppression)
+            {
+                CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+            }
+
+            JsonData responseBody = JsonData.FromBytes(result.Value.ToMemory());
+            Assert.AreEqual("raw", (string)responseBody["received"]);
+        }, ignoreScenario: true);
```
annelo-msft, line 139: Curious: why `ignoreScenario`?

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
annelo-msft: My understanding is that the scope should have the same name as the public API call, per https://github.com/Azure/autorest.csharp/issues/2233#issuecomment-1127819652.  It would be worth understanding why this isn't happening here to learn whether it is a bug or there is a reason to make an exception in this case.

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -93,68 +93,344 @@ await foreach (Page<Product> page in allPages.AsPages())
...
+        {
+            using var diagnosticListener = new ClientDiagnosticListener("dpg_customization_LowLevel", asyncLocal: true);
+            CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+
+            Operation result = await new DPGClient(Key, host, null, nestedScopeSuppression).LroAsync(WaitUntil.Started, "raw");
+            diagnosticListener.AssertAndRemoveScope("DPGClient.Lro");
+            CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+
+            await result.WaitForCompletionResponseAsync();
+            diagnosticListener.AssertAndRemoveScope("DPGClient.Lro.WaitForCompletion");
```
annelo-msft, line 154: Same comment re: span name matching public method name.

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -93,68 +93,344 @@ await foreach (Page<Product> page in allPages.AsPages())
...
+
+            while (!lro.HasCompleted)
+            {
+                await lro.UpdateStatusAsync();
+                diagnosticListener.AssertAndRemoveScope("DPGClient.LroValue.UpdateStatus");
+                if (nestedScopeSuppression)
+                {
+                    CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+                }
+            }
```
annelo-msft, line 190: Should there be a final check of the LRO value after completion?

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
```diff
@@ -93,68 +93,344 @@ await foreach (Page<Product> page in allPages.AsPages())
...
+                CollectionAssert.IsEmpty(diagnosticListener.Scopes);
+            }
             Assert.AreEqual("model", $"{lro.Value.Received}");
         });
 
         [Test]
-        public Task HandwrittenModelLro_ManualIteration() => Test(async (host) =>
+        [TestCase(true)]
+        [TestCase(false)]
+        public Task HandwrittenModelLroResponseOperationOfT(bool nestedScopeSuppression) => Test(async (host) =>
```
annelo-msft, line 289: This is a lower-priority comment, but given the duplication of complex logic across these tests, I'm wondering if there's a way to implement the scope-checking once, and reuse it across the protocol and convenience methods, rather than re-implementing it by hand?

test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs
m-redding: I added it because the tests won't pass unless they either have ignoreScenario set to true or they're set [in the test server](https://github.com/Azure/autorest.testserver/blob/main/src/test-routes/dpg/customization.ts#L4), it seems to me like adding these scenarios to the test server would imply that the other languages should also test these cases as well, but I believe for now the plan was to only add these tests to .NET?

Issue #2935, opened 2022-12-12
Move logic of clear-output-folder from emitter to codegen


pshao25:
**Background:**
We will decide the method signature according to the provious generated ones. Example: https://github.com/Azure/autorest.csharp/issues/2887.
Therefore, we cannot let users pass in `true` to this parameter and clear existing code arbitrarily,

**Expectation:**
We need to move this clear logic to codegen. What's more, we need to check if the signature changes, we can only do the clear if there is no signature change.

lirenhe:
@chunyu3, what do you think of this proposal?

chunyu3:
I think the previous code is the one already existed in the azure-sdk-for-net, Not the one iteratively generated by customer.

So eventually the `Generated` folder need be clear before we start to generate the code, you can move the codegen after you load the existing code if needed, then clear it and then re-generate. But we need to  pay attention to whether we need to load the code or not.

PR #2963, opened 2022-12-22
Move clear-output-folder from emitter to codegen

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/2935


src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -84,6 +89,29 @@ await foreach (var file in workspace.GetGeneratedFilesAsync())
...
+
+            foreach (DirectoryInfo directory in directoryInfo.GetDirectories())
+            {
+                DeleteDirectory(directory.FullName, keepFiles);
+            }
+
+            if (directoryInfo.GetFiles().Length == 0 && directoryInfo.GetDirectories().Length == 0)
+            {
+                directoryInfo.Delete();
+            }
```
ArcturusZhang, line 112: I suggest to use the `directoryInfo.EnumerateFiles()` instead of `GetFiles` especially when you are detecting if the directory is empty.
GetFiles returns an array which is calculated ahead of usage. EnumerateFiles returns an IEnumerable which is calculated when iterated.
If you want to check if the directory is empty, I believe on check `!directoryInfo.EnumerateFileSystemEntries().Any()` could do the same and it is simpler (this I am not quite sure). If this is not the case, I suggest we use `!directoryInfo.EnumerateFiles().Any() && !directoryInfo.EnumerateDirectories().Any()` instead

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -70,6 +70,11 @@ public static async Task RunAsync(CommandLineOptions options)
                 throw new InvalidOperationException($"Neither CodeModel.yaml nor cadl.json exist in {outputPath} folder.");
             }
 
+            if (options.ClearOutputFolder)
+            {
+                var keepFiles = new string[] { "CodeModel.yaml", "Configuration.json", "cadl.json" };
```
chunyu3, line 75: We should keep this files in SDK library. Only for debug. 

Since we clear output folder after generate, we should not need them

ArcturusZhang: I believe this is OK because we only write them when the `save-inputs` option is on.
We should not remove them if the option is on, and if the option is not on, we will not have these files.
Therefore adding these files as exceptions makes sense

pshao25: These files are controlled by "save-inputs", not "clear-output-folder".

chunyu3 (approved):

ArcturusZhang (approved):

PR #2968, opened 2022-12-27
feat(cadl): support `@deprecate` for enum, model and operation

### LINKED ISSUE #2967: [CADL]: Support `@deprecate` decorator
## Description

Per [`@deprecate` decorator introduction](https://microsoft.github.io/cadl/standard-library/built-in-decorators#deprecated), it can be used in models, operations and namespaces. Since we only support two namespaces per SDK, it's not suitable to support deprecation of namespace. So, let's support `@deprecate` for models and operations.

archerzz:
- update emitter to emit the value of `@deprecate` decorator on models, enums and operations
- update generator:
  - parse incoming `Deprecated` value from CADL emitter
  - generate CSharp `[Obsolete]` attribute for classes, emuns and client methods if `Deprecated` value exists
- add unit tests

resolve #2967

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
```diff
@@ -287,6 +287,12 @@ private string GetAbstract(ObjectType schema)
 
         protected virtual void AddClassAttributes(CodeWriter writer, ObjectType schema)
         {
+            writer.WriteDeprecatedAttributeIfExists(schema.Deprecated);
+        }
+
+        private void AddClassAttributes(CodeWriter writer, EnumType enumType)
```
chunyu3, line 293: Can we combine those two methods as one?
AddClassAttributes(CodeWriter writer, TypeProvider schema)

chunyu3 (approved):

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
archerzz: `protected virtual void AddClassAttributes(CodeWriter writer, ObjectType schema)` is overridden by https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Mgmt/Generation/ReferenceTypeWriter.cs#L20

That's why I add a separate `AddClassAttribute(CodeWriter writer, EnumType enumType` which is for enum types only, to prevent the behavior for enum types is overridden accidentally.

pshao25:
What would be the output if
- A model is deprecated and the operation who is using this model not marked as deprecated.
- An operation is deprecated and the referenced models are left isolated.

And do you have tests for deprecated operations?

archerzz:
> What would be the output if
> 
> * A model is deprecated and the operation who is using this model not marked as deprecated.
> * An operation is deprecated and the referenced models are left isolated.
* I think a model can be used in various places, so I don't plan to propagate the deprecation semantic to referenced entities.
* Vice versa.

> 
> And do you have tests for deprecated operations?
* It's https://github.com/Azure/autorest.csharp/pull/2968/files#diff-10abf63aa0e19d2305a1b76a446357a1b81c0f0616d06b331462c31936b2ac4cR32

PR #2974, opened 2022-12-30
Generate all model

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2960

- Add Flag to generate all models
- emit Literal as enum with only one value

# Checklist

# Ready to Land?


eng/Generate.ps1
```diff
@@ -194,7 +194,11 @@ if (!($Exclude -contains "TestProjects"))
             continue
         }
         if ($testName.EndsWith("Cadl")) {
-            Add-Cadl $testName $directory "" "--option @azure-tools/cadl-csharp.generate-convenience-methods=false"
+            if ($testName -eq "GenerateAllModels-Cadl") {
```
m-nash, line 197: I think we need to add a cadl-project.yaml file that can take any options.  Then we can generically read and append any number of options instead of specific implementations for every option.

The issue with letting cadl compile directly consume this is we are using a common cadl installation at the root of the repo and the cadl-project.yaml will be inside each test directory.  I opened [this issue](https://github.com/microsoft/cadl/issues/1633) as a future enhancement at which point we can remove our parsing of the cadl-project.yaml file but until then this is probably the best path forward.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -183,14 +190,15 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
         csharpGeneratorPath: resolvedOptions.csharpGeneratorPath,
         "clear-output-folder": resolvedOptions["clear-output-folder"],
         "save-inputs": resolvedOptions["save-inputs"],
-        "model-namespace": resolvedOptions["model-namespace"]
+        "model-namespace": resolvedOptions["model-namespace"],
+        "generate-all-models": resolvedOptions["generate-all-models"]
```
m-nash, line 194: how does this flag differ from this flag [here](https://github.com/Azure/azure-sdk-for-net/pull/34058/files#diff-e8ccddd90220fbc6a949079f96ecfac074affbeb698e4c1bb8a3f850445342ceR114).

eng/Generate.ps1
chunyu3: The cadl compiler does not load cadl-project.yaml. There is some problem in our test framework and we are working it. This is a workaround before we resolve the test framework issue.

pshao25: > The cadl compiler does not load cadl-project.yaml

This is fixed. Would you try again?

chunyu3:
I regen azure-sdk-for-net, and there is no change. https://github.com/chunyu3/azure-sdk-for-net/tree/generateAllModels

eng/Generate.ps1
chunyu3: I add cadl-project.yaml to configure the flag.

src/CADL.Extension/Emitter.Csharp/src/options.ts
```diff
@@ -18,6 +18,7 @@ export type NetEmitterOptions = {
     "clear-output-folder"?: boolean;
     "save-inputs"?: boolean;
     "model-namespace"?: boolean;
+    "generate-all-models"?: boolean;
```
pshao25, line 21: Seems not specify the default value.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -207,13 +211,17 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
...
 function deleteFile(filePath: string) {
-    fs.unlink(filePath, (err) => {
-        if (err) {
-            console.log(`stderr: ${err}`);
-        }
-
-        console.log(`File ${filePath} is deleted.`);
-    });
+    if (fs.existsSync(filePath)) {
+        console.log(`File ${filePath} does not exist. No deletion. `);
```
pshao25, line 215: I don't think we should print this message.

src/CADL.Extension/Emitter.Csharp/src/options.ts
chunyu3: added

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -254,12 +257,16 @@ export function isNeverType(type: Type): type is NeverType {
...
+    return type.kind === "EnumMember";
+}
+
 export function getInputType(
     program: Program,
     type: Type,
     models: Map<string, InputModelType>,
     enums: Map<string, InputEnumType>
-): InputType {
+): any {
```
pshao25, line 269: Why the return value is changed to any. Seems not a good practice.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -516,6 +524,8 @@ export function getInputType(
                     isReadOnly = true;
                 }
                 if (isNeverType(value.type) || isVoidType(value.type)) return;
+                if (isEnumMember(value.type)) {
+                }
```
pshao25, line 528: Looks need refactor.

pshao25:
I'm confused about how you define the model in "Generate all model". Will the models introduced from template be generated?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: We need to check if the file exists or not, then unlink it, otherwise unlink will give out an error.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: Yes. But this function will be used in callback of `navigateTypesInNamespace` which requires the signature of callback can return other types.

chunyu3:
> I'm confused about how you define the model in "Generate all model". Will the models introduced from template be generated?

Yes. all models will be generated even it is not used. This is the side effort of this. But when customer enables this, he expects to have all models and some of those are non-used.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -368,9 +381,15 @@ export function createModelForService(
         }
     }
 
-    const usages = getUsages(program, convenienceOperations);
-    setUsage(usages, modelMap);
-    setUsage(usages, enumMap);
+    if (generateAllModels) {
+        navigateModels(program, serviceNamespaceType, modelMap, enumMap);
+        setUsageForAll(Usage.RoundTrip, modelMap);
+        setUsageForAll(Usage.RoundTrip, enumMap);
```
ArcturusZhang, line 387: Should we use the flag of `RoundTrip` for those unused models? I believe `None` should be sufficient. By definition, they are not round trip models, and the serialization/deserialization makes no sense for models like that, because our generated serialization/deserilaization only serves the payload.
If they want serialization and deserialization, they could add customized code to have those.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -254,12 +257,16 @@ export function isNeverType(type: Type): type is NeverType {
     return type.kind === "Intrinsic" && type.name === "never";
 }
 
+export function isEnumMember(type: Type): boolean {
+    return type.kind === "EnumMember";
+}
```
ArcturusZhang, line 262: We have quite a few functions like this, is it possible that we could make this more strong typed?
Something like
```
export function isKind(type: Type, kind: TypeKind): boolean {
    return type.kind === kind;
}
```
and we define a new type `TypeKind` which collects all the possible values for kind

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
ArcturusZhang: Well we should find another way when putting this into the callback instead of compromising our own APIs.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
pshao25: We can check, but we don't have to `console.log`. It is strange to tell users that an "internal file is not exist, so we don't delete".

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -629,6 +637,27 @@ export function getInputType(
...
+        namespace,
+        {
+            model: (x) =>
+                x.name !== "" && x.kind === "Model" && computeModel(x),
+            scalar: computeModel,
+            enum: computeModel,
+            union: (x) => x.name !== undefined && computeModel(x)
+        },
+        { skipSubNamespaces }
+    );
```
ArcturusZhang, line 658: I believe this is the place that we are discussing in https://github.com/Azure/autorest.csharp/pull/2974/files#r1121103624
can we do this or similar to cast the return value of getInputType to any?
```suggestion
    const computeModel = (x: Type) => getInputType(program, x, models, enums) as any;
    const skipSubNamespaces = isGlobalNamespace(program, namespace);
    navigateTypesInNamespace(
        namespace,
        {
            model: (x) =>
                x.name !== "" && x.kind === "Model" && computeModel(x),
            scalar: computeModel,
            enum: computeModel,
            union: (x) => x.name !== undefined && computeModel(x)
        },
        { skipSubNamespaces }
    );
```
Or explicitly write the type of function `computeModel`? I do not know exactly what does the syntax looks like in typescript, only demo my point:
```suggestion
    const computeModel: Func<Type, any> = (x: Type) => getInputType(program, x, models, enums);
    const skipSubNamespaces = isGlobalNamespace(program, namespace);
    navigateTypesInNamespace(
        namespace,
        {
            model: (x) =>
                x.name !== "" && x.kind === "Model" && computeModel(x),
            scalar: computeModel,
            enum: computeModel,
            union: (x) => x.name !== undefined && computeModel(x)
        },
        { skipSubNamespaces }
    );
```

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
pshao25: emmm...... Can we wrap this method and in that wrapper method we could make it return any?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
pshao25: This method could return `type is EnumMember`

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: Thanks @ArcturusZhang  @pshao25 , I use casting.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: No. If it is `None`, the model will not be generated. And customer want to generate all the models together with serialize and deserialize method. Because they want to use those model in client code.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: Good suggestion. But now it is not enough to check the kind to identify if it is a special type. And now `isKind` cannot replace is***. And the usage of `isKind` is not high. We will plan to implement this in future if needed.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -629,6 +637,28 @@ export function getInputType(
     }
 }
 
+export function navigateModels(
```
lirenhe, line 640: Don't quite understand what this function mean, could you add more comments?

chunyu3: add comment for the functionality of the method.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
ArcturusZhang: The first part is not correct. We still generate the models and enums when its usage is None. Changed in this PR: https://github.com/Azure/autorest.csharp/pull/3046
Since you already ensures the config for `unreferenced-types-handling` hard-coded to `keepAll`, those models will be there.

For second part, it is a consideration of breaking change when growing up. If we generate it as RoundTrip, in some version, if they change their cadl to involve it in an operation, suddenly its usage will be changed to Input or Output.
The usage cannot be changed to a smaller one (RoundTrip => Output will have breaking changes, but Output => RoundTrip will not have breaking changes)

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -629,6 +637,29 @@ export function getInputType(
     }
 }
 
+/* Navigate all the models in the whole namespace. */
+export function navigateModels(
```
ArcturusZhang, line 641: it is lovely that we added this description here, but it is not really describing what this method is doing.
Could you be more specific on this? This method is navigating all the models in the whole namespace, I could assume it is doing this - but what is the purpose? What it is doing to navigate the models?
Looks like this is converting the models inside the namespace one by one and put it into the delegate `computeModel` and output it as an `InputType`?
Is this method filling up the two input `Map`s as output?

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -370,9 +381,15 @@ export function createModelForService(
         }
     }
 
-    const usages = getUsages(program, convenienceOperations);
-    setUsage(usages, modelMap);
-    setUsage(usages, enumMap);
+    if (generateAllModels) {
```
m-nash, line 384: I think we might have a conflict here with what `getAllSchemas` is trying to do in tcgc (typespec-client-generation-core).  Can we make sure we aren't introducing duplicate logic and if the logic in `getAllSchemas` isn't sufficient lets investigate how to make it so other languages can also take advantage of this.

test/TestProjects/GenerateAllModels-Cadl/cadl-project.yaml
```diff
@@ -0,0 +1,3 @@
+options:
+  "@azure-tools/cadl-csharp":
+    generate-all-models: true
```
cataggar, line 3: What is the meaning of `unreferenced-types-handling: keepAll`? Also asked this at https://github.com/Azure/autorest.csharp/issues/2960#issuecomment-1556692418.

Issue #2905, opened 2022-12-01
Support Page<T> in LRO response type in CADL


imakhlin:
I'm unable to generate the CSharp SDK using "@azure-tools/cadl-csharp" emitter when I use Page<T> as response type for an LRO POST action.
Below is an example CADL with LRO POST action that generates a pageable collection of file URIs.
```
  #suppress "@azure-tools/cadl-azure-core/use-standard-operations" "ADP data-plane API uses custom LRO response template the LRO actions."
  @doc("Initiates the process of generating SAS signed URIs for accessing the data-stream files.")
  @extension("x-ms-long-running-operation", true)
  @asyncOperationOptions("location")
  @pollingOperation(LongRunningOperationStatus)
  generate is CustomLongRunningResourceCollectionAction<
    DataStreamFile,
    {},
    Page<DataStreamFile>
  >;
```

Here is example of the output:
> dotnet --roll-forward Major {local-path}/specification/adp/DataManagement/node_modules/@autorest/csharp/AutoRest.CSharp.dll --project-path {local-path}/specification/adp/clients/DataManagement/Adp.DataManagement.Clients.v2022_11 --new-project
Unhandled exception. System.InvalidOperationException: Method GenerateDataStreamFiles has to have a return value
   at AutoRest.CSharp.Common.Output.Builders.ClientBuilder.BuildPagingMethod(String methodName, String nextLinkName, String itemName, RestClientMethod method, RestClientMethod nextPageMethod, TypeDeclarationOptions declaration) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Output\Builders\ClientBuilder.cs:line 112

As temporary workaround I just dropped the Page<T> wrapper and considering move to 204-NoContent response if this issue won't be addressed.

Reference: https://github.com/Azure/azure-rest-api-specs-pr/pull/9634/files#diff-cc086eb3452ab59e3b4bc639d9b016d1ae74932ab1e7f9410891da9f5e28bd18R144

lirenhe:
@imakhlin, I got the information that the Paged LRO definition was removed from the latest commit, so I would like to confirm whether this is still a problem for your service?  Thanks

https://github.com/Azure/azure-rest-api-specs-pr/blob/d1d0190ee459c556815771c7d1516a6cfc0c07aa/specification/adp/DataManagement/measurement/data-stream/operations.cadl#L141

imakhlin:
@lirenhe , yes, the problem persists.
[...]

PR #2975, opened 2022-12-30
feat(dpg): conveniece method for lro pagination operation

### LINKED ISSUE #2969: [DPG]: Convenience Method for Paging LRO

archerzz:
- change two `ProcessMessage` methods in `ProtocolOperationHelpers` to generic methods so that we can deal with model type as return value
- exclude lro pagination methods from normal pagination methods `LowLevelClient.PaginationMethod`
- add `PagingResponseInfo` in `ConvenienceMethod` record
- add implementation of generating convenicemence methods of LRO pagination operations i `LowLevelClientWriter`
- add test cases

resolve #2969, #2905


src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -682,6 +694,83 @@ private static void WritePageableLongRunningOperationMethodBody(CodeWriter write
...
+            using (WriteConvenienceMethodDeclaration(writer, clientMethod.ConvenienceMethod!.Signature, async))
+            {
+                // RequestContext context = FromCancellationToken(cancellationToken);
+                var contextVariable = new CodeWriterDeclaration(KnownParameters.RequestContext.Name);
+                WriteCancellationTokenToRequestContext(writer, contextVariable);
+
+                var convenienceMethod = clientMethod.ConvenienceMethod!;
+                var diagnostic = (convenienceMethod.Diagnostic != null ? convenienceMethod.Diagnostic : clientMethod.ProtocolMethodDiagnostic);
+
+                var createEnumerableMethodSignature = new MethodSignature("CreateEnumerable", null, null, None, new CSharpType(typeof(IEnumerable<>), new CSharpType(typeof(Page<>), clientMethod.ConvenienceMethod!.PagingResponseInfo!.ItemType)), null, new[] { ResponseParameter, NextLinkParameter, PageSizeHintParameter }).WithAsync(async);
```
ArcturusZhang, line 708: Since this new instance statement takes quite a few parameters, could we use positional parameters for each of the parameters here we are using? This will make the ctor invocation more clear, especially we are passing null for some parameters
```suggestion
                var createEnumerableMethodSignature = new MethodSignature(
                Name: "CreateEnumerable", 
                etc: null,
                etc: null,
                etc: None,
                etc: new CSharpType(typeof(IEnumerable<>),
                etc: new CSharpType(typeof(Page<>),
                etc: clientMethod.ConvenienceMethod!.PagingResponseInfo!.ItemType)),
                etc: null,
                Parameters: new[] { ResponseParameter, NextLinkParameter, PageSizeHintParameter }).WithAsync(async);
```

ArcturusZhang: And will we use the `ReturnType` in this created method signature? Because this method signature is using "open form" generic types (such as `typeof(Page<>)` instead of `typeof(Page<string>)` with a concrete type).
Where do we actually make the generic type closed?

archerzz: > And will we use the `ReturnType` in this created method signature? Because this method signature is using "open form" generic types (such as `typeof(Page<>)` instead of `typeof(Page<string>)` with a concrete type). Where do we actually make the generic type closed?

@ArcturusZhang It's used. See the generated code: https://github.com/Azure/autorest.csharp/pull/2975/files#diff-ebfa2940fc7dbefca5bdd3ce030063f23127a440fa5b893328e1d5c167a5ea47R283
```csharp
async IAsyncEnumerable<Page<Project>> CreateEnumerableAsync(Response response, string nextLink, int? pageSizeHint, [EnumeratorCancellation] CancellationToken cancellationToken = default)
```
The second parameter of `CSharpType` constructor is the class template argument type. So the return type is actually `typeof(Page<clientMethod.ConvenienceMethod!.PagingResponseInfo!.ItemType>)`

archerzz:
The codes are much simpler after Alexander's refactoring. However, it's still not decided how an LRO pageable operation is defined in CADL.

archerzz:
close this pr, we can resume when we need to support pageable lro

PR #2979, opened 2023-01-04
fix(dpg): lro sample should use `WaitUntil.Started`

### LINKED ISSUE #2978: [dpg]: lro protocol method sample should use `WaitUntil.Started`
## Problem

Right now, in the generated sample codes for LRO protocol methods, when we invoke the methods, we pass in`WaitUntil.Completed` which is not good because after that we invoke `Operation.WaitForCompletion()` which is not necessary if we pass  in`WaitUntil.Completed` .
This is because when we mock the value of `WaitUntil`, we use the first enum value which is `Completed`. See https://github.com/Azure/autorest.csharp/blob/29cd031d5f89644e997d667181b2452318215fff/src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs#L571

And
https://github.com/Azure/azure-sdk-for-net/blob/d771ba16ab7bc293e3f913324ba9e9b52cb2bd82/sdk/core/Azure.Core/src/WaitUntil.cs#L10-L20

I think we can make an exception here for `WaitUntil`: return `Started`.

archerzz:
- change `LowLevelExampleComposer` to use `WaitUntil.Started` when mocking method parameter values
- update samples/tests

fix #2978

# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs
+++ b/src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs
@@ -566,6 +566,12 @@ namespace AutoRest.CSharp.Generation.Writers
                     return "Guid.NewGuid()";
                 }
 
+                if (type == typeof(WaitUntil))
+                {
+                    // use `Started`, since we will generate `operation.WaitForCompletion()` afterwards
+                    return $"{nameof(WaitUntil)}.{nameof(WaitUntil.Started)}";
+                }
+
                 if (type.IsEnum)
                 {
                     return $"{type.Name}.{Enum.GetNames(type)[0]}";
--- a/samples/Azure.AI.DocumentTranslation/Generated/Docs/DocumentTranslationClient.xml
+++ b/samples/Azure.AI.DocumentTranslation/Generated/Docs/DocumentTranslationClient.xml
@@ -1119,7 +1119,7 @@ var data = new {
     },
 };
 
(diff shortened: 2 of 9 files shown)


ArcturusZhang (approved):

src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs
```diff
@@ -566,6 +566,12 @@ private string MockParameterTypeValue(string parameterName, CSharpType parameter
                     return "Guid.NewGuid()";
                 }
 
+                if (type == typeof(WaitUntil))
+                {
+                    // use `Started`, since we will generate `operation.WaitForCompletion()` afterwards
+                    return $"{nameof(WaitUntil)}.{nameof(WaitUntil.Started)}";
+                }
```
ArcturusZhang, line 573: Well, I suggest it would be better if we could change this huge if into a switch statement.

ArcturusZhang (approved):

src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs
archerzz: > Well, I suggest it would be better if we could change this huge if into a switch statement.

@ArcturusZhang surprisingly, `typeof` cannot be used in `switch...case`: https://stackoverflow.com/questions/1711005/why-doesnt-c-sharp-switch-statement-allow-using-typeof-gettype

Issue #2989, opened 2023-01-09
Rename paging containing model to {ResourceType}ListResult


pshao25:
Currently when using azure core template to generate paged containg model, it would be called as `PagedXXX` or `CustomPage`. We should change it to `XXXListResult`.

pshao25:
Latest code won't generate this mode, so close it.

PR #2990, opened 2023-01-09
Change paging model name to ListResult

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/2989


src/CADL.Extension/Emitter.Csharp/src/lib/templateHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Pagination model should have name {ResourceType}ListResult
+ */
+export function isPagingModel(type: Type) : boolean {
+    return type.kind === "Model" && hasDecorator(type, "$pagedResult");
+}
+
+export function getPagingModelName(model: Model) : string | undefined {
```
lirenhe, line 24: Will other SDKs also need this? How do you think of putting it in Cadl-dpg?

src/CADL.Extension/Emitter.Csharp/src/lib/templateHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (isPagingModel(model)) {
+        if (!model.templateArguments || model.templateArguments[0].kind !== "Model") {
+            throw "Paging model should have a containing model resource";
+        }
+        
+        return `${model.templateArguments[0].name}ListResult`;
```
lirenhe, line 30: Does this name (nameListResult) align with the naming of current HLC or mgmt. SDK?

src/CADL.Extension/Emitter.Csharp/src/lib/templateHandler.ts
pshao25: Others don't need it. It is .net specific.

src/CADL.Extension/Emitter.Csharp/src/lib/templateHandler.ts
pshao25: Yes, we are following the current naming convention.

pshao25:
Don't need that now.

PR #2994, opened 2023-01-10
Support Literal and union Type

chunyu3:
# Description

- emit Literal Type
- emit Union Type
- serialize Literal Type and Union Type

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -34,6 +34,8 @@ public TypeFactory(OutputLibrary library)
 
         public CSharpType CreateType(InputType inputType) => inputType switch
         {
+            InputLiteralType literalType       => CreateType(literalType.LiteralValueType), //TODO -- need to support literal type with the value.
```
archerzz, line 37: This is not the correct type returned for literal or union. Check the generated `Thing.cs` which doesn't follow what we've discussed.

test/TestProjects/FirstTest-Cadl/Generated/Models/Thing.cs
```diff
@@ -15,15 +15,25 @@ public partial class Thing
...
 
             Name = name;
+            RequiredUnion = requiredUnion;
+            RequiredLiteral = requiredLiteral;
         }
 
         /// <summary> Gets or sets the name. </summary>
         public string Name { get; set; }
+        /// <summary> Gets or sets the required union. </summary>
+        public string RequiredUnion { get; set; }
```
archerzz, line 35: I remember we've decided not to support union type property.

test/TestProjects/FirstTest-Cadl/Generated/Models/Thing.cs
```diff
@@ -15,15 +15,25 @@ public partial class Thing
...
+        /// <summary> Gets or sets the required literal. </summary>
+        public string RequiredLiteral { get; set; }
```
archerzz, line 37: This should be read-only with value initialized in constructor with literal value instead of passed-in parameter.

archerzz (changes requested):
Can we by-pass generation of model for now? The result is not correct. We just need to ensure the definition is correctly serialized by emitter. Also we may want to replace the union type property case with union parameter case.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
chunyu3: Codegen need to handle how to generate the literal or Union, there I just add workaround to pass generation.

test/TestProjects/FirstTest-Cadl/Generated/Models/Thing.cs
chunyu3: Yes, this is just to verify that emitter correct emit the type, after codegen support union, we will write another test, and remove here.

test/TestProjects/FirstTest-Cadl/Generated/Models/Thing.cs
chunyu3: Yes. this part will be implemented in codegen as we discussed.

archerzz (approved):

PR #2998, opened 2023-01-11
Fix several issues in pageable generation

AlexanderSher:
- `CancellationToken` is passed directly into pageable method where possible
- For MPG pageables, item JSON property name in lists is used directly (instead of conversion into C# property name).
- Handle pageables that are wrappers around collections

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
@@ -30,7 +30,13 @@ namespace AutoRest.CSharp.Generation.Writers
                     var messageVariable = new CodeWriterDeclaration("message");
                     var nextPageRequest = GetCreateRequestCall(restClientReference, createNextPageRequestMethod);
                     var nextPageRequestVariable = nextPageRequest != null ? new CodeWriterDeclaration("NextPageRequest") : null;
-                    var createPageableParameters = new List<FormattableString>()
+                    var parameters = methodSignature.Parameters;
+                    if (writer.EnsureRequestContextVariable(methodSignature, null, createNextPageRequestMethod))
+                    {
+                        parameters = parameters.Append(KnownParameters.RequestContext).ToList();
+                    }
+
+                    var createPageableParameters = new List<FormattableString>
                     {
                         $"{KnownParameters.WaitForCompletion.Name}",
                         $"{messageVariable:I}",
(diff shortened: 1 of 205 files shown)


src/AutoRest.CSharp/Mgmt/Decorator/MethodExtensions.cs
```diff
@@ -46,10 +46,9 @@ internal static bool IsListMethod(this RestClientMethod method, [MaybeNullWhen(f
             }
             else
             {
-                string pagingItemName = method.Operation.Paging?.ItemName ?? "value";
+                valuePropertyName = method.Operation.Paging?.ItemName ?? "value";
                 var schemaObject = (SchemaObjectType)returnType.Implementation;
-                itemType = GetValueProperty(schemaObject, pagingItemName)?.ValueType.Arguments.FirstOrDefault();
-                valuePropertyName = pagingItemName.ToCleanName();
```
AlexanderSher, line 52: This `ToCleanName` call has been causing the issue when `"Value"` item property name is generated where `"value"` is expected

ArcturusZhang: Previously this depends on the paged list result models in which has a property `Value` therefore we called `ToCleanName` here. Since our new helpers will parse the raw json to get the value of array, we no longer need this.

src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
```diff
@@ -118,23 +118,51 @@ public static CodeWriter WritePageableBody(this CodeWriter writer, IReadOnlyList
...
+            }
+        }
+
+        private static bool EnsureRequestContextVariable(this CodeWriter writer, MethodSignature methodSignature, RestClientMethod? createFirstPageRequestMethod, RestClientMethod? createNextPageRequestMethod)
         {
-            if (parameters.Contains(KnownParameters.RequestContext))
+            if (ContainsRequestContext(methodSignature.Parameters))
             {
-                return true;
+                return false;
```
archerzz, line 141: Why we changed to return `false` now? Previous codes:
> if (parameters.Contains(KnownParameters.RequestContext))
            {
                return true;
            }

archerzz (commented):
Generally, I'm OK with the changes. There are some SDKs being broken by the regression in pagination: https://github.com/Azure/azure-sdk-for-net/pull/33417
@AlexanderSher Can this PR fix those issues?

src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
AlexanderSher: The usage of the method has changed. It now returns `true` if `RequestContext` variable must be added to the list of parameters (as if it was a parameter in the method signature). Then it can be used in `CreateXxxRequest` methods and in pageable itself

m-nash (approved):

Issue #2944, opened 2022-12-14
CodeGen should flattern parameters defined with alias in CADL 


lirenhe:
Based on the email discussion, we should support the following in .NET DPG.

alias Foo = {
  id: string;
  name: string;
};
op M1(...Foo) : void;

// Same as

op M1(id: string, name: string): void;

lirenhe:
As this would impact the SDK for openAI, we should treat this as high priority.

chunyu3:
# with spread model which only contain one target

```cadl
model Foo {
  id: string;
  name: string;
}
alias FooAlias = {
  id: string;
  name: string;
};

op spreadWithModel(...Foo): void

op spreadWithAlias(...FooAlias): void
```

it will generate c# methods:
```
spreadWithModel(RequestContent content, RequestContext context = null);  //protocol method
spreadWithModel(Foo foo, CancellationToken cancellationToken = default); //convenience method
```

```
spreadWithAlias(RequestContent content, RequestContext context = null); //protocol method

spreadWithAlias(Foo foo, CancellationToken cancellationToken = default); //convenience method , option 1
or 
spreadWithAlias(string id, string name, CancellationToken cancellationToken = default); //convenience method, option2
```
For the convenience method, shall we flatten all the body parameters or not (option1 or option2)?

chunyu3:
# with spread model/alias  which contains multiple targets

## model/alias only contain primitive type property

```
model Foo {
  @path id: string;
  id: string;
  name: string;
}

alias FooAlias = {
  @path id: string;
  id: string;
  name: string;
};
```

chunyu3:
If we flatten body model as method paramters, we need to update the convenience method body to combine those flattened parameters to body json object.

There are two approach: 

1. Generate the body model ( as internal), and in convenience method, create the model and call toRequestConent to seraialize. - it may meet `anonymous model` issue.
```
internal partial class FooAlias 
    {
        /// <summary> Initializes a new instance of Friend. </summary>
        /// <param name="name"></param>
        /// <exception cref="ArgumentNullException"> <paramref name="name"/> is null. </exception>
        public FooAlias(string name, int32 age)
        {

            Name = name;
             Age = age;
        }

        /// <summary> Gets or sets the name. </summary>
        public string Name { get; set; }
        public int32 Age { get; set; }
    }
}
public virtual Response spreadWithAlias(string name, Int32 age, CancellationToken cancellationToken = default)
{
   ......
[...]

PR #3006, opened 2023-01-13
Spread alias

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/2944

- flatten alias as parameters for convenience method

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -262,10 +262,34 @@ private void WriteSubClientInternalConstructor(ConstructorSignature signature)
             _writer.Line();
         }
 
+        private void DeclareMethodParameter(CodeWriter writer, ConvenienceMethod convenienceMethod)
+        {
+            foreach (var parameterChain in convenienceMethod.ProtocolToConvenienceParameters)
+            {
+                if (parameterChain.Input?.Kind == InputOperationParameterKind.Flatten)
+                {
+                    var typename = parameterChain.Convenience?.Type.Name;
```
ArcturusZhang, line 271: nit
```suggestion
                    var typeName = parameterChain.Convenience?.Type.Name;
```

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -19,6 +19,7 @@
 using Azure.Core;
 using static AutoRest.CSharp.Output.Models.MethodSignatureModifiers;
 using Configuration = AutoRest.CSharp.Input.Configuration;
+using static System.Net.WebRequestMethods;
```
ArcturusZhang, line 22: where would we use this?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -196,16 +197,35 @@ private ConvenienceMethod BuildConvenienceMethod(ReturnTypeChain returnTypeChain
...
                 ? new[] { new CSharpAttribute(typeof(ObsoleteAttribute), deprecated) }
                 : null;
             var protocolToConvenience = _orderedParameters
                 .Where(p => p.Protocol != null)
-                .Select(p => (p.Protocol!, p.Convenience))
+                .Select(p => (p.Protocol!, p.Convenience, p.Input))
                 .ToArray();
-
-            var convenienceSignature = new MethodSignature(name, _restClientMethod.Summary, _restClientMethod.Description, _restClientMethod.Accessibility | Virtual, returnTypeChain.Convenience, null, parameters, attributes);
+            //var parameters = parameterList.ToArray();
```
ArcturusZhang, line 227: we should remove this line

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -196,16 +197,35 @@ private ConvenienceMethod BuildConvenienceMethod(ReturnTypeChain returnTypeChain
...
+            var convenienceSignature = new MethodSignature(name, _restClientMethod.Summary, _restClientMethod.Description, _restClientMethod.Accessibility | Virtual, returnTypeChain.Convenience, null, parameterList.ToArray(), attributes);
```
ArcturusZhang, line 228: ToArray here is kind of redundant because List implements IReadOnlyList
```suggestion
            var convenienceSignature = new MethodSignature(name, _restClientMethod.Summary, _restClientMethod.Description, _restClientMethod.Accessibility | Virtual, returnTypeChain.Convenience, null, parameterList, attributes);
```

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
```diff
@@ -34,8 +34,21 @@ model Thing {
   age: int32;
 }
 
+alias ThingAlias = {
+  @doc("name of the Thing")
+  name: string;
+  @doc("age of the Thing")
+  age: int32;
+};
```
ArcturusZhang, line 42: Could we create an alias that has a mixture of @path @header and body parameters?
For instance, add another property with `@path` decorator to show if that part could work properly

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
chunyu3: removed.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
chunyu3: resolved

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
chunyu3: removed

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
chunyu3: add a new test with multiple targets.

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
```diff
@@ -34,8 +34,48 @@ model Thing {
...
+
+alias ThingAliasWithModel = {
+  @path id: string;
+  ...Thing;
+  @header top: int32;
+};
```
ArcturusZhang, line 48: Could we also have a case exactly like this but the `Thing` here is another alias? Something like
```
alias Alias1 = {
   @path id: string;
   ...Alias2;
   @header top: int32;
};
```
We should expect all the properties in Alias1 and Alias2 are parameters of the same method signature.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -262,10 +262,34 @@ private void WriteSubClientInternalConstructor(ConstructorSignature signature)
...
+        private void DeclareMethodParameter(CodeWriter writer, ConvenienceMethod convenienceMethod)
+        {
+            foreach (var parameterChain in convenienceMethod.ProtocolToConvenienceParameters)
+            {
+                if (parameterChain.Input?.Kind == InputOperationParameterKind.Flatten)
+                {
+                    var typeName = parameterChain.Convenience?.Type.Name;
+                    var paraName = parameterChain.Convenience?.Name;
+                    writer.Append($"{typeName} {paraName:D} = ");
+                    writer.Append($"new {typeName}(");
```
ArcturusZhang, line 274: I believe we need some assertion to make sure these are not null.
And BTW we could write these like this to ensure the namespace:
```suggestion
                    var type = parameterChain.Convenience?.Type;
                    var paraName = parameterChain.Convenience?.Name;
                    writer.Append($"{type} {paraName:D} = ");
                    writer.Append($"new {type}(");
```

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -262,10 +262,34 @@ private void WriteSubClientInternalConstructor(ConstructorSignature signature)
...
+                    InputType? type = parameterChain.Input?.Type?? null;
```
ArcturusZhang, line 275: we need a format here

src/CADL.Extension/Emitter.Csharp/src/type/InputParameter.ts
```diff
@@ -27,4 +27,5 @@ export interface InputParameter {
     Explode: boolean;
     ArraySerializationDelimiter?: string;
     HeaderCollectionPrefix?: string;
+    SpreadSource?: string;
```
ArcturusZhang, line 30: Do we need to update the models in the `Input` namespace accordingly to this change?

ArcturusZhang (approved):

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
chunyu3: Yes. This is possible scenario. I will add one. Thanks

src/CADL.Extension/Emitter.Csharp/src/type/InputParameter.ts
chunyu3: I remove this now. It is for spread model with multiple target.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -262,10 +262,34 @@ private void WriteSubClientInternalConstructor(ConstructorSignature signature)
...
+                        foreach (var prop in modelType.Properties)
+                        {
+                            writer.Append($"{prop.Name},");
+                        }
+                    }
+                    writer.RemoveTrailingComma();
+                    writer.Line($");");
+                }
+            }
+        }
```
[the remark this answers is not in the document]
chunyu3: Do you mean to add a empty line in the generated sdk? I think it is not needed.

src/AutoRest.CSharp/Common/Input/Inputs.cs
```diff
@@ -216,7 +216,8 @@ internal enum InputOperationParameterKind
         Client = 1,
         Constant = 2,
         Flattened = 3,
-        Grouped = 4,
+        Flatten = 4,
```
lirenhe, line 219: what is the difference between Flattened = 3,  and Flatten = 4,

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
```diff
@@ -34,8 +34,60 @@ model Thing {
...
 @convenienceAPI
-op spreadModel( ...Thing ): void;
\ No newline at end of file
+op spreadModel( ...Thing ): void;
+
+@route("/spreadAlias")
+@doc("body parameter without body decorator")
+@post
+@convenienceAPI
+op spreadAlias( ...ThingAlias ): void;
```
lirenhe, line 75: Is this a valid case?  

op spreadAlias( ...ThingAlias, ...ThingMultiTargetAlias): void;

lirenhe: This may impact how you name the internal model for alias.

chunyu3: I think the behavior of `op spreadAlias( ...ThingAlias, ...ThingMultiTargetAlias): void; ` is the same as `op spreadAlias( ...ThingAlias ): void`

src/AutoRest.CSharp/Common/Input/Inputs.cs
lirenhe: Is it better to add some comments?  Those 2 names are quite confused.

lirenhe (approved):
Left comment on one minor issue. other parts look good.

src/AutoRest.CSharp/Common/Input/Inputs.cs
pshao25: Why the newly-added optional is not at the end? Isn't this break anything?

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
```diff
@@ -34,8 +34,60 @@ model Thing {
...
 @route("/spreadModel")
 @doc("body parameter without body decorator")
 @post
 @convenienceAPI
-op spreadModel( ...Thing ): void;
\ No newline at end of file
+op spreadModel( ...Thing ): void;
+
+@route("/spreadAlias")
+@doc("body parameter without body decorator")
```
pshao25, line 72: The docs of these operations are not reflecting the expectation.

test/AutoRest.TestServer.Tests/SpreadClientTest.cs
```diff
@@ -35,5 +35,146 @@ public void SpreadModelMethod()
                 Assert.AreEqual("Thing", parameters[0].ParameterType.Name);
             }
         }
+
```
pshao25, line 38: My understanding is this impacts the payload. So we need E2E tests. This is an example: https://github.com/Azure/autorest.csharp/blob/d63b897ea987385356d71e225fa38db9e146e8bd/test/CadlRanchProjects.Tests/LroBasic-Cadl.cs

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
```diff
@@ -2,11 +2,12 @@
...
 using System.Collections.Generic;
+using AutoRest.CSharp.Common.Input;
 using AutoRest.CSharp.Generation.Types;
 using AutoRest.CSharp.Output.Models.Requests;
 using AutoRest.CSharp.Output.Models.Shared;
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal record ConvenienceMethod(MethodSignature Signature, IReadOnlyList<(Parameter Protocol, Parameter? Convenience)> ProtocolToConvenienceParameters, CSharpType? ResponseType, Diagnostic? Diagnostic);
+    internal record ConvenienceMethod(MethodSignature Signature, IReadOnlyList<(Parameter Protocol, Parameter? Convenience, InputParameter? Input)> ProtocolToConvenienceParameters, CSharpType? ResponseType, Diagnostic? Diagnostic);
```
pshao25, line 12: `ConvenienceMethod` should be decoupled with the information from code model.

chunyu3: `Parameter` is just a light present of a model, it just retrieve out the type name, no detail schema information.  In order to support flatten alias to parameters, and combine the parameters as a request content, we need the detail schema information. If we add those into Parameter, then Paramter will be almost the same as `InputParameter`.
Yes, I agree it is better to decouple with input type in the ConvenienceMethod, but now, we will remain the current design of `Parameter` and may think detail about how to design Parameter and decouple this later.

test/AutoRest.TestServer.Tests/SpreadClientTest.cs
chunyu3: hello @pshao25 Yes, the request body payload is important, but now there is no way to get the internal request body payload for test. Do you know any available way?

pshao25: 1. Writing mock api which reflects your cadl under test\CadlRanchMockApis\src.
2. Writing unit tests under test\CadlRanchProjects.Tests, LroBasic-Cadl.cs is an example.

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
chunyu3: updated

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
pshao25: I don't mean to change `Parameter`. I'm saying we decouple `InputParameter` from `ConvenienceMethod`. Seems only type information of `InputParameter` is used? So we could only store the type information to `ConvenienceMethod`?

src/AutoRest.CSharp/Common/Input/Inputs.cs
ArcturusZhang: > Why the newly-added optional is not at the end? Isn't this break anything?

This is a internally used thing, we break ourselves therefore it is fine.
And in fact adding anything into an existing enum is a breaking change.

ArcturusZhang: I strongly suggest we change a name for this. Do not use the word `Flatten` here, it is really confusing.
I believe adding comment is only a workaround, if possible we should avoid any confusion on our code.
How about `Spreaded`?

test/AutoRest.TestServer.Tests/SpreadClientTest.cs
chunyu3: Thanks @pshao25 I added the CadlRanch test.

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
chunyu3: Except the `InputType`, the `Kind` is also needed. And if we add `InputType` and `Kind` into the parameter chain, it is not better to add `InputParameter` directory. Both `InputType` and `Kind` are the in InputModel serial, also

pshao25 (approved):

src/AutoRest.CSharp/Common/Input/Inputs.cs
chunyu3: @ArcturusZhang How about `Spread`? `Spreaded` means the parameter is spreaded from other parameter. `Spread` can mean the parameter is to be spreaded.

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Input/Inputs.cs
ArcturusZhang: I think Spread is a good idea

Issue #2936, opened 2022-12-12
Support the union of types in alias definition in CADL


lirenhe:
in https://github.com/Azure/azure-rest-api-specs/blob/feature/cognitiveservices/openai-cadl/specification/cognitiveservices/OpenAI.Inference/models/embeddings.create.cadl#L22, we have the model defined as 

input: string | string[] | integer[] | integer[][];   in the follow model.

```
alias EmbeddingsRequest = {
    @doc("The ID of the end-user, for use in tracking and rate-limiting.")
    user?: string;

    @doc("input type of embedding search to use")
    input_type?: string;

    @doc("ID of the model to use")
    "model"?: string;

    @doc("""
    An input to embed, encoded as a string, a list of strings, or a list of token
    lists
    """)
    input: string | string[] | integer[] | integer[][];
};
```
We need to support union of types for a property in model definition in CADL

This concept is similar to AnyOf in openapi 3 so I believe this is a valid case.

chunyu3:
possible solution

Union.cs
```
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

using System;
using System.Collections.Generic;
using System.Text;

namespace AutoRest.CSharp.Common.Input
{
    internal partial class Union<T1, T2>
    {
        private T1? _v1;
        private T2? _v2;

        Union(T1 v1)
        {
            _v1= v1;
        }

        Union(T2 v2)
        {
            _v2= v2;
        }
    }
}


//Union<T1, T2, T3>
//Union<T1, T2, T3, T4>
//Union<T1, T2, T3, T4, T5>

```

Union.Serialization.cs
```
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

using System;
using System.Collections.Generic;
using System.Text;
using System.Text.Json;
using Azure.Core;

namespace AutoRest.CSharp.Common.Input
{
    internal partial class Union<T1, T2>: IUtf8JsonSerializable
    {
        void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
        {
            if (_v1!= null)
            {

            } else if(_v2!= null)
            {

            }

            if (_v1 == null && _v2 == null) return;

            switch (_v1 ?? _v2 )
            {
                case IUtf8JsonSerializable serializedModel:
                    serializedModel.Write(writer);
                    break;
                case Int32 int32V:
                    writer.WriteNumberValue((int32V);
                    break;
                case string str:
[...]

PR #3008, opened 2023-01-13
Union support

ArcturusZhang:
Depends on: https://github.com/Azure/autorest.csharp/pull/3025
Fixes https://github.com/Azure/autorest.csharp/issues/2936

# Description

Introduction:
This PR introduces the implementation of union types.
This PR separates the implementation of "flatten spread in cadl" and union types in models. This PR only focuses on union types in models. The emitter treats aliases very similar with models just with an extra flag `Flatten` to let the generator now when generating the method signature, we need to do some extra work (it may or may not be included in this PR, depending on the process of [the spread PR](https://github.com/Azure/autorest.csharp/pull/3006)).

We should always have a solution for the models with union type properties. Once we have this, the spread flatten mechanism just needs to know how to properly flatten a model with union types (by creating method overload).

## Design of models with union types

comparsion: [java's design](https://github.com/Azure/autorest.java/issues/1891#issuecomment-1381285183)

If we have this cadl:
```
model Thing {
    name: string
    input: string | int
}
```
this will generate 3 classes in dotnet SDK code:
```
public abstract class Thing {
}
public class StringThing : Thing {
    public string Input { get; set; }
}
public class IntThing : Thing {
    public int Input { get; set; }
}
```
In this design, we will automatically get the serialization of these models because serialization works on an instance method.
We still need more discussion on the deserialization part. But this PR will not cover that because our current plan is to "only support the union types in input".

Because this design introduces extra inheritance layers, there are a few questions to be answered. I will list a few of them below
1. When the alias contains optional union, for instance
```
alias AliasWithOptionalUnion = {
  name: string;
  input: string | int;
};
op Do( ...AliasWithOptionalUnion ): void;
```
will generate methods like
```
public Response Do(string name, string input = null, CancellationToekn cancellationToekn = default);
public Response Do(string name, int input = null, CancellationToekn cancellationToekn = default);
```
This will be ambiguous when calling it without a value for input (compile error).
2. what should we do if another model is extending this model with union types?
```
model Thing {
    name: string
    input: string | int
}
model DetailedThing extends Thing {
    details: string[]
}
```
3. what should we treat the relationship between the union expanded classes with discriminator?

# Checklist

# Ready to Land?


test/TestProjects/FirstTest-Cadl/FirstTest-Cadl.cadl
```diff
@@ -133,6 +138,12 @@ op patchAction(@body body: Thing): Thing;
 @convenienceAPI
 op anonymousBody( ...Thing ): Thing;
 
+@route("/aliasBody")
+@doc("body parameter without body decorator")
+@post
+@convenienceAPI
+op aliasBody( ...ThingAlias ): Thing;
+
```
lirenhe, line 146: One question related to current implementation:

If we have 
op alias2Body( ...ThingAlias ): Thing;
op alias3Body( ...ThingAlias ): Thing;

It seems we will create Alias2BodyRequest，  Alias3BodyRequest with all their derived classes.
Do we have a way to combine them as all of them are mapping to '...ThingAlias'

ArcturusZhang: That depends on how emitter will output the body parameter for this scenario.
And it does not matter because those models for aliases will always be internal.

pshao25:
I would suggest:
1. Enable 'unions' test in cadl-ranch
2. Remove the duplicate scenario in FirstTest-Cadl, because it is hard to tell a union test now.
3. If there are more tests not covered in cadl-ranch and they are spread related, lets put them into SpreadTest-Cadl.

ArcturusZhang:
Close this PR for now since we did not really decide a conclusion on how to handle the union types

PR #3021, opened 2023-01-18
Upgrade projects to net6.0

heaths:
Use latest LTS and build against net7.0 SDK, which azure-sdk-for-net is now using.


heaths (commented):
@m-nash's changes LGTM. Thanks!

src/assets/TestFramework/ClientDiagnosticListener.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Azure.Core.Tests
 {
-    public class ClientDiagnosticListener : IObserver<KeyValuePair<string, object>>, IObserver<DiagnosticListener>, IDisposable
+    public class ClientDiagnosticListener : IObserver<KeyValuePair<string, object?>>, IObserver<DiagnosticListener>, IDisposable
```
AlexanderSher, line 13: Why do we need this change? This file should be a copy of https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.TestFramework/src/ClientDiagnosticListener.cs

heaths: Because the `OnNext` method has to be declared `public void OnNext(KeyValuePair<string, object?> value)`, so the interface has to match. And something in the body was saying the `Value` (the second tparam) could be null. There was a cascading effect.

As I told @m-nash yesterday, I tried to actual fix as many nullable violations as possible, but it's clear the code wasn't really utilizing it to the fullets. Even though expressions like `var elem = collection.FirstOrDefault(); return elem.Property;` definitely should've been caught before, they weren't. This might have to do with upgrading from netcoreapp3.1 to net6.0 (the new LTS) which had more nullable annotations. Either way, there were a lot of places where the code wasn't checking for null regardless of annoations.

We can discuss better fixes for the code at hand, but what's here should be no worse than what was there before. We need to get autorest.csharp unblocked.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeProperty.cs
```diff
@@ -118,7 +118,7 @@ internal string CreateExtraDescriptionWithManagedServiceIdentity()
         {
             var extraDescription = string.Empty;
             var originalObjSchema = SchemaProperty?.Schema as ObjectSchema;
-            var identityTypeSchema = originalObjSchema?.GetAllProperties().FirstOrDefault(p => p.SerializedName == "type").Schema;
+            var identityTypeSchema = originalObjSchema?.GetAllProperties()?.FirstOrDefault(p => p.SerializedName == "type")?.Schema;
```
m-nash, line 121: Lets use `!` here instead of `?` so that if these are ever null the error is thrown from the same spot as before

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -440,12 +440,17 @@ private bool TryGetPropertyForInputModelProperty(Func<ObjectTypeProperty, bool>
                 discriminatorPropertyName = parent.Discriminator.SerializedName;
                 property = parent.Discriminator.Property;
             }
-            else
+            else if (_derivedTypes != null)
```
m-nash, line 443: lets let the null ref happen like before IF it ever does so that its obvious where the issue is

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -322,7 +322,7 @@ public static string ToPlural(this string single, bool inputIsKnownToBeSingular
         public static string LastWordToPlural(this string single, bool inputIsKnownToBeSingular = true)
         {
             var words = single.SplitByCamelCase();
-            var lastWord = words.LastOrDefault();
+            var lastWord = words.Last();
```
m-nash, line 325: I think we want to leave LastOrDefault here, we never dereference `lastWord` and the comparison below should work with a null string fine.

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -346,7 +346,7 @@ public static string ToSingular(this string plural, bool inputIsKnownToBePlural
         public static string LastWordToSingular(this string plural, bool inputIsKnownToBePlural = true)
         {
             var words = plural.SplitByCamelCase();
-            var lastWord = words.LastOrDefault();
+            var lastWord = words.Last();
```
m-nash, line 349: I think we want to leave LastOrDefault here, we never dereference `lastWord` and the comparison below should work with a null string fine.

m-nash (approved):

Issue #3023, opened 2023-01-19
Handle optional properties when flatten an alias


chunyu3:
When there are optional properties in an alias, should flatten these optional properties as nuallable parameters.
e.g.
```
alias ThingAliasWithOptionalProps = {
  @path id: string;
  @doc("name of the Thing")
  name: string;
  @doc("optional property of the Thing")
  color?: string;
  @doc("age of the Thing")
  age?: int32;
  @header top: int32;
};
op spreadAliasWithOptionalProps( ...ThingAliasWithOptionalProps ): void;
```

The expected generated sdk
```
public virtual Response SpreadAliasWithOptionalProps(string id, int top, string name, string color, int? age, CancellationToken cancellationToken = default)
```

ArcturusZhang:
I believe optional properties should be flattened into optional parameters in the method signature. For instance, the cadl
```
alias ThingAliasWithOptionalProps = {
  @path id: string;
  @doc("name of the Thing")
  name: string;
  @doc("optional property of the Thing")
  color?: string;
  @doc("age of the Thing")
  age?: int32;
  @header top: int32;
};
op spreadAliasWithOptionalProps( ...ThingAliasWithOptionalProps ): void;
```
should be generated as
```
public virtual Response SpreadAliasWithOptionalProps(string id, int top, string name, string color = null, int? age = null, CancellationToken cancellationToken = default)
```
So that the users will not required to pass in a `null` value for those optional properties. But this might introduce issues around overloads of methods, or issues of breaking changes when new properties are added into the alias because we will have to sort the parameters

PR #3025, opened 2023-01-19
fix the types and initializations of the spreaded parameters

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3023
Regen: nothing changed

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
```diff
@@ -6,8 +6,13 @@
 using AutoRest.CSharp.Generation.Types;
 using AutoRest.CSharp.Output.Models.Requests;
 using AutoRest.CSharp.Output.Models.Shared;
+using AutoRest.CSharp.Output.Models.Types;
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal record ConvenienceMethod(MethodSignature Signature, IReadOnlyList<(Parameter Protocol, Parameter? Convenience, InputParameter? Input)> ProtocolToConvenienceParameters, CSharpType? ResponseType, Diagnostic? Diagnostic);
+    internal record ConvenienceMethod(MethodSignature Signature, IReadOnlyList<ProtocolToConvenienceParameterConverter> ProtocolToConvenienceParameters, CSharpType? ResponseType, Diagnostic? Diagnostic);
```
archerzz, line 13:
```suggestion
    internal record ConvenienceMethod(MethodSignature Signature, IReadOnlyList<ProtocolToConvenienceParameterConverter> ProtocolToConvenienceParameterConverters, CSharpType? ResponseType, Diagnostic? Diagnostic);
```

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -464,6 +472,6 @@ private ReferenceOrConstant CreateReference(InputParameter? operationParameter,
 
         private record ReturnTypeChain(CSharpType Convenience, CSharpType Protocol, CSharpType? ConvenienceResponseType);
 
-        private record ParameterChain(InputParameter? Input, Parameter? Convenience, Parameter? Protocol, Parameter? CreateMessage);
+        private record ParameterChain(Parameter? Convenience, Parameter? Protocol, Parameter? CreateMessage, bool IsSpreadParameter = false);
```
chunyu3, line 475: ParameterChain contains parameters in different usage senarios, it may be a little confuse to add a `Kind` in such a chain.

ArcturusZhang: I did not add a `Kind`. I just spread some property you might need later in this class to decouple the Input out of this

ArcturusZhang: You could regard this as a follow up change on this comment: https://github.com/Azure/autorest.csharp/pull/3006#discussion_r1071133460

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
```diff
@@ -36,7 +36,9 @@ alias ThingAlias = {
   @doc("name of the Thing")
   name: string;
   @doc("age of the Thing")
-  age: int32;
+  age?: int32;
```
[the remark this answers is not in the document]
ArcturusZhang: I copied the test cases you added in your PR here

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
ArcturusZhang: resolved

archerzz (approved):

test/TestProjects/Spread-Cadl/Generated/SpreadClient.cs
```diff
@@ -496,6 +498,134 @@ public virtual Response SpreadAliasWithSpreadAlias(string id, int top, RequestCo
...
+        /// <param name="top"> The Int32 to use. </param>
+        /// <param name="name"></param>
+        /// <param name="items"></param>
+        /// <param name="color"></param>
+        /// <param name="age"></param>
+        /// <param name="elements"></param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="id"/>, <paramref name="name"/> or <paramref name="items"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="id"/> is an empty string, and was expected to be non-empty. </exception>
+        public virtual async Task<Response> SpreadAliasWithOptionalPropsAsync(string id, int top, string name, IEnumerable<int> items, string color = null, int? age = null, IEnumerable<string> elements = null, CancellationToken cancellationToken = default)
```
pshao25, line 512: What is the order of required parameters? i.e. why top is before name?

test/TestProjects/Spread-Cadl/Generated/Models/SpreadAliasWithOptionalPropsRequest.Serialization.cs
```diff
@@ -0,0 +1,65 @@
...
+                if (Age != null)
+                {
+                    writer.WritePropertyName("age");
+                    writer.WriteNumberValue(Age.Value);
+                }
+                else
+                {
+                    writer.WriteNull("age");
+                }
+            }
```
pshao25, line 36: Both color and age are optional, why they have different behaviors? i.e., why there is no null check insider Color block.

test/TestProjects/Spread-Cadl/Generated/SpreadClient.cs
ArcturusZhang: This logic exists before. Order of parameters:
Required path -> Required query -> body -> optional query (we cannot have an optional path therefore it is skipped here)

In this PR, we spread body into a list of parameters, they replace body in the above list.
Therefore
id (required path), top (required query), name (required spread body), items (required spread body), others (optional spread body)

test/TestProjects/Spread-Cadl/Generated/Models/SpreadAliasWithOptionalPropsRequest.Serialization.cs
ArcturusZhang: This is not something I introduced in this PR.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
chunyu3: Sorry. I mean it is a little confuse to add a bool `IsSpreadParameter` to a parameter chain.

ArcturusZhang: We cannot add it to the `Parameter` because it stands for a parameter of dotnet method in which "spread" or not makes no sense.
We do not want things in `Input` to couple with other things, therefore this is the only option I could think of.
Speaking of confusion, I believe the name of this record is more confusing.

chunyu3: Ok. We may refine the name later.

chunyu3 (approved):

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -262,38 +262,34 @@ private void WriteSubClientInternalConstructor(ConstructorSignature signature)
             _writer.Line();
         }
 
-        private void DeclareMethodParameter(CodeWriter writer, ConvenienceMethod convenienceMethod)
+        private void WriteSpreadBodyConverter(CodeWriter writer, ConvenienceMethod convenienceMethod, FormattableString? spreadVariable)
```
m-nash, line 265: The writer should have no idea what a spread is.  The writer should simply get a list of things to write, this feels like we need to move some logic back into the output types OR perhaps even input if this information can be known at translation time between cadl and cadl.json

ArcturusZhang: I believe we kind of have to have this somewhere because "convert the spread parameter back into a body model" is not always done by a one-line command (especially when there are arrays in them, you could find one inside our test project), it is actually a block of code.
We could consider to add a delegate inside `ConvenienceMethod` data record so that it could know how to write this thing, and the writer could call it if there is one, in this way, the writer does not have to know about anything about "spread parameters". Is this OK?

ArcturusZhang: And we should not put this inside the emitter, because eventually we still need to write the part about "constructing body model back" (this way) or "adhoc serialize the body parameter from the spread parameters"
In addition, if we put the "spread parameter logic" in emitter, it will force us to put the logic related with unions (spread into overload of convenience method) also in the emitter which makes the code in emitter a lot complicated than it is right now, and we agree that we should not do that to make the emitter complicated.

test/TestProjects/Spread-Cadl/Spread-Cadl.cadl
```diff
@@ -78,25 +78,25 @@ alias ThingAliasWithOptionalProps = {
 @route("/spreadModel")
 @doc("spread a model as body.")
 @post
-@convenientAPI(true)
+@convenienceAPI
```
m-nash, line 81: this is the old decorator which we shouldn't use anymore we need to remove support for this

test/TestProjects/Spread-Cadl/Generated/Models/SpreadAliasWithOptionalPropsRequest.Serialization.cs
```diff
@@ -15,26 +15,26 @@ internal partial class SpreadAliasWithOptionalPropsRequest : IUtf8JsonSerializab
...
-                writer.WritePropertyName("color");
+                writer.WritePropertyName("color"u8);
                 writer.WriteStringValue(Color);
             }
             if (Optional.IsDefined(Age))
             {
                 if (Age != null)
                 {
-                    writer.WritePropertyName("age");
+                    writer.WritePropertyName("age"u8);
```
m-nash, line 29: what is the `u8` at the end here?

m-nash:
Can we get the regen PR added?

ArcturusZhang:
> Can we get the regen PR added?

Nothing changes in the SDK repo

m-nash (approved):

PR #3036, opened 2023-01-28
feat(cadl): support literal type model property

### LINKED ISSUE #3035: [Feat]: Support Literal Type Model Property
## Description

Part of #2972  For CADL model with literal property like below:

```typescript
model Foo {
  name: string;
  type: "accepted"
}
```

We will generate the model codes like:
```csharp
public class Foo
{
        internal Foo(string name)
        {
            Argument.AssertNotNull(name, nameof(name));

            Name = name;
            Type = "accepted";
        }

        /// <summary> Gets the name. </summary>
        public string Name { get; }
        /// <summary> Gets the age. </summary>
        internal string Type { get; }
    }

    void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
        {
            writer.WriteStartObject();
            writer.WritePropertyName("name");
            writer.WriteStringValue(Name);
             writer.WritePropertyName("type");
            writer.WriteStringValue(Type);
            writer.WriteEndObject();
        }
        internal static Foo DeserializeFoo(JsonElement element)
        {
            string name = default;
            foreach (var property in element.EnumerateObject())
            {
                if (property.NameEquals("name"))
                {
                    name = property.Value.GetString();
                    continue;
                }
            }
            return new Foo(name);
        }

        /// <summary> Deserializes the model from a raw response. </summary>
        /// <param name="response"> The response to deserialize the model from. </param>
        internal static Foo FromResponse(Response response)
        {
            using var document = JsonDocument.Parse(response.Content);
            return DeserializeThing(document.RootElement);
        }
}
```

- the property doesn't show up in the constructor parameter list, but it's initialized in constructor
- the property is read-only
- the property is internal

archerzz:
- change json coverter for `InputModelProperty` to set the `DefaultValue` property if the property type is a literal type
  - update `ObjectTypeProperty` so that `DefaultValue` is set if it's from literal type
- change `ModelTypeProviderFields`:
  - exclude literal properties from constructor parameter list
  - literal properties are `internal readonly`
- change `ModelTypeProvider`, update the criteria on properties to be serialized, now literal type property will be serialized regardless it's read-only
- fix a issue in emitter that boolean literal type is not identified
- minor refactoring
- update test projects

resolve #3035

# Description

# Checklist

# Ready to Land?


test/TestProjects/FirstTest-Cadl/Generated/Models/Thing.cs
```diff
@@ -16,24 +16,27 @@ public partial class Thing
...
         }
 
         /// <summary> Gets or sets the name. </summary>
         public string Name { get; set; }
         /// <summary> Gets or sets the required union. </summary>
         public string RequiredUnion { get; set; }
-        /// <summary> Gets or sets the required literal. </summary>
-        public string RequiredLiteral { get; set; }
+        /// <summary> Gets the required literal string. </summary>
+        internal string RequiredLiteralString { get; } = "accept";
```
ArcturusZhang, line 34: well in mgmt plane or hlc, this is implemented inside the constructor using "propertyInitializer" like this: https://github.com/Azure/autorest.csharp/blob/80418efe80837f8748227f19ccf32871c12e7391/src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs#L226

but in our implementation, we are not using this way by adding its initialization to the initializer of the corresponding property while building the ctor.

Do we need to align?

ArcturusZhang (commented):
We have a `InputConstant` in our input types, which is really similar to the literal type.

Do we need to use two things to represent a very similar case?

Currently the `InputConstant` class is only used in `InputParamter`s, but I believe this literal type has very few differences from the LiteralType.

This is what `InputConstant` and `InputLiteralType` looks like:
```csharp
internal record InputConstant(object? Value, InputType Type);
internal record InputLiteralType(string Name, InputType LiteralValueType, object Value, bool IsNullable = false)
```
The `Name` in literal is constant `Literal`, and `IsNullable` makes no sense on literal types (you can only assign one value for it, how could it be null? if it could be null, it only longer only accepts one value).

Therefore I suggest we combine the `InputConstant` and `InputLiteralType` and things would align with how we deal with the same thing about ConstantSchema back using swaggers

test/TestProjects/FirstTest-Cadl/Generated/Models/Thing.cs
archerzz: This is another option. Previously I found initializers which are bound to constructor parameters, which is not feasible.

archerzz: We have the following codes in `ModelWriter`, that's why I choose to use `DefaultValue`:
https://github.com/Azure/autorest.csharp/blob/80418efe80837f8748227f19ccf32871c12e7391/src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs#L211-L219

archerzz: Because for literal types, the values are constants. So maybe that's enough.

archerzz:
A couple of changes since last review:
- Modify emitter to add basic support of differentiating integer from float. We'll not try to identify the full list of CADL number types. Instead, we just support `int32` for integer value and `double` for float value.
- Remove literal type properties from generated deserialization codes
- Add E2E test case

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeProperty.cs
```diff
@@ -167,5 +168,7 @@ private static string CreateExtraPropertyDiscriminatorSummary(CSharpType valueTy
             }
             return updatedDescription;
         }
+
+        public bool ShouldSkipDeserialization => InputModelProperty?.Type is InputLiteralType;
```
AlexanderSher, line 172: I think we can move this into `ModelTypeProvider` and call inside `CreatePropertySerializations`. I wonder maybe we should have two sets of serialization properties - for serialization and for deserialization.

archerzz: Previously I implemented this property in `ObjectTypeProperty` because there are other 2 places (`SerializationBuilder.GetPropertySerializationsFromBag` and `JsonAddtiontalPropertiesSerialization`) which handle `JsonPropertySerliaization`. I rechecked and confirmed that we won't skip deserialization in those 2 places. So I move this check into `ModelTypeProvider`.

I think currently `JsonPropertySerialization` should be sufficient to represent the serialization constraints.

archerzz:
@m-nash I've committed change to treat literal type operation parameters as constant parameters (see the last 2nd commit). However, I think it's still better to handle that in the emitter, because:
- By the meaning, it's a constant. It's more like a compiler optimization, v.s. type/name mapping which should be handled in generator.
- Previously it's handled outside generator. When autorest passes in the parameters, their types are already decided.
- We don't have a neat encapsulation on this issue. I have to change a few places in the codes, instead of one line code change in the emitter.
- Another option is to write a Json converter to transform the `Kind` property. But that's almost equal to change in the emitter.

m-nash (approved):

m-nash (approved):

PR #3037, opened 2023-01-29
Add a new configuration to convert things to scope

ArcturusZhang:
Regen: https://github.com/Azure/azure-sdk-for-net/pull/33745

# Description

This PR adds a new configuration to allow you to mark a specific request path as a "scope" with a certain resource type.

For more details about this, please check the new doc changes in this PR.

This PR also fixes a using issue when the samples invoke the extension methods (added the missing using statement)

# Checklist

# Ready to Land?


docs/mgmt/readme.md
```diff
@@ -689,6 +710,64 @@ public static partial class ResourcesExtension
...
+    {
+        if (!scope.ResourceType.Equals("Microsoft.Compute/virtualMachines"))
+            throw new InvalidOperationException(string.Format("Invalid resource type {0} expected Microsoft.Compute/virtualMachines", scope.ResourceType));
+        /* ... */
+    }
++
++   public static VmGuestConfigurationCollection GetVmGuestConfigurations(this ArmResource armResource)
++   {
++       /* ... */
++   }
```
fengzhou-msft, line 767: Theorectically, the combination of the two configurations work fine. But the extension method would only work for `VirtualMachineResource`, others would throw exception, right? Then I wonder if there are real cases that need to use both configurations. We'd better not mention it in the doc if no real usages.

ArcturusZhang: OK sorry I misunderstood your comment in the first place.
Yes, I planned to add validations before the ArmResource extension method as those scope methods so that they could only call it without exceptions on VirtualMachineResource.
I also planned to add more information in the generated xml document to emphasize that.

fengzhou-msft: I mean do we ever need to add an extension method to `ArmResource` if the extension method only works for `VirtualMachineResource`? From functionality perspective, it works. Keeping the case in the tests is also fine. My suggestion is that if we do not have real usage case of ever needing such extension methods, we may omit the case of using both configurations in the doc. Or if there may be real cases, please mention when we need to consider adding the `ArmResource` extension methods in the doc.

ArcturusZhang: In this case, no. The configuration just provide the possibility.
In the real case, we do not add this `generate-arm-resource-extensions` configuration unless we could confirm that the scope could be a lot of resources.

ArcturusZhang: let me add something in the document to explain that a bit more by telling the reader that despite we could do this, but it is not recommended.

fengzhou-msft (approved):

PR #3044, opened 2023-01-31
Enable raw array as input and output

### LINKED ISSUE #2673: Support array payload in convenience method
Returning an array (of string is the most always case) exists on real life, but we haven't support is now.

Cadl:
```
op world(): string[];
```

What we generate now is:
```
public virtual async Task<Response<IList<string>>> WorldValueAsync(CancellationToken cancellationToken = default)
{
    using var scope = ClientDiagnostics.CreateScope("HelloClient.WorldValue");
    scope.Start();
    try
    {
        Response response = await WorldAsync(context).ConfigureAwait(false);
        return Response.FromValue(IList<string>.FromResponse(response), response);
    }
    catch (Exception e)
    {
        scope.Failed(e);
        throw;
    }
}

```
Pay attention to `IList<string>.FromResponse(response)` which is invalid.

Epic: https://github.com/Azure/cadl-azure/issues/1943

pshao25:
Fix #2673 

Expectation: https://github.com/Azure/autorest.csharp/issues/2673#issuecomment-1422253225

SDK change: No change


src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -255,6 +255,7 @@ internal static void LoadConfiguration(string? projectPath, string outputPath, s
                 ReadOption(root, Configuration.Options.SingleTopLevelClient),
                 ReadOption(root, Configuration.Options.SkipSerializationFormatXml),
                 ReadOption(root, Configuration.Options.DisablePaginationTopRenaming),
+                ReadOption(root, Configuration.Options.ArrayToPage),
```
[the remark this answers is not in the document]
ArcturusZhang: OK I get it since this configuration is only used in cadl projects, and for cadl projects the configurations are written by the emitter, we do not need to write the configuration in the generator.

But should we align? or add some comments to explain?

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -121,6 +124,7 @@ public enum UnreferencedTypesHandlingOption
         public static bool SingleTopLevelClient { get; private set; }
         public static bool SkipSerializationFormatXml { get; private set; }
         public static bool DisablePaginationTopRenaming { get; private set; }
+        public static bool ArrayToPage { get; private set; }
```
m-nash, line 127: I don't think this will be a bool.  It will be specific to an individual operation.  It will most likely be a decorator in cadl that only the dotnet emitter will care about.

pshao25: like `@PageResponse(boolean)`?

How could we not confuse other languages?

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -85,6 +85,11 @@ public static FormattableString GetIdentifiersFormattable(this IEnumerable<strin
 
         public static FormattableString GetConversionFormattable(this Parameter parameter, CSharpType toType)
         {
+            if (TypeFactory.IsList(parameter.Type) && toType.EqualsIgnoreNullable(typeof(RequestContent)))
+            {
+                return $"{typeof(RequestContent)}.{nameof(RequestContent.Create)}({parameter.Name})";
+            }
+
             var conversionMethod = GetConversionMethod(parameter.Type, toType);
```
ArcturusZhang, line 93: This new `if` is very similar with the `GetConversionMethod` method, is it possible that we combine these two methods?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -120,18 +120,20 @@ private bool IsParameterTypeSame(Parameter? first, Parameter? second)
...
+            var operationBodyTypes = operation.Responses.Where(r => !r.IsErrorResponse).Select(r => r.BodyType).Distinct().ToArray();
+            return operationBodyTypes.Length != 0 ? operationBodyTypes[0] : null;
+        }
+
         private ReturnTypeChain BuildReturnTypes()
         {
-            var operationBodyTypes = Operation.Responses.Where(r => !r.IsErrorResponse).Select(r => r.BodyType).Distinct().ToArray();
             CSharpType? responseType = null;
-            if (operationBodyTypes.Length != 0)
+            var firstBodyType = GetFirstBodyType(Operation);
```
ArcturusZhang, line 132: this is quite weird, could we have multiple body types?

pshao25: Yes, emitter passes multiple types to generator and generator only takes the first one.

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
pshao25: But it would be just copy exact content to `GetConversionMethod` which is not that valuable I think, because I don't want to break the switch logic in thta function.

pshao25: We want to return `RequestContent.Create("name")`, which name is `parameter.Name`, so we cannot return a string from `GetConversionMethod`

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
chunyu3: Is it request body or response body?
if it is response body, maybe we need to get the body according to the status code, such as response body for 200.

pshao25: It is response body, and yes we already do that.

pshao25: Aactually I don't change any logic here, I just extract that logic into a method for better usage. If you don't like I could revert that part.

ArcturusZhang: well to be clear please add some comments here to emphasize this is taking the first of the "response body type"

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
ArcturusZhang: That is why I suppose we could see some refactors here

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
chunyu3: Then maybe need to change the function name `GetFirstBodyType` to avoid confusion. e.g. GetResponseBody

pshao25: I will revert this change, it is irrelevant to the issue I want to fix.

chunyu3 (approved):

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
pshao25: Synced we will keep it for now.

PR #3061, opened 2023-02-02
Minimizing writer logic

AlexanderSher:
This PR is an attempt to minimize the amount of duplicating code in writers and to join back DPG, MPG and HLC code branches as much as possible, with the final goal to have writers handle only universal set of model objects, while differences in flavors represented in specifics of output model tress.

It is not in a final state, hence should be treated as a proof of concept. However, it is fully functional. 

#### Main types
PR introduces two type groups: value expressions (derived from `ValueExpression`) and statements (derived from `MethodBodyStatement`). While according to C# official documentation, [expressions are statements](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/statements#expression-statements), in PR they are a separate group so that we don't have statements (that don't return any value) inside expressions.

Statements include foreach loops, if/else blocks, method calls that don't return a value. Expressions include variable and parameter references, string literals, property access, method calls that return a value.

A special type of expression is `FormattableStringToExpression`, which is used a temporary shim between existing writers and new model. At the end of consolidation process, it should be removed completely.

#### Known value expressions
Latest version of the PR introduces `TypedValueExpression` abstract record. It is a wrapper around any other `ValueExpression` that also contains type of the value returned by expression. Types derived from `TypedValueExpression` aren't specially handled by code writer, hence they don't affect the generated code. Instead, they provide a way to build output model trees so that code for model output looks like the generated code. For example, model code:
```cs
public static Method BuildFromResponse(SerializableObjectType type, MethodSignatureModifiers modifiers)
{
    var fromResponse = new Parameter("response", "The response to deserialize the model from.", new CSharpType(typeof(Response)), null, Validation.None, null);
    return new Method
    (
        new MethodSignature("FromResponse", null, "Deserializes the model from a raw response.", modifiers, type.Type, null, new[]{fromResponse}),
        new MethodBodyStatement[]
        {
            UsingVar("document", JsonDocumentExpression.Parse(new ResponseExpression(fromResponse).Content), out var document),
            Return(SerializableObjectTypeExpression.Deserialize(type, document.RootElement))
        }
    );
}
```
would produce generated code
```cs
/// <summary> Deserializes the model from a raw response. </summary>
/// <param name="response"> The response to deserialize the model from. </param>
internal static ModelWithCustomizedProperties FromResponse(Response response)
{
    using var document = JsonDocument.Parse(response.Content);
    return DeserializeModelWithCustomizedProperties(document.RootElement);
}
```

#### Debugging
PR introduces `DebuggerCodeWriter` that helps rendering expressions and statements while debugging. For example:
image
Writer can be improved to include offsets. However, that would make its output different than the one we use in generation


samples/AnomalyDetector/Generated/AnomalyDetectorClient.cs
```diff
@@ -894,7 +894,7 @@ public virtual AsyncPageable<AnomalyDetectionModel> GetMultivariateModelValuesAs
             RequestContext context = cancellationToken.CanBeCanceled ? new RequestContext { CancellationToken = cancellationToken } : null;
             HttpMessage FirstPageRequest(int? pageSizeHint) => CreateGetMultivariateModelsRequest(skip, maxCount, context);
             HttpMessage NextPageRequest(int? pageSizeHint, string nextLink) => CreateGetMultivariateModelsNextPageRequest(nextLink, skip, maxCount, context);
-            return PageableHelpers.CreateAsyncPageable(FirstPageRequest, NextPageRequest, AnomalyDetectionModel.DeserializeAnomalyDetectionModel, ClientDiagnostics, _pipeline, "AnomalyDetectorClient.GetMultivariateModels", "models", "nextLink", context);
+            return PageableHelpers.CreateAsyncPageable(FirstPageRequest, NextPageRequest, AnomalyDetectionModel.DeserializeAnomalyDetectionModel, ClientDiagnostics, _pipeline, "AnomalyDetectorClient.GetMultivariateModelValues", "models", "nextLink", context);
```
AlexanderSher, line 897: These are the only changes in generated code, and they are correct, cause method name is `GetMultivariateModelValuesAsync`, not `GetMultivariateModelsAsync`.

test/TestProjects/Pagination-Cadl/Generated/PaginationClient.cs
```diff
@@ -132,7 +132,7 @@ public virtual AsyncPageable<LedgerEntry> GetLedgerEntryValuesAsync(Cancellation
             RequestContext context = cancellationToken.CanBeCanceled ? new RequestContext { CancellationToken = cancellationToken } : null;
             HttpMessage FirstPageRequest(int? pageSizeHint) => CreateGetLedgerEntriesRequest(context);
             HttpMessage NextPageRequest(int? pageSizeHint, string nextLink) => CreateGetLedgerEntriesNextPageRequest(nextLink, context);
-            return PageableHelpers.CreateAsyncPageable(FirstPageRequest, NextPageRequest, LedgerEntry.DeserializeLedgerEntry, ClientDiagnostics, _pipeline, "PaginationClient.GetLedgerEntries", "value", "nextLink", context);
+            return PageableHelpers.CreateAsyncPageable(FirstPageRequest, NextPageRequest, LedgerEntry.DeserializeLedgerEntry, ClientDiagnostics, _pipeline, "PaginationClient.GetLedgerEntryValues", "value", "nextLink", context);
```
pshao25, line 135: This is a breaking change. And this makes me think whether your last refactor causes similar breaking changes. Could you check?

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -620,5 +617,200 @@ public static IDisposable WriteCommonMethod(this CodeWriter writer, MethodSignat
...
+                case MethodBodyLines lines:
+                    foreach (var line in lines.MethodBodySingleLine)
+                    {
+                        writer.Line(line);
+                    }
+                    break;
+            }
+        }
+
+        public static CodeWriter Line(this CodeWriter writer, MethodBodyLine line)
```
chunyu3, line 653: There already is function Line to write a normal line. It is better to give out a more clear name such as WriteBodyLine.

chunyu3 (commented):
Where do we create the ClientMethodBodyLines? in the buildMethod or buildLibray?  Thanks

test/TestProjects/Pagination-Cadl/Generated/PaginationClient.cs
AlexanderSher: - No, it is not a breaking change. For the list of things we recognize as a breaking change, see: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md

- Yes, this bug was introduced by the previous refactoring of pageables.

src/AutoRest.CSharp/Common/Generation/Writers/ClientWriter.cs
```diff
@@ -25,7 +25,7 @@ internal class ClientWriter
 
         protected static string CreateMethodName(string name, bool async) => async ? $"{name}Async" : name;
 
-        protected void WriteClientFields(CodeWriter writer, RestClient client, bool writePipelineField)
+        protected void WriteClientFields(CodeWriter writer, DataPlaneRestClient client, bool writePipelineField)
```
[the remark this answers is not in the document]
AlexanderSher: Management plane doesn't use these methods yet. It is an upcoming work to join these two code paths together.

test/TestProjects/Pagination-Cadl/Generated/PaginationClient.cs
pshao25: It is not an SDK breaking change, but a relemetry breaking change. I agree to make this change to fix the bug, but we need to be careful that the telemetry are not broken.

src/AutoRest.CSharp/Common/Generation/Writers/ClientWriter.cs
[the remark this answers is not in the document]
m-nash: Is the plan to use this in the future for mgmt plane?  Or is the plan that both dataplane and mgmt plane will change to some 3rd thing?

The proposed final destination shouldn't have data or mgmt in the name.

samples/AnomalyDetector/Generated/AnomalyDetectorClient.cs
m-nash: This is actually an interesting question.  Is telemetry intended to track client side or service side?

If its service side then `GetMultivariateModels` and `GetMultivariateModelValues` should all share the same scope similar to how we don't have 2 scopes to differentiate between sync and async since they both end up calling the same service operation.

@tg-msft do you have an opinion here?

src/AutoRest.CSharp/Common/Output/Models/ClientMethodBodyLines.cs
```diff
@@ -0,0 +1,216 @@
...
+
+    internal record Method(MethodSignature Signature, MethodBody Body);
+    internal record MethodBody(IReadOnlyList<MethodBodyBlock> Blocks);
+
+    internal record MethodBodyBlock
+    {
+        public static MethodBodyBlock Create(params MethodBodyLine[] lines) => new MethodBodyLines(lines);
+    }
+
+    internal record ParameterValidationBlock(IReadOnlyList<Parameter> Parameters) : MethodBodyBlock;
```
m-nash, line 204: Can we please have each class represented by a file, its very difficult especially in github pr review to figure out where these things are defined.

src/AutoRest.CSharp/Common/Output/Models/ClientMethodBodyLines.cs
```diff
@@ -0,0 +1,216 @@
...
+                var newInstance = new NewInstanceExpression(model.Type, arguments);
+                return new DeclareVariableLine(model.Type, variable, newInstance);
+            }
+        }
+    }
+
+    internal record Method(MethodSignature Signature, MethodBody Body);
+    internal record MethodBody(IReadOnlyList<MethodBodyBlock> Blocks);
+
+    internal record MethodBodyBlock
```
m-nash, line 199: I typically see a code block as the set of lines that go inside a for, if, else etc.  If this is the case would we not represent this as an having an opening line as well as a set of lines that go inside?

usage would be something like
```c#
public static WriteMethodBodyBlock(this CodeWriter writer, MethodBodyBlock methodBodyBlock)
{
    writer.Scope(methodBodyBlock.OpeningLine)
    {
        foreach(var line in methodBodyBLock.Lines)
        {
            writer.Line(line);
        }
    }
}
```

If instead this isn't intended to represent a block with this definition can we change it to `MethodBodySnippet` to simply represent some subset of the entire method body?

src/AutoRest.CSharp/Common/Output/Models/ClientMethodBodyLines.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                return new SetValueLine(GetResponseValueId(response), from);
+            }
+        }
+
+        public static class Declare
+        {
+            public static MethodBodyLine FirstPageRequest(ValueExpression? restClient, string methodName, IEnumerable<ValueExpression> arguments, out CodeWriterDeclaration localFunctionName)
+            {
+                var requestMethodCall = new InstanceMethodCallExpression(restClient, methodName, arguments.ToList(), false);
```
m-nash, line 154: If we go this route it starts to get harder to read the code and determine what the code will look like in the end.

I wonder if we should consider something along the lines of what exists with [ModelMethodDefinition](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Output/Models/Types/ModelMethodDefinition.cs).

The idea here is that each writer can create simple templated methods like [FromRequest and ToResponseContent](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs#L24-L25) methods and then the output types can [combine](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs#L393) the signature with the body implementation they want.

This maintains a fairly easy visual translation between the generator code and the resulting dotnet code for the project.  Yes there will be some duplication around some minor things but as we find significant ones we can abstract those out so they can get called.

At a minimum we shouldn't have two ways to do this since this is a consolidation effort.

samples/AnomalyDetector/Generated/AnomalyDetectorClient.cs
tg-msft: My intuition is we want the scope name to match the method name, modulo the `Async` suffix since that's usually a shared or at least identical code path.  If we have different names between protocol and convenience methods, I think that would imply different scope names.  +@lmolkova to give you a definitive answer though.

lmolkova: your intuition is absolutely correct @tg-msft, the public-surface API telemetry represents client-side  (how long logical call took in their app and how it ended). Having method name in the Activity/span name helps users match telemetry to their code.  This Async suffix is not that helpful and I believe we decided not to populate it. 

So this change looks great to me: scope matches method name as per guidelines

src/AutoRest.CSharp/Common/Generation/Writers/ClientWriter.cs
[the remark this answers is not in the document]
AlexanderSher: Yes, the plan is to have one type instead of two (`RestClient`) that would be shared by MPG and HLC

src/AutoRest.CSharp/Common/Output/Models/ClientMethodBodyLines.cs
AlexanderSher: Sure.

src/AutoRest.CSharp/Common/Output/Models/ClientMethodBodyLines.cs
AlexanderSher: Unlike with `ValueExpression` where we have basically any type of supported C# expression, we only need a very small subset of code blocks:
 - foreach in deserialization
 - try/catch/finally for DiagnosticsScope
 - switch for status code handling

So we don't really need fully generalized method body block support. Maybe call them `MethodBodyLinesSnippets`?

samples/AnomalyDetector/Generated/AnomalyDetectorClient.cs
m-nash: @lmolkova  thanks for the clarification.

src/AutoRest.CSharp/Mgmt/Decorator/ParameterMappingBuilder.cs
```diff
@@ -197,18 +195,18 @@ private static IEnumerable<List<Segment>> SplitDiffIntoPairs(RequestPath diff)
...
             if (TypeFactory.IsStringLike(type))
-                return rawExpression;
+                return valueExpression;
 
             if (!type.IsFrameworkType)
             {
                 if (type.Implementation is EnumType enumType && !enumType.IsExtensible)
                 {
-                    return $"{rawExpression}.To{enumType.Declaration.Name}()";
+                    return new InstanceMethodCallExpression(valueExpression, $"To{enumType.Declaration.Name}", Array.Empty<ValueExpression>(), false);
```
ArcturusZhang, line 207: Could we also add the method for enum serialization and deserialization to the `EnumType` class? We are kind of constructing the names of these two methods in multiple places (this is one of the places) and assuming those names.

For this deserialization method, the name is kind of stable, but for serialization, it varies from the underlying type. Therefore I believe it should benefit if we put these names inside `EnumType`.

src/AutoRest.CSharp/Common/Output/Models/ValueExpressions.cs
```diff
@@ -0,0 +1,291 @@
...
+                    return ProcessMessage(pipeline, message, clientDiagnostics, scopeName, finalStateVia, async, methodName);
+                }
+
+                private static ValueExpression ProcessMessage(CodeWriterDeclaration pipeline, CodeWriterDeclaration message, CodeWriterDeclaration clientDiagnostics, string scopeName, OperationFinalStateVia finalStateVia, bool async, string methodName)
+                {
+                    var arguments = new List<ValueExpression> {
+                        new VariableReference(pipeline),
+                        new VariableReference(message),
+                        new VariableReference(clientDiagnostics),
+                        new FormattableStringToExpression($"{scopeName:L}"),
```
ArcturusZhang, line 238: Should we use the `Literal` method here?

src/AutoRest.CSharp/Common/Output/Models/ValueExpressions.cs
```diff
@@ -0,0 +1,291 @@
...
+                {
+                    var arguments = new List<ValueExpression>
+                    {
+                        new ParameterReference(KnownParameters.WaitForCompletion),
+                        new VariableReference(message),
+                        createNextPageRequest is not null ? new VariableReference(createNextPageRequest) : Null,
+                        GetValueFactory(pageItemType),
+                        new VariableReference(clientDiagnostics),
+                        new VariableReference(pipeline),
+                        new FormattableStringToExpression($"{typeof(OperationFinalStateVia)}.{finalStateVia}"),
```
ArcturusZhang, line 178: Should we create a method like `PropertyReference` or `PropertyInvocation` to handle case like this?
like 
```
public static ValueExpression PropertyReference(CSharpType type, string propertyName)
```

src/AutoRest.CSharp/Common/Output/Models/ValueExpressions.cs
```diff
@@ -0,0 +1,291 @@
...
+    {
+        public IReadOnlyDictionary<string, ValueExpression> Properties { get; init; } = new Dictionary<string, ValueExpression>();
+    }
+
+    internal record TernaryConditionalOperator(ValueExpression Condition, ValueExpression Consequent, ValueExpression Alternative) : ValueExpression;
+    internal record TypeReference(CSharpType Type) : ValueExpression;
+    internal record MemberReference(ValueExpression Inner, string MemberName) : ValueExpression;
+    internal record ParameterReference(Parameter Parameter) : ValueExpression;
+    internal record VariableReference(CodeWriterDeclaration Name) : ValueExpression;
+    internal record FormattableStringToExpression(FormattableString Value) : ValueExpression; // Shim between formattable strings and expressions
```
ArcturusZhang, line 286: I believe the purpose of this record here is for compatibility only, but I think we should not abuse this class.
The purpose of this PR is trying to give our writers "semantic meanings" so that we could more likely to write code that could compile and maybe more easy to read. But this class is kind of following the "old ways" - I do not have a meaning, just write whatever I have here.

Therefore I suggest for everywhere we are using the `FormattableStringToExpression` record, we might need to take a look and see if we could use an existing static method or come up to a static method to describe that usage, and eventually remove this.

ArcturusZhang (commented):
I think this PR introduces a great concept that we could finally give our writers to write code with semantic meanings.
But I have a bit concern that we should not abuse the `FormattableStringToExpression`, for every place we are using it, we might need to use an existing method instead or come up a new static method to describe that scenario

src/AutoRest.CSharp/Common/Output/Models/ClientMethodBodyLines.cs
m-nash: I don't think `Lines` adds anything we can probably shrink to just `MethodBodySnippet`

src/AutoRest.CSharp/Common/Generation/Writers/ClientWriter.cs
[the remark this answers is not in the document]
m-nash: I also think it would be good to see what the mgmt plane changes would look like before finalizing here since the devil is always in the details.

src/AutoRest.CSharp/Mgmt/Decorator/ParameterMappingBuilder.cs
AlexanderSher: Done.

src/AutoRest.CSharp/Common/Output/Models/ValueExpressions.cs
AlexanderSher: Here it is not a property reference, but rather an enum value. A snippet is created to handle this and similar cases

src/AutoRest.CSharp/Common/Output/Models/ValueExpressions.cs
AlexanderSher: Yes

PR #3071, opened 2023-02-03
Use UTF-8 constants in serialization/deserialization code

AlexanderSher:
https://github.com/Azure/azure-sdk-for-net/ repo is updated to C# 11, so now we can use [UTF-8 string literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/utf8-string-literals). As @chunyu3 has found, utf8 spans give us 30% performance boost in serialization/deserialization comparing to strings.


src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -246,7 +246,7 @@ private static void ToSerializeCall(this CodeWriter writer, FormattableString wr
                 if (property.ValueSerialization == null)
                 {
                     // Flattened property
-                    writer.Line($"{writerName}.WritePropertyName({property.SerializedName:L});");
```
archerzz, line 249: Maybe we can change here to append `u8`? https://github.com/Azure/autorest.csharp/blob/b8ebef729c64db45fe033830699ccb4ac0fe425a/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriter.cs#L385

archerzz (commented):
LGTM. Just wonder if we could update `CodeWriter` to append `u8` to all string literals.

PR #3088, opened 2023-02-09
Improve management plane lro implementation

### LINKED ISSUE #3076: Bug when the operation is lro and the body parameter is of type Stream
This issue comes from https://github.com/Azure/azure-sdk-for-net/issues/33275.

After investigation, it turns out that the root cause is the body parameters is a stream and we use this parameter twice. At first, we will create the content from this stream and read it when sending the request. After that this stream will be closed. However, as this method is a lro operation, we will call the CreateXXXRequest again when constructing ArmOperation and it leads to that we will try to create a RequestContent from stream with `CanSeek` is equal to false, which will throw the exception [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/RequestContent.cs#L130).

Considering that the ArmOperation only needs the Uri and Method of the request, I think one of the solutions could be that we add an additional parameter `isContentSkipped` to the CreateXXXRequest method when the operation is lro and has a stream parameter. Then we can just skip content when we construct the request that ArmOperation needs.

Yao725:
Fix #3076 

# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
+++ b/src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
@@ -743,9 +743,20 @@ namespace AutoRest.CSharp.Mgmt.Generation
 
         protected virtual void WriteLROMethodBranch(MgmtRestOperation operation, IEnumerable<ParameterMapping> parameterMapping, bool async)
         {
+            var restClientName = GetRestClientName(operation);
+            if (!operation.IsFakeLongRunningOperation)
+            {
+                _writer.Append($"using var message = {restClientName}.{RequestWriterHelpers.CreateRequestMethodName(operation.Method.Name)}(");
+                WriteArguments(_writer, parameterMapping);
+                _writer.RemoveTrailingComma();
+                _writer.Line($");");
+            }
             _writer.Append($"var response = {GetAwait(async)} ");
-            _writer.Append($"{GetRestClientName(operation)}.{CreateMethodName(operation.Method.Name, async)}(");
-            WriteArguments(_writer, parameterMapping);
+            _writer.Append($"{restClientName}.{CreateMethodName(operation.Method.Name, async)}(");
+            if (operation.IsFakeLongRunningOperation)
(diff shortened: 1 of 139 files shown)


src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -743,9 +743,20 @@ protected virtual void WriteLROMethodBody(MgmtClientOperation clientOperation, D
 
         protected virtual void WriteLROMethodBranch(MgmtRestOperation operation, IEnumerable<ParameterMapping> parameterMapping, bool async)
         {
+            var restClientName = GetRestClientName(operation);
```
m-nash, line 746: I would love for this to be consolidated into a common writer for dpg / mpg.  If we want to do that in a separate PR please open an issue to track it for future reference.

m-nash (approved):

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
Yao725: Opened an issue here: https://github.com/Azure/autorest.csharp/issues/3106

lirenhe:
Please hold the merge of this PR. 
Could you generate the SDK PR so that we could understand the impact?   cc @ArthurMa1978

Yao725:
> Please hold the merge of this PR. Could you generate the SDK PR so that we could understand the impact? cc @ArthurMa1978 Arthur Ma FTE

The regen is [here](https://github.com/Azure/azure-sdk-for-net/pull/34066) and this PR will not be merged at present as many tests in management plane will be affected.

m-nash:
> > Please hold the merge of this PR. Could you generate the SDK PR so that we could understand the impact? cc @ArthurMa1978 Arthur Ma FTE
> 
> The regen is [here](https://github.com/Azure/azure-sdk-for-net/pull/34066) and this PR will not be merged at present as many tests in management plane will be affected.

Can we add that link to the top level description so its easy to find

PR #3089, opened 2023-02-09
feat(cadl): support multiple level discriminators

### LINKED ISSUE #3081: Support multiple discriminators
Sample cadl:
https://github.com/Azure/cadl-ranch/blob/68ac3ccc553645284f39254c558bc0e4f24c85c8/packages/cadl-ranch-specs/http/models/inheritance/main.cadl

We could support this like HLC, I have a trial here: https://github.com/pshao25/autorest.csharp/tree/inheritance/test/TestProjects/AdditionalPropertiesEx/Generated

archerzz:
- change `SerializationWriter` to generate correct deserialization codes for base classes with multiple level discriminators
- change `ModelTypeProvider` to generate correct initialization constructor parameters and initializers
- fix an issue that we use child class name instead of discriminator value as the constants to compare with discriminator property value
- enable cadl ranch native `models/inheritance` project, and remove modified local copy
- fix corresponding tests for cadl ranch `models/inheritance`
- add multiple level discriminator case for hlc

resolve #3081

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -443,7 +445,7 @@ private bool TryGetPropertyForInputModelProperty(Func<ObjectTypeProperty, bool>
             else
             {
                 //only load implementations for the base type
-                implementations = _derivedTypes!.Select(child => new ObjectTypeDiscriminatorImplementation(child.Name, _typeFactory.CreateType(child))).ToArray();
```
archerzz, line 446: We should use the `DiscriminatorValue`, not `Name`. Here is what we previously created:
```
case "Shark": return Shark.DeserializeShark(element);
case "Salmon": return Salmon.DeserializeSalmon(element);
```

That's wrong. It should be:
```
case "shark": return Shark.DeserializeShark(element);
case "salmon": return Salmon.DeserializeSalmon(element);
```

archerzz:
For cadl ranch definition, see: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/models/inheritance/main.cadl

archerzz:
Per discussion, we will not support multiple level discriminators. See https://github.com/Azure/cadl-azure/issues/2219

Issue #2802, opened 2022-10-21
[Feature Req] Change the way we handle empty string value in deserialization


xboxeer:
**Describe the issue or request**
In this github issue, https://github.com/Azure/azure-sdk-for-net/issues/31701, the keyvault id is an empty string, and SDK is throwing exception that the value cannot be null for property resourceId

**Describe your ideas for solutions**
We should either not throwing exception when keyvault id is empty string, or we should throw the exception before we are trying to constructing the resource identifior  throw an empty string, throwing the exception in the resource identifior ctor will miss a lot of context information. 
The keyvault id here can applied to any kind of id property in raw response that SDK is trying to desterilize to a resource identifior 

**Add labels**
As appropriate, select a label to describe how hard the issue is to work
around or how hard it would be to do without a particular feature that
could help make it easier. These labels start with "workaround".

Also select a label that describes how many instances of the workaround
you would have to do without the bug being fixed or feature implemented.
These labels start with "instances".

These labels will help priority bug fixes and feature requests.

If this bug or feature request is for older versions of autorest, please
remove the v3 label and add the v2 label as appropriate.

xboxeer:
CC @ArcturusZhang

xboxeer:
@ArcturusZhang do we have any update on this?

Yao725:
I believe these issues are also related to this feature request:

- https://github.com/Azure/azure-sdk-for-net/issues/33245
- https://github.com/Azure/azure-sdk-for-net/issues/33388

fengzhou-msft:
We decided to use a config both for mgmt palne and data plane to specify the models that should treat empty string as null. The intial types to check for empty strings could be: `Uri`, `Guid`, `ResourceIdentifier` or plus others that we may add format in `autorest.md` for the original string in swagger.

fengzhou-msft:
Previously we fixed the issue with custom code for several SDKs. After the generator fix, we should remove those custom code and fix with the configs for easier maintenance in the future.

PR #3111, opened 2023-02-14
Treat empty string as null during deserialization with configs

fengzhou-msft:
# Description
Mitigate #2802

This PR adds 2 new configs and example values as below:
```
models-to-treat-empty-string-as-null:
- ExactMatchModel11

additional-intrinsic-types-to-treat-empty-string-as-null:
- ResourceType
- IPAddress
- AzureLocation
```

The default types to treat empty string as null are`Uri`, `ResourceIdentifier`, `Guid` and `DateTimeOffset`. Values for `additional-intrinsic-types-to-treat-empty-string-as-null` will be added to this list.

A local regeneration of SDKs showed this PR caused no changes without adding the configs.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -263,6 +272,8 @@ internal static void LoadConfiguration(string? projectPath, string outputPath, s
                 projectPath ?? ReadStringOption(root, Configuration.Options.ProjectFolder),
                 protocolMethods,
                 suppressAbstractBaseClasses,
+                treatEmptyStringAsNullForModels,
+                treatEmptyStringAsNullForTypes,
```
archerzz, line 276: better to align with the same name:
```suggestion
                treatEmptyStringAsNullForExtraTypes,
```

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -37,6 +38,8 @@ public static class Options
             public const string UnreferencedTypesHandling = "unreferenced-types-handling";
             public const string ModelFactoryForHlc = "model-factory-for-hlc";
             public const string GenerateModelFactory = "generate-model-factory";
+            public const string TreatEmptyStringAsNullForModels = "treat-empty-string-as-null-for-models";
+            public const string TreatEmptyStringAsNullForExtraTypes = "treat-empty-string-as-null-for-extra-types";
```
archerzz, line 42: from its usage, it's about intrinsic types (e.g. framework types, azure core types):
```suggestion
            public const string TreatEmptyStringAsNullForIntrinsicTypes = "treat-empty-string-as-null-for-intrinsic-types";
```

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
archerzz: ah, so here you make a union of models and extra types

src/CADL.Extension/Emitter.Csharp/src/options.ts
```diff
@@ -64,7 +76,9 @@ const defaultOptions = {
     "generate-protocol-methods": true,
     "generate-convenience-methods": true,
     "package-name": undefined,
-    debug: undefined
+    debug: undefined,
+    "treat-empty-string-as-null-for-models": undefined,
```
archerzz, line 80: why one is undefined, and one is empty array?

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -370,6 +372,22 @@ private static void DeserializeIntoObjectProperties(this CodeWriter writer, IEnu
...
+        {
+            FormattableString result = $"";
+            if (!shouldTreatEmptyStringAsNull)
+                return result;
+            if (property.ValueSerialization is JsonValueSerialization valueSerialization
+                && valueSerialization.Type.IsFrameworkType)
+            {
+                if (Configuration.TreatEmptyStringAsNullForTypes.Contains(valueSerialization.Type.FrameworkType.Name))
+                {
+                    result = $" || {itemVariable}.Value.GetString().Length == 0";
```
archerzz, line 385: What if the `JsonValueKind` is not a string? Right now `treat-empty-string-as-null` is specified by human. If someone made a mistake (e.g. specifying a wrong property which is not of string value type), which behavior is better?
- `Value.GetString()` throw an exception
- We add extra check on the `JsonValueKind` so that we let the code continue

fengzhou-msft: The config is supposed to be used for the types that are originally string in swagger but converted to other types by the generator. If we have a way to know the original type in swagger, then we can add check for the values of this config.

src/CADL.Extension/Emitter.Csharp/src/options.ts
fengzhou-msft: Originally, both default values could be undefined as if users don't set them, we don't want to store them in the `Configuration.json` and undefined is the right value to check if users have set them. 

But for `treat-empty-string-as-null-for-extra-types`, we need to union it with the default set of types, using `[]` is easier without additional check. 

Besides, whether we should store the full list of types depends on whether `treat-empty-string-as-null-for-models` is undefined since if the model list is empty then the types apply to no models and there is also no use to only store the default list of types. 

Therefore, the default value for the types config doesn't need to be undefined.

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
fengzhou-msft: Yeah, users set the extra types through autorest.md configuration, but we save the full list to `Configuration.json` for less surprise.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
fengzhou-msft: Extra means that we also have a list of default types that will treat empty string as null. Intrinsic type seems not very clear to me (just like framework type), is it a cadl concept? I found this [code snippet](https://github.com/Azure/autorest.csharp/blob/7ec42556b6515821f1f1f02b654c98b583a1cf61/src/CADL.Extension/Emitter.Csharp/src/lib/model.ts#L244-L267), it looks like number and boolean are not included in intrinsic types? But we may need to support them as long as we may convert the original string type in swagger to those types.

archerzz: How about `ForCoreTypes`?

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
archerzz: How about this check?
```suggestion
                    result = $" || (({itemVariable}.Value.ValueKind == {typeof(JsonValueKind)}.{nameof(JsonValueKind.String}} && {itemVariable}.Value.GetString().Length == 0)";
```

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
ArcturusZhang: The name of this config sounds like it is a toogle of a feature, whether we want to treat empty string as null for those types, instead of meaning this is a list that you want to do that.
How about we change the order of words like `intrinsic-types-to-treat-empty-string-as-null`

chunyu3: `TreatEmptyStringAsNullForExtraTypes` is to select the types from an intrinstic type lists to treat empty string as null. Right?
If so, I don't see that we need to separate handle this as regular type.
I think it is better to eliminate the number of options as possible.

fengzhou-msft: It looks like you guys are all familiar with the term `intrinsic types`, then I'm fine to use this name. I will use `additional-intrinsic-types-to-treat-empty-string-as-null` as the option name.

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
fengzhou-msft: Updated.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
fengzhou-msft: > `TreatEmptyStringAsNullForExtraTypes` is to select the types from an intrinstic type lists to treat empty string as null. Right? If so, I don't see that we need to separate handle this as regular type. I think it is better to eliminate the number of options as possible.

The models config is to specify which models, the types config is to specify which properties of the specified models based on type match. 

This is a balance of not making too many properties (let's say based on the type match within an RP) to treat empty string as null (there is a risk that services may actually treat the two differently) and not make the config too granular (let's say specify each properties with a single config) since that may lead to fixing deserialzation issues one property by one property multiple times for the same model relying on customer report.

archerzz (approved):

ArthurMa1978 (approved):

Issue #3084, opened 2023-02-08
Support for `nullable` primitive type array


jpalvarezl:
**Describe the issue or request**
We have JSON that is a list of floats that may or may not contain a `null` item. Roughly speaking, something that looks like `[null, 0.1, 0.2, ...]`. It seems like it is not possible to have an array of nullable primitive types at the moment. 

We would like to have C# generated from this sort of cadl:

```cadl
alias nullableFloat = float32 | null;

model MyModel {
  listOfFloats?: nullableFloat[];   
}
```

nguerrera:
The code you have there should work.

OpenAPI emitter has some issues (*) with it and the getTypeName is omitting parens when it shouldn't but there should actually be the right structure in the type graph for an emitter to do what you want here (**), which is also the correct interpretation. You can also express this without the alias using `(float32 | null)[]`

(*) https://github.com/microsoft/cadl/issues/356#issuecomment-1287278240
(**) verifying this now to be sure

nguerrera:
Oh sorry, I thought this was filed against Cadl itself.

chunyu3:
We will set the nullable and optional as following:
```
name? : type
    - if type is Model
        - optional: true
        - nullable: true
    - if type is primitive
        - optional: true
        - nullable: true
name? : type | null
    - if type is Model
        - optional: true
        - nullable: true
    - if type is primitive
        - optional: true
        - nullable: true
name : type
    - if type is Model
        - optional: false
        - nullable: true
    - if type is primitive
        - optional: false
        - nullable: false
name : type | null
    - if type is Model
        - optional: false
        - nullable: true
    - if type is primitive
        - optional: false
        - nullable: true
 ```

Generate code:
```
if primitive && nullable
    - primitivate? i.e. int? bool?
```
serialization logic
```
if readonly
    - don't write the serailize line
if null
    - skip serialize at runtime
 ```

PR #3112, opened 2023-02-14
feat(cadl): support nullabe primitive type array in deserialization

archerzz:
- add `InputIntrinsicType` and its child types in emitter and generator
- update the `TypeFactory.CreateType`, for union types, return nullable input type if the union is composed of only `InputNullType` and a primitive or model type
- add test cases

part of #3084
regen PR https://github.com/Azure/azure-sdk-for-net/pull/34278

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -98,6 +98,22 @@ public TypeFactory(OutputLibrary library)
...
+            }
+
+            var nullableTypeCandidate = unionType.UnionItemTypes.FirstOrDefault(t => t is InputPrimitiveType or InputModelType);
+            if (nullableTypeCandidate == null)
+            {
+                return CreateType(unionType.UnionItemTypes[0]);
+            }
+
+            return CreateType(nullableTypeCandidate with { IsNullable = true });
+        }
```
ArcturusZhang, line 115: I kind of disagree this part, we should not treat something like `int | null` as a `union` despite it is in typescript.
In our original plan for unions of concrete types, the union part should throw an exception because dotnet does not have a concept of union, we will never generate a type for InputUnionType directly. Instead, we will do some kind of transformation, which transforms unions to some fundamental things that dotnet could express.
Instead I believe here our DOTNET emitter should do something to hide the fact of `int | null` is a union, instead it should just emit an int with `nullable = true` by producing a `InputPrimitiveType` instead.

ArcturusZhang (commented):
As we previously mentioned in sync ups, the emitter is not handling the "nullable" property correctly. For instance, we could constantly see we have models with `nullable=false`. Models are reference types therefore their nullability should always be true.
In the scenario this PR is trying to solve, instead of introducing a new type, we could just fix the nullability issue maybe only for primitive types, just emit `int | null` into a `InputPrimitiveType {Nullable = true}` should be able to get what we want

m-nash (approved):

PR #3128, opened 2023-02-16
build(common): split ci and pr build definition

### LINKED ISSUE #3126: build(common): split pipeline definition for `autorest.csharp` and `AutoRest C# - Publish`
## Problem

Pipelines `autorest.csharp` and `AutoRest C# - Publish` share the same `eng\pipelines\build.xml`, which cause excessive job run triggered in `auorest.csharp`.

See [this job run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2191194) in `autorest.csharp`. It's the same as [the one](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2191192) in `AutoRest C# - Publish`. We should try to avoid that.

The root cause is that `autorest.csharp` is supposed to run only for PR, and `AutoRest C# - Publish` is supposed to run only for commits to main branch (e.g. `feature/v3`). So when they share the same build definition YAML file, ``autorest.csharp` is also triggered when there is commit to main branch.

archerzz:
- PR build (e.g. `autorest.csharp` pipeline): still uses `build.xml` which now:
  - only contains build and test steps
  - only triggered by PR, e.g. disable CI build
- CI build (e.g. `AutoRest C# Publish` pipeline): now changes to use `publish.yml` which:
  - only triggered by commits on `feature/v3` branch
  - in the builds we'll publish new versions of `autorest.csharp` and update `azure-sdk-for-net` repo
- extract build and test steps into a template `build-test.yml` which is used by `build.xml` and `publish.yml`

resolve #3126 

# Checklist

# Ready to Land?


archerzz:
For the CI build of `Autorest C# publish`, see the playground run: https://dev.azure.com/azure-sdk/playground/_build/results?buildId=2193980&view=results

mikeharder:
@archerzz, @m-nash: Reading issue #3126, it sounds like what you are trying to do is disable pipeline "autorest.csharp" from running CI builds (when there are new commits to the `feature/v3` branch).  If this is all you want to do, you don't need to change any pipeline YML, you can simply disable the CI trigger in the "autorest.csharp" pipeline settings in DevOps.

But if you think it's better to refactor your pipeline YML in this way, it's also fine with me.

m-nash:
> @archerzz, @m-nash: Reading issue #3126, it sounds like what you are trying to do is disable pipeline "autorest.csharp" from running CI builds (when there are new commits to the `feature/v3` branch). If this is all you want to do, you don't need to change any pipeline YML, you can simply disable the CI trigger in the "autorest.csharp" pipeline settings in DevOps.
> 
> But if you think it's better to refactor your pipeline YML in this way, it's also fine with me.

I do like that this makes two clean pipelines vs having the if/else statements which makes it easy to affect both when you really only wanted to affect one.

@archerzz what are your thoughts on the comment above?

archerzz:
@mikeharder @m-nash Maybe I don't have enough privilege. But I could not find the setting to disable CI trigger, as described in [that stackoverflow Q&A](https://stackoverflow.com/questions/65249158/azure-devops-how-to-disable-ci-trigger-on-a-yaml-template-pipeline). I can only disable the pipeline. That's why I choose to disable it from YAML definition.

Since that also brings cleaner build YAML contents, let's move on this way.

m-nash (approved):

mikeharder:
> @mikeharder @m-nash Maybe I don't have enough privilege. But I could not find the setting to disable CI trigger, as described in [that stackoverflow Q&A](https://stackoverflow.com/questions/65249158/azure-devops-how-to-disable-ci-trigger-on-a-yaml-template-pipeline). I can only disable the pipeline. That's why I choose to disable it from YAML definition.
> 
> Since that also brings cleaner build YAML contents, let's move on this way.

@weshaggard: Any thoughts on this?  Refactoring the YAML in this way seems fine to me, but EngSys could also disable the CI trigger on this pipeline, or we could maybe grant the necessary permissions to @archerzz.

eng/pipelines/publish.yml
```diff
@@ -0,0 +1,70 @@
+trigger:
+  branches:
+    include:
+    - feature/*
+
+variables:
+  BlobFeedUrl: https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json
```
weshaggard, line 7: Why is this needed? This is an old feed which is no longer used.

eng/pipelines/publish.yml
```diff
@@ -0,0 +1,70 @@
+trigger:
+  branches:
+    include:
+    - feature/*
```
weshaggard, line 4: should this only be feature/3?

eng/pipelines/build.yml
```diff
@@ -1,8 +1,4 @@
-trigger:
-  branches:
-    include:
-    - feature/*
-
+trigger: none
 pr:
   branches:
     include:
```
weshaggard, line 4: BlobFeedUrl property is not needed.

eng/pipelines/build-test.yml
```diff
@@ -0,0 +1,123 @@
...
+          - script: |
+              npm run test --prefix src/CADL.Extension/Emitter.Csharp
+            displayName: 'Unit Test'
+            workingDirectory: $(Build.SourcesDirectory)/autorest.csharp
+          - task: PublishBuildArtifacts@1
+            condition: succeededOrFailed()
+            displayName: 'Publish Artifacts'
+            inputs:
+              ArtifactName: packages
+          - ${{if and(ne(variables['Build.Reason'], 'PullRequest'), eq(variables['System.TeamProject'], 'internal'))}}:
```
weshaggard, line 81: What is this publish doing? Should it also be moved to the publish.yml?

eng/pipelines/build.yml
```diff
@@ -20,177 +16,6 @@ resources:
       name: Azure/azure-sdk-tools
```
weshaggard, line 16: This repository reference doesn't look to be used in the build steps anywhere and should be removed.

eng/pipelines/build-test.yml
```diff
@@ -0,0 +1,123 @@
...
+  - stage: 'Build_and_Test'
+    jobs:
+      - job: Build
+        timeoutInMinutes: 120
+        pool:
+          name: azsdk-pool-mms-win-2022-general
+          vmImage: MMS2022Compliant
+        steps:
+          - checkout: self
+          - checkout: azure-sdk-tools
```
weshaggard, line 11: Do we need to checkout this repository any longer? I wouldn't expect so.

eng/pipelines/publish.yml
weshaggard: Doesn't look like this variable is used any longer in our template https://dev.azure.com/azure-sdk/internal/_git/azure-sdk-build-tools?path=/pipelines/stages/net-release-to-feed.yml so it should be safe to remove here and in the build.yml file as well.

weshaggard:
I don't mind the split up to help organize things but your split doesn't end up changing the amount of things that run in PR or in CI because the publish steps are guarded by conditions that make them not run during PR. 

As for the split as I mentioned I'm fine with but I did leave some comments for things to clean-up.

eng/pipelines/publish.yml
archerzz: Will delete it.

archerzz:
> I don't mind the split up to help organize things but your split doesn't end up changing the amount of things that run in PR or in CI because the publish steps are guarded by conditions that make them not run during PR.
> 
> As for the split as I mentioned I'm fine with but I did leave some comments for things to clean-up.

@weshaggard Actually the main purpose of this PR is to prevent duplicated job run. In [`autorest.csharp` pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=4974), each merge to `feature/v3` will trigger the same job as that in [`Autorest C# Publish` pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=1375&_a=summary). The root cause is in the original YAML file, both CI trigger and PR trigger are defined. That's why I want to split them apart.
image

eng/pipelines/build-test.yml
archerzz: In the `PublishRelease.ps1` will publish artifacts to github/npm, and it will also publish test reports. The logic is tightly coupled. I've created a follow-up issue to move publish artifacts steps to publish: https://github.com/Azure/autorest.csharp/issues/3169

eng/pipelines/build.yml
archerzz: removed

eng/pipelines/build-test.yml
archerzz: removed

weshaggard:
@archerzz if that is the main reason you don't need to split the yml files to fix that. You can simply edit on of the pipelines to disable the CI triggers in the DevOps UI. So that might be an easier option if we don't want split these.

archerzz:
> @archerzz if that is the main reason you don't need to split the yml files to fix that. You can simply edit on of the pipelines to disable the CI triggers in the DevOps UI. So that might be an easier option if we don't want split these.

Thanks. I found now I can edit the pipeline triggers. I've disabled the CI trigger for `autorest.csharp` pipeline. So this PR is unnecessary. Closing.

PR #3134, opened 2023-02-16
Change pools from Windows to Ubuntu

mikeharder:


mikeharder:
Currently, builds fail on Linux due to forward-slash vs backward-slash in generated code.

Issue #3131, opened 2023-02-16
[Bug] When return object, we should generate as Response<BinaryData>


pshao25:
Cadl:
```
@route("/test")
@doc("...")
@convenientAPI(true)
op hello(@body parameter: Thing): object;
```

Actual:
```C#
public virtual Azure.Response<Models.object> Hello(Thing parameter, CancellationToken cancellationToken = default)
{
    Argument.AssertNotNull(parameter, nameof(parameter));

    RequestContext context = FromCancellationToken(cancellationToken);
    Response response = Hello(parameter.ToRequestContent(), context);
    return Response.FromValue(Models.object.FromResponse(response), response);
}
```

Expected:
```C#
public virtual Azure.Response<BinaryData> Hello(Thing parameter, CancellationToken cancellationToken = default)
{
    Argument.AssertNotNull(parameter, nameof(parameter));

    RequestContext context = FromCancellationToken(cancellationToken);
    Response response = Hello(parameter.ToRequestContent(), context);
    // Change to BinaryData and return.
}
```

ArcturusZhang:
I believe this could be an easy fix in this way: https://github.com/Azure/autorest.csharp/pull/3137

chunyu3:
@chunyu3 will investigate emitter return.

chunyu3:
We will not support object type.

chunyu3:
build-in type `object` is deprecated.

chunyu3:
codegen will not support it now. close it.

PR #3137, opened 2023-02-17
Change the return type to `BinaryData` when in the cadl it is returning `object`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3131

# Description

# Checklist

# Ready to Land?


pshao25:
Would you add tests?

lirenhe:
Could you also add the regen PR or add the message to indicate that there are no changes

m-nash (commented):
Do we have the regen pr yet?

ArcturusZhang:
This is more complicated than I thought - and I believe the real root cause is that the emitter is not returning a flag indicating we are returning `object` therefore the generator just thinks it gets an ordinary model with the name of `object`.
We will investigate this further and open another PR

PR #3139, opened 2023-02-17
Simplify NamedTypeSymbolExtensions

AlexanderSher:
This PR is a follow-up to https://github.com/Azure/autorest.csharp/pull/3112

Refactoring only, no changes in generated code.


test/TestProjects/MgmtCollectionParent/Generated/Extensions/SubscriptionResourceExtensionClient.cs
```diff
@@ -5,6 +5,7 @@
 
 #nullable disable
 
+using System;
```
ArcturusZhang, line 8: any ideas why we have these back? This using statement is actually unused.

ArcturusZhang (approved):

m-nash (approved):

PR #3141, opened 2023-02-17
Update Azure Core Shared Codes 2023-02-16-19-55-06

archerzz:
Update Azure Core Shared Codes 2023-02-16-19-55-06


archerzz:
This PR originally contains a change of Azure core shared codes. See the commit history. After merging the `feature/v3`, the change doesn't show up because it's already merged.

Issue #3144, opened 2023-02-17
The generator is not picking up the constructors defined in customized code


ArcturusZhang:
# Description of the issue

Some of the SDKs have some customized code to customize the constructor by changing its accessibility and/or its body. For instnace: https://github.com/Azure/azure-sdk-for-net/blob/e322ae0a28f46e8e631598ce8951ec79a3e859c3/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs#L98
But this is not caught by the generator, therefore the generator still generates an overload for the asp.net extension method, not knowing that the corresponding constructor has been customized to be internal, and this leads to a compilation error.

# Some analysis

By checking our implementation, our all "ExistingType" stories work on the attributes during generation.
During the generation, our `SourceInputModel` and `ModelTypeMapping` class will gather the information on the attribute to rename things or replace things, and still generate those members as defined in the customized code.

After we output all the code to files, we have a `MemberRemoveRewriter` to remove the duplicate implementations from the generated documents when there is already a copy inside customized code.

Right now we do not have an attribute for the constructors, therefore currently in our code there is no way to let the generator know anything in the customized code (it is handled after the generation is done).

Therefore to fix this vulnerability, we might need to:
- introduce an attribute specifically to replace the constructor, and we could handle that attribute explicitly in the generator. (we already have ⁠[one](https://github.com/Azure/autorest.csharp/blob/7bd024de629af5a1d35352a5f6e5d576e615536e/src/assets/Generator.Shared/CodeGenMemberAttribute.cs#L11) for property and field)
- support this functionality without introducing an attribute, just read all the existing constructors and see if any of them matches the one that the generator is about to write, and writes the constructor in the same way as it is customized to take up the accessibility (the generator still writes it with its default body, but this will be removed by `MemberRemoveRewriter` since it has identical signature with the customized one)

In addition, we might also need to taking the `CodeGenSuppress` into account when writing the constructor, this also results in invalid code in asp.net extensions.

PR #3146, opened 2023-02-17
Picking up customized constructor

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3144
Regen: https://github.com/Azure/azure-sdk-for-net/pull/35065

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/Source/SourceMemberMapping.cs
```diff
@@ -1,20 +0,0 @@
...
-    {
-        public SourceMemberMapping(string originalName, ISymbol existingMember)
-        {
-            OriginalName = originalName;
-            ExistingMember = existingMember;
-        }
-
-        public string OriginalName { get; }
-        public ISymbol ExistingMember { get; }
-    }
```
ArcturusZhang, line 19: This type is not removed, I renamed this to `SourcePropertyMapping` and changed it into a `record` because apperantly this only applies to properties.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs
```diff
@@ -831,7 +839,13 @@ private static IReadOnlyList<MethodSignatureBase> GetClientInvocationChain(LowLe
 
                 client = client.ParentClient;
             }
-            callChain.Push(client.SecondaryConstructors.Where(c => c.Modifiers == MethodSignatureModifiers.Public).OrderBy(c => c.Parameters.Count).First());
+            var ctor = client.SecondaryConstructors.Where(c => c.Modifiers.HasFlag(MethodSignatureModifiers.Public)).OrderBy(c => c.Parameters.Count).FirstOrDefault();
+
+            // if we do not have a public secondary constructor, we have to return null and compose nothing
+            if (ctor == null)
+                return null;
```
ArcturusZhang, line 846: We can only return null here because to demo how to use a method on a client, the first step is to create a client instance, but if all the constructors are changed to internal, we are not able to write any example (if we do not change the logic here)

m-nash (commented):
Can we add a test case validating this?

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -163,24 +163,12 @@ internal static CSharpType FromSystemType(Type type, string defaultNamespace, So
...
-        {
-            if (!IsFrameworkType)
-                return false;
-
-            return FrameworkType.Equals(typeof(IList<>)) ||
-                FrameworkType.Equals(typeof(IEnumerable<>)) ||
-                FrameworkType == typeof(IReadOnlyList<>) ||
-                FrameworkType.Equals(typeof(IDictionary<,>)) ||
-                FrameworkType == typeof(IReadOnlyDictionary<,>);
-        }
```
ArcturusZhang, line 176: I remove this method because we already have a static method `IsCollectionType` in `TypeFactory` doing the same thing and it is already widely used and this method only have one usage.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -163,24 +163,12 @@ internal static CSharpType FromSystemType(Type type, string defaultNamespace, So
...
-
         public CSharpType GetNonNullable()
         {
             if (!IsNullable)
                 return this;
 
-            return new CSharpType(Implementation, Namespace, Name, IsValueType, IsEnum, false, Arguments);
+            return IsFrameworkType ? new CSharpType(FrameworkType, false, Arguments) : new CSharpType(Implementation, Namespace, Name, IsValueType, IsEnum, false, Arguments);
```
ArcturusZhang, line 171: I change this because it throws when we call this method on a framework type.

ArcturusZhang:
Hi @m-nash I have added a use case in our test project, please take a look.

test/TestProjects/Customizations-Typespec/Generated/Docs/CustomizationsInCadlClient.xml
```diff
@@ -1,309 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <doc>
   <members>
-    <member name="RoundTripAsync(RequestContent,RequestContext)">
```
m-nash, line 4: why is this doc now empty?

ArcturusZhang: Because this client now does not have a public ctor, we do not know a way to construct it and all the content in this doc is depending on the public ctor, therefore the doc is empty by design.

m-nash: Does this mean it has a method from its parent to retrieve it?  If so are we not able to get that information to still provide an example?

ArcturusZhang: The generator will not know that. The generator is taking the accessibility on customized code for the constructors, it only sees the client no longer has public constructors now. There might be a way to get the client in other ways, like those are written in customization code, but the generator will not know that.
Therefore in order not to generate invalid example code (since the ctor is not public), I choose we just generate nothing for the documents.

m-nash: Let see if we can find this information.  It seems doable to determine if there is a method or property on any other client that returns this client and if so we know how its constructed.

Issue #3057, opened 2023-02-01
Support @projectedName for CADL+DPG


lirenhe:
There is a feedback from archboard review meeting to rename a model in OpenAI.
We need to enable to support for @projectedName to handle this requirement.

Please check https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/projection/main.cadl for the test project.

CADL-dpg support is already done in https://github.com/Azure/cadl-azure/pull/2462.

ArcturusZhang:
Some useful document: https://microsoft.github.io/cadl/standard-library/projected-names

pshao25:
Please add tests by:
1. Add "projection" to `$cadlRanchProjectPaths` in `eng\generated.ps1` and generate the SDK.
2. Add corresponding E2E tests under test\CadlRanchProjects.Tests.

PR #3147, opened 2023-02-17
Support `@projectedName`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3057
Regen: no changes in azure-sdk-for-net

# Description

# Checklist

# Ready to Land?


archerzz (approved):

pshao25 (changes requested):
Please add tests from cadl-ranch, not make it a made-up test.

archerzz (approved):

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -512,15 +511,25 @@ export function getInputType(
                     isReadOnly = true;
                 }
                 if (isNeverType(value.type) || isVoidType(value.type)) return;
+                const jsonKey = "json";
+                const csharpKey = "csharp";
+                const clientKey = "client";
```
pshao25, line 516: Let's create a new constant file and put them in.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -512,15 +511,25 @@ export function getInputType(
...
+                    projectedNamesMap?.get(csharpKey) ??
+                    projectedNamesMap?.get(clientKey) ??
+                    value.name;
+                const serializedName =
+                    projectedNamesMap?.get(jsonKey) ?? value.name;
                 const inputProp = {
-                    Name: value.name,
-                    SerializedName: value.name,
+                    Name: name,
+                    SerializedName: serializedName,
```
pshao25, line 526: What if projectName and friendlyname exist together, I think friendlyname wins.

ArcturusZhang: Do we have a place to search the "target" of those decorator? Currently we only handles `@friendlyName` on **models**. But `@projectedName` works on **properties**.
I do not know if `@friendlyName` could work on properties, or `@projectedName` could work on models.
If both not, we do not need to worry about it.

pshao25: The definition of `@friendlyName` is 
```
export function $friendlyName(
  context: DecoratorContext,
  target: Type,
  friendlyName: string,
  sourceObject: Type | undefined
)
```,
so it could be applied on any type.

Comment (lirenhe) [reply src/CADL.Extension/Emitter.Csharp/src/lib/model.ts:526]: @pshao25,  could you share the doc link of @friendlyName?    I am unable to find it in the latest spec.

Comment (ArcturusZhang) [reply src/CADL.Extension/Emitter.Csharp/src/lib/model.ts:526]: > 

this makes sense but this is not the goal of my PR, therefore I will not add support friendlyName for properties in this PR.
We should do not in a separate PR, and we handle "what should happen if a property has both" then.

File: src/CADL.Extension/Emitter.Csharp/src/emitter.ts [hunk 3]
```diff
@@ -305,98 +297,84 @@ export function createModelForService(
...
-            ApiVersions: Array.from(apiVersions.values()),
-            Enums: Array.from(enumMap.values()),
-            Models: Array.from(modelMap.values()),
-            Clients: clients,
-            Auth: auth
-        } as CodeModel;
-        return clientModel;
-    } catch (err) {
-        if (err instanceof ErrorTypeFoundError) {
-            return;
```
ArcturusZhang, line 394: I removed this try catch because I found that outside this function, we never check if the code model returned in this function is undefined or not therefore we just throw the error out when something wrong happens instead of return undefined. This could also enable this function to return with a concrete type `CodeModel` which previously is `any`

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
pshao25: Double checked we only applied on models, so agreed to not consider this corner case now.

pshao25: > @pshao25, could you share the doc link of @friendlyName? I am unable to find it in the latest spec.

https://github.com/microsoft/typespec/blob/ec21edc66926da0541a4d645be4ccd9f260add01/packages/compiler/lib/decorators.ts#L708-L712

pshao25 (approved):
Looks good to me.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
ArcturusZhang: Resolving this because we currently did not support friendlyName on properties.

lirenhe (approved):

Issue #3068, opened 2023-02-03
build(common): split SDK code generation into multiple tasks


archerzz:
## Description

Our job to generate all SDKs on `azure-sdk-for-net` take too much time. To reduce the timeout failure and boost the speed, we could consider splitting the job into multiple tasks, like what we did for testing.

PR #3153, opened 2023-02-19
build(publish): parallel code gen on azure-sdk-for-net

### LINKED ISSUE #3067: build(common): add generate sample into a separate task
## Description

To reduce the impact, we should separate sample generation from code generation. And we can consider:
- Either we generate sample in a separate task inside the `Autorest C# publish` pipeline, in parallel with code generation
- Or we create a cron job to generate samples periodically

archerzz:
- split job `Update_azure_sdk_for_net` into 3 stages:
  - `Update_Generator_Versions`: change the generator versions only, create the first commit to new branch `auto-update-autorest-${Version}`
  - `Update_Codes_X_Y` and `Update_Samples_X_Y`: update SDK codes or sample codes from service of which name starts with `X` to `Y`.
  - `Create_PR`: create PR from branch `auto-update-autorest-${Version}` to `main`
- add retry for `Update_Codes_X_Y` and `Update_Samples_X_Y`

resolve #3067 #3068 

- trial run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2201624&view=results
- PR generated: https://github.com/Azure/azure-sdk-for-net/pull/34361

---

image

# Checklist

# Ready to Land?


archerzz:
Other benefits:
- Re-running the failed tasks will be much easier. We don't need to start over from the very beginning again.
- For PR with thousands of changes, people can choose to review each commit (there are 12 commits in total).
- Each change on `auotrest.csharp` is now separately published to `azure-sdk-for-net`. So if a change screws the `azure-sdk-for-net`, people can still merge all previous changes.

eng/UpdateAzureSdkCodes.ps1
```diff
@@ -0,0 +1,12 @@
+param([string]$Branch, [string]$SdkRepoRoot, [string[]]$ServiceDirectoryFilters)
+
+Write-Host "Generating Azure SDK Codes..."
+foreach ($filter in $ServiceDirectoryFilters) {
+    Write-Host 'Generating projects under service directory ' -ForegroundColor Green -NoNewline
+    Write-Host "$filter" -ForegroundColor Yellow
+    dotnet msbuild /restore /t:GenerateCode /p:ServiceDirectory=$filter "$SdkRepoRoot\eng\service.proj"
+    if ($LastExitCode -ne 0) {
+        Write-Error 'Generation error'
```
ArcturusZhang, line 9: is there a way that we could emphasize the lines that reports the errors on the script finishes?

eng/UpdateAzureSdkForNet.ps1
```diff
@@ -1,20 +1,18 @@
...
+Write-Host "Generating Azure SDK Codes..."
+foreach ($filter in $ServiceDirectoryFilters) {
+    Write-Host 'Generating projects under service directory ' -ForegroundColor Green -NoNewline
+    Write-Host "$filter" -ForegroundColor Yellow
+    dotnet msbuild /restore /t:GenerateCode /p:ServiceDirectory=$filter "$SdkRepoRoot\eng\service.proj"
+    if ($LastExitCode -ne 0) {
+        Write-Error 'Generation error'
+        exit 1
+    }
+}
```
ArcturusZhang, line 18: This part is identical with the script `eng/UpdateAzureSdkCodes.ps1`, should we call that script instead here?

eng/UpdateAzureSdkGeneratorVersions.ps1
```diff
@@ -0,0 +1,7 @@
+param([string]$AutorestCSharpVersion, [string]$CadlEmitterVersion, [string]$SdkRepoRoot)
+
+Import-Module "$PSScriptRoot\UpdateGeneratorMetadata.psm1" -DisableNameChecking -Force;
+
+$SdkRepoRoot = Resolve-Path $SdkRepoRoot
+
+UpdateGeneratorMetaData $AutorestCSharpVersion $CadlEmitterVersion $SdkRepoRoot
```
ArcturusZhang, line 7: This script also looks identical with the first part of `eng/UpdateAzureSdkForNet.ps1`, should we call this script in that?

eng/pipelines/build.yml
```diff
@@ -163,34 +166,83 @@ stages:
...
         - Release
       jobs:
-        - job: Update
-          timeoutInMinutes: 180
-          pool:
-            vmImage: ubuntu-20.04
+        - template: update-generator-versions.yml
+# update codes
+        - template: update-azure-sdk-for-net-codes.yml
+          parameters:
```
Yao725, line 172: I'm curious about how we determined the filter. Is it determined by the number of RPs under different filter? For example, the number of RPs that the name starts with `a,b,c` is equal to the RPs that starts with `d,e`.

archerzz: > I'm curious about how we determined the filter. Is it determined by the number of RPs under different filter? For example, the number of RPs that the name starts with `a,b,c` is equal to the RPs that starts with `d,e`.

It's just a guess. I actually tuned the division once. We can further optimize.

Yao725: Got it, thanks, I think we can do something like [here](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/scripts/splittestdependencies/generate-dependency-functions.ps1#L4) in the future to split the projects under SDK directory and dynamically generate the corresponding jobs accordingly.

archerzz: @Yao725 file https://github.com/Azure/autorest.csharp/issues/3165 to track

eng/UpdateAzureSdkCodes.ps1
archerzz: It's a known issue (or by design) that `dotnet msbuild` is different from `msbuild`. You need to either set the verbose level to normal (default is minimal) or use a file logger. https://github.com/dotnet/msbuild/issues/4288

m-nash (approved):
@mikeharder can you please take a look as well since you are helping in the same area.  This gets us from 3+ hours to roughly 35 minutes on the regen portion of the publish pipeline.

If you have even more improvements lets open those as trackers so we can make incremental progress but if you see any major issues with this let us know.

eng/UpdateAzureSdkCodes.ps1
ArcturusZhang: well at least we have a clear information to tell us it failed. Previously it just outputs nothing

mikeharder:
@m-nash, @archerzz: Are there example builds before and after this change I can look at?

eng/UpdateAzureSdkCodes.ps1
archerzz: I only enabled it for CI, since we split the build. Each job will not print too many logs.

archerzz:
> @m-nash, @archerzz: Are there example builds before and after this change I can look at?
> 
> I see an example build at the top of this PR: [dev.azure.com/azure-sdk/internal/_build/results?buildId=2201624&view=results](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2201624&view=results). I assume I should just compare this to a similar build from another PR?

@mikeharder Recently we don't change the pipeline except this PR. So any build on `feature/v3` branch should before this change, like [this one](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2204670&view=results).

eng/pipelines/update-azure-sdk-for-net-codes.yml
```diff
@@ -0,0 +1,53 @@
...
+        -ServiceDirectoryFilters ${{ parameters.filter}}
+        -ShowSummary
+      failOnStderr: false
+      workingDirectory: $(Build.SourcesDirectory)/autorest.csharp
+    retryCountOnTaskFailure: 3
+#   - pwsh: |
+#       git restore eng/Packages.Data.props
+#       git restore eng/emitter-package.json
+#     workingDirectory: $(Build.SourcesDirectory)/azure-sdk-for-net
+  - template: /eng/common/pipelines/templates/steps/git-push-changes.yml@azure-sdk-tools
```
weshaggard, line 45: Why are each of these templates pushing changes? It looks like they are pushing changes to the same branch even. I wouldn't expect we would push any changes unless all the validation has completed.

eng/pipelines/build.yml
```diff
@@ -163,34 +166,83 @@ stages:
...
+            name: Update_Codes_S_U
+            filter: "s*,t*,u*"
+        - template: update-azure-sdk-for-net-codes.yml
+          parameters:
+            name: Update_Codes_V_Z
+            filter: "v*,w*,x*,y*,z*"
+# update samples
+        - template: update-azure-sdk-for-net-samples.yml
+          parameters:
+            name: Update_Samples_A_F
```
weshaggard, line 202: We have support for creating a dynamic matrix so we shouldn't need to statically create these jobs. Have you looked at our matrix generation scripts? https://github.com/Azure/azure-sdk-tools/blob/main/doc/common/matrix_generator.md

As an example we use this to dynamically create our test jobs for the core pipeline in .NET, see https://github.com/Azure/azure-sdk-for-net/pull/29957 for more details.

eng/pipelines/update-azure-sdk-for-net-codes.yml
m-nash: We are doing this partially to support PRs with 5000+ files changed.  By splitting these into commits we can still review each commit since as we know reviewing a PR with 5000+ files doesn't work in pretty much any tool out there.

eng/pipelines/build.yml
m-nash: @weshaggard do you see this as something we can do in a future enhancement PR?  The reason I ask is this hard coded matrix gets us from 3+ hours to 35 minutes which is a massive improvement and if possible I would like to check this in ASAP and increment on it over time.

eng/pipelines/update-azure-sdk-for-net-codes.yml
weshaggard: I believe I'm starting to understand why you are pushing for each job but I will add that this potentially runs the risk of conflicts if files under a shared path change.

m-nash: Agreed that risk is there, but that would break our conventions so if it fails in that case I think I am ok with that.  Worst case we can hard code common paths to run first.

eng/pipelines/build.yml
weshaggard: I think it can be a future improvement. It will definitely help us more easily tune the number of jobs vs the number of libraries in job just by tweaking some values in a setup script instead of trying to manually partition.

eng/pipelines/update-azure-sdk-for-net-codes.yml
archerzz: > I believe I'm starting to understand why you are pushing for each job but I will add that this potentially runs the risk of conflicts if files under a shared path change.

> Agreed that risk is there, but that would break our conventions so if it fails in that case I think I am ok with that. Worst case we can hard code common paths to run first.

AFAIKS, there are 2 ways to resolve that:
1. If it's a rare case, we can ask developers to manually create a PR to update `azure-sdk-for-net`, which we have done before.
2. Or we change to use our own scripts to push changes to github. The overlapped part should have the same changes. I did a test on local, such conflict can be resolved by `git pull --rebase`.

eng/pipelines/build.yml
archerzz: I've already filed an issue to track this: https://github.com/Azure/autorest.csharp/issues/3165

archerzz:
Here is the comparison of existing pipeline and new pipeline.

## Existing Pipeline

- latest run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2210555&view=results
- PR created (549 files changed): https://github.com/Azure/azure-sdk-for-net/pull/34465

## New Pipeline

- latest run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2211526&view=results
- PR created (549 files changed): https://github.com/Azure/azure-sdk-for-net/pull/34495

PR #3158, opened 2023-02-21
Collection format

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3114

- emit `    ArraySerializationDelimiter` and `Explode` according to the collectionFormat
- add sample test

# Checklist

# Ready to Land?


chunyu3:
Flash the azure-sdk-for-net. there is no any changes. The sdk branch: https://github.com/chunyu3/azure-sdk-for-net/tree/collectionFormat

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -781,11 +787,37 @@ function loadOperation(
...
+            Kind: kind,
+            ArraySerializationDelimiter: getArraySerializationDelimiter(format)
         } as InputParameter;
     }
 
+    function getArraySerializationDelimiter(
+        format: string | undefined
+    ): string | undefined {
+        if (!format) return undefined;
+        switch (format) {
```
ArcturusZhang, line 805: Can switch in js work on `undefined`? If so we do not need to check if format is undefined or not, it just fall into the case `default`

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -781,11 +787,37 @@ function loadOperation(
...
             IsEndpoint: false,
             SkipUrlEncoding: false, //TODO: retrieve out value from extension
-            Explode: false,
-            Kind: kind
+            Explode:
+                (inputType as InputListType).ElementType &&
+                format &&
+                format === "multi"
+                    ? true
+                    : false,
```
ArcturusZhang, line 795: looks like this condition is just: `format === "multi" && inputType is InputListType`.
Could we simplify this?
As far as I know, if `format` is null or undefined, `format === "multi" will be false, therefore the check for `format` itself is redundant.

ArcturusZhang (approved):

pshao25 (dismissed):
We already have cadl-ranch tests: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/collectionFormat/main.cadl

Let's use this.

src/CADL.Extension/Emitter.Csharp/src/type/CollectionFormat.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+export enum CollectionFormat {
```
pshao25, line 4: We have constant.ts. Can we put it into that file?

src/AutoRest.CSharp/CommandLineOptions.cs
```diff
@@ -28,7 +28,7 @@ internal class CommandLineOptions
         [Option(longName: "server", Required = false, Default = null, HelpText = "Server argument.")]
         public string? Server { get; set; }
 
-        [Option('c', "clear-output-folder", Required = false, Default = false, HelpText = "Clear the output folder before generating code.")]
+        [Option('X', "clear-output-folder", Required = false, Default = false, HelpText = "Clear the output folder before generating code.")]
```
pshao25, line 31: Strictly, it is breaking change. But I actually don't have any other ideas except break this.

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: updated

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: We cannot use `inputType is InputListType` because of build failure. I removed `format` check to simplify.

src/CADL.Extension/Emitter.Csharp/src/type/CollectionFormat.ts
chunyu3: It is the different type of collection format, it is enum, not a constant.

chunyu3:
> We already have cadl-ranch tests: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/collectionFormat/main.cadl
> 
> Let's use this.

We need to update the cadl ranch collectionFormat mockapi. It will take time to release.

src/CADL.Extension/Emitter.Csharp/src/type/CollectionFormat.ts
ArcturusZhang: should we have a dictionary from the enum values here to the delimiter as a constant?
Currently our delimiters might be hard-coded in the code.

src/AutoRest.CSharp/CommandLineOptions.cs
ArcturusZhang: I believe this should be acceptable because it never worked previously

m-nash: should we go with `x` since all the others are lower case?

chunyu3: updated.

m-nash (approved):

src/CADL.Extension/Emitter.Csharp/src/type/CollectionFormat.ts
chunyu3: updated

PR #3161, opened 2023-02-21
Use StringComparer.OrdinalIgnoreCase.Equals for ASCII values

AlexanderSher:
`StringComparer.OrdinalIgnoreCase.Equals` is 7 to 8 times faster than `string.Equals` with `StringComparison.InvariantCultureIgnoreCase`. 

It should be safe to use it for ASCII enum values parsing instead of invariant culture comparison.


chunyu3 (approved):

m-nash (approved):

m-nash (changes requested):
Can you add the regen PR?

AlexanderSher:
> Can you add the regen PR?

Sure: https://github.com/Azure/azure-sdk-for-net/pull/34470

m-nash (approved):

PR #3162, opened 2023-02-21
Simplify the generation script

ArcturusZhang:
# Description

I noticed that we have too many configurations inside our `eng/Generate.ps1` script. Such as, some projects need to have some special flags to work, the name list of the projects we are going to generate and run. These things make our generate script cumbersome and hard to read.

In this PR, 
- I removed the special options for some cadlRanch projects, instead, I introduced `tspconfig.yaml` file in those projects to hold their special options.
- I extracted the list of cadl ranch project name list and test server test name list into a json file, so that in the future if we have a new test to add, we do not have to modify the script.

# Checklist

# Ready to Land?


ArcturusZhang:
Hi @m-nash could you take a look at this PR?

test/CadlRanchProjects/inheritance/Generated/Configuration.json
```diff
@@ -5,5 +5,6 @@
   "SharedSourceFolders": [
     "../../../../artifacts/bin/AutoRest.CSharp/Debug/net6.0/Generator.Shared",
     "../../../../artifacts/bin/AutoRest.CSharp/Debug/net6.0/Azure.Core.Shared"
-  ]
+  ],
+  "unreferenced-types-handling": "keepAll"
```
pshao25, line 9: Why we have this change in configuration? Do we need this?

pshao25 (approved):

test/CadlRanchProjects/arrays/item-types/cadl-project.yaml
```diff
@@ -0,0 +1,3 @@
+options:
```
m-nash, line 1: Where is this file getting consumed?  I don't see it referenced in the generate script and why are we only adding this for CadlRanchProjects and not the TestProjects that are also cadl like FirstTest-Cadl

pshao25: > Where is this file getting consumed? I don't see it referenced in the generate script

We run emitter from path test/CadlRanchProjects/arrays/item-types, it will automatically load the cadl-project.yaml from that where it is called. In this case, it is test/CadlRanchProjects/arrays/item-types.

test/CadlRanchProjects/inheritance/Generated/Configuration.json
ArcturusZhang: I was told that this `inheritance` project is not a cadl ranch project, therefore I should not add this config in its cadl-project.yaml. I will remove this.

test/CadlRanchProjects/arrays/item-types/cadl-project.yaml
ArcturusZhang: I am not quite sure exactly how is the mechanism this is loaded.
Indeed we can add this for other cadl test projects and it should work fine? I could have a try.
But the reason I did not add this file for other test projects in cadl is just because they do not have extra configurations I assume.

m-nash: yeah I would love to have the ability to have test projects that target various combinations of cadl-project.yaml settings so if we can get this working for all that would be ideal

ArcturusZhang: Hi @m-nash I had a local test that adds the `cadl-project.yaml` file to a random project with some special options, and those new options could take effect. Therefore this way works for all test projects. I did not add this file for other projects because they do not need any extra options

test/CadlRanchProjects/inheritance/Generated/Configuration.json
ArcturusZhang: reverted

test/CadlRanchProjects/arrays/item-types/cadl-project.yaml
ArcturusZhang: Well I now took a step forward that I added an empty cadl-project.yaml to all the cadl test project and samples, and also changed the script to require a `readme.md` or a main json with the same name as its directory name, or a `cadl-project.yaml` file, otherwise the script will just throw an error requiring it.

m-nash (approved):

test/CadlRanchProjects/arrays/item-types/cadl-project.yaml
pshao25: I don't quite follow why it will fail if there is no cadl-project.yaml, but if we have to generate an empty one, we could put a comment like readme.me as:
```
# Run `dotnet build /t:GenerateCode` to generate code.
```

chunyu3: `cadl-project.yaml` is not required, so we should not throw exception if this file does not exist. And `cadl-project.yaml` is not an specific configure for sdk generation, it is not need to add c# sdk generation command in it.

pshao25: Yes, I agree "we should not throw exception". @ArcturusZhang could give more details? What's the error?

For the c# sdk generation command, it just because when I generate code from sdk repo, I always have to go to another SDK folder to find that command. It would be better if we could find somewhere to put this command.

ArcturusZhang: @pshao25 @chunyu3 this is only for the test projects. Nothing to do with the real RPs in azure-sdk-for-net.
The thought here from Michael is that we ensure every test project to have a `cadl-project.yaml` file (this is the thing that the script is checking).
This helps people understand, if they want some special parameters in their test projects, where they could put them

m-nash: I think a test requirement only is fine as long as this doesn't impose any requirements when we generate against azure-sdk-for-net which I don't think it will.

PR #3163, opened 2023-02-21
feat(common): support null value for reference type collection elements

### LINKED ISSUE #3159: feat(common): support null value of reference type collection element in deserialization
## Description

This is the follow-up of #3084. Right now, our generated deserialization logic will not handle null value of reference type collection element. This will fail the OpenAI service which will return `null` for [CompletionsLogProbability.CompletionsLogProbability](https://github.com/Azure/azure-sdk-for-net/blob/bc5930762b8f33ce0df7929baf0662748c96b719/sdk/openai/Azure.AI.OpenAI/src/Custom/CompletionsLogProbability.cs#LL20C58-L20C58).

### Expected Deserialization Logic

In the customization codes, we add null value handling: https://github.com/Azure/azure-sdk-for-net/blob/bc5930762b8f33ce0df7929baf0662748c96b719/sdk/openai/Azure.AI.OpenAI/src/Custom/CompletionsLogProbability.Serialization.cs#L70-L87

```csharp
if (item.ValueKind == JsonValueKind.Null)
{
    array.Add(null);
}
```

### Actual Generated Logic

See previous version which is generated: https://github.com/Azure/azure-sdk-for-net/blob/a4392f3e9da4864fb98457ba41fc93fe82a11d97/sdk/openai/Azure.AI.OpenAI/src/Generated/CompletionsLogProbability.Serialization.cs#L70-L80

It doesn't handle null value.

It's a known issue for MPG as well.

archerzz:
- update `SerializationBuilder` to set `ValueSerialization.IsNullable : true` for collection elements of reference type
  - that will cause null value handling in serialization & deserialization logic to be added for collection elements of reference type
- update `SerializationWriter` so that in generated model deserialization methods, null value handling (return null if pass in null json value) is added at the beginning
- update `Utf8JsonWriterExtension.WriteBase64StringValue()` to handle null json value in serialization
- update `JsonElementExtension.GetBytesFromBase64()` to handle null json value in deserialization
---
update `JsonCodeWriterExtension`, for collection items:
- In list serialization, instead of generating null value handling logic for all nullable value serialization, only generating the logic for nullable value types. That can eliminate a lot null handling codes in serialization method. And for those without such logic:
  - string type can be handled by `Utf8JsonWriter.WriteStringValue()` which has null value support
  - `byte[]` is handled by the modified `Utf8JsonWriterExtension.WriteBase64StringValue()` (see the change above)
  - model classes and other reference types are handled by `Utf8JsonWriterExtension.WriteObjectValue()` which has null value support
- In dictionary serialization, add the missing null value handling for values
- In list & dictionary deserialization, change the generated codes to:
  - model class null value check is handled by their own deserialization method (see the change above)
  - string null value check in handled by `JsonElement.GetString()`
  - `byte[]` is handled by `JsonElementExtension.GetBytesFromBase64()`
  - nullable value type, list/dictionary, other framework reference will have generated null value check codes

---
- fix 3 test cases about null value in collection elements in deserialization

resolve #3159

### Serialization

| Serialization        | Null check in                                     | example           | comments                                             |
| -------------------- | ------------------------------------------------- | ----------------- | ---------------------------------------------------- |
| nullable value type  | generated codes in parent serialization method                       |                   | `if (value==null) { writer.WriteNullValue()}` |
| SDK model            | Utf8JsonWriterExtensions.WriteObjectValue()       |                   |                                                      |
| string               | Utf8JsonWriterExtensions.WriteStringValue()       |                   |                                                      |
| byte[]               | Utf8JsonWriterExtensions.WriteBase64StringValue() |                   | need to add null check in WriteBase64StringValue     |
| array, dictionary  | generated codes in parent serialization method                       |                   | `if (value==null) { writer.WriteNullValue()}` |
| other reference type | generated codes in parent serialization method       |          |   `if (value==null) { writer.WriteNullValue()}`      |

### Deserialization

| Serialization                  | Null Check in                             | example   | comments                                                                       |
| ------------------------------ | ----------------------------------------- | --------- | ------------------------------------------------------------------------------ |
| nullable value type            | generated codes in parent deserialization method             | int?      | `if (element.Value.ValueKind == JsonValueKind.Null) { array.Add(null)}` |
| SDK model                      | model deserialization method              | IPAddress | need to add null check in deserialization method                               |
| string                         | JsonElement.GetString()                   |           |                                                                                |
| byte[]                         | JsonElementExtension.GetBytesFromBase64() |           |      need to add null check in GetBytesFromBase64                  |
| array, dictionary              | generated codes in parent deserialization method             |           | `if (element.Value.ValueKind == JsonValueKind.Null) { array.Add(null)}` |
| other framework reference type | generated codes in parent deserialization method             |     | `if (element.Value.ValueKind == JsonValueKind.Null) { array.Add(null)}` |

# Checklist

# Ready to Land?


archerzz:
The last commit adds an optimization: put null value handling logic in deserialization into each model deserialization method, so that less `if (element.ValueKind == NullKind)` is generated.
Also it contains a minor optimization which will not generate null value handling for string elements, because `JsonElement.GetString()` will handle null value.

m-nash:
I believe we want to get your parallel work in before uploading the regen PR?  If so lets mark this as draft until we get ready for that.

samples/Azure.Management.Storage/Generated/Models/BlobInventoryPolicyFilter.Serialization.cs
```diff
@@ -51,6 +51,10 @@ void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
 
         internal static BlobInventoryPolicyFilter DeserializeBlobInventoryPolicyFilter(JsonElement element)
         {
+            if (element.ValueKind == JsonValueKind.Null)
```
m-nash, line 54: It appears we have 2 pattern changes in here?  I would like to keep this uber large regens to a single pattern if possible as it makes it much easier to review.  Lets keep this one focused on the null element in dictionary / list and we can add this in a separate PR.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -139,23 +139,23 @@ private static bool IsManagedServiceIdentityV3(Schema schema, CSharpType type)
...
+                CodeModelType codeModelType => BuildSerialization(codeModelType.Schema, valueType, isCollectionElement),
+                InputListType listType => new JsonArraySerialization(TypeFactory.GetImplementationType(valueType), BuildJsonSerialization(listType.ElementType, TypeFactory.GetElementType(valueType), true), valueType.IsNullable || (isCollectionElement && !valueType.IsValueType)),
+                InputDictionaryType dictionaryType => new JsonDictionarySerialization(TypeFactory.GetImplementationType(valueType), BuildJsonSerialization(dictionaryType.ValueType, TypeFactory.GetElementType(valueType), true), valueType.IsNullable || (isCollectionElement && !valueType.IsValueType)),
+                _ => new JsonValueSerialization(valueType, GetSerializationFormat(inputType, valueType), valueType.IsNullable || (isCollectionElement && !valueType.IsValueType)) // nullable CSharp type like int?, Etag?, and reference type in collection
             };
         }
 
-        private static JsonSerialization BuildSerialization(Schema schema, CSharpType type)
+        private static JsonSerialization BuildSerialization(Schema schema, CSharpType type, bool isCollectionElement)
```
m-nash, line 158: Does this need to be a separate property?  Is the bool reflecting whether or not `type` is a collection?  If so can we add a property to CSharpType to return this for us?

archerzz: It represents whether the `JsonSerialization` is for element inside a collection (list or dictionary), because here we only handle the null value issue for collection element, excluding singular properties.

`JsonSerialzation.IsNullable` is different from `CSharpType.IsNullable`. The former is actually `AllowNull`, and the latter is about whether the type is nullable (e.g. `int` v.s. `int?`)

samples/Azure.Management.Storage/Generated/Models/BlobInventoryPolicyFilter.Serialization.cs
archerzz: Right. The simplest solution is for every `JsonSerialization.IsNullable : true`, add null check in serialization/deserialization codes. The side effect is we'll have a lot unnecessary null check.

- `string`: both `Utf8JsonWriterExtensions.WriteStringValue()` and `JsonElement.GetString()` can handle null value correctly
- `byte[]`: we can modify it's serialization/deserialization method to handle null value correctly
- custom models: add the null value handling in deserialization method, which could also benefit for future work if we want to support nullable optional properties

For the 1st stage, we could roll back to the simplest solution.

m-nash (approved):

archerzz:
Here is the generated preview PR: https://github.com/Azure/azure-sdk-for-net/pull/34624

lirenhe (approved):

PR #3166, opened 2023-02-22
add dotnet build parameter to set cadl options

chunyu3:
# Description
Fix https://github.com/Azure/azure-sdk-for-net/issues/34357

- add dotnet build parameter to set cadl emitter options

# Checklist

# Ready to Land?


ArcturusZhang (approved):
This is very helpful, we no longer need to manually change the build script to add the extra flag

src/AutoRest.CSharp/build/CodeGeneration.targets
```diff
@@ -16,6 +16,13 @@
       use the following command line (remove the space between minus minus): dotnet msbuild /t:GenerateCode /p:AutoRestAdditionalParameters="- -interactive"
     -->
 
+    <!--
+      <CadlAdditionalOptions></CadlAdditionalOptions>
+     
+      Allows passing additional cadl command line arguments, for example to run special emitter 
+      use the following command line: dotnet msbuild /t:GenerateCode /p:CadlAdditionalOptions="@azure-tools/cadl-csharp.csharpGeneratorPath=/absolute/path/to/artifacts/bin/AutoRest.CSharp/Debug/net6.0/AutoRest.CSharp.dll"
```
ArcturusZhang, line 23: One question: how do we append multiple options using this new parameter?

chunyu3: CadlAddiotnalOptions will accept multiple options  as one string and sperate by semicolon

readme.md
```diff
@@ -98,17 +104,11 @@ When the automatic PR is created for azure-sdk-for-net if there are any issues f
...
 - At this point there are several ways to generate but the idea is to generate and test all projects that will be affected by your PR.  There are tools such as [this](https://github.com/ArcturusZhang/Regen) which are written to handle specific cases.  If you are unsure you should run against all projects to be safe.
 
 To regen and test everything in azure-sdk-for-net after you have updated to use your new local build do the following:
 
-- First generate all projects in the repo by executing `dotnet build [RepoRoot]/eng/service.proj /t:GenerateCode`
+- First generate all projects in the repo by executing `dotnet build [RepoRoot]/eng/service.proj /t:GenerateCode`. If your generator PR needs to apply to typespec/cadl projects, or your generator PR changes anything in our emitter (the `src\CADL.Extension\Emitter.Csharp` project), you will need to run `dotnet build [RepoRoot]/eng/service.proj /t:GenerateCode  /p:CadlAdditionalOptions="@azure-tools/cadl-csharp.csharpGeneratorPath=/absolute/path/to/artifacts/bin/AutoRest.CSharp/Debug/net6.0/AutoRest.CSharp.dll`.
```
m-nash, line 111: I think we should just have key/value pairs here and have the script wrap them into cadl syntax.

`dotnet build /t:GenerateCode /p:CadlAdditionalOptions=csharpGeneratorPath=/absolute/path/to/artifacts/bin/AutoRest.CSharp/Debug/net6.0/AutoRest.CSharp.dll`

Inside the script we should wrap this using the existing methods to get the emitter name.

m-nash: I also wonder if we should have something "simpler" like `dotnet build /t:GenerateCode /p:UseLocalBuild` which can look at the `emitter-package.json` file and get the local build path from there since it will be in the same location relatively?

ArcturusZhang: > I also wonder if we should have something "simpler" like `dotnet build /t:GenerateCode /p:UseLocalBuild` which can look at the `emitter-package.json` file and get the local build path from there since it will be in the same location relatively?

That would be very convenient, vote for that

m-nash: I should have thought about the name before making the suggestion but if we go that route it should be more precise like `UseLocalGenerator` or something like that.

chunyu3: We may need to override other options beside 'csharpGeneratorPath'. And it is common library for all languages. There will be special option for other language to override.
So I think we shall give out a way to set options.

chunyu3: > I also wonder if we should have something "simpler" like `dotnet build /t:GenerateCode /p:UseLocalBuild` which can look at the `emitter-package.json` file and get the local build path from there since it will be in the same location relatively?

I update to remove emitter name in the parameter, just need to set key-value pair.

chunyu3: > I think we should just have key/value pairs here and have the script wrap them into cadl syntax.
> 
> `dotnet build /t:GenerateCode /p:CadlAdditionalOptions=csharpGeneratorPath=/absolute/path/to/artifacts/bin/AutoRest.CSharp/Debug/net6.0/AutoRest.CSharp.dll`
> 
> Inside the script we should wrap this using the existing methods to get the emitter name.


I update to remove emitter name in the parameter, just need to set key-value pair.

ArcturusZhang (approved):

readme.md
```diff
@@ -10,12 +10,18 @@ ___
...
-- [Customizing the generated code](#customizing-the-generated-code)
+- [C# code generator for AutoRest V3](#c-code-generator-for-autorest-v3)
+  - [Prerequisites](#prerequisites)
+  - [Build](#build)
+  - [Test](#test)
+    - [Testing Details](#testing-details)
+    - [Validate generator changes against Azure SDK before merging your autorest.csharp PR](#validate-generator-changes-against-azure-sdk-before-merging-your-autorestcsharp-pr)
+  - [Use in `azure-sdk-for-net` repo](#use-in-azure-sdk-for-net-repo)
+    - [PR Integration with Azure SDK Repository](#pr-integration-with-azure-sdk-repository)
+  - [Use outside of the `azure-sdk-net` repo](#use-outside-of-the-azure-sdk-net-repo)
```
ArcturusZhang, line 21: should be
```suggestion
  - [Use outside of the `azure-sdk-for-net` repo](#use-outside-of-the-azure-sdk-net-repo)
```

lirenhe (approved):

PR #3175, opened 2023-02-27
Split extension client to ensure the mocking

Yao725:
Regen: https://github.com/Azure/azure-sdk-for-net/pull/35294

# Description

## Splitting rules
Previously our extension clients are generated by "which resource it is extending" (ie, by the type of `this` parameter of the extension method), for instance
```csharp
public static Response<XX> Foo(this ResourceGroupResource resourceGroupResource, Bar bar)
```
will go to the `ResourceGroupResourceExtensionClient`.
This PR changes the splitting logic, now we put the operation into its corresponding extension client by its `Resource` property.
In the generator, each operation is represented by `MgmtClientOperation` and it has a property `Resource` indicating which resource this operation is manipulating. An extension method could be manipulating another resource.
For instance, if this operation:
```csharp
public static Response<XX> Foo(this ResourceGroupResource resourceGroupResource, Bar bar)
```
has a `Resource` property of `BarResource`, we will now put it into a new `BarResourceExtensionClient`. Of course, as fall back, if the `Resource` property on an operation is null, we will still put it into the extension client of its `this` parameter.

## Refactors around the "Extension" type providers and writers in the generator

Previously our extension type providers and writers are a bit messy, this PR also introduces some refactoring on those.

### Naming of the extension related type providers and writers

Previously some of them have a plural name (like MgmtExtensions) some do not (like `MgmtExtensionClient`), and plural words makes it difficult when we want to describe a list of MgmtExtensions, and introduces misleading around it.

This PR unified them to all use singular words: MgmtExtension, MgmtExtensionWrapper, MgmtExtensionClient and ArmClientExtension.

### Logic between ArmCore and others

In `Azure.ResourceManager` (armCore) we do not have the big extension class, instead, we output all the extensions individually into their own files as non-static classes with the same name, so that those methods are instance methods because those things we would like to extend on, are defined in this library (for instance, ResourceGroupResource).

Extension Clients are patches on the extension methods because extension methods are static, therefore since those methods are not static, ArmCore does not ned extension clients.

Therefore we do not write the `MgmtExtensionWrapper` or `MgmtExtensionClient` when it is ArmCore. We only write the individual MgmtExtension.

In other libraries, we do not write the individual `MgmtExtension` and we write the big `MgmtExtensionWrapper` into a big static classes. And we write the extension clients to enable mocking for our customers.

Summary:
| type provider | writes in ArmCore? | writers in others? |
| --- | --- | --- |
| `MgmtExtensionWrapper` | No | Yes |
| List of `MgmtExtension` | Yes | No |
| List of `MgmtExtensionClient` | No | Yes |

They never overlaps, therefore this PR changes the logic around that to be clear.

# Checklist

# Ready to Land?


Yao725:
This PR is still in progress, and I think we still have two things this PR to do:

- Improve the way that we split the extension client. Now I just will move the method from the old extension client to a new extension client when the return type of this method is one of the resources, but the results don't look as good as they should. We might could also move the `GetCollection` and `GetResource` method to further slim down the original extension client.
- Update the xml documentation for the public method in [RP]Extensions so the customer could know how to mock extension method.

Yao725:
One more thing here is that I'm wondering that if we could do something such as providing a detailed doc to improve the user experience when they want to mock the extension method. For example, if we want to mock the extension method, we need to mock twice as shown in below based on our current implementation, it's not easy for the user to do that without any education.
```csharp
public void MockingSample()
{
        var subscriptionClientMock = new Mock<SubscriptionResourceExtensionClient>();
        CdnNameAvailabilityContent content = new CdnNameAvailabilityContent("falseName", CdnResourceType.Endpoints);
        CdnNameAvailabilityResult result = new CdnNameAvailabilityResult(false, "name already exists.", "name already exists.");
        var response = Response.FromValue(result, null);
        subscriptionClientMock.Setup(subscriptionClient => subscriptionClient.CheckCdnNameAvailabilityWithSubscription(content, default)).Returns(response);

        var subscriptionMock = new Mock<SubscriptionResource>();
        subscriptionMock.Setup(sub => sub.GetCachedClient(It.IsAny<Func<ArmClient, SubscriptionResourceExtensionClient>>())).Returns(subscriptionClientMock.Object);

        var val = subscriptionMock.Object.CheckCdnNameAvailabilityWithSubscription(content);
        Assert.AreEqual(val.Value.NameAvailable, result.NameAvailable);
}
```

samples/Azure.Management.Storage/Generated/Extensions/StorageAccountResourceExtensionClient.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        /// <summary> Initializes a new instance of the <see cref="StorageAccountResourceExtensionClient"/> class. </summary>
+        /// <param name="client"> The client parameters to use in these operations. </param>
+        /// <param name="id"> The identifier of the resource that is the target of operations. </param>
+        internal StorageAccountResourceExtensionClient(ArmClient client, ResourceIdentifier id) : base(client, id)
+        {
+        }
+
+        private ClientDiagnostics StorageAccountClientDiagnostics => _storageAccountClientDiagnostics ??= new ClientDiagnostics("Azure.Management.Storage.Mock", StorageAccountResource.ResourceType.Namespace, Diagnostics);
```
ArcturusZhang, line 35: Should we use `Azure.Management.Storage` here instead of `Azure.Management.Storage.Mock`?
```suggestion
        private ClientDiagnostics StorageAccountClientDiagnostics => _storageAccountClientDiagnostics ??= new ClientDiagnostics("Azure.Management.Storage", StorageAccountResource.ResourceType.Namespace, Diagnostics);
```

src/AutoRest.CSharp/Mgmt/Generation/MgmtExtensionWrapperWriter.cs
```diff
@@ -1,36 +1,105 @@
...
+                    null,
+                    new[] { armClientParameter, scopeParameter });
+                using (_writer.WriteMethodDeclaration(scopeExtensionClientSignature))
+                {
+                    using (_writer.Scope($"return {armClientParameter.Name}.GetResourceClient(() => ", newLine: false))
+                    {
+                        _writer.Line($"return new {extensionClientSignature.ReturnType}({armClientParameter.Name}, {scopeParameter.Name});");
+                    }
+                    _writer.LineRaw(");");
+                }
```
ArcturusZhang, line 87: To be refactored - move these method signatures into the MgmtClientExtension class

src/AutoRest.CSharp/Mgmt/Generation/ArmResourceExtensionWriter.cs
```diff
@@ -147,44 +135,39 @@ protected override void WriteChildResourceGetMethod(ResourceCollection resourceC
 
         private void WriteResourceEntry(ResourceCollection resourceCollection, bool isAsync, ICollection<FormattableString>? types)
         {
-            if (IsArmCore)
```
ArcturusZhang, line 150: by design as I described in the description of this PR, Azure.ResourceManager (ArmCore) will never have the `ArmResourceExtension` written, therefore nothing will be output for ArmResourceExtension. We will never go into this method when it is ArmCore, therefore we no longer need this check

src/AutoRest.CSharp/Mgmt/Generation/MgmtExtensionWriter.cs
```diff
@@ -35,37 +46,15 @@ protected override WriteMethodDelegate GetMethodDelegate(bool isLongRunning, boo
...
-            }
-        }
+            var extensionClient = This.GetExtensionClient(resourceType);
 
-        private void WriteMethodBodyWrapper(MethodSignature signature, bool isAsync, bool isPaging)
-        {
             _writer.AppendRaw("return ")
                 .AppendRawIf("await ", isAsync && !isPaging)
-                .Append($"GetExtensionClient({This.ExtensionParameter.Name}).{CreateMethodName(signature.Name, isAsync)}(");
+                .Append($"Get{extensionClient.Type.Name}({This.ExtensionParameter.Name}).{CreateMethodName(signature.Name, isAsync)}(");
```
ArcturusZhang, line 57: should use `extensionClient.FactoryMethodName` here

samples/Azure.Management.Storage/Generated/Extensions/StorageAccountResourceExtensionClient.cs
ArcturusZhang: This is fixed

ArcturusZhang:
Hi @m-nash I believe this PR is ready for review now. 
Also I have fixed the issues in the corresponding regen PR: https://github.com/Azure/azure-sdk-for-net/pull/35792 all the pipelines in it could pass now.

m-nash:
@ArcturusZhang can we add a test case demonstrating how a customer would use this to mock responses?

ArcturusZhang:
Hi @m-nash in [this PR](https://github.com/Azure/azure-sdk-for-net/pull/36029) I introduced a simplest sample for mocking the extension method.

One of the biggest inconvenience is exactly what Yao said here: https://github.com/Azure/autorest.csharp/pull/3175#issuecomment-1445638770
We not only have to mock the "extension clients", we have to mock the `GetCachedClient` method on whatever resource is carrying the extension (in the sample, it is the TenantResource). This setup is kind of un-intuitive, one of the reasons could be the `GetCachedClient` is actually hidden on IDEs.

m-nash:
> One more thing here is that I'm wondering that if we could do something such as providing a detailed doc to improve the user experience when they want to mock the extension method. For example, if we want to mock the extension method, we need to mock twice as shown in below based on our current implementation, it's not easy for the user to do that without any education.
> 
> ```cs
> public void MockingSample()
> {
>         var subscriptionClientMock = new Mock<SubscriptionResourceExtensionClient>();
>         CdnNameAvailabilityContent content = new CdnNameAvailabilityContent("falseName", CdnResourceType.Endpoints);
>         CdnNameAvailabilityResult result = new CdnNameAvailabilityResult(false, "name already exists.", "name already exists.");
>         var response = Response.FromValue(result, null);
>         subscriptionClientMock.Setup(subscriptionClient => subscriptionClient.CheckCdnNameAvailabilityWithSubscription(content, default)).Returns(response);
> 
>         var subscriptionMock = new Mock<SubscriptionResource>();
>         subscriptionMock.Setup(sub => sub.GetCachedClient(It.IsAny<Func<ArmClient, SubscriptionResourceExtensionClient>>())).Returns(subscriptionClientMock.Object);
> 
>         var val = subscriptionMock.Object.CheckCdnNameAvailabilityWithSubscription(content);
>         Assert.AreEqual(val.Value.NameAvailable, result.NameAvailable);
> }
> ```

T

m-nash (changes requested):
Before we merge this lets nail down the user scenario.  Thank you @Yao725  for putting together the example lets discuss this in our next sync.

ArcturusZhang:
Change namespace from Mock to Testing

ArcturusZhang:
Close in favor of https://github.com/Azure/autorest.csharp/pull/3483 because of a change of design

Issue #3056, opened 2023-02-01
Support multiple content types as input


pshao25:
**Sample swagger:**
We want 4 content types: "text/plain", "application/json", "image/jpeg", "image/png"
```
"post": {
  "operationId": "params_postParameters",
  "description": "POST a JSON or a JPEG",
  "consumes": ["text/plain", "application/json", "image/jpeg", "image/png"],
  "parameters": [
    {
      "name": "parameter",
      "in": "body",
      "description": "I am a body parameter with a new content type.",
      "schema": {
        "$ref": "#/definitions/PostInput"
      },
      "required": true
    }
  ],
  "responses": {
    "200": {
      "description": "Answer from service",
      "schema": {
        "type": "object"
      }
    }
  }
}
```

**Cadl:**
Example in cadl-ranch: https://github.com/Azure/cadl-ranch/pull/201
```
// main.cadl
@route("/serviceDriven/parameters")
@doc("POST a JSON or a JPEG")
@convenienceAPI
op postParameters(parameter: PostInput | bytes, @header contentType: "text/plain" | "application/json" | "image/jpeg" | "image/png"): object;
```
We expect user writes this. We could have internal mapping like M4 (how it maps needs to confirm with Tim):
- bytes: "image/jpeg", "image/png", ...
- PostInput: "application/json", ...
- string: "text/plain", ...

**Generated SDK:**
```C#
// protocol method: ContentType is `Azure.Core.ContentType` containing all types of content type
public virtual Response PostParameters(RequestContent content, ContentType contentType, RequestContext context = null);

// convenience method for model
public virtual Response<BinaryData> PostParameters(PostInput parameter, ContentType contentType, CancellationToken cancellationToken = default);
// convenience method for bytes
public virtual Response<BinaryData> PostParameters(BinaryData parameter, ContentType contentType, CancellationToken cancellationToken = default);
// convenience method for string
public virtual Response<BinaryData> PostParameters(string parameter, ContentType contentType, CancellationToken cancellationToken = default);
```
Above examples are under the condition if there are more than 1 content types for the specific type. If there is only one content type, we omit `ContentType` parameter. For example, this case we will generate:
```C#
// protocol method: ContentType is `Azure.Core.ContentType` containing all types of content type
public virtual Response PostParameters(RequestContent content, ContentType contentType, RequestContext context = null);

// convenience method for "application/json"
[...]

PR #3176, opened 2023-02-27
Support multiple content types with only one body type

pshao25:
This is the first step of https://github.com/Azure/autorest.csharp/issues/3056#issuecomment-1445755382

No SDK changes.


lirenhe:
@pshao25, could you help to regen the SDK so that we could understand the impact of this change. Thanks

pshao25:
@m-nash could you review this PR?

src/CADL.Extension/Emitter.Csharp/src/type/InputType.ts
```diff
@@ -20,11 +20,19 @@ export interface InputLiteralType extends InputType {
...
+export function isInputLiteralType(type: InputType): type is InputLiteralType {
+    return type.Name === "Literal";
+}
+
 export interface InputUnionType extends InputType {
     Name: "Union";
     UnionItemTypes: InputType[];
 }
 
+export function isInputUnionType(type: InputType): type is InputUnionType {
```
chunyu3, line 32: Can we not check the name, instead to check the core property `UnionItemTypes`?

e..g if ( (type as InputUnionType).UnionItemTypes !== undefined)

test/TestProjects/MediaTypes-Cadl/Generated/MultipleMediaTypesClient.cs
```diff
@@ -0,0 +1,347 @@
...
+        /// <param name="contentType"> The Union to use. Allowed values: &quot;application/json&quot; | &quot;application/octet-stream&quot; | &quot;text/plain&quot;. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="body"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="body"/> is an empty string, and was expected to be non-empty. </exception>
+        public virtual Response OneStringBodyThreeContentTypes(string body, ContentType contentType, CancellationToken cancellationToken = default)
+        {
+            Argument.AssertNotNullOrEmpty(body, nameof(body));
+
+            RequestContext context = FromCancellationToken(cancellationToken);
+            Response response = OneStringBodyThreeContentTypes(body, contentType, context);
```
chunyu3, line 159: Can `string` implicitly cast to `RequestContent`? I know there is a function `public static RequestContent Create(string content) `

src/CADL.Extension/Emitter.Csharp/src/type/InputType.ts
pshao25: See the definition of `InputUnionType` above, it always has the name of "Union". So it is the best to check by name. Another reason is: if we use the cast in the body, we lose the meaning of adding this function whose return type is type predicate. I add this function to avoid casting.

test/TestProjects/MediaTypes-Cadl/Generated/MultipleMediaTypesClient.cs
pshao25: Yes. `string` and `BinaryData` can both implicitly cast to `RequestContent`

chunyu3 (approved):

lirenhe (approved):

PR #3193, opened 2023-03-07
refact emitter

chunyu3:
# Description

- split operation, athentication, client model from emitter.

# Checklist

# Ready to Land?


src/CADL.Extension/Emitter.Csharp/src/lib/ServiceAuthentication.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
pshao25, line 1: Can we make the first letter small case?

ArcturusZhang: some of our files start with an upper case letter, the others start with a lower case letter. Could we align?

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -0,0 +1,446 @@
...
+            program,
+            operation,
+            resourceOperation
+        ),
+        Paging: paging,
+        GenerateProtocolMethod: generateProtocol,
+        GenerateConvenienceMethod: generateConvenience
+    } as InputOperation;
+
+    function loadOperationParameter(
```
ArcturusZhang, line 218: instead of defining these functions inside this function, could we just define these in this file instead? We will have less indent and make this function more clear

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -0,0 +1,446 @@
...
+        program: Program,
+        parameter: HttpOperationParameter
+    ): InputParameter {
+        const { type: location, name, param } = parameter;
+        let format = undefined;
+        if (parameter.type !== "path") {
+            format = parameter.format;
+        }
```
ArcturusZhang, line 226: could we refine this?
```suggestion
        const format = parameter.type === "path" ? undefined : parameter.format;
```
Or we might could change it to just `const format = parameter.format;` because we will never have a `format` defined on a path parameter from the compiler. Even if we have, we will not use it below

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -0,0 +1,446 @@
...
+        return {
+            StatusCodes: status,
+            BodyType: type,
+            BodyMediaType: BodyMediaType.Json,
+            Headers: responseHeaders,
+            IsErrorResponse: isErrorModel(program, response.type)
+        } as OperationResponse;
+    }
+
+    function loadOperationLongRunning(
```
ArcturusZhang, line 366: should we rename this to `loadLongRunningOperation`?

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        return loadOperationResponse(
+            program,
+            op.responses[0],
+            resourceOperation
+        );
+    }
+
+    function isLroOperation(program: Program, op: Operation) {
```
ArcturusZhang, line 417: since `Lro` stands for long running operation, we might need to rename this to either `isLro` or `isLongRunningOperation`

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -0,0 +1,446 @@
...
+import { capitalize } from "./utils.js";
+
+export function loadOperation(
+    context: EmitContext<NetEmitterOptions>,
+    operation: HttpOperation,
+    uri: string,
+    urlParameters: InputParameter[] | undefined = undefined,
+    serviceNamespaceType: Namespace,
+    models: Map<string, InputModelType>,
+    enums: Map<string, InputEnumType>
```
ArcturusZhang, line 72: I saw a lot of functions are taking these two maps, shall we refactor the whole thing into a class and put these things as properties or fields of that class, and fucntions like this could be instance method and it does not have to take these maps as parameters, they will be internal state of the `builder` if you will

lirenhe:
@chunyu3, please help to check what the impact on the SDK repo, thanks

src/CADL.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -0,0 +1,370 @@
...
+export function createModelForService(
+    context: EmitContext<NetEmitterOptions>,
+    service: Service
+): CodeModel {
+    const program = context.program;
+    const title = service.title;
+    const serviceNamespaceType = service.type;
+    const apiVersions: Set<string> = new Set<string>();
+    let version = service.version;
+    if (version && version !== "0000-00-00") {
```
lirenhe, line 81: Can we move those hard coded value into constants.ts?

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -0,0 +1,446 @@
...
+): InputOperation {
+    const program = context.program;
+    const dpgContext = createDpgContext(context);
+    const {
+        path: fullPath,
+        operation: op,
+        verb,
+        parameters: cadlParameters
+    } = operation;
+    console.log(`load operation: ${op.name}, path:${fullPath} `);
```
lirenhe, line 82: As you are working on the refactor, one general comment is to add more logging information during the emitter run.
We could by default turn off the log, but if some issues happen, we could enable more logging for trouble shooting.

src/CADL.Extension/Emitter.Csharp/src/lib/ServiceAuthentication.ts
archerzz: The TS official code style doesn't mention file name case. It only says the class should be upper camel case. However, I found a code guide of a framework, in which the file name should be uppder camel case: https://www.itwinjs.org/learning/guidelines/typescript-coding-guidelines/#:~:text=Files%201%20Use%20the%20.ts%20file%20extension%20for,files%202%20TypeScript%20file%20names%20should%20be%20PascalCase

image

chunyu3: consistent naming convention, the File name is cameCase. Update the file names.

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
chunyu3: updated.

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
chunyu3: add logger to do this. and an emitter option `logLevel` to set the level.

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
chunyu3: those functions are not used outside, and use internal function can reuse some variables instead of passing them as parameter. I prefer to keep it as internal function as now.

src/CADL.Extension/Emitter.Csharp/src/lib/operation.ts
chunyu3: @ArcturusZhang `models` and `enums` are golbal variables for the whole codeModel. It can be only be the fields of the codeModel.  And then every function such as loadOperation, getInputModel will be the function of builder. There is only one class, one file, it is not good, either. 
or, we can split a lot of builder, operationBuilder, modelBuilder, and those two will be set in their constructor. But now they are parameters of constuctor instead of function.

or, we will define an global varaible for these, and it will be imported in any ts files. this approach I don't think it is good for encapsulation.

We can discuss this, and maybe delay to implement it.

ArcturusZhang (approved):

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -207,22 +132,21 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
...
+                    if (error.stderr) logger.error(error.stderr);
+                    if (error.stdout) logger.verbose(error.stdout);
                     throw error;
                 }
             }
 
             if (!options["save-inputs"]) {
                 // delete
-                deleteFile(resolvePath(generatedFolder, "cadl.json"));
+                deleteFile(resolvePath(generatedFolder, "cadl2.json"));
```
ArthurMa1978, line 149: are this file same as ln154 which is still `cadl.json`?

ArthurMa1978: consider to predefine this as a const

src/CADL.Extension/Emitter.Csharp/src/options.ts
```diff
@@ -60,6 +62,11 @@ export const NetEmitterOptionsSchema: JSONSchemaType<NetEmitterOptions> = {
             type: "array",
             nullable: true,
             items: { type: "string" }
+        },
+        logLevel: {
+            type: "string",
+            enum: ["error", "warn", "info", "debug", "verbose"],
```
ArthurMa1978, line 68: Should the values of this enum come from `logger.ts`?

chunyu3: Yes

src/CADL.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: I updated to use constant.

ArthurMa1978 (approved):

PR #3200, opened 2023-03-14
Align implementation of model generation in mgmt and DPG

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3417
Fixes https://github.com/Azure/autorest.csharp/issues/3406

<del>Depends on https://github.com/Azure/autorest.csharp/pull/3207</del>
<del>Depends on https://github.com/Azure/autorest.csharp/pull/3237</del>
<del>Depends on https://github.com/Azure/autorest.csharp/pull/3360</del>
<del>Depends on https://github.com/Azure/autorest.csharp/pull/3419</del>
Depends on https://github.com/Azure/autorest.csharp/pull/3373

# Description

This PR is trying to align the implementations between `ModelTypeProvider` and `SchemaObjectType` (and derived types like `MgmtObjectType`, etc). This is making preparations for the work that before we generate any model in mgmt/HLC, we first convert the code model schemas into input types and we directly use `ModelTypeProvider` to generate the models, which deprecates `SchemaObjectType`.
In this way we will finally "combine our generator" among the different targets (mgmt, HLC and DPG)

The diff in generated code could roughly categorize into 3 categories:
1. Escaping changes in the xml documents. I have extracted the changes into [this PR](https://github.com/Azure/autorest.csharp/pull/3490), if we could merge that PR first, these changes will vanish in this PR's generated code.
2. Internal ctor added for those "input-only" models. Since those ctors are internal, there is no harm of generating them. This unblocks us to introduce more models into model factory. Now model factory only includes the models with deserialization because of this.
3. The assignment of constant values for the property moved from the public ctor to the same way as currently in DPG - the property initializer.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/CmcRestClientBuilder.cs
```diff
@@ -79,16 +79,15 @@ private static string GetRequestParameterName(RequestParameter requestParameter)
...
         /// </summary>
         /// <param name="operation"></param>
         /// <param name="httpRequest"></param>
         /// <param name="requestParameters"></param>
-        /// <param name="responseHeaderModel"></param>
         /// <param name="accessibility"></param>
         /// <param name="returnNullOn404Func"></param>
         /// <returns></returns>
-        public RestClientMethod BuildMethod(Operation operation, HttpRequest httpRequest, IEnumerable<RequestParameter> requestParameters, DataPlaneResponseHeaderGroupType? responseHeaderModel, string accessibility, Func<string?, bool>? returnNullOn404Func = null)
+        public RestClientMethod BuildMethod(Operation operation, HttpRequest httpRequest, IEnumerable<RequestParameter> requestParameters, string accessibility, Func<string?, bool>? returnNullOn404Func = null)
```
ArcturusZhang, line 90: I removed this parameter because now the only usage of this method comes from mgmt plane, and that removed parameter is always getting null.

ArcturusZhang:
Now we are kind of having a difficulty: the constants are not included in the constructors, but the intiializers in ctor are determined from the parameters of the ctor. One of these fact needs to be changed.
We will resolve this once we consolidate the constant/literal issue.

test/TestProjects/MgmtOperations/Generated/Models/AvailabilitySetGrandChildData.Serialization.cs
```diff
@@ -102,7 +102,7 @@ internal static AvailabilitySetGrandChildData DeserializeAvailabilitySetGrandChi
                     continue;
                 }
             }
-            return new AvailabilitySetGrandChildData(id, name, type, systemData.Value, Optional.ToDictionary(tags), location, bar.Value);
+            return new AvailabilitySetGrandChildData(id, name, type, systemData.Value, Optional.ToDictionary(tags), location, bar);
```
ArcturusZhang, line 105: This should be expected. The same case in DPG will generate the same code.
And this actually does not change the result - `.Value` is always called when passing to the ctor by the implicit operator

PR #3204, opened 2023-03-16
Improve DateTimeOffset parsing for headers

AlexanderSher:
Follow-up the changes in https://github.com/dotnet/aspnetcore/pull/47040
Scoped to headers only to avoid potential issues in JSON / XML deserialization

No changes in generated code, recorded tests in azure-sdk-for-net repo have passed


src/assets/Generator.Shared/ResponseHeadersExtensions.cs
```diff
@@ -12,6 +12,34 @@ namespace Azure.Core
 {
     internal static class ResponseHeadersExtensions
     {
+        private static readonly string[] KnownFormats =
```
annelo-msft, line 15: It looks like we have some redundancy with formats in multiple places, including https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Generator.Shared/TypeFormatters.cs#L14.  Would it make sense to centralize, or is there a reason to keep DateTime formatting separately?

src/assets/Generator.Shared/ResponseHeadersExtensions.cs
```diff
@@ -40,7 +68,16 @@ public static bool TryGetValue(this ResponseHeaders headers, string name, out Da
         {
             if (headers.TryGetValue(name, out string? stringValue))
             {
-                value = TypeFormatters.ParseDateTimeOffset(stringValue, "");
+                if (DateTimeOffset.TryParseExact(stringValue, "r", DateTimeFormatInfo.InvariantInfo, DateTimeStyles.None, out var dto) ||
```
christothes, line 71: Is there any value in converting the `name` string to Span first?

christothes (approved):

src/assets/Generator.Shared/ResponseHeadersExtensions.cs
annelo-msft, line 71: Do we have a sense of how broadly this change will take affect (which libraries/operations), and whether there are any possible breaking behaviors?

AlexanderSher: No, `System.Net.Http.Headers` doesn't have any span-based API

AlexanderSher: It will affect most of the headers. There won't be any breaking behavior cause we fallback to the old code path if `TryParseExact` fails.

annelo-msft: Is it possible that a header could ever parse to a different value than it did previously?

KrzysztofCwalina (approved):
It would be good to add more source code comments around such perf optimizations. After years, it will be difficult for devs to understand why this was done.

Issue #3206, opened 2023-03-17
[DPG] The derived models with all required properties do not have proper internal constructor


ArcturusZhang:
For this typespec:
```
@discriminator("name")
@doc("Base model with discriminator property.")
model DeliveryRuleAction {
    foo?: string;
}

@doc("Deriver model with discriminator property.")
model DeliveryRuleCacheExpirationAction extends DeliveryRuleAction {
    name: "A";

  @doc("Required parameters.")
  parameters: CacheExpirationActionParameters;
}

model CacheExpirationActionParameters {
    foo: string;
    bar?: int32;
}
```
Now on the latest commit e07e879768792aea9b30c2a8b65abf8f3d596616 generates the code like this for class `DeliveryRuleCacheExpirationAction`:
```csharp
    /// <summary> Deriver model with discriminator property. </summary>
    public partial class DeliveryRuleCacheExpirationAction : DeliveryRuleAction
    {
        /// <summary> Initializes a new instance of DeliveryRuleCacheExpirationAction. </summary>
        /// <param name="parameters"> Required parameters. </param>
        /// <exception cref="ArgumentNullException"> <paramref name="parameters"/> is null. </exception>
        public DeliveryRuleCacheExpirationAction(CacheExpirationActionParameters parameters) : base("A")
        {
            Argument.AssertNotNull(parameters, nameof(parameters));

            Parameters = parameters;
        }

        /// <summary> Required parameters. </summary>
        public CacheExpirationActionParameters Parameters { get; set; }
    }
```
and corresponding deserialization looks like this:
```csharp
        internal static DeliveryRuleCacheExpirationAction DeserializeDeliveryRuleCacheExpirationAction(JsonElement element)
        {
            if (element.ValueKind == JsonValueKind.Null)
            {
                return null;
            }
            CacheExpirationActionParameters parameters = default;
            string name = default;
            Optional<string> foo = default;
            foreach (var property in element.EnumerateObject())
            {
                if (property.NameEquals("parameters"u8))
                {
                    parameters = CacheExpirationActionParameters.DeserializeCacheExpirationActionParameters(property.Value);
                    continue;
                }
                if (property.NameEquals("name"u8))
                {
                    name = property.Value.GetString();
                    continue;
                }
                if (property.NameEquals("foo"u8))
                {
                    foo = property.Value.GetString();
                    continue;
                }
            }
            return new DeliveryRuleCacheExpirationAction(parameters);
        }
```
[...]

Issue #3178, opened 2023-03-01
The internal constructors for a model in DPG and mgmt/HLC are different


ArcturusZhang:
The difference happens when the model has a discriminator.
In mgmt/HLC, the discriminator property will show up as a parameter in the internal constructor, for example:
https://github.com/Azure/autorest.csharp/blob/9151a2ec37f140ffbc53c8a5acda3e3280314afd/test/TestProjects/MgmtDiscriminator/Generated/Models/DeliveryRuleQueryStringCondition.cs#L31
In DPG, the discriminator property will not show up as parameter in the internal constructor, for example:
https://github.com/Azure/autorest.csharp/blob/9151a2ec37f140ffbc53c8a5acda3e3280314afd/test/CadlRanchProjects/inheritance/Generated/Models/Salmon.cs#L30

We should align the behavior considering that we cannot quickly align the implementation.

ArcturusZhang:
We are going to align the behavior in mgmt/HLC to the way in DPG by not exposing the discriminator property in their internal constructors.

ArcturusZhang:
Another inconsistency: DPG models will put the discriminator property in their protected constructor, but mgmt does not.
Mgmt (discriminator property is `name`):
https://github.com/Azure/autorest.csharp/blob/8ee2ca051a2ac5f4f79fb2d82209bfa076f93cbd/test/TestProjects/MgmtDiscriminator/Generated/Models/DeliveryRuleAction.cs#L18
DPG (discriminator property is `discriminatorProperty`):
https://github.com/Azure/autorest.csharp/blob/8ee2ca051a2ac5f4f79fb2d82209bfa076f93cbd/test/TestProjects/Models-Typespec/Generated/Models/BaseModelWithDiscriminator.cs#L20

ArcturusZhang:
Differences between the discriminator in mgmt/HLC and DPG:

| category | mgmt/HLC | DPG |
| :-- | :-- | :-- |
| accessibility | internal | public |
| has setter? | no | yes |
| type | wrapped in extensible enum | always string |
| show up in internal ctor of base? | yes | no |
| show up in public ctor of base? | Always no | Always yes |
| show up in internal ctor of derived? | yes | no |
| show up in public ctor of derived? | Always no | Always no |

ArcturusZhang:
My proposal is like this:

1. We do not expose the discriminator on the public/protected ctor.
2. We make the discriminator property internal (and the related extensible enum if any).
[...]

PR #3207, opened 2023-03-17
Fix dpg model generation issue

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3206
Fixes DPG part of https://github.com/Azure/autorest.csharp/issues/3178
Regen: https://github.com/Azure/azure-sdk-for-net/pull/35424

# Description

1. Fixed the issue that in DPG, the model deserialization is not putting the properties from the base model
2. changed ctor implementation according to decision here: https://github.com/Azure/autorest.csharp/issues/3178#issuecomment-1508154021
3. added a configuration `publicDiscriminatorProperty` to control if we want the discriminator property to be public, defaults to `false`, because it is not a breaking change if we change the discriminator from internal to public, but it will be a breaking change on the other way around

# Checklist

# Ready to Land?


chunyu3 (approved):

lirenhe (approved):

test/TestProjects/Models-Typespec/Generated/Models/DerivedModel.cs
```diff
@@ -29,7 +29,7 @@ public DerivedModel(IEnumerable<CollectionItem> requiredCollection)
         /// <param name="requiredCollection"> Required collection. </param>
         internal DerivedModel(IList<CollectionItem> requiredCollection)
         {
-            RequiredCollection = requiredCollection.ToList();
+            RequiredCollection = requiredCollection;
```
ArcturusZhang, line 32: we never need this extra `ToList` therefore I removed it in this PR

ArcturusZhang: BTW mgmt models never have this ToList

samples/AnomalyDetector/Generated/Models/AnomalyState.cs
```diff
@@ -31,7 +30,7 @@ internal AnomalyState(DateTimeOffset timestamp, AnomalyValue value, IReadOnlyLis
         {
             Timestamp = timestamp;
             Value = value;
-            Errors = errors.ToList();
+            Errors = errors;
```
ArcturusZhang, line 33: we never need this `ToList()` therefore I removed it.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -113,9 +113,27 @@ private ConstructorSignature EnsurePublicConstructorSignature()
...
-                ? CreateSerializationConstructorSignature(Declaration.Name, Fields.PublicConstructorParameters, Fields.SerializationParameters) ?? InitializationConstructorSignature
-                : InitializationConstructorSignature;
+            // the property bag never needs deserialization, therefore we return the initialization constructor here so that we do not write it in the generated code
+            if (IsPropertyBag)
+                return InitializationConstructorSignature;
+
+            var serializationCtorSignature = CreateSerializationConstructorSignature(Declaration.Name, Fields.PublicConstructorParameters, Fields.SerializationParameters);
+
+            // verifies if this new ctor is the same as the public one
+            if (!serializationCtorSignature.Parameters.Any(p => TypeFactory.IsList(p.Type)) && InitializationConstructorSignature.Parameters.SequenceEqual(serializationCtorSignature.Parameters, new ParameterByTypeEqualityComparer()))
```
AlexanderSher, line 123: Do we have any requirements about optional parameters?

ArcturusZhang: What do you mean by "requirements"?
This check is here just because we need to ensure we did not write two identical constructors which leads to compile errors.
We will even write something like this:
```
public Foo(IEnumerable<string> bar) {}
internal Foo(IReadOnlyList<string> bar) {}
```
just because they are different. The compiler should be smart enough to figure out which to choose, and mgmt/hlc is always following this way.

For optional parameters, the public constructor will never have them, and serialization constructor will have them therefore they will never be identical constructors.

ArcturusZhang:
@m-nash could you have a review on this PR? This is ready to review.
This PR changes the behavior of discriminator on the DPG side to align with our decision: https://github.com/Azure/autorest.csharp/issues/3178#issuecomment-1486583589

samples/AnomalyDetector/Generated/Models/AnomalyState.cs
```diff
@@ -31,7 +30,7 @@ internal AnomalyState(DateTimeOffset timestamp, AnomalyValue value, IReadOnlyLis
         {
             Timestamp = timestamp;
             Value = value;
-            Errors = errors.ToList();
```
m-nash, line 34: @AlexanderSher I believe this ToList had something to do with us owning the reference memory?  Can you help elaborate on the reasoning?

@ArcturusZhang if the ToList is needed we can change to
```c#
Errors = errors?.ToList() ?? new ChangeTrackingList<ErrorResponse>();
```

ArcturusZhang: No, this `ToList` comes from `System.Linq` therefore it did nothing.

We ensure the `ChangeTrackingList` in the deserialization method where we call this constructor using `Optional.ToList`, you could refer it here: https://github.com/Azure/autorest.csharp/blob/ab0215d968d7d35a47e3f3d429732974aaf3559d/samples/AnomalyDetector/Generated/Models/AnomalyState.Serialization.cs#L60
Therefore we do not need this `ToList` because we have already did the `ToList` and the parameter here will never get a null value.

m-nash:
Can we get the regen PR updated so I can take one final look?

ArcturusZhang:
> Can we get the regen PR updated so I can take one final look?

@m-nash the latest regen is here: https://github.com/Azure/azure-sdk-for-net/pull/35412

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -227,12 +235,12 @@ public static void Initialize(IPluginCommunication autoRest)
...
 
-        private static bool GetOptionBoolValue(IPluginCommunication autoRest, string option)
+        private static bool GetOptionBoolValue(IPluginCommunication autoRest, string option, bool isAzureArm)
         {
-            return autoRest.GetValue<bool?>(option).GetAwaiter().GetResult() ?? GetDefaultBoolOptionValue(option)!.Value;
+            return autoRest.GetValue<bool?>(option).GetAwaiter().GetResult() ?? GetDefaultBoolOptionValue(option, isAzureArm)!.Value;
         }
 
-        public static bool? GetDefaultBoolOptionValue(string option)
+        public static bool? GetDefaultBoolOptionValue(string option, bool isAzureArm)
```
lirenhe, line 243: we already have Options.AzureArm, why do we need this additonal flag?

ArcturusZhang: Because this PR adds a new configuration `publicDiscriminatorProperty` which has different default value for mgmt plane and others, therefore I have to add an input here so that we could return different default values.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -65,6 +66,7 @@ public enum UnreferencedTypesHandlingOption
             bool skipSerializationFormatXml,
             bool disablePaginationTopRenaming,
             bool generateModelFactory,
+            bool publicDiscriminatorProperty,
```
lirenhe, line 69: Not quite understand what true & false mean for this. Do you mean  EnablePublicDiscriminatorProperty?  Maybe you should find a better name.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
lirenhe: Can we optimize the logic here?  Also seems we only need IsAzureArm for Options.PublicDiscriminatorProperty, but we passed IsAzureArm for all other properties.

ArcturusZhang: No that is not an optimization - because potentially we could have different "default values" for all the options, this change enables us to do it. If the default value is the same regardless which plane we are on, that option could just ignore this flag.

Otherwise we have to make another method to give default values and we will have two methods for assigning default values. I will never call that an "optimization".

It is a more valid ask that "why we have to have different default values for mgmt plane and others". I do not like this design as well and I hope we could have a unified default value.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
ArcturusZhang: We use this configuration like this:
```
public-discriminator-property: true
```

which means we will have a public discriminator property, otherwise we do not. Which part do we have the confusion?

ArcturusZhang: And in the generator code, we use this like this:
```
var accessibilityOfDiscriminatorProperty = Configuration.PublicDiscriminatorProperty ? Public : Internal;
```
I think this is clear enough

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
ArcturusZhang: Nevertheless, I just pushed a commit that removes this different default value part because I do not like it either.
Now the default value of this configuration is always false. Because it will not be a breaking change if we change it from false to true but it is a breaking change the other way around

ArcturusZhang:
> Can we get the regen PR updated so I can take one final look?

The latest regen is here: https://github.com/Azure/azure-sdk-for-net/pull/35424

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -150,6 +153,9 @@ public enum UnreferencedTypesHandlingOption
         public static bool SkipSerializationFormatXml { get; private set; }
         public static bool DisablePaginationTopRenaming { get; private set; }
         public static bool GenerateModelFactory { get; private set; }
+
+        public static bool PublicDiscriminatorProperty { get; private set; }
```
lirenhe, line 157: It would be good to describe the meaning of each property.

ArcturusZhang: I added some comments

ArcturusZhang:
@m-nash latest regen is here: https://github.com/Azure/azure-sdk-for-net/pull/35584

ArcturusZhang:
Hi @m-nash please take a look at the latest regen of this here: https://github.com/Azure/azure-sdk-for-net/pull/35720

m-nash (approved):

PR #3229, opened 2023-03-23
Fix optional constant issue

ArcturusZhang:
# Description

We generate wrong initialization code in constructor when we have a constant int (or other type that is not string), for instance
```csharp
RequiredIntConstant = new IntConstant("0"); // this is wrong
RequiredIntConstant = IntConstant._0; // this is correct
```
To see the issue and problematic generated code, please check the first commit.
To see how the issue is fixed, please exclude the first commit.

# Checklist

# Ready to Land?


ArcturusZhang:
@m-nash could you have a review?

m-nash (approved):

ArcturusZhang:
We could discuss further on the case "whether we should wrap int-based constant into an extensible enum", but this fix is always required and harms nothing even if we decide not to wrap int-constant into extensible enums

PR #3239, opened 2023-03-29
Use overloads when there are ambiguity between protocol and convenience

pshao25:
Core part of #2887 

Regen PR: https://github.com/Azure/azure-sdk-for-net/pull/35219/files


pshao25:
We already have this feature.

Issue #3240, opened 2023-03-30
Extensible enum based on int is not serializing into integer


ArcturusZhang:
Example: https://github.com/Azure/autorest.csharp/blob/f57c4da3593858e1af113b11d7dad7635773cb9c/test/TestProjects/MgmtOptionalConstant/Generated/Models/ModelWithOptionalConstant.Serialization.cs#L26

PR #3241, opened 2023-03-30
Fix the serialization issue in extensible enum

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3240
Regen: https://github.com/Azure/azure-sdk-for-net/pull/35293

# Description

# Checklist

# Ready to Land?


test/TestServerProjects/non-string-enum/Generated/Models/FloatEnum.cs
```diff
@@ -38,6 +38,9 @@ public FloatEnum(float value)
         public static FloatEnum FourHundredSix2 { get; } = new FloatEnum(FourHundredSix2Value);
         /// <summary> 429.1. </summary>
         public static FloatEnum FourHundredTwentyNine1 { get; } = new FloatEnum(FourHundredTwentyNine1Value);
+
+        internal float ToSerialSingle() => _value;
```
chunyu3, line 42: How do you define the serialization method name? why is the serialization method name of floatEnum `ToSerailSingle`?

Can we just call it `ToSerialize()`?

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
```diff
@@ -55,15 +55,21 @@ public EnumType(InputEnumType input, string defaultNamespace, string defaultAcce
...
             IsExtensible = isExtensible;
         }
 
         public CSharpType ValueType { get; }
         public bool IsExtensible { get; }
-        public bool IsStringValueType => ValueType.Equals(typeof(string));
-        public bool IsIntValueType => ValueType.Equals(typeof(Int32)) || IsLongValueType;
-        public bool IsLongValueType => ValueType.Equals(typeof(Int64));
+        public bool IsStringValueType { get; }
+        public bool IsIntValueType { get; }
```
chunyu3, line 68: No `IsFloatValueType` ? Do we still need special handle for Float?

test/TestServerProjects/non-string-enum/Generated/Models/FloatEnum.cs
ArcturusZhang: first of all, this is internal, we do not need to care what the name is this.
second, this aligns the behaviour how we get a name of this for fixed enum. We could have fixed enum based on float as well, and there, it is called ToSerialSingle.
The name `ToSerialize` does not make sense. This means to convert this to something else, for instance for int we have `ToSerialInt`.

We get this name using this expression $"ToSerial{ValueType.Name}". In dotnet, float is alias of `System.Single` therefore we see the word single here.

In summary,
1. this aligns with the serialization method name for fixed enum
2. this is internal therefore we do not need really care about this name here.

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
ArcturusZhang: We do not have that flag because we do not need it.
I only add a flag when it is necessary not because there is such a case.

IsStringValueType flag is used when we want to know if we need the extra method
IsIntValueType flag is used in fixed enum's serialization to determine if we need to use cast instead of calling `ToSerialXX` method.

In summary, for fixed enums, when IsIntValueType is true, we have one case, and when it is false, we have the other case (string and float make no difference).
For extensible enums, when IsStringValueType is true, we have one case, and when it is false, we have the other case (int and float make no difference).
Therefore we do not need any more flags. We could add them when we need some.

chunyu3:
hello @ArcturusZhang Can we add test for enum array to check if the enum array is correct serialzed?

ArcturusZhang:
> hello @ArcturusZhang Can we add test for enum array (int enum array and float enum array) to check if the enum array is correct serialzed?

I just added some properties in FirstTest-Typespec to show cases of enums

chunyu3 (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
```diff
@@ -55,15 +55,21 @@ public EnumType(InputEnumType input, string defaultNamespace, string defaultAcce
...
+
+        private string? _serializationMethodName;
+        public string SerializationMethodName => _serializationMethodName ??= IsStringValueType && IsExtensible ? "ToString" : $"ToSerial{ValueType.Name.FirstCharToUpperCase()}";
```
AlexanderSher, line 71: Both `IsStringValueType` and `IsExtensible` values are set in constructor, hence `SerializationMethodName` can be set in constructor as well.

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
```diff
@@ -425,6 +427,14 @@ public void WriteExtendableEnum(CodeWriter writer, EnumType enumType)
                         writer.Append($"public static {cs} {choice.Declaration.Name}").AppendRaw("{ get; }").Append($" = new {cs}({fieldName});").Line();
                     }
 
+                    // write ToSerial method, only write when the underlying type is not a string
+                    if (!enumType.IsStringValueType)
+                    {
+                        writer.Line();
+                        writer.Line($"internal {enumType.ValueType} {enumType.SerializationMethodName}() => _value;");
```
m-nash, line 434: nit: any reason we didn't just use ToInt32 vs ToSerialInt32?  Is there any actual difference between the two if they existed side by side?

m-nash (approved):
Approve with one comment

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
ArcturusZhang: make sense. Let me change this

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
ArcturusZhang: I choose this name to align the existing similar methods in fixed enum. If we have a fixed enum based on float, it will have an extension method `ToSerialSingle`. Therefore I am choosing the same pattern here.

Issue #3143, opened 2023-02-17
Support the invalid model property names


pshao25:
Example:
```
model DerivedModel {
  "model.kind": "derived";
  "derived.name": string;
  for: string;
}
```
"model.kind" and "derived.name" are not valid.

Solution 1:
Add lineter to compiler. I don't think cadl team will agree.

Solution 2:
Add linter in emiter and let compiler call it.

Solution 3:
Codegen remove the invalid character and make it camelcase.

I prefer solution 2.

chunyu3:
What's the name will be in the json payload?

chunyu3:
Those invalid model should be supported. Apply naming conversion in autorest.csharp/common, and apply to all targets(mgmt, hlc, dpg).

For keyword, (e.g.  for: string;), we will add waring in apiview and ask arch board to review this. If they don't accept, we may add @projectName to override.

PR #3246, opened 2023-04-03
Support invalid model property name on discriminator

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/3143

No SDK change.


chunyu3 (commented):
This PR only support invalid property name for discriminator. How about normal property with invalid name?

ArcturusZhang (approved):

chunyu3 (approved):

lirenhe (approved):

PR #3247, opened 2023-04-03
Handle non-model return type

### LINKED ISSUE #3110: Handle non-model return type
Cadl:
```
op world(): string;
```

It returns a string. Generated convenience method:
```C#
public virtual async Task<Response<string>> WorldValueAsync(CancellationToken cancellationToken = default)
{
    ...
    Response response = await WorldAsync(context).ConfigureAwait(false);
    return Response.FromValue(string.FromResponse(response), response);
}
```

Option1: add a linter to avoid return non-model type.
Option2: we change the last line of our convenience method to 
```C#
return Response.FromValue((string)JsonData.FromBytes(response.Content.ToMemory()), response);
```

pshao25:
Fix #3110


test/CadlRanchProjects.Tests/hello.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Reflection.PortableExecutable;
```
lirenhe, line 4: where do we need this using?

ArcturusZhang (approved):

PR #3249, opened 2023-04-04
Fix NRE when more than one document is updated by MarkInternal

AlexanderSher:
After `project` is changed, it can no longer find a document by the syntax tree from the unchanged project. We should use `documentId` instead.


ArcturusZhang (approved):

ArcturusZhang:
@m-nash could you please have a review on this? These changes LGTM

lirenhe:
@AlexanderSher, could you share the SDK regen result?

AlexanderSher:
> @AlexanderSher, could you share the SDK regen result?

No generated code changes here: https://github.com/Azure/azure-sdk-for-net/pull/35404/files

lirenhe (approved):

PR #3250, opened 2023-04-04
Fix the property bag issue

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/35000

Regen https://github.com/Azure/azure-sdk-for-net/issues/35413

# Description

## Current issue

Currently one operation with and without the property bag may generate different request when one of the parameter is array because in the property bag, we always assign an empty list when there is no element given for the array.

In the method without a property bag, like:
```csharp
public virtual Response DoSomething(IEnumerable<string> items = null)
```
we have the ability to distinguish between the case that `items` is null and the `items` is not null but empty.
When there is a property bag, we have
```csharp
public virtual Response DoSomething(ResourceDoSomethingOptions options)
{
return DoSomething(options.items);
}
```
but in the ctor of `ResourceDoSomethingOptions`
```csharp
public partial class ResourceDoSomethingOptions 
{
public ResourceDoSomethingOptions()
{
Items = new ChangeTrackingList<string>();
}

public IList<string> Items {get;} = null;
}
```
No matter how you construct the instance, the `Items` will never be null.

I added a few test cases in this PR, and when the items is null, the test fails indicating that in that case, we are generating different requests.
image

## The fix in this PR

Previously when we are building the request, we are using this when a query parameter is array and optional:
```csharp
if (items != null)
{
    uri.AppendQueryDelimitered(items, ", ", true);
}
```
In this PR, we changed the condition to
```csharp
if (items != null && Optional.IsCollectionDefined(items))
{
    /* the same */
}
```
The method `Optional.IsCollectionDefined` only returns true when the type of `items` is `ChangeTrackingList` and its `IsUnDefined` returns `true`. If the type of `items` is not `ChangeTrackingList` (for instance, it is an array or other kind of list implementation), this always returns true, even if `items` is `null`, this will return `true`.

Given the fact that `ChangeTrackingList` or `ChangeTrackingDictionary` is internal, our customers can never construct an instance of these types, and it is only constructed when a property bag is introduced (as claimed above), we have the following cases:
1. When there is no property bag, our customer will be sending an instance directly in the method signature, therefore it will never be of type `ChangeTrackList` or `ChangeTrackingDictionary` therefore `Optional.IsCollectionDefined` will always return true and the condition is the same as before.
2. When there is a property bag, the `items` we have here will always be constructed from the property bag therefore it is always a `ChangeTrackingList` and it is never null. Therefore in this case we are actually checking `Optional.IsCollectionDefined(items)` only. This is only `true` when the customer actually touches the list like by adding an item or calling `Clear` if they want to explicitly send an empty array.

In conclusion, this implementation will cover both cases when there is a property bag and when there is not a property bag and both created correct requests.

# Checklist

# Ready to Land?


ArcturusZhang:
Hi @m-nash could you have a look at this PR as well?

ArcturusZhang:
We kind of have two solutions:
1. Change the implementation of the property bag, including removing the initializer of collections in ctor, adding setter to collection properties. Downside of this is that this breaks the design pattern in other models, making models inconsistent because only the collection in property bags have setters.
2. Change the implementation of the place we create requests, replacing [`items != null` check](https://github.com/Azure/autorest.csharp/blob/2c80ef8204fe0db5f8d82da8daa7ae60e0eb1df7/test/TestProjects/MgmtPropertyBag/Generated/RestOperations/BarsRestOperations.cs#L264) with `Optional.IsCollectionDefined(items)`. The downside of this is that if we really want to pass in an empty array, we will have to touch the array for one time (by calling `Items.Clear()` maybe)

This PR is taking the second solution, because it should be super super rare case that we really want a query parameter with an empty array producing requests like this:
```
<originalRequest>/items=
```

m-nash:
Do we have the regen PR for this?

ArcturusZhang:
> Do we have the regen PR for this?

@m-nash it is here: https://github.com/Azure/azure-sdk-for-net/pull/35413

samples/Azure.ResourceManager.Sample/Generated/Models/SubscriptionResourceGetVirtualMachineImagesOptions.cs
```diff
@@ -40,12 +40,10 @@ public SubscriptionResourceGetVirtualMachineImagesOptions(AzureLocation location
         /// <summary> A valid image SKU. </summary>
         public string Skus { get; }
         /// <summary> The expand expression to apply on the operation. </summary>
-        public string Expand { get; set; } = null;
```
ArcturusZhang, line 43: These are removed on purpose because they are meaningless.

For collections, this initialization to `null` never worked.
For non-collection types, this initialization is pointless because they will always be initialized as default values when not explicitly assigned

ArcturusZhang:
> Do we have the regen PR for this?

The regen is here: https://github.com/Azure/azure-sdk-for-net/pull/35423

ArcturusZhang:
Hi @m-nash the regen is here: https://github.com/Azure/azure-sdk-for-net/pull/35728

test/TestProjects/MgmtPropertyBag/Generated/Models/FooCollectionGetAllOptions.cs
```diff
@@ -18,21 +18,16 @@ public FooCollectionGetAllOptions()
         }
 
         /// <summary> The filter to apply on the operation. </summary>
-        public string Filter { get; set; } = null;
-
+        public string Filter { get; set; }
         /// <summary> The Integer to use. </summary>
-        public int? Top { get; set; } = 10;
```
m-nash, line 24: we are losing the default value here, how are we accomodating for this?

ArcturusZhang: well this is also an inconsistency just like the null value in arrays.
When we do not have the option bag, we will always have this kind of method:
```csharp
public Response DoSomething(int? top = null)
```
therefore this default value also changes something

m-nash: hmm, shouldn't have the default in both the method and property bag?

ArcturusZhang: Sorry I just noticed that this default value is provided by `x-ms-client-default` therefore we need to keep this. I will investigate this to see if without the property bag, it is still there. If it is still there, I will keep it here.

ArcturusZhang: I fixed this issue - all the `null` value initializations are removed because they are just redundant.
The `10` is kept as before

eng/DownloadSharedSource.ps1
```diff
@@ -15,7 +15,7 @@ function DownloadAll([string[]]$files, [string]$baseUrl, [string]$downloadPath)
 $downloadPath = Resolve-Path (Join-Path $PSScriptRoot '..' 'src' 'assets' 'Azure.Core.Shared')
 Get-ChildItem $downloadPath -Filter *.cs | Remove-Item;
 $files = @('AsyncLockWithValue.cs', 'ClientDiagnostics.cs', 'DiagnosticScope.cs', 'DiagnosticScopeFactory.cs', 'ContentTypeUtilities.cs', 'HttpMessageSanitizer.cs',
-    'OperationInternalBase.cs', 'OperationInternal.cs', 'VoidValue.cs', 'OperationInternalOfT.cs', 'TaskExtensions.cs', 'Argument.cs', 'Multipart/MultipartFormDataContent.cs',
```
ArcturusZhang, line 18: I change this because we have the `VoidValue.cs` twice

m-nash (approved):

PR #3262, opened 2023-04-11
Mitigate CA1825 error in azure-sdk-for-net

### LINKED ISSUE #3261: CA1825 code analysis error on CodeGenSuppressAttribute
From whenever CA1825 ("Avoid zero-length array allocations") was added, `CodeGenSuppressAttribute` now yields a compiler error and wants you to pass `Array.Empty<Type>()` which can't work for attributes because it's not a constant expression or one of the few supported allocation expressions which, ironically, includes the `new Type[0]` expression the analyzer wants you to avoid.

This forces us to add `#pragma warning disable CA1825` whenever used.

This should be easily fixed by adding a constructor overload that only takes a `string member` parameter.

heaths:
Fixes #3261

### DIFF
--- a/eng/CodeAnalysis.ruleset
+++ b/eng/CodeAnalysis.ruleset
@@ -214,4 +214,7 @@
         <Rule Id="SA1650"  Action="None" />          <!-- Element documentation should be spelled correctly -->
         <Rule Id="SA1651"  Action="None" />          <!-- Do not use placeholder elements -->
     </Rules>
+    <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.NetAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.CSharp.NetAnalyzers">
+        <Rule Id="CA1825" Action="Error" />          <!-- Avoid unnecessary zero-length array allocations -->
+    </Rules>
 </RuleSet>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -4,6 +4,8 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
+    <PackageReference Update="Microsoft.CodeAnalysis.CSharp" Version ="3.9.0" PrivateAssets="all" />
+    <PackageReference Update="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.1" PrivateAssets="All" />
     <AdditionalFiles Include="$(RepoEngPath)\stylecop.json">
       <Visible>false</Visible>
     </AdditionalFiles>
--- a/src/AutoRest.CSharp/Common/AutoRest/Plugins/MemberRemoverRewriter.cs
+++ b/src/AutoRest.CSharp/Common/AutoRest/Plugins/MemberRemoverRewriter.cs
@@ -114,14 +114,14 @@ namespace AutoRest.CSharp.AutoRest.Plugins
                 if (arguments.Length == 0)
                 {
                     var fullName = typeSymbol.ToDisplayString(_fullyQualifiedNameFormat);
-                    ErrorHelpers.ThrowError($"CodeGenSuppress attribute on {fullName} must specify a method name as its first argument.");
+                    ErrorHelpers.ThrowError($"CodeGenSuppress attribute on {fullName} must specify a member name as its first argument.");
                 }
 
-                if (arguments.Length == 1 || arguments[0].Kind != TypedConstantKind.Primitive || arguments[0].Value is not string)
+                if (arguments.Length == 1 && arguments[0].Kind != TypedConstantKind.Primitive && arguments[0].Value is not string)
                 {
                     var attribute = attributeData.ApplicationSyntaxReference.GetText();
                     var fullName = typeSymbol.ToDisplayString(_fullyQualifiedNameFormat);
-                    ErrorHelpers.ThrowError($"{attribute} attribute on {fullName} must specify a method name as its first argument.");
+                    ErrorHelpers.ThrowError($"{attribute} attribute on {fullName} must specify a member name as its first argument.");
                 }
 
                 if (arguments.Length == 2 && arguments[1].Kind == TypedConstantKind.Array)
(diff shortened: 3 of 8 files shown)


heaths:
@m-nash is there a way to test injection before we do it? Being unable to repro the analyzer error here (see #3261 comments) - but having "fixed" this locally in the nupkg cache and it worked - this *should* work but I would rather double check and remove some CA1825 suppressions in the Azure/azure-sdk-for-net repo.

eng/CodeAnalysis.ruleset
```diff
@@ -214,4 +214,7 @@
         <Rule Id="SA1650"  Action="None" />          <!-- Element documentation should be spelled correctly -->
         <Rule Id="SA1651"  Action="None" />          <!-- Do not use placeholder elements -->
     </Rules>
+    <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.NetAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.CSharp.NetAnalyzers">
+        <Rule Id="CA1825" Action="Error" />          <!-- Avoid unnecessary zero-length array allocations -->
```
annelo-msft, line 218: Curious: is this adding the rule so it is applied to generated libraries?

heaths: No. That's only for projects that reference this ruleset, which are only those projects in this repo. Besides, with `WarningsAsErrors=true` in our azure-sdk-for-net repo, it's already an error for all those projects unless suppressed explicitly.

heaths:
Adding the constructor overload showed a bug in our Client SDK analyzers that wasn't caught before because the compiler would generate a `new Type[0]` expression:

```C#
                if (arguments.Length == 0)
                {
                    var fullName = typeSymbol.ToDisplayString(_fullyQualifiedNameFormat);
                    ErrorHelpers.ThrowError($"CodeGenSuppress attribute on {fullName} must specify a method name as its first argument.");
                }
```

At some point we added support to suppress generating properties, and this code was never removed (or perhaps never should've been added if properties were always supported).

m-nash:
@heaths can you run the [regen preview against this PR](https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#preview-changes-in-azure-sdk-for-net)?

heaths:
> No pipelines are associated with this pull request.

@weshaggard what would it take to make `/azp run` work here?

weshaggard:
@heaths it should already support those pipelines.

heaths:
How exactly does a GitHub repo associate with an AzDO pipeline? Does it need it listed as an artifact?

heaths:
The specific pipeline Michael wanted is not listed: [Autorest Regen Preview](https://dev.azure.com/azure-sdk/internal/_build?definitionId=5990&_a=summary)

weshaggard:
>How exactly does a GitHub repo associate with an AzDO pipeline? Does it need it listed as an artifact?

It is done by hooking up a pipeline in ADO to a repo in github and it hooks it up. Note that it only supports being associated with one ADO project (not sure if that is the case here or not). Also the pipeline needs to support PR triggers.

weshaggard:
By default we disable PR triggers for internal pipelines. If we want to support this one then we will need to enable it and make sure the comments are required so it doesn't automatically trigger.

heaths:
Thanks for the clarification. CI and PR triggers are disabled (not that we need the former), but I like the "Requires /azp comment" idea. @m-nash you see any problem with enable that? Then one can just comment `/azp run Autorest Regen Preview` without all the extra steps.

m-nash:
There are some api changes in the regen PR that I didn't expect.  @heaths  do you know why these are there?

heaths:
> There are some api changes in the regen PR that I didn't expect. @heaths do you know why these are there?

@m-nash should just be this one: https://github.com/Azure/autorest.csharp/pull/3262/files#diff-0d2afd3e69419b5a1ad3c162cfab5e48dbb96827394aba2f9c4f87ef37ab6a8bR14 , which was pretty much the whole point of this PR. Code Analysis was erring on the compiler emitting `new object[] {}` instead of `Array.Empty<object>()` for the `params` argument, but the latter can't be encoded in metadata like the former. So I added an overload that just handles a member name (a property, specifically). While I was at it, I fixed a bunch of the other CA1825 warnings (as errors).

m-nash:
We should hunt down why the other api changes are in those libraries as I don't think they are planned by this PR.  It does say this branch is out of date so perhaps lets sync and rerun the regen preview

heaths:
> I'll do that, but where are you seeing other API changes? I don't see any comments or anything on this PR.

Oh, I see. In the insertion PR e.g., https://github.com/Azure/azure-sdk-for-net/pull/35806/checks?check_run_id=13016034656

I can't imagine why some of those changes are in there, especially since I rebased on this on upstream/feature/v3 right before force pushing. And it seems there's a couple different type of changes going on with the generated source. Any thoughts?

heaths:
Ah, I may have an idea. Looking at KV sources, since I own those, it looks like my suppression is being ignored now: https://github.com/Azure/azure-sdk-for-net/blob/50025bb8f51968e543aeb884d7173ad6614bb845/sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultSetting.cs#L15

May be the problem with the other files.

pshao25:
> @heaths I would be fine with that, @chunyu3, @ArcturusZhang, @archerzz, @pshao25, @lirenhe do you guys see any issues with Heath's suggestion?

@heaths we don't want Autorest Regen Preview to be triggered automatically when a PR is created or a new commit pushed. We hope it should be started manually. If your suggestion still follows those, I'm OK with that.

heaths:
> > @heaths I would be fine with that, @chunyu3, @ArcturusZhang, @archerzz, @pshao25, @lirenhe do you guys see any issues with Heath's suggestion?
> 
> @heaths we don't want Autorest Regen Preview to be triggered automatically when a PR is created or a new commit pushed. We hope it should be started manually. If your suggestion still follows those, I'm OK with that.

No, it won't trigger automatically. A member of the repo can issue a comment with only `/azp run Autorest Regen Preview` to run it. Much easier.

heaths:
I've got this nearly working, but for some reason the *.Serialization.cs is still trying to serialize the property when it's not even defined. The `ModelTypeProvider` seems to completely ignored suppressed members, though it was working before my change. It never consulted suppressed members, though.

Supposedly, this problem was fixed in https://github.com/dotnet/roslyn-analyzers/pull/6114 but when I updated our Code Analysis packages, I still see it. Honestly, I'd rather just have the fix to the warning than trying to hack Autorest, even if the latter is also a good thing i.e., you're trying to suppress a property and not a method - or even a method with no parameters (unlikely for generated code, though).

heaths:
Given the bug tail that seems to be growing, now that we upgraded to NetAnalyzers that has the fix, I'm closing this in favor of Azure/azure-sdk-for-net#35885.

Issue #3181, opened 2023-03-02
The generator is not generating ModelFactory methods for some of the models


ArcturusZhang:
Related example and occurrence of this issue see this comment: https://github.com/Azure/azure-sdk-for-net/pull/33110#discussion_r1121266069

apattath:
@ArcturusZhang, @m-nash - is anyone looking into this issue? The issue was originally raised from our PR (linked above) for Release candidate SDK. We're now planning to make our SDK release for GA (GA is at the end of March)

ArcturusZhang:
> @ArcturusZhang, @m-nash - is anyone looking into this issue? The issue was originally raised from our PR (linked above) for Release candidate SDK. We're now planning to make our SDK release for GA (GA is at the end of March)

this is not a blocker for your GA. We could always add the missing entries back using customization code.
If you believe some of them are needed in model factory, please add them in customization code.

m-nash:
@ArcturusZhang that is true if this is purely a "missing models" issue.  If that is just the tip of the iceberg and some of the items that are there are not correct then it could have a breaking change post GA.  I would like us to do this investigation for root cause so we know the answer to this before they GA if possible.

ArcturusZhang:
@m-nash I just debugged into this library to see what is happening for those public models.
EmailSendResult: internal property.
EmailMessage: its `IncludeDeserialize` is false (currently by design)
EmailContent: same as above
EmailRecipients: same as above
EmailAddress: same as above
EmailAttachment: same as above
Above are all the public models we have in this RP (others are internal)

Let me know if you would like to loose the constraint. We are doing bunch of checks before we generate an entry for a model: https://github.com/Azure/autorest.csharp/blob/524aa32ab8bb4196da1c9800cae9c839603d0ba5/src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs#L217
1. it must be public
2. it must have a deserializer (usage contains output)
3. it must not be abstract unless it has a discriminator
4. it must not have internal properties except for the safe flattened properties
5. it must at least have one readonly property (exclude IList and IDictionary)
6. its internal constructor must not have a parameter with internal type

We are doing the extra checks to minimize the size of our model factory. Let me know if any of these checks do not make sense, so that we could change them.
Some of them exist before I change anything on the model factories.

m-nash:
I think we should remove check 2.

[...]

PR #3266, opened 2023-04-12
Loose the condition of model factory

ArcturusZhang:

<del>This depends on the feature that unifies the constructors of mgmt and DPG: https://github.com/Azure/autorest.csharp/issues/3178</del>
Depends on https://github.com/Azure/autorest.csharp/pull/3200

Reason:
Now in mgmt/HLC, we always have a serialization constructor for every model (it is different from the public initialization constructor), but it is not written into the generated code when a model does not need deserialization. In DPG, we will not construct a different internal constructor when the model does not need deserialization, instead we will just return the public initialization constructor when an internal constructor is asked.

# Description

We currently have the following conditions for a model to be included in the model factory.
1. it must be public
2. it must have a deserializer (usage contains output)
3. it must not be abstract unless it has a discriminator
4. it must not have internal properties except for the safe flattened properties
5. it must at least have one readonly property (exclude IList and IDictionary)
6. its internal constructor must not have a parameter with internal type

This PR removes condition 2.

# Checklist

# Ready to Land?


ArcturusZhang:
Superseded by https://github.com/Azure/autorest.csharp/pull/3574

Issue #3265, opened 2023-04-12
Serialization file not generated for fixed enum when not setting unreferenced-types-handling


pshao25:
Repro steps:
1. Go to `node_modules\@azure-tools\cadl-ranch-specs\http\enums\fixed\main.cadl`, replace `DaysOfWeekEnum` to 
```
@fixed
enum DaysOfWeekEnum {
  Monday: "A",
  Tuesday: "B",
  Wednesday: "C",
  Thursday: "D",
  Friday: "E",
  Saturday: "F",
  Sunday: "G",
}
```
2. Add `"enums/fixed",` to `eng\testProjects.json`.
3. Run `.\eng\Generate.ps1 enums/fixed`

PR #3268, opened 2023-04-12
fix issue in post processor

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3265

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/PostProcessing/ReferenceMapBuilder.cs
```diff
@@ -127,11 +127,22 @@ private async Task ProcessExtensionSymbol(INamedTypeSymbol extensionClassSymbol,
...
+                // this is to hook the extension class like this:
+                // internal static class FooExtensions
+                // {
+                //     public static string ToSerialString(this Foo foo) => foo.ToString();
+                //     public static Foo ToFoo(this string foo) => // omit body
+                // }
+
+                // if this is an extension method, we add it to the reference map of the type it is extending to pretend that this class is a part of that type
+                // handle the first method above
+                if (methodSymbol.Parameters[0].Type is INamedTypeSymbol typeSymbol)
```
ArthurMa1978, line 139: Is that possible that the method has no parameter at all?

ArcturusZhang: make sense. Thanks for pointing out! Let me fix this

ArcturusZhang: well after a thought - we have a check whether this method is extension method above, and we only get here when it is an extension method, therefore it is guaranteed to have at least one parameter.
But nevertheless I still made the change

PR #3276, opened 2023-04-17
introduce a doc for mitigating breaking change

ArcturusZhang:
# Description

This PR introduces a guidance for our service partners to keep our generated libraries away from breaking changes when they are using newer swagger to update the generated SDKs

# Checklist

# Ready to Land?


ArcturusZhang:
@m-nash @ArthurMa1978 I believe introducing documents guiding how to mitigate breaking changes is a long term work because we have so many cases and practices therefore I would like to invoke you to have a review on this small piece first so that we could put the link of this document in the Mgmt-Release-Process wiki first and we could improve to include other more complicated cases later.

docs/mgmt/mitigate_breaking_changes.md
```diff
@@ -0,0 +1,108 @@
...
+## Overview
+
+Whenever an update is applied to an existing SDK library, there is a chance that the update will introduce breaking changes. Breaking changes are changes that are not backwards compatible and may require client code to be updated. For example, a method may be removed, a method may be renamed, or a method may have its signature changed. Breaking changes are inconvenient for clients that need to update their code in order to consume the latest version of the SDK. Therefore we would like to eliminate breaking changes whenever possible.
+
+To enforce this, we have a check in `dotnet build` that will fail if a breaking change is detected. This check is performed by the APIComPat tool. The tool is run as part of the build process and will compare the current assembly with the previous assembly (if it exists). If any breaking changes are detected, the build will fail, and you will see the following error message:
+```
+error : MembersMustExist : Member '<MemberName>' does not exist in the implementation but it does exist in the contract.
+```
+
+## General Principles
```
ArthurMa1978, line 12: Can refer https://azure.github.io/azure-sdk/general_introduction.html

ArcturusZhang: Do we have a section for breaking changes in this doc? I only find the SDK design guidelines which are not very relevant to the topic here

ArthurMa1978: Here https://dev.azure.com/msazure/AzureWiki/_wiki/wikis/AzureWiki.wiki/37684/Breaking-Changes

ArcturusZhang: Since this is a public repo and this is an internal material, I do not like to include this in this document.
I am kind of focusing on "how do we resolve the build failures reported by ApiCompat tool in azure-sdk-for-net" in this document

docs/mgmt/mitigate_breaking_changes.md
```diff
@@ -0,0 +1,108 @@
...
+1. We should try our best to avoid behavioral breaking changes. The way we mitigate breaking changes should not change the behavior of the API.
+1. The SDK is designed to support multiple API versions, therefore we should ensure that the mitigation steps do not introduce breaking changes for older API versions.
+
+## Mitigation Steps
+
+### `MembersMustExist` Errors
+
+`MembersMustExist` errors from ApiCompat are usually seen when a member is removed in the new version of the SDK. This can happen when a method/property/class is removed, or when a method/property/class is renamed. To mitigate this, we should introduce the removed member back and add the `[EditorBrowsable(EditorBrowsableState.Never)]` attribute to the member. This will hide the member from intellisense, and will prevent the member from being used in client code. This will not be a breaking change for clients that are using the member, but it will prevent new clients from using the member.
+
+To mitigate a `MembersMustExist` error for removed **properties** or **methods**, follow these steps:
```
m-nash, line 24: Can we specify the actual scenario here?  MembersMustExist can happen for multiple reasons.

- Property was renamed
- Property was removed
- Property had its type changed
- etc

It is not necessary to document all cases on the first PR but lets at least specify which case is being handled here.

docs/mgmt/mitigate_breaking_changes.md
```diff
@@ -0,0 +1,108 @@
...
+}
+```
+
+If the property still works for old API versions, and it no longer works for newer API versions, please **DO NOT** add the `Obsolete` attribute to this property. In addition, to make sure our SDK still works for old API versions, please add a new internal constructor that takes all the properties including this removed property, and call the base generated internal constructor, and we need to add the removed property to its serialization method and deserialization method (if any).
+
+```csharp
+// in file `src/Customization/TheTypeContainingRemovedProperty.cs`
+public partial class TheTypeContainingRemovedProperty
+{
+    internal TheTypeContainingRemovedProperty(string propertyThatDefinedInOtherPartialClassFiles, string removedProperty) : base(propertyThatDefinedInOtherPartialClassFiles)
```
live1206, line 52: Doubt about the `base` constructor here, it would be from `Object` not the constructor within the current type.

ArcturusZhang: Good point - it should be `this` instead of `base`

docs/mgmt/mitigate_breaking_changes.md
```diff
@@ -0,0 +1,108 @@
...
+To mitigate a `MembersMustExist` error for removed **properties** or **methods**, follow these steps:
+
+1. Create a file outside the `src/Generated` directory with the same name of its containing class. Usually they should go into the `src/Customization` or `src/Custom` directory, if there is not such a directory, please create one.
+1. Create a partial class in the file with the same name as the containing class.
+1. Introduce the property or method back (**including its original implementation and xml documentation**) with the `[EditorBrowsable(EditorBrowsableState.Never)]` attribute.
+1. Introduce everything needed in the original implementation back as well, including any helper methods, classes, etc.
+
+In addition, for removed properties, we need to do some additional steps to make sure our SDK still works for old API versions. Please see the section below for more details.
+
+If the property no longer works in all API versions, please add the `[Obsolete("<Put a message here>", false)]` attribute to the property and just leave the property with its auto getter and setter.
```
KrzysztofCwalina, line 33: I don't think we should just leave the auto properties. Auto properties will create a false impression that they work. We should add the property and throw NotSupportedException from both the getter and the setter.

docs/mgmt/mitigate_breaking_changes.md
```diff
@@ -0,0 +1,108 @@
...
+public partial class TheTypeContainingRemovedProperty
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [Obsolete("This property no longer works in all API versions. Please use the <NewProperty> instead.", false)]
+    public string RemovedProperty { get; } // if it previously does not have a setter
+    //public string RemovedProperty { get; set; } // if it previously has a setter
+}
+```
+
+If the property still works for old API versions, and it no longer works for newer API versions, please **DO NOT** add the `Obsolete` attribute to this property. In addition, to make sure our SDK still works for old API versions, please add a new internal constructor that takes all the properties including this removed property, and call the base generated internal constructor, and we need to add the removed property to its serialization method and deserialization method (if any).
```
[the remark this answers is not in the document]
KrzysztofCwalina: We should throw NotSupportedException if an API does not work. If a property does not work only in some API versions, we should throw conditionally.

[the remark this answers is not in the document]
KrzysztofCwalina: Also, we should discuss the point about not obsoleting. I think it would be better to obsolete if the newest API version does not work. If only older versions don't work, then I agree that we should not obsolete.

jsquire:
@ArcturusZhang:  Can we please update the PR description to discuss the goals and audience for the guide?   My comments assume that this is intended to guide service partners generating management libraries, based on the content.

docs/mgmt/mitigate_breaking_changes.md
m-nash: This deviates slightly from our old recommendations.  Original we were going to use Obsolete and eventually remove since anyone treating warnings as errors would immediately get a compilation error (better than the throw runtime error) and when we turn it to an error later everyone would get this.

Happy to change this if we think throwing on these is better but wanted to make sure.

ArcturusZhang:
> @ArcturusZhang: Can we please update the PR description to discuss the goals and audience for the guide? My comments assume that this is intended to guide service partners generating management libraries, based on the content.

Yes exactly. I will update the description accordingly

Issue #3277, opened 2023-04-17
[Feature Request] Introduce new attribute to let the generator could include the customized properties into serialization and deserialization


ArcturusZhang:
**Describe the issue or request**

Currently, if we would like to change anything in the serialization/deserialization, we may also would like those customized properties to be included in the serialization and deserialization.
Now, to do this, we have to write a customization method that overrides the whole process of serialization and deserialization. Something like:
```csharp
// in file `src/Customization/TheTypeContainingRemovedProperty.cs`
public partial class TheTypeContainingRemovedProperty
{
    internal TheTypeContainingRemovedProperty(string propertyThatDefinedInOtherPartialClassFiles, string removedProperty) : base(propertyThatDefinedInOtherPartialClassFiles)
    {
        RemovedProperty = removedProperty;
    }

    [EditorBrowsable(EditorBrowsableState.Never)]
    public string RemovedProperty { get; } // if it previously does not have a setter
    //public string RemovedProperty { get; set; } // if it previously has a setter
}

// in file `src/Customization/TheTypeContainingRemovedProperty.Serialization.cs`
public partial class TheTypeContainingRemovedProperty : IUtf8JsonSerializable
{
    // the full method name here is required to make sure the generate do not generate another method with the same name
    void global::Azure.Core.IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
    {
        /* here we just copy the generated IUtf8JsonSerializable.Write method content here */
        writer.WriteStartObject(); // this is copied here
        writer.WritePropertyName("propertyThatDefinedInOtherPartialClassFiles"); // this is copied here
        writer.WriteStringValue(PropertyThatDefinedInOtherPartialClassFiles); // this is copied here
        // these lines are the serialization of the removed property - it should change based on the type of the property
        writer.WritePropertyName("removedProperty");
        writer.WriteStringValue(RemovedProperty);
        writer.WriteEndObject(); // this is copied here
    }

    internal static TheTypeContainingRemovedProperty DeserializeTheTypeContainingRemovedProperty(JsonElement element)
    {
        /* here we just copy the generated DeserializeTheTypeContainingRemovedProperty method content here */
        Optional<string> propertyThatDefinedInOtherPartialClassFiles = default; // this is copied here
        Optional<string> removedProperty = default;
        foreach (var property in element.EnumerateObject()) // this is copied here
        {
            if (property.NameEquals("propertyThatDefinedInOtherPartialClassFiles")) // this is copied here
            {
[...]

PR #3278, opened 2023-04-17
New attribute to add customized property into serialization/deserialization

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3277

A rough design could be found here: https://github.com/Azure/autorest.csharp/issues/3277#issuecomment-1514177614

# Description

Current behavior:
1. Adding a setter to a property using customization code changes nothing in serialization.
2. Adding a new property that does not exist in the spec changes nothing in ctor or serialization.

Since we have quite a few refactor PRs that may change the schema based model generation code a lot, I put the POC inside the DPG model generation to show how it works.

# Checklist

# Ready to Land?


annelo-msft (commented):
Per @lirenhe's observation, it might be possible to use [MutableJsonDocument](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.cs) for this.  

For example, if you had the base payload and wanted to add a property, you could potentially do:

```csharp
MutableJsonDocument mdoc = MutableJsonDocument.Parse(response.Content);
mdoc.RootElement.SetProperty(“requiredInt”, RequiredIntValue);
mdoc.WriteTo(writer);
```

ArcturusZhang:
> Per @lirenhe's observation, it might be possible to use [MutableJsonDocument](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.cs) for this.
> 
> For example, if you had the base payload and wanted to add a property, you could potentially do:
> 
> ```cs
> MutableJsonDocument mdoc = MutableJsonDocument.Parse(response.Content);
> mdoc.RootElement.SetProperty(“requiredInt”, RequiredIntValue);
> mdoc.WriteTo(writer);
> ```

Well this looks different from the goal that this PR is trying to achieve - maybe it is related to your that original comment thread in openai SDK because it is appending something in the request. The goal here is trying to make the serialization part generated from the generator configurable by using customized code, previously when we want to do that we have to override the whole serialization or deserialization method.

Changing it inside the request is also achievable but it gets more complicated if the model we would like to change is not the root, for instance, if we have
```
model Body
{
foo: Foo;
}
model Foo
{
bar: Bar
}
model Bar
{
}
```
And we would like to add or change the serialization of one property in Bar, using the `JsonDocument` thing is not very convenient because when we have a chance to change it, the request has been finished, if we want to change it in `Bar`, we will have to know which property it is in Foo and Body.

The way we are trying to achieve in this PR is that we just change things in Bar only, we never touch things outside of Bar (like the property name in Foo that uses Bar).

Therefore I believe the `MutableJsonDocument` is for a different purpose of this POC

annelo-msft:
> Therefore I believe the MutableJsonDocument is for a different purpose of this POC

That makes sense - I agree with your assessment @ArcturusZhang.  It might be able to be used, but it would require a lot of changes to the existing structure of the code, and I don't think it is worth doing that.  Thanks for looking at this!

ArcturusZhang:
@m-nash could you please take a review on this PR? I believe this PR is working perfectly now.

test/TestProjects/Customizations-Typespec/src/Customizations/ModelToAddAdditionalSerializableProperty.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// This property is mocking this scenario:
+        /// In the SDK, this property is defined as int, but in the actual traffic, this property comes as a string.
+        /// We use this attribute to fix its serialization and deserialization using the following two methods
+        /// </summary>
+        [CodeGenMemberSerializationHooks(nameof(WriteRequiredInt), nameof(DeserializeRequiredInt))]
+        public int RequiredInt { get; set; }
+
+        private void WriteRequiredInt(Utf8JsonWriter writer)
+        {
+            writer.WriteStringValue(RequiredInt.ToString());
```
m-nash, line 29: We should use aggressive inlining on these methods but since they are hand written we should add an analyzer that ensures this.  Please create a follow up work item in azure-sdk-tools to add this analyzer.

m-nash:
Please add the regen after merging with feature/v3 just to verify there should be zero changes in azure-sdk-for-net.

ArcturusZhang:
@m-nash the regen is here and it has no change in the libraries: https://github.com/Azure/azure-sdk-for-net/pull/36327

m-nash (approved):
Please add an issue to update the docs in this file https://github.com/Azure/autorest.csharp#customizing-the-generated-code to include this an example on how to use this new attribute

ArcturusZhang:
> Please add an issue to update the docs in this file https://github.com/Azure/autorest.csharp#customizing-the-generated-code to include this an example on how to use this new attribute

thanks for the heads up. I will update it soon

PR #3283, opened 2023-04-18
Support raw dictionary as input and output

### LINKED ISSUE #3127: [cadl-ranch]Support dictionary as input and output
cadl:
```
op get(): Record<int32>;
op post(@body body: Record<int32>): void;

op get(): Record<InnerModel>;
op post(@body body: Record<InnerModel>): void;
```

proposal:
```C#
// int
public virtual Response<IReadOnlyDictionary<string, int>> GetInt32ValueValue(CancellationToken cancellationToken = default)
{
    ...
        Response response = GetInt32Value(context);
        IReadOnlyDictionary<string, int> value = default;
        using var document = JsonDocument.Parse(response.ContentStream);
        Dictionary<string, int> dictionary = new Dictionary<string, int>();
        foreach (var property in document.RootElement.EnumerateObject())
        {
            dictionary.Add(property.Name, property.Value.GetInt32());
        }
        value = dictionary;
        return Response.FromValue(value, response);
    ...
}
public virtual async Task<Response> PutAsync(IDictionary<string, int> body, CancellationToken cancellationToken = default)
{
    Argument.AssertNotNull(body, nameof(body));

    var content = new Utf8JsonRequestContent();
    content.JsonWriter.WriteStartObject();
    foreach (var item in body)
    {
        content.JsonWriter.WritePropertyName(item.Key);
        content.JsonWriter.WriteObjectValue(item.Value);
    }
    content.JsonWriter.WriteEndObject();
    Response response = await PutAsync(content, context).ConfigureAwait(false);
    return response;
}

// Model
public virtual Response<IReadOnlyDictionary<string, InnerModel>> GetModelValueValue(CancellationToken cancellationToken = default)
{
    ...
        Response response = GetModelValue(context);
        IReadOnlyDictionary<string, InnerModel> value = default;
        using var document = JsonDocument.Parse(response.ContentStream);
        Dictionary<string, InnerModel> dictionary = new Dictionary<string, InnerModel>();
        foreach (var property in document.RootElement.EnumerateObject())
        {
            dictionary.Add(property.Name, InnerModel.DeserializeInnerModel(property.Value));
        }
        value = dictionary;
        return Response.FromValue(value, response);
    ...
}
public virtual async Task<Response> PutAsync(IDictionary<string, InnerModel> body, CancellationToken cancellationToken = default)
{
    Argument.AssertNotNull(body, nameof(body));

    var content = new Utf8JsonRequestContent();
    content.JsonWriter.WriteStartObject();
    foreach (var item in body)
    {
        content.JsonWriter.WritePropertyName(item.Key);
        content.JsonWriter.WriteObjectValue(item.Value);
    }
    content.JsonWriter.WriteEndObject();
    Response response = await PutAsync(content, context).ConfigureAwait(false);
    return response;
}
```

pshao25:
Fix #3127


src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -266,10 +266,21 @@ private void WriteConvenienceMethod(LowLevelClientMethod clientMethod, Convenien
...
+                if (requestBody is RequestContentRequestBody body)
+                {
+                    var bodyInputParamter = clientMethod.RequestMethod.Operation.Parameters.First(p => p.Location == Common.Input.RequestLocation.Body);
+                    var bodyConvenienceParameter = convenienceMethod.ProtocolToConvenienceParameterConverters.First(c => c.Protocol.Name == KnownParameters.RequestContent.Name).Convenience;
+                    if (TypeFactory.IsDictionary(bodyConvenienceParameter.Type))
+                    {
+                        var serialization = SerializationBuilder.Build(clientMethod.RequestMethod.Operation.RequestBodyMediaType, bodyInputParamter.Type, bodyConvenienceParameter.Type);
+                        bodyOverrideVariable = RequestWriterHelpers.WriteSerializeContent(_writer, null, serialization, $"{bodyConvenienceParameter.Name}");
+                    }
+                }
```
ArcturusZhang, line 280: I believe this part is writing code like this:
```csharp
            var content = new Utf8JsonRequestContent();
            content.JsonWriter.WriteStartObject();
            foreach (var item in body)
            {
                content.JsonWriter.WritePropertyName(item.Key);
                content.JsonWriter.WriteBooleanValue(item.Value);
            }
            content.JsonWriter.WriteEndObject();
```
and it should belong to the `_converter` callback (maybe?) because this is converting the body parameter (a dictionary) to a request content instance. It looks really cumbersome that we have these here.

I am thinking that could we introduce a helper method here to help us? We could have it in the `Generator.Shared`
```csharp
internal static RequestContent FromDictionary<T>(IDictionary<string, T> dict)
```

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -83,8 +83,13 @@ public static FormattableString GetIdentifiersFormattable(this IEnumerable<strin
             return conversion == null ? constantFormattable : $"{constantFormattable}{conversion}";
         }
 
-        public static FormattableString GetConversionFormattable(this Parameter parameter, CSharpType toType)
+        public static FormattableString GetConversionFormattable(this Parameter parameter, CSharpType toType, CodeWriterDeclaration? bodyOverrideVariable = null)
         {
+            if (bodyOverrideVariable != null)
+            {
+                return $"{bodyOverrideVariable}";
+            }
```
ArcturusZhang, line 91: If we have the converter helper method, we will never need this "override", we could just check the toType and the type of parameter like below:
```
if (TypeFactory.IsDictionary(parameter.Type) && toType.EqualsIgnoreNullable(typeof(RequestContent)))
return $"<the helper method>";
```

ArcturusZhang (commented):
I think if we could support dictionary as input type in this way, we could support array as input as well.
Now we are explicitly sabotage the "array input" case but are not doing the same for the "dictionary input" case, which just looks weird.

And I think it could make our generator a lot cleaner if we could have a static helper method in `Generator.Shared` which converts from Dictionary to RequestContent. I believe the performance impact by extracting this method out should be minimum

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
pshao25: The point here is to pass in the content variable.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
pshao25: The problem is: we rely this part to create a `content` variable, which is used in `GetParameterValues.PrepareConvenienceMethodParameters`, and the `converter` depends on `GetParameterValues.PrepareConvenienceMethodParameters`. So we cannot move this part to `converter`.

ArcturusZhang:
@pshao25 you did not get my point.
Here is a commit that shows another approach which I believe is cleaner: https://github.com/Azure/autorest.csharp/blob/3ca2bf076b8e91a04343ca6e552959e67c72d43e/src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs#L71
You will find one test case for duration is failing because [this method](https://github.com/Azure/autorest.csharp/blob/3ca2bf076b8e91a04343ca6e552959e67c72d43e/src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs#L71) is not adding `TimeSpan` as one of its case and we could add it to this method, or create more overload of `RequestContentHelper.FromDictionary` (for instance add a `IDictionary<string, TimeSpan>` version) to solve it.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -266,10 +266,21 @@ private void WriteConvenienceMethod(LowLevelClientMethod clientMethod, Convenien
...
-                var contextVariable = new CodeWriterDeclaration(KnownParameters.RequestContext.Name);
-
-                var (parameterValues, converter) = convenienceMethod.GetParameterValues(contextVariable);
+                var requestBody = clientMethod.RequestMethod.Request.Body;
+                CodeWriterDeclaration? bodyOverrideVariable = null;
+                if (requestBody is RequestContentRequestBody body)
+                {
+                    var bodyInputParamter = clientMethod.RequestMethod.Operation.Parameters.First(p => p.Location == Common.Input.RequestLocation.Body);
+                    var bodyConvenienceParameter = convenienceMethod.ProtocolToConvenienceParameterConverters.First(c => c.Protocol.Name == KnownParameters.RequestContent.Name).Convenience;
+                    if (TypeFactory.IsDictionary(bodyConvenienceParameter.Type))
```
chunyu3, line 275: Do we support nested dictionary, e.g. IReadOnlyDictionary<int, IReadOnlyDictionary<int, int>> in this solution?
or the element of the dictionary is another collection?

pshao25: Yes, we have tests "Dictionary_RecursiveModelValue_get" and "Dictionary_RecursiveModelValue_put".

chunyu3 (approved):

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -85,6 +85,11 @@ public static FormattableString GetIdentifiersFormattable(this IEnumerable<strin
 
         public static FormattableString GetConversionFormattable(this Parameter parameter, CSharpType toType)
         {
+            if (TypeFactory.IsDictionary(parameter.Type) && toType.EqualsIgnoreNullable(typeof(RequestContent)))
```
ArcturusZhang, line 88: Since our helper method does not accept `IReadOnlyDictionary`, could we change the condition to `IsReadWriteDictionary`?

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
ArcturusZhang, line 1: since we no longer need the changes in this file, maybe we could revert these changes

src/assets/Generator.Shared/RequestContentHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace Azure.Core
+{
+    internal static class RequestContentHelper
+    {
+        public static RequestContent FromDictionary<T>(IDictionary<string, T> dictionary) where T : notnull
```
ArcturusZhang, line 14: Could we create an issue to track the format story we discussed offline?

pshao25: I don't think we will take any actions on that format issue until we run into it. So we could open up then.

test/AutoRest.Shared.Tests/RequestContentHelperTests.cs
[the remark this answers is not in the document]
pshao25: Actually, I think the pupose of this test is to test different framework. As for the different types of cases, we already have it in the cadl-ranch tests. But since there is no harm to add more, I still add two more cases in this test class.

lirenhe (approved):

PR #3290, opened 2023-04-19
Fix #3192: Move generated protocol method samples to use DynamicJson

### LINKED ISSUE #3192: Move generated protocol method samples to use DynamicJson
The [DynamicJson](https://learn.microsoft.com/en-us/dotnet/api/azure.core.dynamic.dynamicjson?view=azure-dotnet-preview) type allows users to work with raw JSON more easily through a dynamic layer.

We would like to update the generated samples for protocol methods to improve their readability by using DynamicJson.  For example, [this sample](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/src/Generated/Docs/OpenAIClient.xml#L229-L245) from the OpenAI library would change to:

```
Console.WriteLine((string)result.Id);
Console.WriteLine((string)result.Object);
Console.WriteLine((long)result.Created);
Console.WriteLine((string)result.Model);
Console.WriteLine((string)result.Choices[0].text);
Console.WriteLine((int)result.Choices[0].index);
Console.WriteLine((string)result.Choices[0].logprobs.tokens[0]);
Console.WriteLine((double)result.Choices[0].logprobs.token_logprobs[0]);
Console.WriteLine((double)result.Choices[0].logprobs.top_logprobs[0]["<test>"]);
Console.WriteLine((int)result.Choices[0].logprobs.text_offset[0]);
Console.WriteLine((string)result.Choices[0].finish_reason);
Console.WriteLine((int)result.Usage.completion_tokens);
Console.WriteLine((int)result.Usage.prompt_tokens);
Console.WriteLine((int)result.Usage.total_tokens);
```

Note: a functional sample using recorded test JSON can be found at [openai_sample.cs](https://gist.github.com/annelo-msft/c29ddc33b63e0db02bfae949a6fffc26#file-openai_sample-cs).

AlexanderSher:


samples/Azure.AI.OpenAI/Generated/Docs/OpenAIClient.xml
```diff
@@ -0,0 +1,505 @@
...
+var endpoint = new Uri("<https://my-service.azure.com>");
+var client = new OpenAIClient(endpoint, credential);
+
+var data = new {
+    input = new {},
+};
+
+Response response = await client.GetEmbeddingsAsync("<deploymentId>", RequestContent.Create(data));
+
+dynamic result = response.Content.ToDynamicFromJson();
```
annelo-msft, line 18: We will need to pass options into the `ToDynamicFromJson()` call, or PascalCase getters will fail when the JSON is camelCased.  Could we pass [DynamicDataNameMapping.PascalCaseGettersCamelCaseSetters](https://apiview.dev/Assemblies/Review/b67e68cfee4841bda045f895c90ad383#Azure.Core.Dynamic.DynamicDataNameMapping)?

src/assets/Azure.Core.Shared/OperationPoller.cs
```diff
@@ -15,51 +17,52 @@ namespace Azure.Core
     /// </summary>
     internal sealed class OperationPoller
     {
-        private readonly DelayStrategyInternal _delayStrategy;
+        private readonly DelayStrategy _delayStrategy;
```
annelo-msft, line 20: Curious: Are LRO changes required for sample updates?

samples/Azure.AI.OpenAI/Generated/Docs/OpenAIClient.xml
AlexanderSher: Is it different than `DynamicJsonOptions.Default`?

src/assets/Azure.Core.Shared/OperationPoller.cs
AlexanderSher: Build will fail without this update. I can create a separate PR for it.

AlexanderSher: In fact, I think I must do it.

annelo-msft (commented):
Is there a way to run the samples to validate correctness?  Ideally we would use these to test the DynamicData type to improve our confidence in its quality before we release it GA.

samples/Azure.AI.OpenAI/Generated/Docs/OpenAIClient.xml
annelo-msft: It has the same behavior.  I am considering removing `DynamicJsonOptions` from the final API when we GA the type.

src/assets/Azure.Core.Shared/OperationPoller.cs
AlexanderSher: https://github.com/Azure/autorest.csharp/pull/3281 - @JoshLove-msft already created a PR

Issue #3291, opened 2023-04-19
[BUG] Different CultureInfo can cause wrong URL query in AppendQueryDelimited() in RequestUriBuilderExtensions.cs 


annelo-msft:
We will eventually want to move RequestUriBuilderExtensions.cs into Azure.Core shared source, but in order to address [this customer issue](https://github.com/Azure/azure-sdk-for-net/issues/35431) in time for their next release, it would be good to make the fix in generator shared source in the meantime.

PR #3311, opened 2023-04-25
fix culture issue in query appending

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3291

# Description

# Checklist

# Ready to Land?


lirenhe (approved):

lirenhe:
@ArcturusZhang,  could you regen the SDK?

ArcturusZhang:
Well it has some issues on net461 target. I need to investigate it further.

src/assets/Generator.Shared/RequestUriBuilderExtensions.cs
```diff
@@ -6,120 +6,137 @@
...
         {
-            builder.AppendQuery(name, value.ToString(), escape);
+            builder.AppendQuery(name, ConvertToString(value), escape);
         }
 
-        public static void AppendQueryDelimited<T>(this RequestUriBuilder builder, string name, IEnumerable<T> value, string delimiter, bool escape = true)
+        public static void AppendQueryDelimited<T>(this RequestUriBuilder builder, string name, IEnumerable<T> value, string delimiter, bool escape = true) where T : notnull
         {
-            builder.AppendQuery(name, string.Join(delimiter, value), escape);
+            var stringValues = value.Select(v => ConvertToString(v));
```
annelo-msft, line 138: Using Linq sometimes has fairly negative perf implications - have you looked at the difference in performance from using this approach vs. adding the overloads we discussed in the issue?  Do we know which approach is favorable from a performance perspective?

src/assets/Generator.Shared/RequestUriBuilderExtensions.cs
```diff
@@ -6,120 +6,137 @@
...
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Xml;
 
 namespace Azure.Core
 {
     internal static class RequestUriBuilderExtensions
     {
+        private static string ConvertToString(object value, string? format = null)
```
annelo-msft, line 16: If all these extension methods go through the same code path, what is the value of having the separate extension methods?  Would it make sense to remove them and add a single `AppendPath(object value, string? format, bool escape)` method?  Why or why not?

test/AutoRest.Shared.Tests/RequestUriBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,389 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading;
+using NUnit.Framework;
+
+namespace Azure.Core.Tests
+{
+    public class RequestUriBuilderExtensionsTests
```
annelo-msft, line 12: It's good to see a dedicated test for this class -- thanks for adding it!  Would it make sense to add tests for the new cases to the test `PathsClient` and `PathsRestClient` as well?

src/assets/Generator.Shared/RequestUriBuilderExtensions.cs
ArcturusZhang: I did not do this change because I would like to minimize the change - this method is private therefore if we change the AppendPath to apply on `object`, we might have new cases get into this path, but if we do in this way, we will never have extra cases.

test/AutoRest.Shared.Tests/RequestUriBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,389 @@
...
+            Assert.AreEqual(expected, uri.ToUri().ToString());
+        }
+
+        [TestCase(null, new double[] { 3.14, -3.14 }, ",", "http://localhost/something?double=3.14%2C-3.14")]
+        [TestCase(null, new double[] { 3.14, -3.14 }, " ", "http://localhost/something?double=3.14 -3.14")]
+        [TestCase(null, new double[] { 3.14, -3.14 }, "|", "http://localhost/something?double=3.14|-3.14")]
+        [TestCase("fr-FR", new double[] { 3.14, -3.14 }, ",", "http://localhost/something?double=3.14%2C-3.14")]
+        [TestCase("fr-FR", new double[] { 3.14, -3.14 }, " ", "http://localhost/something?double=3.14 -3.14")]
+        [TestCase("fr-FR", new double[] { 3.14, -3.14 }, "|", "http://localhost/something?double=3.14|-3.14")]
+        public void AppendQueryDelimited_DoubleValue(string? culture, double[] value, string delimiter, string expected)
```
ArcturusZhang, line 376: add a test case for AppendQueryDelimited with a string array with null values.

test/AutoRest.Shared.Tests/RequestUriBuilderExtensionsTests.cs
ArcturusZhang: Unfortunately we must remove this test case because of this issue: https://github.com/Azure/azure-sdk-for-net/issues/35828
The escaping logic is running differently in net6.0 and net461 therefore some of the test cases here will never pass.
Because of the same reason, I have to say that we may not add a case for `PathsClient` or `PathsRestClient` because they might have the same issue.
Therefore to minimize the change, we will add a new test case for the `ConvertToString` method because
- The `RequestUriBuilder` in `Azure.Core` only accepts strings or `Span<char>` for `AppendPath` and `AppendQuery`
therefore the responsibility of the code in `Generator.Shared` is just making sure we have the correct conversion from various types to string, and then `RequestUriBuilder` in Azure.Core should handle them correctly.

src/assets/Generator.Shared/RequestUriBuilderExtensions.cs
```diff
@@ -6,120 +6,137 @@
...
         }
 
         public static void AppendQuery(this RequestUriBuilder builder, string name, Guid value, bool escape = true)
         {
-            builder.AppendQuery(name, value.ToString(), escape);
+            builder.AppendQuery(name, ConvertToString(value), escape);
         }
 
-        public static void AppendQueryDelimited<T>(this RequestUriBuilder builder, string name, IEnumerable<T> value, string delimiter, bool escape = true)
+        public static void AppendQueryDelimited<T>(this RequestUriBuilder builder, string name, IEnumerable<T> value, string delimiter, bool escape = true) where T : notnull
```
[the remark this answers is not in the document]
ArcturusZhang: I will remove this notnull constraint.
But since our generated code does not have the nullable static check enabled, this will not cause any difference.

src/assets/Generator.Shared/RequestUriBuilderExtensions.cs
annelo-msft: Sounds good.

annelo-msft (commented):
It would be good to understand the Linq question before merging this change.  Thanks!

src/assets/Generator.Shared/RequestUriBuilderExtensions.cs
ArcturusZhang: we could introduce a string builder here instead

ArcturusZhang: Well I cannot find a way to do this without linq - 
we could do it in an array but it need us to know when we are getting the last since we need to omit an extra `,`
I do not think we could do that and check if the string ends with a comma because you will never know if the input values contain comma at the end.
when we are trying to identify if the element is the last, we will have to use linq method - either ToArray, ToList or Count.
Therefore I believe linq method here is inevitable.
Do you have a recommendation?

ArcturusZhang: Well since this discussion is here for quite a while, and if we have better solution for this linq issue, we could do that later.
Therefore I would like to merge this PR for now, and then we could introduce some other changes to polish

PR #3331, opened 2023-04-27
Add instructions to protocol methods

### LINKED ISSUE #3326: [Doc] add instruction of convenience method in the summary of protocol method
When there is convenience method, add a instruction in the summary of protocol method to tell customer there is convenience method to use, 
something like:

This is the advanced method use <see cref="OtherMethod"> for an overload that provides fully qualified model types

pshao25:
Fix #3326

### DIFF
--- a/samples/AnomalyDetector/Generated/Docs/AnomalyDetectorClient.xml
+++ b/samples/AnomalyDetector/Generated/Docs/AnomalyDetectorClient.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <doc>
   <members>
-    <member name="DetectUnivariateEntireSeriesAsync(RequestContent,RequestContext)">
+    <member name="DetectUnivariateEntireSeriesAsync(global::Azure.Core.RequestContent,global::Azure.RequestContext)">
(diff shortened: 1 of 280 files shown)


samples/AnomalyDetector/Generated/Docs/AnomalyDetectorClient.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <doc>
   <members>
-    <member name="DetectUnivariateEntireSeriesAsync(RequestContent,RequestContext)">
+    <member name="DetectUnivariateEntireSeriesAsync(global::Azure.Core.RequestContent,global::Azure.RequestContext)">
```
m-nash, line 4: we want this to go in the summary on the method itself not in the xml doc file.

ArcturusZhang: Or maybe we could have both?

chunyu3: MSDoc will retrieve <summary> in client file, and the external xml doc. We may need add this one place, othewise, msdoc may have duplicate instruction.

@m-nash I have create a PR (https://github.com/Azure/autorest.csharp/pull/3334) to add summary in the method itself, and follow the format described in the email. Would you please help to review that.

PR #3334, opened 2023-04-28
add convenience method instruction in summary

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3326

- When there is convenience method, add a instruction in the summary of protocol method to tell customer there is convenience method to use,

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -261,10 +261,10 @@ public static IDisposable WriteMethodDeclaration(this CodeWriter writer, MethodS
             });
         }
 
-        public static CodeWriter WriteMethodDocumentation(this CodeWriter writer, MethodSignatureBase methodBase)
+        public static CodeWriter WriteMethodDocumentation(this CodeWriter writer, MethodSignatureBase methodBase, FormattableString? summaryText = null)
```
ArcturusZhang, line 264: A general comment: could we put those extra information into the method signature when we are building them instead of adding a patch after everything has been built?

chunyu3: The content depends on if there is any convenience method, and need the information of convenience method. If we want to put this method signature, we need to pass convenience method as parameter, it will couple one with other, and need to teach user that the parameter passed must be the convenience method of this client method. It may be error-prone.

m-nash (approved):

PR #3345, opened 2023-05-03
Generate samples as code as well as doc to validate compilation

m-nash:
# Description

# Checklist

# Ready to Land?


jsquire (approved):

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
```diff
@@ -40,6 +41,10 @@ public static async Task ExecuteAsync(GeneratedCodeWorkspace project, InputNames
                 lowLevelClientWriter.WriteClient();
                 project.AddGeneratedFile($"{client.Type.Name}.cs", codeWriter.ToString());
                 project.AddGeneratedDocFile($"Docs/{client.Type.Name}.xml", xmlDocWriter.ToString());
+
+                var exampleCompileCheckWriter = new ExampleCompileCheckWriter(client);
+                exampleCompileCheckWriter.Write();
+                project.AddGeneratedFile($"../../tests/Generated/Samples/Samples_{client.Type.Name}.cs", exampleCompileCheckWriter.ToString());
```
chunyu3, line 47: In each SDK, there are manually written samples under `tests\Samples`, Shall we put the auto-generated samples under `tests\Samples\Generated` ? In that way, all the samples (manual or auto) are in the same directory.

eng/Generate.ps1
```diff
@@ -119,7 +129,9 @@ function Add-Directory ([string]$testName, [string]$directory, [boolean]$forTest
     }
 
     if ($forTest) {
-        Add-Swagger-Test $testName $directory $testArguments
+        if(Test-Path "$directory/autorest.md") {
```
chunyu3, line 132: Current the configure file is `readme.md`
```suggestion
        if(Test-Path "$directory/readme.md") {
```

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
m-nash: Our convention for generated files is always `subDir/Generated`

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -328,14 +329,14 @@ internal MethodSignatureBase GetEffectiveCtor()
...
+        private bool IsParamMatch(IReadOnlyList<Parameter> methodParameters, INamedTypeSymbol[] suppressionParameters)
         {
             if (methodParameters.Count != suppressionParameters.Length)
                 return false;
 
             for (int i = 0; i < methodParameters.Count; i++)
             {
-                var namedSymbol = ((INamedTypeSymbol)suppressionParameters[i].Value!);
+                var namedSymbol = suppressionParameters[i];
                 var suppressionParamType = namedSymbol.GetCSharpType(_typeFactory);
```
ArcturusZhang, line 340: We have a method to assert if a symbol and a CSharpType are the same type: https://github.com/Azure/autorest.csharp/blob/4fd31dcd869853538d7f47f6cc98b2fe5b022c3a/src/AutoRest.CSharp/Common/Utilities/NamedTypeSymbolExtensions.cs#L35

heaths:
@jsquire when you reviewed, did you look at the preview insertion API? I think these need to be open long enough - and during typical business hours - for impacted owners to review.

For example, now the code generator is unneccessarliy adding `new RequestContext()` to every call for samples, thus making samples unnecessarily harder (and misleading) for devs to copy. My understanding is that we want samples to be as simple as possible, which is why we also don't show async in samples despite it being a best practice for network calls. Do you agree, @KrzysztofCwalina or @tg-msft?

tg-msft:
Isn't this part of the change to make `RequestContext` required on bodyless endpoints so we can drop the confusing `Value` suffix on evolved conveniences?  I don't think the generator is adding it unnecessarily here, but it's admittedly a little hard to tell without any context in the PR description.  😄 

It's also being rushed through for the next UX study so the short PR window is necessary, but any feedback is always welcome and we can address it in a follow-up PR.

heaths:
> It's also being rushed through for the next UX study so the short PR window is necessary, but any feedback is always welcome and we can address it in a follow-up PR.

@tg-msft certainly, and @m-nash and I discussed just that offline. In general, I think we need to refine the process of autorest.csharp changes a bit and started a Teams thread in our internal team channel with our team. Please have a look.

PR #3354, opened 2023-05-05
DPG confidence level mechanism

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3356
Fixes https://github.com/Azure/autorest.csharp/issues/3581 by preventing the generator to generate convenience method for it.

# Description

This PR adds a `IsConfident` property to all input type, which usually defaults to `true`. This property only exists in the emitter right now.
For all the types we currently have:
- `InputLiteralType`: literal types are not confident when the underlying type is numbers, which causes its underlying `InputEnumType` to be not confident
- `InputUnionType`: union types are always not confident (for now, there might be some exceptions for this but we did not implemented yet)
- `InputModelType`: model types are not confident when any type of its properties is not confident
- `InputEnumType`: enum types are not confident when it is used as the underlying type of an `InputLiteralType`. For other cases, `InputEnumType` should be confident.
- `InputListType` and `InputDictionaryType`: these container types do not have its own confident levels, or it should stay the same confident level as its `ElementType` or its `ValueType`
- Other types are always confident

This PR adds a `ConvenienceMethodOmitReason` property to the `InputOperation` type as well. This is calculated in the emitter and populated to the generator. 

If any of the types used directly or indirectly by this operation (including parameters, responses) is not confident (`IsConfident: false`), the `GenerateConvenienceMethod` of the operation is still true, but the `ConvenienceMethodOmitReason` will have the value `TypeNotConfident`. And in the generator, we will generate the convenience method as an internal method.
There are three values for `ConvenienceMethodOmitReason` enum:
1. `TypeNotConfident`: explained above.
2. `SuppressedInTypeSpec`: operation will have this reason when the convenienceAPI decorator is not configured.
3. `PatchOperation`: operation will have this reason when it has a http verb of PATCH.

In the generator, we will add an extra line of comment above the protocol method if its corresponding convenience method is not generated based on its reason.

This PR also changes the generated type of unions to `object`.
Previously the union types in the model will generate from its first type, for instance, if a union is `string | int32`, in the generated code, we will have `string`. This PR changes this behavior to generate them as `object`.
Since it is `object`, when serializing, it should work properly depending on what type the customer assigns. Some of them are unsupported and exceptions will be thrown.
When deserializing, our SDK has the functioanlity to properly deserialize it to its correct type based on the value in json. If it is primitive type, a primitive type will return. If it is an array, an array will return. If it is a complex object or dictionary, a `Dictionary<string, object>` will return.

Considering the first part of this change will make the operation with unions internal, and the post processor will automatically make the models in use of the operation internal as well, therefore the change to type of object should be acceptable.
In real life, if a service team wants to introduce SDKs with union types, the operation involved with unions will not be public. They need to write their own method to make it public, and customize the corresponding properties and the related serializations (use `CodeGenMemberSerializationHooks` for instance)

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -116,6 +116,59 @@ private bool IsConvenienceMethodMeaningful()
...
+                InputDictionaryType dictType => IsLowConfident(dictType.KeyType, visitedModels) || IsLowConfident(dictType.ValueType, visitedModels),
+                InputModelType modelType => IsModelLowConfident(modelType, visitedModels),
+                _ => type.IsLowConfident
+            };
+            return result;
+        }
+
+        private static bool IsModelLowConfident(InputModelType modelType, HashSet<InputModelType> visitedModels)
+        {
+            if (visitedModels.Contains(modelType)) // if this has been checked or is being checked, we just return false and let others to decide this
```
pshao25, line 152:
```C#
class NestedModel
{
    public AUnion{set; get;}
}
class Model
{
    public NestedModel {get; set;}
}
```

After NestedModel checked, the Model will return false, which acutually should return true.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -116,6 +116,59 @@ private bool IsConvenienceMethodMeaningful()
                 || !_returnType.Convenience.Equals(_returnType.Protocol);
         }
 
+        private bool IsLowConfident()
+        {
+            var visitedModels = new HashSet<InputModelType>();
```
pshao25, line 121: I'm thinking this method do once per method. But models are shared across library. So there will be some redundant calculation.

pshao25:
Now that we put the flag at model, maybe we could skip generating models when we see this flag instead of we generate it and then remove by post processor?

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
```diff
@@ -3,4 +3,4 @@
 
 namespace AutoRest.CSharp.Common.Input;
 
-internal abstract record InputType(string Name, bool IsNullable = false) { }
+internal abstract record InputType(string Name, bool IsLowConfident, bool IsNullable = false) { }
```
chunyu3, line 6: Can `IsLowConfident` has a default value `false` so that by default we will not set it, only when it has `IsLowConfident` value in the json, we will set it?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
ArcturusZhang: Yes there is duplicated calculation.
Now the IsLowConfident flag does not propagate. I will see if we could make it propagates on the emitter so that we could just read them here.

chunyu3 (commented):
When a model has a property whose model is not confident, then this model will also be marked as lowConfident. Do you implement this?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
ArcturusZhang: make sense. let me see how to resolve this

chunyu3 (changes requested):
When a model has a property whose model is not confident, then this model will also be marked as lowConfident. Do you implement this?

lirenhe:
@ArcturusZhang, in the PR description, I saw a couple of 'maybe' for different types, could you be more specific about what are high confidence and what are low confidence?

ArcturusZhang:
> @ArcturusZhang, in the PR description, I saw a couple of 'maybe' for different types, could you be more specific about what are high confidence and what are low confidence?

Updated. Previously it does not consider the propagation. "May be" means it will be determined by the propagation.
I updated it so that propagation is ALWAYS explicitly explained

samples/AnomalyDetector/src/Generated/tspCodeModel.json
```diff
@@ -77,7 +77,8 @@
    ],
    "IsExtensible": false,
    "IsNullable": false,
-   "Usage": "Input"
+   "Usage": "Input",
+   "IsLowConfident": false
```
lirenhe, line 81: I am wondering whether we could use enum instead and call it ConfidentLevel.
Then we could support more values in the future.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
```diff
@@ -146,17 +150,5 @@ private static IReadOnlyList<InputEnumTypeValue> NormalizeValues(IReadOnlyList<I
 
             return concreteValues;
         }
-
```
lirenhe, line 149: Can we remove those code directly?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -60,39 +65,70 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
...
                 {
-                    model = CreateInputModelTypeInstance(id, name, ns, accessibility, deprecated, description, usageString, discriminatorValue, discriminatorPropertyName, baseModel, properties, resolver);
+                    model = CreateInputModelTypeInstance(id, name, ns, accessibility, deprecated, description, usageString, discriminatorValue, discriminatorPropertyName, baseModel, properties, isLowConfident, isNullable, resolver);
                     reader.Read();
                     CreateProperties(ref reader, properties, options);
-                    if (reader.TokenType != JsonTokenType.EndObject)
-                    {
-                        throw new JsonException($"{nameof(InputModelType)}.{nameof(InputModelType.Properties)} must be the last defined property.");
-                    }
+                    //if (reader.TokenType != JsonTokenType.EndObject)
```
lirenhe, line 71: Any future use of them?  Can we remove those code directly?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -60,39 +65,70 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
...
                 resolver.AddReference(id, model);
             }
             return model;
         }
 
+        private static bool TryReadArrayWithConverter<T>(ref Utf8JsonReader reader, string propertyName, JsonSerializerOptions options, ref ICollection<T> array)
+        {
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException();
```
lirenhe, line 105: Could you provide some information when throw exception?  That will be helpful for troubleshooting.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -60,39 +65,70 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
...
-            var model = new InputModelType(name, ns, accessibility, deprecated, description, usage, properties, baseModel, new List<InputModelType>(), discriminatorValue, discriminatorPropertyValue);
+            var model = new InputModelType(name, ns, accessibility, deprecated, description, usage, properties, baseModel, new List<InputModelType>(), discriminatorValue, discriminatorPropertyValue, isLowConfident, isNullable);
             if (id != null)
             {
                 resolver.AddReference(id, model);
             }
             return model;
         }
 
+        private static bool TryReadArrayWithConverter<T>(ref Utf8JsonReader reader, string propertyName, JsonSerializerOptions options, ref ICollection<T> array)
```
lirenhe, line 101: Where is this function used?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecSerialization.cs
```diff
@@ -3,13 +3,65 @@
...
 {
     internal static class TypeSpecSerialization
     {
+        /// <summary>
+        /// Since we need to parse the json using references, we require the "Properties" property in the json to appear the last after all other properties of the object.
+        /// This method here is a workaround that reads all the content of the json and reorders the properties so that "Properties" is the last property.
+        /// </summary>
+        /// <param name="json"></param>
+        /// <returns></returns>
+        internal static string NormalizeJson(string json)
```
lirenhe, line 20: How is this function related to the LowConfidence features?

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/propertyType.test.ts
```diff
@@ -198,6 +199,7 @@ describe("Test GetInputType for enum", () => {
                         }
                     ],
                     IsExtensible: false,
+                    IsLowConfident: false,
```
lirenhe, line 202: From the PR description, there are multiple combination of LowConfident cases, could you add more testcases to ensure more scenarios are covered?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
ArcturusZhang: These are experimental changes. Will revert in the future

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
ArcturusZhang: The magic behind `JsonException` makes it could automatically append some useful information about line number, property name, etc, therefore this should be fine.
But this is experimental code, will remove in the future.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
ArcturusZhang: this is experimental, will remove later

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecSerialization.cs
ArcturusZhang: this is only experimental

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/propertyType.test.ts
ArcturusZhang: will add more test cases in the unit test of the emitter
not covered right now

samples/AnomalyDetector/src/Generated/tspCodeModel.json
ArcturusZhang: I changed it to a boolean IsConfident since we only have two values to control whether we will have convenience method or not.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
ArcturusZhang: this should be resolved now.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
ArcturusZhang: I changed the implementation to calculation this in the emitter so that there should be no more duplicated calculations

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/propertyType.test.ts
ArcturusZhang: unit tests added

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -38,6 +39,8 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
             string? usageString = null;
             string? discriminatorPropertyName = null;
             string? discriminatorValue = null;
+            bool isNullable = false;
+            bool isLowConfident = false;
```
chunyu3, line 43:
```suggestion
            bool isConfident = false;
```

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs
```diff
@@ -105,11 +108,11 @@ public static InputPrimitiveType ReadPrimitiveType(ref Utf8JsonReader reader, st
             return primitiveType;
         }
 
-        public static InputPrimitiveType CreatePrimitiveType(string? inputTypeKindString, bool isNullable)
+        public static InputPrimitiveType CreatePrimitiveType(string? inputTypeKindString, bool isLowConfident, bool isNullable)
```
chunyu3, line 111:
```suggestion
        public static InputPrimitiveType CreatePrimitiveType(string? inputTypeKindString, bool isConfident, bool isNullable)
```

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -130,7 +130,8 @@ private void CreateModels(IDictionary<InputModelType, ModelTypeProvider> models,
                         actualBase,
                         Array.Empty<InputModelType>(),
                         "Unknown", //TODO: do we need to support extensible enum / int values?
-                        null)
+                        null,
+                        false)
```
chunyu3, line 134: What's the unknowDeriveModel for? Will we generate this model? If we set it isConfident = false, we will not generate it.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
```diff
@@ -0,0 +1,78 @@
...
+function getModelConfident(
+    type: InputModelType,
+    visitedModels: Set<InputModelType>
+): boolean {
+    if (visitedModels.has(type)) return true; // this means this has been calculated before or being calculated right now, return
+    visitedModels.add(type);
+
+    if (type.IsConfident !== null) return type.IsConfident;
+
+    const isBaseLowConfident = type.BaseModel
```
chunyu3, line 35:
```suggestion
    const isBaseConfident = type.BaseModel
```

chunyu3 (changes requested):
change `XXXLowConfident` to `isConfident`

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    return type.IsConfident ?? true; // by default they are confident
+}
+
+function getModelConfident(
+    type: InputModelType,
+    visitedModels: Set<InputModelType>
+): boolean {
+    if (visitedModels.has(type)) return true; // this means this has been calculated before or being calculated right now, return
```
chunyu3, line 30:
```suggestion
    if (visitedModels.has(type)) return type.IsConfident; // this means this has been calculated before or being calculated right now, return
```

pshao25: And we don't need `visitedModels` anymore. `type.IsConfident` itself is already a cache.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
```diff
@@ -0,0 +1,78 @@
...
+    visitedModels.add(type);
+
+    if (type.IsConfident !== null) return type.IsConfident;
+
+    const isBaseLowConfident = type.BaseModel
+        ? getModelConfident(type.BaseModel, visitedModels)
+        : true;
+
+    let isModelLowConfident = isBaseLowConfident;
```
pshao25, line 39: We could short circuit here.

pshao25, line 39: I think you don't save the reaults of properties right? If so we could short circuit here.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
```diff
@@ -0,0 +1,78 @@
...
+    return (
+        getConfident(type.KeyType, visitedModels) &&
+        getConfident(type.ValueType, visitedModels)
+    );
+}
+
+// we use this method to get the confident level value of a type
+// TODO -- in the future, we could consider if we could convert the InputType and all its derived types into a concrete type instead of interfaces, so that they could contain method and method override. If so, we no longer need to do this.
+export function isConfident(type: InputType | undefined): boolean {
+    if (type === undefined) return true; // if undefined, return true
```
pshao25, line 67: If `null`, we throw error?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -221,6 +223,30 @@ export function createModelForService(
...
+        for (const op of client.Operations) {
+            let isOperationConfident = true;
+            for (const parameter of op.Parameters) {
+                // skip the special parameters
+                if (
+                    parameter.IsApiVersion ||
+                    parameter.IsEndpoint ||
+                    parameter.IsContentType
+                ) {
+                    parameter.Type.IsConfident = true; // we never recognize these as low confident level
```
pshao25, line 237: Why are we setting values in the propagation process?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -221,6 +223,30 @@ export function createModelForService(
...
+                    continue;
+                }
+                isOperationConfident &&= isConfident(parameter.Type);
```
pshao25, line 240: Some short circuit could be applied here.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -113,6 +113,7 @@ export function createModelForService(
         Type: {
             Name: "String",
             Kind: InputTypeKind.String,
+            IsConfident: true,
```
pshao25, line 116: There are many places have such pattern like "we know it should be true so set true". So scattered, hard to maintain. Could we put all of them into same `getConfident` function?

src/TypeSpec.Extension/Emitter.Csharp/src/type/inputType.ts
```diff
@@ -8,6 +8,7 @@ import { InputTypeKind } from "./inputTypeKind.js";
 export interface InputType {
     Name: string;
     IsNullable: boolean;
+    IsConfident: boolean | null;
```
pshao25, line 11: I'm think the value on `InputType` is never used in the generator. Maybe we could set it to undefined after all the calculation so the it doesn't need to pass to the downstream?

src/AutoRest.CSharp/Common/Input/InputTypes/InputPrimitiveType.cs
```diff
@@ -3,37 +3,37 @@
 
 namespace AutoRest.CSharp.Common.Input;
 
-internal record InputPrimitiveType(InputTypeKind Kind, bool IsNullable = false) : InputType(Kind.ToString(), IsNullable)
+internal record InputPrimitiveType(InputTypeKind Kind, bool IsConfident, bool IsNullable = false) : InputType(Kind.ToString(), IsConfident, IsNullable: IsNullable)
```
lirenhe, line 6: Can we have the default value of 'IsConfident' since most of the cases, this value is true.

lirenhe: In general, as we are confident about the most of the features, can we have the default value of true in all inputTypes?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -74,25 +79,56 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
                 }
             }
 
-            return model ?? CreateInputModelTypeInstance(id, name, ns, accessibility, deprecated, description, usageString, discriminatorValue, discriminatorPropertyName, baseModel, properties, resolver);
+            return model ?? CreateInputModelTypeInstance(id, name, ns, accessibility, deprecated, description, usageString, discriminatorValue, discriminatorPropertyName, baseModel, properties, isLowConfident, isNullable, resolver);
```
lirenhe, line 82: Same above,  we should use isConfident now.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
```diff
@@ -0,0 +1,78 @@
+import {
```
lirenhe, line 1: Please add the header for this new file.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -221,6 +223,30 @@ export function createModelForService(
...
+                    parameter.IsEndpoint ||
+                    parameter.IsContentType
+                ) {
+                    parameter.Type.IsConfident = true; // we never recognize these as low confident level
+                    continue;
+                }
+                isOperationConfident &&= isConfident(parameter.Type);
+            }
+            for (const response of op.Responses) {
+                isOperationConfident &&= isConfident(response.BodyType);
```
lirenhe, line 243: If the response.BodyType is the base type and one of its derived type is lowConfidence, will this operation also be considered as low confidence?  Same for the request.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
ArcturusZhang: should be true. I will fix this.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
ArcturusZhang: > And we don't need `visitedModels` anymore. `type.IsConfident` itself is already a cache.

this is not true, we need that cache in a case of a model referencing itself.

ArcturusZhang: > 

this I agree

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
ArcturusZhang: because they are constructed elsewhere separately

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
ArcturusZhang: This function is not only calculating this value for this model, it is also "propagating" the calculation to every model it is referencing, therefore we should not short circuit here to ensure every type of the property has been gone through.
The property itself has a cache to ensure there is no duplicated calculations.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
ArcturusZhang: This is for type, not the isConfident property.

There is some cases that when we call this, the type might be undefined (for instance, when we call this for the response), I do not want to put a if statement on the caller, therefore I put it here.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
ArcturusZhang: If you remember, in the first iteration of this PR, I want to calculate the `isConfident` in one place exactly in the way you are saying here, and you are the person that says we should calculate this property when we construct them.
I do not want to change it back

src/AutoRest.CSharp/Common/Input/InputTypes/InputPrimitiveType.cs
ArcturusZhang: No. Making a property optional has more possibilities to get it wrong values when we construct them.
We have done that for the `IsNullable` property at many places.

These constructors will nearly never be used other than deserialization, therefore making all of them "REQUIRED" has more benefits than making it optional as you said to have a default value which ensures we ALWAYS explicitly set them

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
ArcturusZhang: Yes. The calculation part of IsConfident answers your question, code here is USING the calculated values.

This is how we calculate isConfident for a model:

1. check if it has baseModel. Set baseModel?.IsConfident as the intial value of `IsConfident` of this model.
2. go through all the properties to get the IsConfident value of this model.

Therefore if base is false, the initial value is false, and the result will be false as well.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
ArcturusZhang: Instead of changing back and forth, I am thinking that we should change our interfaces to classes so that it has constructors, and we could put those values in the constructor or the implementation of the property.
For instance, an issue the current design cannot resolve is that for the container types such as list and dictionary, the IsConfident should always redirect the value to its element type or value type, but now we cannot do that because during the calculation and the propagation, we always get to the container first, and then get to its element or value type. During this when we set the `IsConfident` of the element, we can no longer set the IsConfident of the container because it is very likely to be out of scope now (taking the `model Error { details: Error[] }` as an example to think this)

src/TypeSpec.Extension/Emitter.Csharp/src/type/inputType.ts
ArcturusZhang: I agree with your first comment but I do not see a value of doing it.
I do not see why making it undefined could declare it to be `IsConfident: boolean`. If we could do that, we should declare it as `IsConfident?: boolean` or `IsConfident: boolean | undefined` and we have discussed why this cannot resolve the issue that `IsConfident` property show up after `Properties`.

If you are thinking that we could do:
1. Serialize the code model into a json string
2. Modify that thing to remove unnecessary IsConfident: true, and move IsConfident in front of Properties if this happens
This is so ugly that I would not like to do it.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
ArcturusZhang: This is not propagation process, per your suggestion, now propagation and calculation happen together when we constructing them.

These parameters are constructed elsewhere, and this is a kind of workaround because we still recognize the "union" of literal values as union types therefore sometimes, the content type will become "IsConfident: false"
This is a easy workaround of that issue - but in the meantime, let me see if I could resolve it from its root cause.

src/AutoRest.CSharp/Common/Input/InputTypes/InputPrimitiveType.cs
ArcturusZhang: BTW I changed IsNullable to required as well in this PR: https://github.com/Azure/autorest.csharp/pull/3373
The reason of that is that these values are important therefore we should always ensure we are setting them explicitly even if in most cases they are having the same value to avoid some hidden issues which I discovered a few in the nullable PR.
For instance [this issue](https://github.com/Azure/autorest.csharp/issues/3600) is exactly caused by the default value we set, and we forgot to deserialize its `IsNullable` value when deserializing the tspCodeModel and the compiler does not give us any warning about that.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
ArcturusZhang: Fixed but we should have a CI for that.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    return type.IsConfident ?? true; // by default they are confident
+}
+
+function getModelConfident(
+    type: InputModelType,
+    visitedModels: Set<InputModelType>
+): boolean {
+    if (visitedModels.has(type)) return type.IsConfident ?? true; // this means this has been calculated before or being calculated right now, return
```
chunyu3, line 33: Shall we make sure that, if the model is visited, the IsConfident should be calculated? It may cause problem, we set the true as isConfident if it is not calculated.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
ArcturusZhang: This is resolved.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
ArcturusZhang: The fact is that, if the model is visited, the IsConfident is calculated or being calculated right now.
Returning true here is not setting true.

The isConfident is always calculating in this way:
```
isConfident &&= returnedValue;
```
There here we return true is not making that to be true, this is skipping it based on this simply fact:
the expression `isConfident &&= true` will not change the value of `isConfident`.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
ArcturusZhang: Make sense, we should short circuit here

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
ArcturusZhang: to reduce the possible confusions, I will add some more comments here.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
```diff
@@ -0,0 +1,81 @@
...
+    visitedModels.add(type);
+
+    if (type.IsConfident !== null) return type.IsConfident;
+
+    const isBaseConfident = type.BaseModel
```
chunyu3, line 48: Just little suggestion: when we meet `false` once, we can directly return `false` and can skip further calculation.

chunyu3 (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
ArcturusZhang: Yes I already did this now. I should see it if you refresh

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
ArcturusZhang: as I said in [this comment](https://github.com/Azure/autorest.csharp/pull/3354#discussion_r1270184726) below, using optional properties here is not quite a good idea, it could lead to hidden problems.
But I agree we could omit the values in the tspCodeModel file if it is using its default values. But I do not see quite much value for that

ArcturusZhang:
Hi @m-nash could you also have a review on this PR?
There is no API changes in the regen PR even if for openai service.
I locally tried it by adding the save-inputs parameter in generation and found out that it now contains no "IsConfident: false" operations nor Union types.

ArcturusZhang:
> Now that we put the flag at model, maybe we could skip generating models when we see this flag instead of we generate it and then remove by post processor?

I would like to keep that, because when the low confident happens, there are usually possibilities that they still want the convenience methods, and they will customize them. When doing that, they still need the models, for the implementation now, they do not have to do that from scratch, they could just change the option of post processor to `internalize` so that they could get the involved models and move them to customized code so that they do not have to do those models from scratch. If we omit the generation when the model is not confident, they must do these from scratch.

src/AutoRest.CSharp/Common/Input/InputTypes/ConvenienceMethodOmitReason.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AutoRest.CSharp.Common.Input;
+
+internal enum ConvenienceMethodOmitReason
+{
+    SuppressedInTypeSpec = 0,
```
lirenhe, line 8: Ususally, the enum value of 0 is considered as a place holder or 'None'.  What is the best practise in our codebase?

ArcturusZhang: The default value of the corresponding property is null because it is nullable, therefore I did not put a None here.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/confidentLevels.ts
```diff
@@ -0,0 +1,94 @@
...
+    } else if (isInputListType(type)) {
+        return getListConfident(type, visitedModels);
+    } else if (isInputDictionaryType(type)) {
+        return getDictionaryConfident(type, visitedModels);
+    }
+
+    return type.IsConfident ?? true; // by default they are confident
+}
+
+function getModelConfident(
```
lirenhe, line 35: As those functions return boolean, can we rename them to 'isModelConfident', 'isListConfident' etc,

ArcturusZhang:
@m-nash could you please have a review?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -544,6 +549,26 @@ private void WriteProtocolMethodDocumentationWithExternalXmlDoc(LowLevelClientMe
             }
         }
 
+        private void WriteConvenienceMethodOmitReason(ConvenienceMethodOmitReason reason)
+        {
+            // TODO -- create wiki links to provide guidance here: https://github.com/Azure/autorest.csharp/issues/3624
+            switch (reason)
+            {
+                case ConvenienceMethodOmitReason.SuppressedInTypeSpec:
+                    _writer.Line($"// The convenience method of this operation is not generated because the convenience method is not turned on in typespec file.");
```
m-nash, line 558: We should consider a readonly struct which has a description, wiki link, etc and then the writer can simply write what is in the struct and no case statement is needed.

src/TypeSpec.Extension/Emitter.Csharp/samples/CadlFirstTest/Generated/tspCodeModel.json
```diff
@@ -29,6 +29,7 @@
     }
    ],
    "IsExtensible": false,
+   "IsConfident": true,
```
m-nash, line 32: I would have expected the emitter to emit an enum value for what the reason is?  Otherwise don't we have to detect it twice once in emitter and once in generator?

Or we don't have this here at all and do all the detection in the generator might be better?

ArcturusZhang: The description should explained this.
How this is implemented:
1. Every type in the emitter will have a `boolean IsConfident`.
2. The `InputOperation` has a property `ConvenienceMethodOmitReason` (enum)
3. The emitter will propagate the IsConfident value on the tree.
4. The generator does nothing, it just reads the reason from the `tspCodeModel.json`. No calculations are done in the generator. Everything is calculated in the emitter.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
ArcturusZhang: I created a class but since we do not have those entries now, I just created a new class with a `string Message` property.
We could add other things when we have those.

ArcturusZhang:
Will move the detection/calculation of the confident level (boolean or enum) to the generator (csharp part)

lirenhe:
@ArcturusZhang,  could you check the CI failure in this PR?

ArcturusZhang:
Superseded by https://github.com/Azure/autorest.csharp/pull/3643

PR #3355, opened 2023-05-05
Support internal decorator

### LINKED ISSUE #3242: [Feature]: Support the decorator of @internal
Please check the definition of this decorator:
https://azure.github.io/typespec-azure/docs/libraries/azure-dpg/reference/decorators#@Azure.ClientGenerator.Core.internal

The test was already added into Cadl-ranch:
https://github.com/Azure/cadl-ranch/pull/237

Please note: this decorator may cause breaking change for the SDK.

pshao25:
Fix #3242


src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -52,7 +52,7 @@
         "@autorest/csharp": "3.0.0-beta.20230309.1",
         "@azure-tools/typespec-autorest": "0.29.0",
         "@azure-tools/typespec-azure-core": "0.29.0",
-        "@azure-tools/typespec-client-generator-core": "0.29.0",
+        "@azure-tools/typespec-client-generator-core": "0.30.0-dev.13",
```
lirenhe, line 55: Shall we wait for the official build of TCGC?

pshao25: Then we don't have tests for this feature. We could change to official one after it releases.

src/CADL.Extension/Emitter.Csharp/src/options.ts
```diff
@@ -53,6 +53,7 @@ export const NetEmitterOptionsSchema: JSONSchemaType<NetEmitterOptions> = {
         "model-namespace": { type: "boolean", nullable: true },
         "generate-protocol-methods": { type: "boolean", nullable: true },
         "generate-convenience-methods": { type: "boolean", nullable: true },
+        "filter-out-core-models": {type: "boolean", nullable: true},
```
lirenhe, line 56: I didn't find the usage of this const,  so where do we use this?

src/CADL.Extension/Emitter.Csharp/src/type/inputType.ts
```diff
@@ -49,6 +49,7 @@ export interface InputModelType extends InputType {
     BaseModel?: InputModelType;
     DiscriminatorPropertyName?: string;
     DiscriminatorValue?: string;
+    IsInternal: boolean;
```
chunyu3, line 52: Use `Accessibility` which is to define if the operation is public/internal/private. No need to add a new property.

chunyu3 (approved):

lirenhe (approved):

Issue #3363, opened 2023-05-08
Incorrect Deserialize logic for a nullable array


pshao25:
```
model Model {
  nullableProperty: bytes[]| null;
}
```
Actual: it takes every element in that array as nullable
```C#
foreach (var property in element.EnumerateObject())
{
    if (property.NameEquals("nullableProperty"u8))
    {
        List<BinaryData> array = new List<BinaryData>();
        foreach (var item in property.Value.EnumerateArray())
        {
            if (item.ValueKind == JsonValueKind.Null)
            {
                array.Add(null);
            }
            else
            {
                array.Add(BinaryData.FromString(item.GetRawText()));
            }
        }
        nullableProperty = array;
        continue;
    }
}
```
Expected: it should take the whole array as nullable
```C#
foreach (var property in element.EnumerateObject())
{
    if (property.NameEquals("nullableProperty"u8))
    {
        if (property.Value.ValueKind == JsonValueKind.Null)
        {
            nullableProperty = null;
            continue;
        }

        List<BinaryData> array = new List<BinaryData>();
        foreach (var item in property.Value.EnumerateArray())
        {
             array.Add(BinaryData.FromString(item.GetRawText()));
        }
        nullableProperty = array;
        continue;
    }
}
```

Issue #2340, opened 2022-06-21
Generate Nullable properties on models from Cadl


annelo-msft:
[Description TBD]

annelo-msft:
Pending discussion regarding when to allow nullable properties in standard models vs. models used for JSON merge patch.

annelo-msft:
Address the following in the C# generated client design document:
- https://github.com/Azure/azure-sdk-for-net/pull/30176/files#r932512758
- https://github.com/Azure/azure-sdk-for-net/pull/30176/files#r932504877

lirenhe:
Just FYI.  I reviewed some PR from mgmt. SDK and I am told that 'null' and 'property not present in the payload' has different meaning in their service side.  https://github.com/Azure/azure-rest-api-specs/pull/19885

lirenhe:
@archerzz, as you are investigating this item, could you update with your findings?  Thanks

chunyu3:
.dotnet cannot primitively support Nullable properties.

chunyu3:
Move to DPG-Backlog

joseharriaga:
The topic of nullable properties came up in the following conversation while reviewing this TypeSpec PR for OpenAI:
🔗 https://github.com/Azure/azure-rest-api-specs/pull/23206/files/04e3094c7183fc7084f5eda3b3b6d27896da7c13#r1161981141

In sum, the `Choice` model has a required nullable property called `logprobs` of type `CompletionsLogProbabilityModel` (which is another model defined in the TypeSpec). Therefore, `logprobs` is defined as:
```text
logprobs: CompletionsLogProbabilityModel | null
```

When we generate C# code from this TypeSpec, the generated constructor includes an `Argument.AssertNotNull` for this property, which throws an exception when the value is null:

```csharp
internal Choice(string text, int index, CompletionsLogProbabilityModel logProbabilityModel, CompletionsFinishReason finishReason)
{
    Argument.AssertNotNull(text, nameof(text));
    Argument.AssertNotNull(logProbabilityModel, nameof(logProbabilityModel));

    Text = text;
    Index = index;
    LogProbabilityModel = logProbabilityModel;
    FinishReason = finishReason;
}
```
In other words, the property is not being respected as nullable.

m-nash:
@lirenhe @chunyu3 we might need to pull this out of backlog since it appears to be affecting current services if they a property that is required and nullable.

chunyu3:
> backlog
[...]

PR #3373, opened 2023-05-09
fix the required + nullable issue

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/2340
Fixes https://github.com/Azure/autorest.csharp/issues/3363
Fixes https://github.com/Azure/autorest.csharp/issues/3440
Fixes https://github.com/Azure/autorest.csharp/issues/3600

# Description

This PR fixes multiple issues around nullable properties, especially when the property is required.

This PR also updates the `InputType` to make `IsNullable` property no longer optional. This optional property `IsNullable` could be problematic because of its optionality, sometimes we forget to give it a value and it causes hidden issues.

# Checklist

# Ready to Land?


chunyu3 (approved):

chunyu3 (approved):

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/CollectionsByteProperty.cs
```diff
@@ -18,14 +18,13 @@ public partial class CollectionsByteProperty
         /// <summary> Initializes a new instance of CollectionsByteProperty. </summary>
         /// <param name="requiredProperty"> Required property. </param>
         /// <param name="nullableProperty"> Property. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="requiredProperty"/> or <paramref name="nullableProperty"/> is null. </exception>
+        /// <exception cref="ArgumentNullException"> <paramref name="requiredProperty"/> is null. </exception>
         internal CollectionsByteProperty(string requiredProperty, IEnumerable<BinaryData> nullableProperty)
```
ArcturusZhang, line 22: This corresponds to the typespec:
```
model CollectionsByteProperty {
nullableProperty: bytes[] | null;
}
```
But in our design, an array should never be null. Should we change the behavior here back?

m-nash: Good catch let me get some more information on this.

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/CollectionsByteProperty.cs
```diff
@@ -18,14 +18,13 @@ public partial class CollectionsByteProperty
...
         {
             Argument.AssertNotNull(requiredProperty, nameof(requiredProperty));
-            Argument.AssertNotNull(nullableProperty, nameof(nullableProperty));
 
             RequiredProperty = requiredProperty;
-            NullableProperty = nullableProperty.ToList();
+            NullableProperty = nullableProperty?.ToList();
```
m-nash, line 27: This needs to be initialized as a ChangeTrackingList if its `nullable` or `optional`.  If we use Optional<T> here the .ToList takes care of that for us.

ArcturusZhang: I agree that we should initialize this as a ChangeTrackingList and never have a list to be null anywhere
But here we cannot use Optional.ToList because this is in the public constructor. The `Optional<T>` only appear in the deserialization method.

ArcturusZhang:
The assignment for nullable arrays is breaking test cases like this: https://github.com/Azure/autorest.csharp/blob/0df9bf7948af9d87cd8ff18071465cef5d665e3e/test/AutoRest.TestServer.Tests/ModelShapeTests.cs#L183

test/TestProjects/ModelShapes/Generated/Models/InputModel.Serialization.cs
```diff
@@ -81,7 +81,7 @@ void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
             {
                 writer.WriteNull("RequiredNullableInt");
             }
-            if (RequiredNullableStringList != null)
+            if (RequiredNullableStringList != null && Optional.IsCollectionDefined(RequiredNullableStringList))
```
ArcturusZhang, line 84: I have to add this additional check to check undefined in order to write `null`

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/CollectionsByteProperty.Serialization.cs
```diff
@@ -32,12 +32,17 @@ internal static CollectionsByteProperty DeserializeCollectionsByteProperty(JsonE
...
+                        nullableProperty = null;
+                        continue;
+                    }
                     List<BinaryData> array = new List<BinaryData>();
                     foreach (var item in property.Value.EnumerateArray())
                     {
                         if (item.ValueKind == JsonValueKind.Null)
                         {
-                            array.Add(null);
+                            continue;
```
pshao25, line 45: Just continue? Looks not correct to me. I thought either throw error or just add null to array.

pshao25:
Let's remove all the Ignore in test\CadlRanchProjects.Tests\type-property-nullable.cs to see if it passes.

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/CollectionsByteProperty.Serialization.cs
ArcturusZhang: we should never throw

pshao25:
Could you add tests in cadl-ranch for cases that the element of array is nullable and non-nullable?

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/CollectionsByteProperty.Serialization.cs
ArcturusZhang: I will revert this

ArcturusZhang:
Current behavior for required nullable arrays:

# API and serialization side

Schema:
```
model Model
{
array: T[] | null;
}
```
where `T` is either a reference type or value type, we have no difference.

Generated code:
```csharp
public class Model
{
      // public constructor
      public Model(IEnumerable<T> array)
      {
           Array = array?.ToList() as IList<T> ?? new ChangeTrackingList<T>(); // ensure this is never null
      }
}
```

Serialization:
- if `null` is passed into the ctor, `var model = new Model(null)`, after serialization, we have:
```
{
    "array": null
}
```
- if empty array is passed into the ctor, `var model = new Model(Array.Empty<T>())`, after serialization, we have:
```
{
    "array": []
}
```

# Deserialization side

Here we have differences between value types and reference types, but in general, in the DeserializationModel static method, in the last step that it is calling the internal ctor, we will create an instance of the ChangeTrackingList to ensure it is still not null: https://github.com/Azure/autorest.csharp/blob/fa50b20e0e86d40adec0d7dd7b87f3aeae4b7eb7/test/CadlRanchProjects/type/property/nullable/src/Generated/Models/CollectionsByteProperty.Serialization.cs#L56

In the deserializing property part, for int32 and other null-unaccepting element types
```
model Model
{
array: int32[] | null;
}
```
We get code like this:
```csharp
List<int> array = new List<int>();
foreach (var item in property.Value.EnumerateArray())
{
    array.Add(item.GetInt32());
}
```
if there is a `null` in the array in the response, we will get an exception.

if the element is a reference type, we have the following code:
```csharp
List<BinaryData> array = new List<BinaryData>();
foreach (var item in property.Value.EnumerateArray())
{
    if (item.ValueKind == JsonValueKind.Null)
    {
        array.Add(null);
    }
    else
    {
        array.Add(BinaryData.FromBytes(item.GetBytesFromBase64("D")));
    }
}
```
if there is a `null` in the array in the response, we will add it into the array.

if the element is a complex model, we have the following code:
```csharp
List<CollectionItem> array = new List<CollectionItem>();
foreach (var item in property.Value.EnumerateArray())
{
    array.Add(CollectionItem.DeserializeCollectionItem(item));
}
```
if there is a `null` in the array in the response, we will add it into the array.

test/TestProjects/ModelShapes/Generated/Models/InputModel.Serialization.cs
```diff
@@ -81,34 +81,20 @@ void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
...
-                writer.WriteStartArray();
-                foreach (var item in RequiredNullableIntList)
-                {
-                    writer.WriteNumberValue(item);
-                }
-                writer.WriteEndArray();
+                writer.WriteStringValue(item);
             }
-            else
+            writer.WriteEndArray();
```
ArcturusZhang, line 90: Need to investigate this - this actually changes the serialization result

ArcturusZhang:
I am finding more and more issues when I am trying to "ensure the collection property can never be null".
Here is one related issue: https://github.com/Azure/autorest.csharp/issues/3440

ArcturusZhang:
Add back the null check for collections

ArcturusZhang:
Note - need to revert the changes on property bags since it is working right now

ArcturusZhang:
Hi @m-nash this is ready for review.

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/CollectionsByteProperty.Serialization.cs
```diff
@@ -32,6 +32,11 @@ internal static CollectionsByteProperty DeserializeCollectionsByteProperty(JsonE
                 }
                 if (property.NameEquals("nullableProperty"u8))
                 {
+                    if (property.Value.ValueKind == JsonValueKind.Null)
```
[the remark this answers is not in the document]
ArcturusZhang: I believe we should not do this.
We will get into this if when we get this json in response:
```
{ "nullableProperty": null }
```
If we want a model to serialize  into the json above, we will need to ensure the value of this collection is `null`. And we should do the same here when it is in the response, this makes the serialization and deserialization round trip.

Here are the model definition:
```csharp
public partial class Foo
{
    public Foo(IEnumerable<string> requiredNullableList)
    {
		RequiredNullableList = requiredNullableList?.ToList();
	}

	public IList<string> RequiredNullableList { get; set; }
}
```
If we have this code:
```csharp
var model = new Foo(null);
```
it will serialize into:
```json
{ "requiredNullableList": null }
```
And if you deserialize from the above json, you will get an instance of Foo, and its `RequiredNullableList` property is `null` again, something like:
```csharp
var model = new Foo(null);
var json = Serialize(model); // json is `{ "requiredNullableList": null }`
var deserializedModel = Deserialize(json);
Console.WriteLine(model.RequiredNullableList == deserializedModel.RequiredNullableList); // true
```
Same thing applies to optional list. If we have:
```csharp
var model = new Foo();
model.OptionalNullableList = null;
```
it serializes into
```json
{ "optionalNullableList": null }
``` 
and when we deserialize above json, we get an instance of `Foo` whose `OptionalNullableList` property is null.
If we have:
```csharp
var model = new Foo();
```
it serializes into nothing:
```json
{ }
```
and when we deserialize above json, we get an instance of `Foo` whose `OptionalNullableList` property is the ChangeTrackingList with `IsUndefined = true`.
I will add some more test cases in `Models-TypeSpec` to demo these.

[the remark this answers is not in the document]
ArcturusZhang: We resolved this in offline discussion

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -244,7 +244,8 @@ private InputModelType GetOrCreateModel(ObjectSchema schema, SchemaUsageProvider
                     : null,
                 DerivedModels: derived,
                 DiscriminatorValue: schema.DiscriminatorValue,
-                DiscriminatorPropertyName: schema.Discriminator?.Property.CSharpName());
+                DiscriminatorPropertyName: schema.Discriminator?.Property.CSharpName(),
```
ArcturusZhang, line 247: This is very likely to be wrong - we should use serialized name of the property because that will never change.

ArcturusZhang: Fixed in another PR

ArcturusZhang:
> @ArcturusZhang it appears that the changes are "no-op" meaning the functional effect is the same that we end up with an default changetrackinglist?

@m-nash 
This is how things work:
For optional lists, we will do this:
```csharp
Optional<IList<string>> optionalList = default;
foreach (var property in root.EnumerateObjects())
{
    if (property.NameEquals("optionalList"u8))
    {
        // this is the part handling null values
        if (property.Value.ValueKind == JsonValueKind.Null)
        {
			// we do nothing here so that the optional struct now is still "default"
			continue;
		}
    }
	// others
	return new Model(Optioanl.ToList(optionalList)); // when optionalList is default, this will return a changeTrackingList instance, and when it is not default, it returns that list instance (null included)
}
```
Code of Optional.ToList: https://github.com/Azure/autorest.csharp/blob/671929c1d9fbd74dc5bf53862bd1599ddd4af164/src/assets/Generator.Shared/Optional.cs#L73
When `optionalList` is default, `optionalList.HasValue` is false, therefore we return new ChangeTrackingList.

For required lists, we will do this:
```csharp
IList<string> requiredList = default; // which is `null`
foreach (var property in ...)
{
	if (property.NameEquals("requiredList"u8))
	{
        // this is the part handling null values
        if (property.Value.ValueKind == JsonValueKind.Null)
        {
			// we assign an instance of change tracking list here so that it is not null
			requiredList = new ChangeTrackingList<string>();
			continue;
		}
	}
	return new Model(requiredList); // no magic here
}
```

Hope this answers your question

ArcturusZhang:
Hi @m-nash I added a configuration `deserialize-null-collection-as-null-value` in this PR and it defaults to false.
When set to true, we will have previous behavior.
I have updated the GA libraries in the [corresponding regen PR](https://github.com/Azure/azure-sdk-for-net/pull/37553) by adding this flag to them, also I fixed issues in openai library, please also take a look.

The changes in generated files of openai I believe is expected, here is the corresponding typespec file: https://github.com/Azure/azure-rest-api-specs/blob/a89c3854f2d1cca8c996e00c97af3179ecb9dd72/specification/cognitiveservices/OpenAI.Inference/models/chat.completions.tsp#L196

ArcturusZhang:
@m-nash 
Last time decision on sync up meeting:
since changing the behavior of nullable collections is a behavioral breaking change, we should add a configuration to control that behavior, and I added a new configuration `deserialize-null-collection-as-null-value` which defaults to false (true corresponds to the old behavior).

Therefore in the regen PR, 
- There are quite a few RPs only have changes on autorest.md or tspConfig.yml without code changes, this is adding the config to stay on the old behavior to avoid behavior breaking changes.
- For those RPs with code changes, they are all in beta or preview therefore I did not add the new flag and breaking changes for them should be fine.
Other RPs might have API changes (for instance OpenAI) which is expected, because the initial purpose of this PR is fixing some issue around the required nullable properties. I added the corresponding typespec link in a comment in the regen to describe this change is expected.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -116,21 +116,41 @@ public ModelTypeProviderFields(InputModelType inputModel, TypeFactory typeFactor
         public IEnumerator<FieldDeclaration> GetEnumerator() => _fields.GetEnumerator();
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 
+        private static bool ShouldPropertyOmitSetter(InputModelType inputModel, InputModelProperty property, CSharpType type)
+        {
+            if (property.IsDiscriminator)
+            {
+                // Discriminator properties should be writeable
+                return false;
```
lirenhe, line 124: Could you explain why we need the discriminator property to be writeable?

ArcturusZhang: Sure

test/TestProjects/FirstTest-TypeSpec/FirstTest-TypeSpec.tsp
```diff
@@ -111,6 +111,24 @@ model Thing {
...
     requiredBadDescription: string;
+
+    @doc("optional nullable collection")
+    optionalNullableList?: int32[] | null;
+
+    @doc("required nullable collection")
+    requiredNullableList: int32[] | null;
+}
+
+@doc("A model with a few required nullable properties")
```
pshao25, line 122: Why these not in cadl-ranch?

ArcturusZhang: I do not know, maybe we could add some

AlexanderSher:
Can we please add a test that will do alias spread with required collections?

ArcturusZhang:
> Can we please add a test that will do alias spread with required collections?

Added

m-nash (approved):
@ArcturusZhang please create a follow up issue to reduce the granularity of the new flag `deserialize-null-collection-as-null-value`.  I imagine there is a small set of models / properties that have this issue and we don't want to proliferate it any further.

The granularity should probably be a list of `ModelName.PropertyName` to grandfather in.

ArcturusZhang:
> @ArcturusZhang please create a follow up issue to reduce the granularity of the new flag `deserialize-null-collection-as-null-value`. I imagine there is a small set of models / properties that have this issue and we don't want to proliferate it any further.
> 
> The granularity should probably be a list of `ModelName.PropertyName` to grandfather in.

issue created for this: https://github.com/Azure/autorest.csharp/issues/3613

test/TestProjects/FirstTest-TypeSpec/FirstTest-TypeSpec.tsp
ArcturusZhang: We could add some new test cases in cadl ranch

PR #3389, opened 2023-05-11
Fix custompage issue when there are two

### LINKED ISSUE #3386: [ContentSafety] The return type of pageable method is not correct
This method should return TextBlockItem instead of TextBlocklist：

```
public virtual Pageable<TextBlocklist> GetTextBlocklistItems(string blocklistName, int? maxCount = null, int? skip = null, int? maxpagesize = null, CancellationToken cancellationToken = default)
```

The [corresponding typespec code](https://github.com/Azure/azure-rest-api-specs-pr/blob/bdd03b95e0ccf52c3578fac731bc392734d2ea44/specification/cognitiveservices/ContentSafety/routes.tsp#L87) is:

 @summary("Get All BlockItems By blocklistName")
  @doc("Get all blockItems in a text blocklist")
  listTextBlocklistItems is Azure.Core.ResourceList<
    TextBlockItem,
    ListQueryParametersTrait<StandardListQueryParameters>
  >;

pshao25:
Fix #3386


src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -380,7 +380,7 @@ export function getInputType(
 
     function getInputModelForModel(m: Model): InputModelType {
         m = getEffectiveSchemaType(context, m) as Model;
-        const name = getFriendlyName(program, m) ?? m.name;
+        const name = getFriendlyName(program, m) ?? templateRename(m);
```
chunyu3, line 383: we need to discuss how to support template class, and how to generate it.

m-nash:
It appears the current regen shows some odd swapping from ctor to Create which doesn't seem right.  https://github.com/Azure/azure-sdk-for-net/pull/36206/files#diff-a076be848338a356eff6ca0db1f64dfb29bd595405264856797b6041c97e2ad2R123

Please send an updated regen PR.  Did we manually check this against this [branch](https://github.com/Azure/azure-sdk-for-net/pull/34933) to see that the two lists now generate as expected?

test/TestProjects/Authoring-Typespec/src/Generated/Docs/AuthoringClient.xml
```diff
@@ -619,17 +619,7 @@ var client = new AuthoringClient(endpoint);
 await foreach (var item in client.GetDeploymentsAsync("<projectName>", new RequestContext()))
 {
     JsonElement result = JsonDocument.Parse(item.ToStream()).RootElement;
-    Console.WriteLine(result.GetProperty("projectName").ToString());
```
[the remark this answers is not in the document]
pshao25: This change is expected. `GetDeploymentsAsync` should retrun `Deployment` which is 
```C#
@parentResource(Project)
model Deployment {
  @key("deploymentName")
  @visibility("read")
  @segment("deployments")
  name: string;
}
```

Previous it returns `Project` which is wrong.

pshao25:
> It appears the current regen shows some odd swapping from ctor to Create which doesn't seem right. https://github.com/Azure/azure-sdk-for-net/pull/36206/files#diff-a076be848338a356eff6ca0db1f64dfb29bd595405264856797b6041c97e2ad2R123
> 
> Please send an updated regen PR. Did we manually check this against this [branch](https://github.com/Azure/azure-sdk-for-net/pull/34933) to see that the two lists now generate as expected?

This is already updated. We have that change in regen PR is because @JoshLove-msft did this change manually https://github.com/Azure/azure-sdk-for-net/pull/36192 and testcommon doesn't have a `ci.yml` there so we cannot regen after that manual change. @JoshLove-msft I think the autogenerated code is expected, we shouldn't have manually change them. Is there any reason you did this?

Yes, I checked their entire spec and the generated code is expected.

JoshLove-msft:
> 

The testcommon library is pinned to a specific version of AutoRest so it doesn't pick up new changes. Should we try to unpin it?

pshao25:
> > 
> 
> The testcommon library is pinned to a specific version of AutoRest so it doesn't pick up new changes. Should we try to unpin it?
> 
> https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/testcommon/Azure.Graph.Rbac/src/Azure.Graph.Rbac.csproj#L16

@m-nash Seems you did this change. Any background here? I create an issue here: https://github.com/Azure/azure-sdk-for-net/issues/36235

src/CADL.Extension/Emitter.Csharp/src/lib/utils.ts
```diff
@@ -1,3 +1,16 @@
+import { Model } from "@typespec/compiler";
+
 export function capitalize(str: string): string {
     return str.charAt(0).toUpperCase() + str.slice(1);
 }
+
+export function getNameForTemplate(model: Model): string {
```
lirenhe, line 7: Can we add unit test for this?  I believe this part is easy to verify in emitter

m-nash:
> > 
> 
> The testcommon library is pinned to a specific version of AutoRest so it doesn't pick up new changes. Should we try to unpin it?
> 
> https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/testcommon/Azure.Graph.Rbac/src/Azure.Graph.Rbac.csproj#L16

We can't unpin it, we would either need to sync the release label that is pinned now and create a fix in a hotfix branch and release from there and update the version, OR we need to move this to manual code, we cannot change generated files manually.

src/CADL.Extension/Emitter.Csharp/src/lib/utils.ts
m-nash: Added a follow up since this is high priority change if you can add this in when you are back next week https://github.com/Azure/autorest.csharp/issues/3395.

m-nash:
Commented out the download shared source check until this gets resolved https://github.com/Azure/autorest.csharp/pull/3392

m-nash (approved):
Validated the fix in the content saftey pr

PR #3396, opened 2023-05-12
Use git clone instead of http download for shared files

hallipr:
To avoid throttle issues in github's usercontent api, use sparse checkout to download the azure-sdk-for-net shared files


ArcturusZhang (approved):
I personally agree that we put this temp directory inside this repo to avoid accidentally change anything outside this repo especially when we run this script locally.
Putting it inside a `gitignored` directory is really a good idea.

eng/DownloadSharedSource.ps1
```diff
@@ -2,29 +2,54 @@
...
-Get-ChildItem $downloadPath -Filter *.cs | Remove-Item;
-$baseUrl = 'https://raw.githubusercontent.com/Azure/azure-sdk-for-net/main/sdk/resourcemanager/Azure.ResourceManager/src/Shared/'
-DownloadAll $files $baseUrl $downloadPath
+$sourcePath = "$clonedPath/sdk/resourcemanager/Azure.ResourceManager/src/Shared"
+$destinationPath = "$repoRoot/src/assets/Management.Shared"
+
+Get-ChildItem $destinationPath -Filter *.cs | Remove-Item;
+CopyAll $files $sourcePath $destinationPath
+
+Start-Sleep -Seconds 1
```
pshao25, line 54: What is this sleep for?

pshao25 (approved):

eng/DownloadSharedSource.ps1
weshaggard: Alson wondering about this sleep.

weshaggard (approved):

eng/DownloadSharedSource.ps1
```diff
@@ -2,29 +2,54 @@
...
+
+function CopyAll([string[]]$files, [string]$source, [string]$destination)
 {
     foreach ($file in $files)
     {
-        Write-Host "Downloading" $file
-        $text = (Invoke-WebRequest -Uri "$baseUrl/$file").Content
-        $text.Trim() | Out-File (Join-Path $downloadPath $file)
+        Write-Host "Copying $file to $destination"
+        Copy-Item (Join-Path $source $file) (Join-Path $destination $file)
```
weshaggard, line 31: Are there any existing changes in the shared source? I think there were some EOF newlines missing so we should probably make sure we update them with this script.

m-nash:
@hallipr Can we turn the download source check back on in this PR as it was temporarily turned off due to this issue.

https://github.com/Azure/autorest.csharp/blob/feature/v3/eng/SharedCodeCheck.ps1#L6-L9

m-nash (approved):

eng/DownloadSharedSource.ps1
hallipr: The folder can be locked for delete. I used "-ErrorAction SilentlyContinue" to prevent an error from bubbling when this occors, but it only happens when there was recent io action on the folder.  I'm thinking it's probable Windows Defender scanning the files.

I would have just left the folder in place and skipped the delete, but this script is expected to be run on dev machines (presumably Windows with Defender).  The 1 second sleep appears to be enough time to let all the other consumers release the file.

eng/DownloadSharedSource.ps1
hallipr: Out-File was appending a new line to the end of the piped in string

I've corrected the 2 files that didn't have trailing new lines in the azure-sdk-for-net repo:
https://github.com/Azure/azure-sdk-for-net/pull/36250

eng/DownloadSharedSource.ps1
weshaggard: OK please add a comment in the code to that affect otherwise everyone that looks at this will wonder the same thing and likely end up deleting it.

hallipr:
To avoid adding shared code to this PR, I've opened another to bring the repo up to date.
https://github.com/Azure/autorest.csharp/pull/3409

Issue #3384, opened 2023-05-10
Model property not showing up as Uri when typespec specifies url


m-nash:
AnomalyDetector is pointed at [this sha](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/anomalydetector/Azure.AI.AnomalyDetector/tsp-location.yaml#L2) which shows MultivariateBatchDetectionOptions.DataSource as a [url](https://github.com/Azure/azure-rest-api-specs/blob/07be669ac8e83c8766f10e99c59dbd2375614776/specification/cognitiveservices/AnomalyDetector/multivariate/models.tsp#L129) yet in the generated code it comes up as a [string](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/anomalydetector/Azure.AI.AnomalyDetector/src/Generated/MultivariateBatchDetectionOptions.cs#L80).

We need to make sure our TypeFactory is correctly translating `url` in tsp to `Uri` in dotnet.

Issue #3343, opened 2023-05-03
TypeSpec emitter should not be throwing errors/exceptions


timotheeguerin:
From https://github.com/microsoft/typespec/issues/1891 and https://github.com/microsoft/typespec/issues/1894

Errors stop the compilation and prevent diagnostics from showing up. An error showing up will be presented to the user as a bug in the emitter.

**Errors should be reported using the compiler diagnostic API instead**

https://microsoft.github.io/typespec/next/extending-typespec/diagnostics.html

Issue #3340, opened 2023-05-03
support for format uuid in typespec-csharp


cataggar:
Using `tsp compile` with `@azure-tools/typespec-csharp": "^0.2.0-beta.20230501.3"` 

TypeSpec model property:
``` tsp
  @projectedName("json", "SubscriptionId")
  @format("uuid")
  subscriptionId: string;
```

Output openapi3:

``` yaml
        SubscriptionId:
          type: string
          format: uuid
```

Output C#:
`string subscriptionId`

Expected C#:
`Guid subscriptionId`

[MS internal Teams thread](https://teams.microsoft.com/l/message/19:906c1efbbec54dc8949ac736633e6bdf@thread.skype/1683108216859?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1683108216859&teamName=Azure%20SDK&channelName=TypeSpec%20(Cadl)%20Discussion%20%F0%9F%90%AE&createdTime=1683108216859&allowXTenantAccess=false)

PR #3400, opened 2023-05-15
Format support

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3384
Fix https://github.com/Azure/autorest.csharp/issues/3343
Fix https://github.com/Azure/autorest.csharp/issues/3340

- Use typespec diagnostics for typespec compile issues in emitter
- Support string format

# Checklist

# Ready to Land?


src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -193,6 +208,7 @@ export function isNeverType(type: Type): type is NeverType {
 export function getInputType(
     context: SdkContext,
     type: Type,
+    format: string | undefined,
```
ArcturusZhang, line 211: what is the difference between `format?: string` and `format: string | undefined`?
we might need to align with line 109

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -463,11 +505,21 @@ export function getInputType(
...
+                const inputType = getInputType(
+                    context,
+                    value.type,
+                    getFormat(program, value),
+                    models,
+                    enums
+                );
+                // if (format && (inputType as InputPrimitiveType)) {
+                //     applyFormat(inputType as InputPrimitiveType, format);
+                // }
```
ArcturusZhang, line 517: we should remove this

ArcturusZhang (commented):
A refactor suggestion:
I noticed that we are always passing two parameters into some functions: type: Type, format: string | undefined, maybe we could unify these two things and create a new type maybe?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: @ArcturusZhang  If you define format?: string, it is optional parameter, it must be put in the last one. Here It is not good to pass as the last one.

chunyu3:
> 

Thanks @ArcturusZhang  Do you mean we need to define an type alias, alias stringAndUndefined = string | undefined, and use this for every where we need to apply two types for a parameter?
Yes, if the occurrence of the combine of more types is high, we can do this refactor. for low occurrence, maybe we will not define such alias.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: removed

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -119,7 +124,17 @@ function getCSharpInputTypeKindByIntrinsicModelName(
...
-            return InputTypeKind.String;
+            switch (format?.toLowerCase()) {
+                case "date":
+                    return InputTypeKind.DateTime;
+                case "uri":
+                case "url":
+                    return InputTypeKind.Uri;
+                case "uuid":
+                    return InputTypeKind.Guid;
+                default:
```
ArcturusZhang, line 135: can we explicitly handle the case `undefined` and let the `default` branch (not date, uri, url, uuid, or undefined values) to send a warning that this format is unhandled?

ArcturusZhang:
> > 
> 
> Thanks @ArcturusZhang Do you mean we need to define an type alias, alias stringAndUndefined = string | undefined, and use this for every where we need to apply two types for a parameter? Yes, if the occurrence of the combine of more types is high, we can do this refactor. for low occurrence, maybe we will not define such alias.

Well I mean that since we have quite a few methods that has this signature:
```typescript
function doSomething(
 ...,
 type: Type,
 format: string | undefined,
 ...
)
```
and maybe we could have a type or interface like this:
```typescript
interface FormattedType {
  type: Type,
  format: string | undefined
}
```
and those methods could be simplified to
```typescript
function doSomething(
  ...,
  type: FormattedType,
  ...
)
```
and we could have a method to get this thing:
```typescript
function getFormattedType(program: Program, type: Type)
{
return {type: type, format: getFormat(program, type)} as FormattedType;
}
```

I say this because I noticed the pattern when we are calling those method:
```
// on model.ts L387
getInputType(
    context,
    elementType,
    getFormat(program, elementType),
    models,
    enums
),
// on model.ts L401
KeyType: getInputType(
    context,
    key,
    getFormat(program, key),
    models,
    enums
),
ValueType: getInputType(
    context,
    value,
    getFormat(program, value),
    models,
    enums
),
// on model.ts L455
getInputType(
    context,
    dm,
    getFormat(program, dm),
    models,
    enums
);
// and more
```

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -119,7 +124,17 @@ function getCSharpInputTypeKindByIntrinsicModelName(
         case "float64":
             return InputTypeKind.Float64;
         case "string":
-            return InputTypeKind.String;
+            switch (format?.toLowerCase()) {
```
pshao25, line 127: Does format only applies to string?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: We need to return `InputTypeKind.String` if format is `undefined` or other invalid format. Invalid format will be ignore. We may not add a new branch, but give out a warning message.

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: Yes. @format is only applied on string, its string format.

chunyu3:
> ```ts
> FormattedType 
> ```

I like this ideal. I will do this.

src/CADL.Extension/Emitter.Csharp/src/type/inputTypeKind.ts
```diff
@@ -3,31 +3,36 @@
...
     ResourceType = "ResourceType",
     Stream = "Stream",
     String = "String",
     Time = "Time",
     Uri = "Uri",
-    UnKnownKind = "UnknownKind",
-    SByte = "int8",
-    Byte = "uint8",
-    Bytes = "bytes"
+    Enum = "Enum",
```
ArcturusZhang, line 34: what does enum mean here?

ArcturusZhang: oh, I see this exists before this PR, but I am curious what does this mean?

src/CADL.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -366,16 +388,37 @@ export function getInputType(
     function getInputTypeForArray(elementType: Type): InputListType {
         return {
             Name: "Array",
-            ElementType: getInputType(context, elementType, models, enums),
+            ElementType: getInputType(
+                context,
+                {
+                    type: elementType,
+                    format: getFormat(program, elementType)
+                } as FormattedType,
```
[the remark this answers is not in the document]
chunyu3: resloved

ArcturusZhang (approved):

lirenhe (approved):

lirenhe:
The code change in regen PR are expected.  @chunyu3,  please follow up with healthInsight team later to decide the right behaviour of the model change

PR #3412, opened 2023-05-17
Generate protocol only with required params when needed even there is no convenience method

### LINKED ISSUE #3401: Generate protocol only with required params when needed even if there is no convenience method
In order to be able to add the convenience method later we need to generate the protocol method the same way regardless of whether its done at the same time or first by itself.

pshao25:
Fix #3401 

SDK change and customization: https://github.com/Azure/azure-sdk-for-net/pull/36351

Description:
Some protocol methods will have ambiguity with its convenience method. 

Currently we resolve this ambiguity by overload (make every parameter in protocol methods required) when both protocol methods and convenience methods are generated. But if there is no convenience method, we don't change the protocol method.

After this change, even there is currently no convenience method, as long as there will be ambiguity after the convenience method is added one day, we make every parameter in protocol methods required as well. But if there is no ambiguity even after the convenience method is added, we still don't change the protocol method.


src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -106,8 +106,13 @@ private bool ShouldGenerateConvenienceMethod()
...
                 && (!Operation.GenerateProtocolMethod
-                ||_orderedParameters.Where(parameter => parameter.Convenience != KnownParameters.CancellationTokenParameter).Any(parameter => !IsParameterTypeSame(parameter.Convenience, parameter.Protocol))
-                || !_returnType.Convenience.Equals(_returnType.Protocol));
+                || IsConvenienceMethodMeaningful());
+        }
+
+        private bool IsConvenienceMethodMeaningful()
+        {
+            return _orderedParameters.Where(parameter => parameter.Convenience != KnownParameters.CancellationTokenParameter).Any(parameter => !IsParameterTypeSame(parameter.Convenience, parameter.Protocol))
+                || !_returnType.Convenience.Equals(_returnType.Protocol);
```
ArcturusZhang, line 115: Shall we add some comments describing what this check is checking and its reasons?

ArcturusZhang: BTW since the return value of this method never changes because the `_orderedParameters` and `_returnType` never changes inside this instance, please consider change this into a property and add a cache

pshao25: > Shall we add some comments describing what this check is checking and its reasons?

I think it is already self-explained. If we add comment, it would be like "Convenience method is not meaningful if all the corresponding parameters and return types have the same type". It is exactly the same as the code!

> BTW since the return value of this method never changes because the _orderedParameters and _returnType never changes inside this instance, please consider change this into a property and add a cache

Emmm, actually I don't think this reason makes sense. It is not completed stateless. It depends on that _orderedParameters and _returnType both built up. I think it is better to keep as a method.

ArcturusZhang: You described this is checking whether the convenience method is meaningful. But why these checks are checking it is meaningful?
From the implementation it is checking whether the convenience method has the same parameter types (excluding cancellation token) and the same return type. Why is this making it meaningful?

test/TestProjects/Parameters-LowLevel/src/Generated/Docs/ParametersLowlevelClient.xml
```diff
@@ -8,17 +8,7 @@ This sample shows how to call NoRequestBodyResponseBodyAsync with required param
...
 
-Response response = await client.NoRequestBodyResponseBodyAsync(1234, 1234);
-
-JsonElement result = JsonDocument.Parse(response.ContentStream).RootElement;
-Console.WriteLine(result.ToString());
-]]></code>
-This sample shows how to call NoRequestBodyResponseBodyAsync with all parameters, and how to parse the result.
-<code><![CDATA[
-var credential = new AzureKeyCredential("<key>");
-var client = new ParametersLowlevelClient(credential);
```
lirenhe, line 19: Why your change would impact the docs?

pshao25: We will generate two sample calls for each method. One only has required parameters and the other has required parameters and optional parameters. Since we change all the parameters to required, then only one sample call left.

test/AutoRest.TestServerLowLevel.Tests/body-array.cs
```diff
@@ -113,7 +113,7 @@ public class BodyArray : TestServerLowLevelTestBase
         [Test]
         public Task GetArrayBooleanWithNullExample() => Test(async (host) =>
         {
-            Response response = await new ArrayClient(Key, host, null).GetBooleanInvalidNullAsync();
+            Response response = await new ArrayClient(Key, host, null).GetBooleanInvalidNullAsync(new());
```
chunyu3, line 116: It is better to specify the type, `new RequestContext()` in our test.

chunyu3 (approved):

test/AutoRest.TestServerLowLevel.Tests/body-array.cs
pshao25: The fact that `new()` might lead to ambiguity problem in the future doesn't mean we should change it to `new RequestContext()`. The reason is the same as why we accept `new()` to introduce ambiguity, i.e., we don't discourage user to  use `new()` just because that might lead to ambiguity when they introduce their own overloads.

lirenhe (approved):

PR #3416, opened 2023-05-17
Add 2 new parameters to the code gen target

m-nash:
To support this issue https://github.com/Azure/azure-sdk-tools/issues/6187 we need to add 2 new parameters to the code gen target.  Both flags are optional and default to false.

Depends on this getting merged first https://github.com/Azure/azure-sdk-tools/pull/6197

- /p:SaveInputs=true  :  This will replace the old `cleanup: false` that we used to put in the tsp-location.yaml and will not only keep the `TempTypeSpecFiles` directory around but also forward the flag to the emitter so that the cadl.json and configuration.json are kept also.
- /p:SkipSync=true  :  This flag allows us to continue to run the `dotnet build` command to regen after making local changes to the `TempTypeSpecFiles` directory without replacing our local changes and without needing to copy and paste the `tsp compile` command from the command line


JoshLove-msft (approved):

src/AutoRest.CSharp/build/CodeGeneration.targets
```diff
@@ -32,6 +32,8 @@
     <UseDefaultNamespaceAndOutputFolder Condition="'$(UseDefaultNamespaceAndOutputFolder)' == ''">true</UseDefaultNamespaceAndOutputFolder>
     <_AutoRestCommand>node $(AutoRestEntryPoint) --max-memory-size=8192 --skip-csproj --skip-upgrade-check --version=$(AutoRestCoreVersion) $(AutoRestInput) $(AutoRestAdditionalParameters) --use=$(MSBuildThisFileDirectory)../tools/net6.0/any/ --clear-output-folder=true --shared-source-folders=&quot;$(AzureCoreSharedCodeDirectory);$(AutoRestSharedCodeDirectory)&quot;</_AutoRestCommand>
     <_AutoRestCommand Condition="'$(UseDefaultNamespaceAndOutputFolder)' == 'true'">$(_AutoRestCommand) --output-folder=$(MSBuildProjectDirectory)/Generated --namespace=$(RootNamespace)</_AutoRestCommand>
+    <_SaveInputs Condition="'$(SaveInputs)' != ''">-SaveInputs</_SaveInputs>
```
chunyu3, line 35: $SaveInputs and $SkipSync are boolean. Maybe we need to check in this way '$(_GenerateCode)' == 'true'

chunyu3, line 35:
```suggestion
    <_SaveInputs Condition="'$(SaveInputs)' == 'true'">-SaveInputs</_SaveInputs>
```

src/AutoRest.CSharp/build/CodeGeneration.targets
```diff
@@ -52,8 +54,9 @@
   </Target>
 
   <Target Name="_GenerateCodeForTypeSpec" Condition="'$(_GenerateCode)' == 'true' AND '$(TypeSpecInput)' != ''" >
-    <Exec Command="$(PowerShellExe) -NoProfile -NonInteractive -executionpolicy Unrestricted -File $(TypeSpecProjectSyncScriptPath) $(MSBuildProjectDirectory)/.."/>
-    <Exec Command="$(PowerShellExe) -NoProfile -NonInteractive -executionpolicy Unrestricted -File $(TypeSpecProjectGenerateScriptPath) $(MSBuildProjectDirectory)/.. $(typespecAdditionalOptions)"/>
+    <Error Text="You used skipped sync but didn't have the TempTypeSpecFiles in your project directory.  Please run dotnet 'build /t:GenerateCode /p:SaveInputs=true' without SkipSync first at least once" Condition="'$(SkipSync)' != '' AND !Exists('$(MSBuildProjectDirectory)/../TempTypeSpecFiles')" />
```
chunyu3, line 57:
```suggestion
    <Error Text="You used skipped sync but didn't have the TempTypeSpecFiles in your project directory.  Please run dotnet 'build /t:GenerateCode /p:SaveInputs=true' without SkipSync first at least once" Condition="'$(SkipSync)' == 'true' AND !Exists('$(MSBuildProjectDirectory)/../TempTypeSpecFiles')" />
```

src/AutoRest.CSharp/build/CodeGeneration.targets
```diff
@@ -52,8 +54,9 @@
...
+    <Exec Condition="'$(SkipSync)' == ''" Command="$(PowerShellExe) -NoProfile -NonInteractive -executionpolicy Unrestricted -File $(TypeSpecProjectSyncScriptPath) $(MSBuildProjectDirectory)/.."/>
```
chunyu3, line 58:
```suggestion
    <Exec Condition="'$(SkipSync)' != 'true'" Command="$(PowerShellExe) -NoProfile -NonInteractive -executionpolicy Unrestricted -File $(TypeSpecProjectSyncScriptPath) $(MSBuildProjectDirectory)/.."/>
```

src/AutoRest.CSharp/build/CodeGeneration.targets
jsquire: Not sure the intent, but I agree with with Crystal that we should consider tightening these up.   If you're trying to catch undefined as well as true, maybe something like:

```xml
<_SaveInputs Condition="'$(SaveInputs)' != 'false'">-SaveInputs</_SaveInputs>
```

Same thought for the other places that Crystal commented.

jsquire (approved):

m-nash:
Waiting for this https://github.com/Azure/azure-sdk-tools/pull/6197 to merge then I will run regen preview to verify no issues.

Issue #3418, opened 2023-05-18
Serialization in DPG models for value type has redundant null check and will write null value


ArcturusZhang:
**Describe the issue or request**

If we have a optional value type property, for instance
```
model Foo
{
   optionalInt?: int32;
}
```
This will generate into
```csharp
public class Foo
{
    public int? OptionalInt {get; set;}
}
```
But the corresponding serialization is not correct.

Current:
```csharp
public class Foo : IUtf8JsonSerializable
{
    void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
    {
        writer.WriteStartObject();
        if (Optional.IsDefined(OptionalInt))
        {
            if (OptionalInt != null)
            {
                writer.WritePropertyName("optionalInt"u8);
                writer.WriteNumberValue(OptionalInt.Value);
            }
            else
            {
                writer.WriteNull("optionalInt");
            }
        }
    }
}
```
it has an extra null check, and it will serialize explicit `null` when the value of this property is not set.
`Optional.IsDefined(OptionalInt)` is the same as `OptionalInt.HasValue` ([code](https://github.com/Azure/autorest.csharp/blob/3e98cb0ea2dbbb6b61c1c5abbf703923acc3747a/src/assets/Generator.Shared/Optional.cs#L29)) which is also the same as `OptionalInt != null`. Therefore the above code is checking the same condition twice.

On the other hand, in mgmt/HLC, and optional int type property has the following serialization code:
```csharp
public class Foo : IUtf8JsonSerializable
{
    void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
    {
        writer.WriteStartObject();
        if (Optional.IsDefined(OptionalInt))
        {
            writer.WritePropertyName("optionalInt"u8);
            writer.WriteNumberValue(OptionalInt.Value);
        }
    }
}
```

There are two differences:
1. No extra null check.
2. <del>Mgmt/HLC will not write `null` into the payload when the value is null.</del> this statement is incorrect, because the code in `else` is actually unreachable since we checked the same condition twice. It works well, but it does not align and we should align

We should align, and I believe the way currently in DPG is incorrect.

ArcturusZhang:
Here is an example in our test project: https://github.com/Azure/autorest.csharp/blob/4b70017e8a5f101ea41bee2509cfd2c23f5529d2/test/TestProjects/Models-Typespec/src/Generated/Models/RoundTripOptionalModel.Serialization.cs#L28

PR #3425, opened 2023-05-23
fix the serialization duplicated check in DPG

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3418

# Description

Now our serialization code in DPG is doing something like this when the type of this property is a nullable value type:
```csharp
        if (Optional.IsDefined(OptionalInt))
        {
            if (OptionalInt != null)
            {
                writer.WritePropertyName("optionalInt"u8);
                writer.WriteNumberValue(OptionalInt.Value);
            }
            else
            {
                writer.WriteNull("optionalInt");
            }
        }
```

The inner check is redundant, and the else branch is unreachable. Because the implementation of Optional.IsDefined for a nullable struct is just
```csharp
        public static bool IsDefined<T>(T? value) where T: struct
        {
            return value.HasValue;
        }
```
`OptionalInt.HasValue` and `OptionalInt != null` are identical, therefore the above code is actually:
```csharp
        if (OptionalInt != null)
        {
            if (OptionalInt != null)
            {
                writer.WritePropertyName("optionalInt"u8);
                writer.WriteNumberValue(OptionalInt.Value);
            }
            else
            {
                writer.WriteNull("optionalInt");
            }
        }
```
And we could simplify it to:
```csharp
        if (Optional.IsDefined(OptionalInt))
        {
                writer.WritePropertyName("optionalInt"u8);
                writer.WriteNumberValue(OptionalInt.Value);
        }
```
This is all the changes we have in the generated code.

The reason why we have this issue is that in mgmt/HLC's model generation, we are distinguish `property.Declaration.Type` and `property.ValueType`. The declaration type is the actual type generated in the API, and the value type is the type of its value.
For an optional int, its declaration type should be `int?` and its value type is `int`.

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs
+++ b/src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs
@@ -8,15 +8,39 @@ using AutoRest.CSharp.Output.Models.Serialization;
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal record FieldDeclaration(FormattableString? Description, FieldModifiers Modifiers, CSharpType Type, CodeWriterDeclaration Declaration, FormattableString? DefaultValue, bool IsRequired, SerializationFormat SerializationFormat, bool IsField = false, bool WriteAsProperty = false, FieldModifiers? GetterModifiers = null, FieldModifiers? SetterModifiers = null)
+    internal record FieldDeclaration(FormattableString? Description, FieldModifiers Modifiers, CSharpType Type, CSharpType ValueType, CodeWriterDeclaration Declaration, FormattableString? DefaultValue, bool IsRequired, SerializationFormat SerializationFormat, bool IsField = false, bool WriteAsProperty = false, FieldModifiers? GetterModifiers = null, FieldModifiers? SetterModifiers = null)
     {
         public string Name => Declaration.ActualName;
         public string Accessibility => (Modifiers & FieldModifiers.Public) > 0 ? "public" : "internal";
 
-        public FieldDeclaration(FieldModifiers modifiers, CSharpType type, string name, bool writeAsProperty = false) : this(null, modifiers, type, name, SerializationFormat.Default, writeAsProperty: writeAsProperty) { }
-        public FieldDeclaration(FieldModifiers modifiers, CSharpType type, string name, FormattableString? defaultValue, SerializationFormat serializationFormat, bool writeAsProperty = false) : this(null, modifiers, type, name, defaultValue, false, serializationFormat, writeAsProperty: writeAsProperty) { }
-        public FieldDeclaration(FormattableString? description, FieldModifiers modifiers, CSharpType type, string name, SerializationFormat serializationFormat, bool writeAsProperty = false) : this(description, modifiers, type, new CodeWriterDeclaration(name), null, false, serializationFormat, false, writeAsProperty) { }
-        public FieldDeclaration(FormattableString? description, FieldModifiers modifiers, CSharpType type, string name, FormattableString? defaultValue, bool isRequired, SerializationFormat serializationFormat, bool isField = false, bool writeAsProperty = false) : this(description, modifiers, type, new CodeWriterDeclaration(name), defaultValue, isRequired, serializationFormat, isField, writeAsProperty) { }
+        public FieldDeclaration(FieldModifiers modifiers, CSharpType type, string name, bool writeAsProperty = false)
+            : this(null, modifiers, type, name, SerializationFormat.Default, writeAsProperty: writeAsProperty) { }
+        public FieldDeclaration(FieldModifiers modifiers, CSharpType type, string name, FormattableString? defaultValue, SerializationFormat serializationFormat, bool writeAsProperty = false)
+            : this(Description: null,
+                  Modifiers: modifiers,
+                  Type: type,
+                  ValueType: type,
+                  Declaration: new CodeWriterDeclaration(name),
+                  DefaultValue: defaultValue,
+                  IsRequired: false,
+                  SerializationFormat: serializationFormat,
+                  IsField: false,
+                  WriteAsProperty: writeAsProperty,
+                  GetterModifiers: null,
+                  SetterModifiers: null)
+        { }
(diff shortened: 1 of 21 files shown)


test/CadlRanchProjects/type/property/optional/src/Generated/Models/DatetimeProperty.Serialization.cs
```diff
@@ -38,14 +31,13 @@ internal static DatetimeProperty DeserializeDatetimeProperty(JsonElement element
...
             }
-            Optional<DateTimeOffset?> property = default;
+            Optional<DateTimeOffset> property = default;
             foreach (var property0 in element.EnumerateObject())
             {
                 if (property0.NameEquals("property"u8))
                 {
                     if (property0.Value.ValueKind == JsonValueKind.Null || property0.Value.ValueKind == JsonValueKind.String && property0.Value.GetString().Length == 0)
                     {
-                        property = null;
```
chunyu3, line 48: After remove implicitly assigning property = null,  The property value will be `DateTimeOffset.MinValue`, not null, is it expected?

src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs
```diff
@@ -8,15 +8,39 @@
...
-        public FieldDeclaration(FormattableString? description, FieldModifiers modifiers, CSharpType type, string name, SerializationFormat serializationFormat, bool writeAsProperty = false) : this(description, modifiers, type, new CodeWriterDeclaration(name), null, false, serializationFormat, false, writeAsProperty) { }
-        public FieldDeclaration(FormattableString? description, FieldModifiers modifiers, CSharpType type, string name, FormattableString? defaultValue, bool isRequired, SerializationFormat serializationFormat, bool isField = false, bool writeAsProperty = false) : this(description, modifiers, type, new CodeWriterDeclaration(name), defaultValue, isRequired, serializationFormat, isField, writeAsProperty) { }
+        public FieldDeclaration(FieldModifiers modifiers, CSharpType type, string name, bool writeAsProperty = false)
+            : this(null, modifiers, type, name, SerializationFormat.Default, writeAsProperty: writeAsProperty) { }
```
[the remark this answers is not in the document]
ArcturusZhang: I missed this one. I will change accordingly.

test/CadlRanchProjects/type/property/optional/src/Generated/Models/DatetimeProperty.Serialization.cs
ArcturusZhang: No this is expected. The value of this property is still null.
Let us sync offline and I will show you how this thing works.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -155,25 +156,26 @@ private static FieldDeclaration CreateFieldFromExisting(ISymbol existingMember,
             CodeWriterDeclaration declaration = new CodeWriterDeclaration(existingMember.Name);
             declaration.SetActualName(existingMember.Name);
 
-            return new FieldDeclaration($"Must be removed by post-generation processing,", fieldModifiers, fieldType, declaration, GetPropertyDefaultValue(originalType, inputModelProperty), inputModelProperty.IsRequired, inputModelProperty.SerializationFormat, existingMember is IFieldSymbol, writeAsProperty);
+            return new FieldDeclaration($"Must be removed by post-generation processing,", fieldModifiers, fieldType, fieldType, declaration, GetPropertyDefaultValue(originalType, inputModelProperty), inputModelProperty.IsRequired, inputModelProperty.SerializationFormat, existingMember is IFieldSymbol, writeAsProperty);
```
archerzz, line 159: A typo?
```suggestion
            return new FieldDeclaration($"Must be removed by post-generation processing,", fieldModifiers, fieldType, fieldValueType, declaration, GetPropertyDefaultValue(originalType, inputModelProperty), inputModelProperty.IsRequired, inputModelProperty.SerializationFormat, existingMember is IFieldSymbol, writeAsProperty);
```

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -134,12 +157,12 @@ private IEnumerable<JsonPropertySerialization> CreatePropertySerializations()
...
                 foreach (var property in objType.Properties)
                 {
-                    if (property.InputModelProperty is null)
+                    if (property.InputModelProperty is not { } inputModelProperty)
                         continue;
 
                     var declaredName = property.Declaration.Name;
                     var serializedName = property.InputModelProperty.SerializedName ?? property.InputModelProperty.Name;
-                    var optionalViaNullability = !property.IsRequired && !property.ValueType.IsNullable && !TypeFactory.IsCollectionType(property.ValueType);
+                    var optionalViaNullability = !inputModelProperty.IsRequired && !inputModelProperty.Type.IsNullable && !TypeFactory.IsCollectionType(property.ValueType);
```
archerzz, line 165: What's the difference between `property.IsRequired` and `property.InputModelProperty.IsRequired`?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
ArcturusZhang: This is creating from an existing member in customization code therefore I believe this does not matter because we cannot get the "value type" from the existing member, therefore I just put them the same value.
This also leaves the code unchanged when there is an existing type.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
ArcturusZhang: They are the same. I changed it because of alignment (mgmt is doing this in this way)

archerzz (approved):
approved with suggested comments being added later

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
ArcturusZhang: Well I added a comment here describing why we are doing this right now and how to improve in the future

lirenhe (approved):

Issue #3426, opened 2023-05-23
typespec-csharp discriminator code does not compile


cataggar:
Similar to #3421, this is a problem with discriminator. The C# code is generated when `kind` is not defined as a field in `Pet`. However, it does not compile.

```
import "@typespec/http";
import "@typespec/versioning";

using TypeSpec.Http;
using TypeSpec.Versioning;

@service({
  title: "Pet API",
})
@versioned(ServiceApiVersions)
namespace Pet;

enum ServiceApiVersions {
  v2023_05: "2023-05",
}

@discriminator("kind")
model Pet {
  // kind: string;
}

model Cat extends Pet {
  kind: "cat";
  meow: boolean;
}

model Dog extends Pet {
  kind: "dog";
  bark: boolean;
}

@route("/pet")
@get
op pet(): Pet;
```

```
PS C:\Users\cataggar\ms\avs\src\conveyor\ConsumptionWorker\client> dotnet build
MSBuild version 17.6.1+8ffc3fe3d for .NET
  Determining projects to restore...
  All projects are up-to-date for restore.
C:\Users\cataggar\ms\avs\src\conveyor\ConsumptionWorker\client\Generated\Cat.cs(17,13): error CS0200: Property or index
er 'Pet.Kind' cannot be assigned to -- it is read only [C:\Users\cataggar\ms\avs\src\conveyor\ConsumptionWorker\client\
ConsumptionWorkerApi.csproj::TargetFramework=netstandard2.0]
C:\Users\cataggar\ms\avs\src\conveyor\ConsumptionWorker\client\Generated\Dog.cs(17,13): error CS0200: Property or index
er 'Pet.Kind' cannot be assigned to -- it is read only [C:\Users\cataggar\ms\avs\src\conveyor\ConsumptionWorker\client\
ConsumptionWorkerApi.csproj::TargetFramework=netstandard2.0]
C:\Users\cataggar\ms\avs\src\conveyor\ConsumptionWorker\client\Generated\Cat.cs(17,13): error CS0200: Property or index
er 'Pet.Kind' cannot be assigned to -- it is read only [C:\Users\cataggar\ms\avs\src\conveyor\ConsumptionWorker\client\
ConsumptionWorkerApi.csproj::TargetFramework=net6.0]
C:\Users\cataggar\ms\avs\src\conveyor\ConsumptionWorker\client\Generated\Dog.cs(17,13): error CS0200: Property or index
[...]

PR #3430, opened 2023-05-24
Fix output discriminator issue

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3426

# Description

Setter of a property is omitted in some cases, but we should never omit the setter for a discriminator property because it causes compilation errors.

# Checklist

# Ready to Land?


chunyu3 (approved):

test/TestProjects/Models-Typespec/src/Generated/Models/OutputBaseModelWithDiscriminator.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary> Initializes a new instance of OutputBaseModelWithDiscriminator. </summary>
+        /// <param name="kind"> Discriminator. </param>
+        internal OutputBaseModelWithDiscriminator(string kind)
+        {
+            Kind = kind;
+        }
+
+        /// <summary> Discriminator. </summary>
+        internal string Kind { get; set; }
```
lirenhe, line 30: From the issue you linked, seems we should generate the code as below.  Even this is for internal usage, do we allow change that value later? 

internal string Kind { get; init; }

cataggar: Pretty sure it should be `init` and not `set`.

ArcturusZhang: @cataggar 
If we do that, we will also change every discriminator in mgmt plane as well
And the generator will write `public string Discriminator { get; protected internal set; }` when the discriminator is changed to public by the config
Therefore the setter will always be internal which means the SDK itself controls when to set the discriminator value and I believe it is safe.
I kind of do not want to make a massive change in our generator, and `init` might be depending a specific langver, we just got [an issue about langver](https://github.com/Azure/autorest.csharp/issues/3433), therefore I prefer we keep using `set` here instead of `init`.

cataggar: `init` is [C# 9 from .NET 5](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/init). That issue is about using [UTF-8 string literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/utf8-string-literals) from C# 11 from .NET 7 and not being able to compile for .NET 6. `init` should be safe to use.

ArcturusZhang: I would buy in that `init` makes more sense and it does not have much risk to change to that, but in the meantime, I will not do it in this PR, because `set` works as well.
We are on the way to eliminate the divergence in our code generator between mgmt, HLC and DPG, therefore we should not introduce another divergence into it. If we want to change it, we should change it for all models in our SDKs.
I will open an issue for this so that we could do it later.

ArcturusZhang: Issue created: https://github.com/Azure/autorest.csharp/issues/3452

lirenhe:
I am good to have a separate issue to track the init/set for discriminator and let's first unblock the major issue.
So I approved this change.

lirenhe (approved):

Issue #3406, opened 2023-05-16
The internal constructor is not properly created in mgmt/HLC when the model does not include deserialization


ArcturusZhang:
The way we build the constructors in mgmt/HLC and DPG is different.

Currently in mgmt/HLC, the constructor is not properly initialized when the model does not need the deserialization.
It works well - but since the model factory needs the internal constructor regardless the model has deserialization or not.

Issue #3353, opened 2023-05-04
TypeSpec: .NET JSON serialization/deserialization for enums -- add generation of a custom converter?


trrwilson:
When defining an enum type in TypeSpec, e.g. 

```
enum Foo {
  first,
  second
}
```

The C# emitter generates a nice "enum-like class" that looks something like (partial):

```
public partial readonly struct Foo : IEquatable<Foo>
{
    private readonly string _value;

    public Foo(string value)
    {
        _value = value ?? throw new ArgumentNullException(nameof(value));
    }
    ...
    public static Foo First { get; } = new Foo(...);
    public static Foo Second { get; } = new Foo(...);
    ...
    public override string ToString() => _value;
}
```

Older versions of .NET don't automatically support `System.Json.Text.Serialization` capabilities for non-property / private fields, so calling `System.Json.Text.Serialization.JsonSerializer.Serialize(new Foo("hello"))` will *not* serialize the `value` string as intended. And subsequent deserialization will of course not be able to reflect the original object.

I think it's fairly easy to address this, fortunately. If the emitter can first create a customer converter (`FooConverter.cs`):

```
internal class FooConverter : JsonConverter<Foo>
{
    public override Foo Read(ref Utf8JsonReader reader, Type _, JsonSerializerOptions __) => new Foo(reader.GetString());
    public override void Write(Utf8JsonWriter writer, Foo value, JsonSerializerOptions _) => writer.WriteStringValue(value.ToString());
}
```

...and then the generated enum-like struct/class is attributed:

```
[JsonConverter(typeof(FooConverter))]
public partial readonly struct Foo: IEquatable<Foo>
{
...
```

...then I think everything ends up happy. This is possible now for emitter consumers via doing the above in custom code, but it'd be nice if were auto-generated.

trrwilson:
Cross-link to progenitor report: https://github.com/Azure/azure-sdk-for-net/issues/35963

ArcturusZhang:
FYI the public serialization is an ongoing feature developing in Azure.Core, here is a PR for it: https://github.com/Azure/azure-sdk-for-net/pull/35946
Once this is finalized, we will update our SDK to make sure every model in the SDK will implement the new `Azure.IJsonSerializable` which allows you to write a generic converter (we may or may not provide a factory method to do that - this is not discussed yet)
But nevertheless, I believe this public serialization feature will solve this issue and currently we should not do anything until the public serialization is ready.

[...]

Issue #2351, opened 2022-06-21
Make serialization of DPG-generated models available in the public API


annelo-msft:
Currently, Azure SDK models are written to/read from wire formats internally in service methods on a client.

This tracks the work to design and implement serialization as part of the public API.  Before starting this work, we will need to formalize requirements for this.

[Description TBD]

ArthurMa1978:
Customer report similar ask here https://github.com/Azure/azure-sdk-for-net/issues/31095

PR #3434, opened 2023-05-24
Update serialization code to use new signatures

m-nash:
Fixes https://github.com/Azure/autorest.csharp/issues/3406
Fixes https://github.com/Azure/autorest.csharp/issues/3353
Fixes https://github.com/Azure/autorest.csharp/issues/2351
Fixes https://github.com/Azure/azure-sdk-for-net/issues/16977

# Description
This change is a draft to support this issue https://github.com/Azure/autorest.csharp/issues/3427.  We want to generate the models used for testing to avoid any copy paste issues in the serialization code.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
```diff
@@ -176,62 +189,278 @@ private static void WriteXmlDeserialize(CodeWriter writer, TypeDeclarationOption
...
+            writer.Line();
+        }
 
-                if (discriminator is not null && discriminator.HasDescendants)
+        private static void WriteInternalDeserialize(CodeWriter writer, TypeDeclarationOptions declaration, JsonObjectSerialization serialization, CSharpType modelSerializableType, bool isUnknownDerivedType, string deserializeMethod)
+        {
+            FormattableString signatureLine = $"internal static {modelSerializableType.Arguments[0]} {deserializeMethod}({typeof(JsonElement)} element, {typeof(ModelSerializerOptions)} options = default)";
+            if (isUnknownDerivedType)
+            {
+                writer.Line($"{signatureLine} => Deserialize{declaration.Name.Remove(0, 7)}(element, options);");
```
ArcturusZhang, line 279: The expression `declaration.Name.Remove(0, 7)` is fragile because it depends on the hard-coded word `Unknown`.
If we would like to remove the word unknown from its name, I assume we want to call the deserialization method on its base class, therefore could we do this:
```suggestion
                writer.Line($"{signatureLine} => Deserialize{serialization.ObjectType!.BaseType.Declaration.Name}(element, options);");
```
I believe it might be better if every `ObjectType` could have a property to store the name of its deserialization method. If we have that, we could have:
```suggestion
                writer.Line($"{signatureLine} => {serialization.ObjectType!.BaseType.DeserializationMethodName}(element, options);");
```

src/assets/Generator.Shared/IXmlSerializable.cs
```diff
@@ -4,6 +4,7 @@
 #nullable enable
 
 using System.Xml;
+using Azure.Core.Serialization;
```
ArcturusZhang, line 7: why we need to add this when the code does not change anything?

src/assets/Generator.Shared/PageableHelpers.cs
```diff
@@ -19,22 +20,24 @@ internal static class PageableHelpers
         private static readonly byte[] DefaultItemPropertyName = Encoding.UTF8.GetBytes("value");
         private static readonly byte[] DefaultNextLinkPropertyName = Encoding.UTF8.GetBytes("nextLink");
 
+        public delegate T ValueFactory<T>(JsonElement element, ModelSerializerOptions? options = default);
```
ArcturusZhang, line 23: well this is good for readability, if we could do the same for other delegates in the Create(Async)Pageable helper methods would also be great.
But I assume we could do that later since everything is good to work

src/assets/Generator.Shared/XmlWriterExtensions.cs
```diff
@@ -5,6 +5,7 @@
 
 using System;
 using System.Xml;
+using Azure.Core.Serialization;
```
ArcturusZhang, line 8: Maybe this is not necessary?

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -33,6 +33,29 @@ protected ObjectType(string defaultNamespace, SourceInputModel? sourceInputModel
...
         }
 
+        private bool? isUnknownDerivedType;
+        public bool IsUnknownDerivedType => isUnknownDerivedType ??= EnsureUnknownDerivedType();
+        public bool EnsureUnknownDerivedType()
+        {
+            if (!Type.Equals(Discriminator?.DefaultObjectType?.Type))
+                return false;
+
+            if (!Declaration.Name.StartsWith("Unknown", StringComparison.Ordinal))
```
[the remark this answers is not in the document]
m-nash: I think the challenge is in the multiple layer tree.  For example the `body-complex` project in TestServerProjects has a salmon which is a base type of SmartSalmon but the unknown type for the tree is `UnknownFish`.  Discriminator.DefaultObjectType in this case comes out to be Salmon, not UnknownFish which means that it will think this is the unknown type even though it is not.  I believe this check is fine as long as the first check above it is there as well.  We should never have a class actually named the same.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeConstructor.cs
```diff
@@ -11,27 +13,79 @@ namespace AutoRest.CSharp.Output.Models.Types
...
                      Initializer: new(isBase: true, baseConstructor?.Signature.Parameters ?? Array.Empty<Parameter>())),
                  initializers,
-                 baseConstructor)
+                 type,
+                 baseConstructor,
+                 createInternal)
         {
         }
 
+        public ObjectTypeConstructor? InternalCtor { get; }
```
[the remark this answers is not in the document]
m-nash: We don't want to bleed the new `RawData` onto the serialization signatures as this is used to do many things including ModelFactory etc.  If we added it to the signatures directly we would need to change many places that consume this property to then ignore it sometimes.  I went this way for now as it was far less code to change.  If you feel we should address this I can open a follow up issue to track but I prefer to not do that as part of this change here.

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
```diff
@@ -42,13 +43,17 @@ public void WriteSerialization(CodeWriter writer, TypeProvider schema)
...
-            => WriteObjectSerialization(writer, model, model.Declaration, model.JsonSerialization, model.XmlSerialization, model.IsStruct, model.IncludeSerializer, model.IncludeDeserializer);
+            => WriteObjectSerialization(writer, model, model.Declaration, model.JsonSerialization, model.XmlSerialization);
 
-        private void WriteObjectSerialization(CodeWriter writer, SerializableObjectType model, TypeDeclarationOptions declaration, JsonObjectSerialization? jsonSerialization, XmlObjectSerialization? xmlSerialization, bool isStruct, bool includeSerializer, bool includeDeserializer)
+        private void WriteObjectSerialization(CodeWriter writer, SerializableObjectType model, TypeDeclarationOptions declaration, JsonObjectSerialization? jsonSerialization, XmlObjectSerialization? xmlSerialization)
         {
             var hasJson = jsonSerialization != null;
             var hasXml = xmlSerialization != null;
 
+            var typeOfT = model.IsUnknownDerivedType ? model.EnumerateHierarchy().Skip(1).First().Type : model.Type;
```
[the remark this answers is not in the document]
m-nash: SerializableObjectType doesn't have a `BaseType` property but we could do this
```c#
            var typeOfT = model.IsUnknownDerivedType ? model.Inherits!.Implementation!.Type : model.Type;
```

src/assets/Generator.Shared/PageableHelpers.cs
m-nash: Do you mind opening a tracking item for this in azure-sdk-for-net since this file has moved with all other shared source?

https://github.com/Azure/azure-sdk-for-net/pull/38399

eng/DownloadSharedSource.ps1
```diff
@@ -19,8 +19,8 @@ Write-Host "git sparse-checkout init"
 git sparse-checkout init
 Write-Host "git sparse-checkout set --no-cone $azCoreSharedPath $armCoreSharedPath"
 git sparse-checkout set --no-cone $azCoreSharedPath $armCoreSharedPath
-Write-Host "git checkout"
-git checkout
+Write-Host "git checkout mnash-updatePageableHelpers"
```
pshao25, line 22: By accident?

samples/AppConfiguration/Generated/AppConfigurationModelFactory.cs
```diff
@@ -10,20 +10,20 @@ namespace AppConfiguration.Models
...
     public static partial class AppConfigurationModelFactory
     {
-        /// <summary> Initializes a new instance of Key. </summary>
+        /// <summary> Initializes a new instance of <see cref="Key"/>. </summary>
         /// <param name="name"></param>
         /// <returns> A new <see cref="Models.Key"/> instance for mocking. </returns>
         public static Key Key(string name = null)
         {
-            return new Key(name);
+            return new Key(name, default);
```
pshao25, line 18: Just curious, isn't it a problem when service add a new property to this model and the constructor becomes like `internal Key(string name, string additionalProperty)`?

samples/AppConfiguration/Generated/Models/Key.Serialization.cs
```diff
@@ -5,29 +5,107 @@
...
+            if (_rawData is not null && options.Format == ModelSerializerFormat.Json)
+            {
+                foreach (var property in _rawData)
+                {
+                    writer.WritePropertyName(property.Key);
+#if NET6_0_OR_GREATER
+				writer.WriteRawValue(property.Value);
+#else
+                    JsonSerializer.Serialize(writer, JsonDocument.Parse(property.Value.ToString()).RootElement);
+#endif
```
pshao25, line 35: Should we align these two lines?

samples/AppConfiguration/Generated/Models/Key.Serialization.cs
```diff
@@ -5,29 +5,107 @@
...
+                }
+            }
+            writer.WriteEndObject();
+        }
+
+        internal static Key DeserializeKey(JsonElement element, ModelSerializerOptions options = default)
         {
+            options ??= ModelSerializerOptions.DefaultWireOptions;
```
pshao25, line 43: Did I miss anything? Because I didn't see any places that you are handling the `DefaultWireOptions` option.

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
```diff
@@ -641,6 +659,10 @@ public static void WriteObjectInitialization(this CodeWriter writer, JsonObjectS
             {
                 writer.Line($"{objAdditionalProperties.Type} {dictionaryVariable:D} = new {objAdditionalProperties.Type}();");
             }
+            else if (serialization.ObjectType is not null && serialization.ObjectType.HasRawDataInHeirarchy)
+            {
+                writer.Line($"{Parameter.RawData.Type} rawData = new {Parameter.RawData.Type}();");
```
pshao25, line 664: Shouldn't the `rawData` be a variable? What if there is a property called `rawData`?

samples/AppConfiguration/Generated/Models/LabelListResult.Serialization.cs
```diff
@@ -5,22 +5,65 @@
...
+        {
+            options ??= ModelSerializerOptions.DefaultWireOptions;
+
             if (element.ValueKind == JsonValueKind.Null)
             {
                 return null;
             }
             Optional<IReadOnlyList<Label>> items = default;
             Optional<string> nextLink = default;
+            Dictionary<string, BinaryData> rawData = new Dictionary<string, BinaryData>();
```
pshao25, line 66: Should we pass options to below `Label.DeserializeLabel(item)` at line 78 as well?

eng/DownloadSharedSource.ps1
m-nash: No this was on purpose because I needed a non merged version but I have updated it back to main since that was merged into main here https://github.com/Azure/azure-sdk-for-net/pull/38399

samples/AppConfiguration/Generated/Models/Key.Serialization.cs
m-nash: For some reason roslyn isn't able to get this right inside #if statements.  We can open a tracking issue for this if we want to try to solve it later but its a long standing formatting issue.

src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs
m-nash: Not sure I follow the question it is a `field` not a `property`

pshao25: You could try 
```
model Test {
  rawData: string;
}
```
and then the deserialize function will have two `rawData`
```C#
string rawData = default;
Dictionary<string, BinaryData> rawData = new Dictionary<string, BinaryData>();
```

m-nash: Gotcha, perhaps we call it `__rawData` and then name the stack variable `_rawDAta` so that the chance of conflict is much lower, nearly zero.

pshao25: Why not we just do something like:
```C#
var rawDataVariable = new CodeWriterDeclaration("rawData");
writer.Line($"{Parameter.RawData.Type} {rawDataVariable:D} = new {Parameter.RawData.Type}();");
```

Issue #3429, opened 2023-05-24
[Doc] Add document for the new serialization related attributes


ArcturusZhang:
We should add some documents describing the usage of the new attributes introduced by https://github.com/Azure/autorest.csharp/pull/3278

PR #3441, opened 2023-05-26
Enable CodeGenMemberSerialization related attributes on mgmt/HLC (SchemaObjectType) and update readme for the new attributes

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3429
Fixes https://github.com/Azure/autorest.csharp/issues/3277

# Description

# Checklist

# Ready to Land?


ArcturusZhang:
Hi @m-nash could you please review this PR?

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -257,26 +257,28 @@ public XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema objectSch
 
         private IEnumerable<JsonPropertySerialization> GetPropertySerializationsFromBag(PropertyBag propertyBag, SchemaObjectType objectType)
```
live1206, line 258: Not related to this PR, but here we refer to a private type `PropertyBag` within this class and there is another public type [`PropertyBag`](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Output/Models/Shared/PropertyBag.cs). It is ambiguous with the same name, consider to rename the private one here.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -296,7 +299,7 @@ public JsonObjectSerialization BuildJsonObjectSerialization(ObjectSchema objectS
                     var schemaProperty = objectTypeProperty.SchemaProperty;
```
live1206, line 299: This line can be removed, `schemaProperty` is not in use anymore.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeProperty.cs
```diff
@@ -32,8 +32,8 @@ public ObjectTypeProperty(FieldDeclaration field, InputModelProperty inputModelP
             InitializationValue = field.DefaultValue;
         }
 
-        public ObjectTypeProperty(MemberDeclarationOptions declaration, string parameterDescription, bool isReadOnly, Property? schemaProperty, CSharpType? valueType = null, bool optionalViaNullability = false)
-            : this(declaration, parameterDescription, isReadOnly, schemaProperty, schemaProperty?.IsRequired ?? false, valueType, optionalViaNullability)
+        public ObjectTypeProperty(MemberDeclarationOptions declaration, string parameterDescription, bool isReadOnly, Property? schemaProperty, CSharpType? valueType = null, bool optionalViaNullability = false, SourcePropertySerializationMapping? serializationMapping = null)
+            : this(declaration, parameterDescription, isReadOnly, schemaProperty, schemaProperty?.IsRequired ?? false, valueType: valueType, optionalViaNullability: optionalViaNullability, serializationMapping: serializationMapping)
```
live1206, line 36: It seems the name assignment is not needed here?

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -310,19 +313,21 @@ public JsonObjectSerialization BuildJsonObjectSerialization(ObjectSchema objectS
...
         }
 
         private static void PopulatePropertyBag(PropertyBag propertyBag, int depthIndex)
         {
-            foreach (Property property in propertyBag.Properties.ToArray())
+            foreach (ObjectTypeProperty property in propertyBag.Properties.ToArray())
             {
-                if (depthIndex >= (property.FlattenedNames?.Count ?? 0) - 1)
+                var schemaProperty = property.SchemaProperty!; // we ensure this is not null when we build the array
+                ICollection<string> flattenedNames = property.SerializationMapping?.SerializationPath as ICollection<string> ?? schemaProperty.FlattenedNames;
```
live1206, line 324: `property.SerializationMapping?.SerializationPath` is `IReadOnlyList`, casting it to `ICollection` basically is to iterate it and turn it into a non-readonly collection.
Another option is to duplicate the logic:
```csharp
string name;
int serializationPathCount = property.SerializationMapping?.SerializationPath?.Count ?? 0;
if (serializationPathCount > 0) 
{
    if (depthIndex >= serializationPathCount - 1)
    {
        {
            continue;
        }
        name = property.SerializationMapping?.SerializationPath!.ElementAt(depthIndex);
    }
}
else
{
    if (depthIndex >= (flattenedNames?.Count ?? 0) - 1)
    {
        continue;
    }

    name = flattenedNames!.ElementAt(depthIndex);
}
```

readme.md
```diff
@@ -541,7 +541,330 @@ namespace Azure.Service.Models
...
+            }
+            return new Cat(name);
+        }
+    }
+}
+```
+
+You could also add the `CodeGenMemberSerializationHooks` attribute to the property to have your own serialization/deserialization logic of the new property. You might have to do this if the type of your new property is an object type or any type that our generator does not natively support.
+
+**NOTE: Adding property to serialization/deserialization methods currently only works for DPG.**
```
live1206, line 863: Does this PR implement for Mgmt as well?

src/AutoRest.CSharp/Common/Input/Source/ModelTypeMapping.cs
```diff
@@ -28,6 +28,8 @@ public ModelTypeMapping(INamedTypeSymbol modelAttribute, INamedTypeSymbol member
 
             foreach (ISymbol member in GetMembers(existingType))
             {
+                string[]? serializationPath = null;
```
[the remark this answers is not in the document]
ArcturusZhang: One member may have both attributes, we need to put the information into the dictionary after we have gone through all the attributes, otherwise we get "key already exist" exception

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
ArcturusZhang: I will consider this.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
ArcturusZhang: This makes code harder to read.
The value of `SerializationMapping?.SerializationPath` is meant to override the existing things in `FlattenedNames` if configured.
We have to have the cast because their types are not the same, and I do not want to call `ToArray` or `ToList` because that will allocate another two collections.

readme.md
ArcturusZhang: Not yet.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeProperty.cs
ArcturusZhang: did not get the point

live1206: Never mind, I saw there are additional optional properties in between.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
live1206: The code does get more complex by duplication, which does not make any copy and no iteration. 
I can take either here.

readme.md
live1206: OK. I thought the change in `test/TestProjects/MgmtCustomizations/Customization/` is showing it works for mgmt.

ArcturusZhang: Oh - I forgot to remove it. I will add a comment to that property emphasizing that this does not work now.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
ArcturusZhang: I renamed it to SerializationPropertyBag

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
ArcturusZhang: removed

readme.md
ArcturusZhang: Actually i already removed the `OldProperty` which is expected to be added into the serialization and deserialization.

readme.md
```diff
@@ -541,7 +541,330 @@ namespace Azure.Service.Models
...
+                    continue;
+                }
+            }
+            return new Cat(name);
+        }
+    }
+}
+```
+
+You can also add multiple element in the attribute and the generator will generate the property in a hierarchy.
```
chunyu3, line 608:
```suggestion
If you want to change the location of the property in the json, you can add all the elements in the location path to this attribute, then the generator will generate the property in correct hierarchy. e.g. following example : move `Name` property under property `properties` and rename to `catName`.
```

readme.md
```diff
@@ -541,7 +541,330 @@ namespace Azure.Service.Models
...
+            }
+            return new Cat(name);
+        }
+    }
+}
+```
+
+</details>
+
+If you want to change the implementation of serialization/deserialization method of one particular property, you could define your own hook methods and assign them to a property using the `CodeGenMemberSerializationHooks` attribute.
```
chunyu3, line 682: Maybe It is better to separate it as another customization.

test/TestProjects/MgmtCustomizations/Customization/Pet.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Drawing;
+using System.Runtime.CompilerServices;
+using System.Text.Json;
+using Azure.Core;
+
+namespace MgmtCustomizations.Models
+{
+    public abstract partial class Pet
+    {
+        /// <summary> The size of the pet. Despite we write type string here, in the real payload of this request, it is actually sending using a number, therefore the type in this swagger here is wrong and we need to fix it using customization code. </summary>
```
chunyu3, line 15:
```suggestion
        /// <summary> The size of the pet. Despite we write type int here, in the real payload of this request, it is actually sending using a string, therefore the type in this swagger here is wrong and we need to write customized serialize and deserialize method. </summary>
```

readme.md
```diff
@@ -541,7 +541,330 @@ namespace Azure.Service.Models
 
 ### Customize serialization/deserialization methods
 
-Use the [Replace any generated member](#replace-any-generated-member) approach to replace Serialize/Deserialize method with a custom implementation.
+If you want to change the property name that serializes into the JSON or deserializes from the JSON, you could define your own property name using the `CodeGenMemberSerialization` attribute.
+
+<details>
+
+Define a partial class and the property you want to change in the partial class, and add `CodeGenMemberSerialization` attribute to it:
```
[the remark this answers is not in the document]
ArcturusZhang: This is a very simple use case therefore it could apply to this pattern.
In the use case now, if we just put all the "generated code before" here, the document will be too long to read.
I could add some more explanation by describing a scenario and how we could do it and how the code CHANGES

readme.md
ArcturusZhang: I believe this is also one of the topics of "customization serialization/deserialization method". I did not have a better name that distinguishing this from the topic above, and this attribute can be used together with the attribute above. Therefore I do not want this to be in another section

readme.md
m-nash: Just to make sure I understand this works for all targets right?

test/TestProjects/MgmtCustomizations/Customization/Pet.cs
ArcturusZhang: Resolved in another PR

readme.md
ArcturusZhang: Resolved in another PR

readme.md
ArcturusZhang: > Just to make sure I understand this works for all targets right?

Adding an extra property to serialization/deserialization using the attribute only works for DPG.

live1206 (approved):

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -255,31 +255,33 @@ public XmlObjectSerialization BuildXmlObjectSerialization(ObjectSchema objectSch
...
 
-        private IEnumerable<JsonPropertySerialization> GetPropertySerializationsFromBag(PropertyBag propertyBag, SchemaObjectType objectType)
+        private IEnumerable<JsonPropertySerialization> GetPropertySerializationsFromBag(SerializationPropertyBag propertyBag, SchemaObjectType objectType)
         {
-            foreach (Property property in propertyBag.Properties)
+            foreach (ObjectTypeProperty property in propertyBag.Properties)
             {
-                var objectProperty = objectType.GetPropertyForSchemaProperty(property, includeParents: true);
-                var parameter = objectType.SerializationConstructor.FindParameterByInitializedProperty(objectProperty);
+                var schemaProperty = property.SchemaProperty!; // we ensured this is never null when constructing the list
```
ArthurMa1978, line 262: If this `property.SchemaPropert` is never null then why it's defined as nullable?

ArthurMa1978 (approved):

m-nash (approved):

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
ArcturusZhang: we need it to be nullable because an `ObjectTypeProperty` can be either constructed from the Property from code model (in this case SchemaProperty is not null, InputModelProperty is null), or constructed from the InputModelProperty from the emitter (in this case InputModelProperty is not null, SchemaProperty is null)

PR #3445, opened 2023-05-29
Refactor and fixes in source input model

ArcturusZhang:
# Description

Since we have quite a few attributes to pass around, this PR introduces a new class to have them all.

This PR also fixes a key conflicts when a member has both serialization attributes (`CodeGenMemberSerialization` and `CodeGenMemberSerializationHooks`)

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
@@ -163,18 +163,6 @@ namespace AutoRest.CSharp.Generation.Types
         internal static CSharpType FromSystemType(BuildContext context, Type type)
             => FromSystemType(type, context.DefaultNamespace, context.SourceInputModel);
 
-        public bool IsCollectionType()
-        {
-            if (!IsFrameworkType)
-                return false;
-
-            return FrameworkType.Equals(typeof(IList<>)) ||
-                FrameworkType.Equals(typeof(IEnumerable<>)) ||
-                FrameworkType == typeof(IReadOnlyList<>) ||
-                FrameworkType.Equals(typeof(IDictionary<,>)) ||
-                FrameworkType == typeof(IReadOnlyDictionary<,>);
-        }
-
         public CSharpType GetNonNullable()
         {
             if (!IsNullable)
--- a/src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
+++ b/src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
@@ -102,7 +102,7 @@ namespace AutoRest.CSharp.Input.Source
                 }
             }
 
-            return true;
+            return usage != null || formats != null;
         }
 
         private static string[]? ToStringArray(ImmutableArray<TypedConstant> values)
--- a/src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
+++ b/src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
@@ -2,8 +2,11 @@
 // Licensed under the MIT License.
 
 using System;
-using Microsoft.CodeAnalysis;
+using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using Azure.Core;
+using Microsoft.CodeAnalysis;
 
 namespace AutoRest.CSharp.Input.Source
 {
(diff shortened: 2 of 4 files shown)


ArcturusZhang:
The changes in [the regen](https://github.com/Azure/azure-sdk-for-net/pull/36572) is the same as the changes in [the auto PR](https://github.com/Azure/azure-sdk-for-net/pull/36574) triggered on feature/v3

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -163,18 +163,6 @@ internal static CSharpType FromSystemType(Type type, string defaultNamespace, So
         internal static CSharpType FromSystemType(BuildContext context, Type type)
             => FromSystemType(type, context.DefaultNamespace, context.SourceInputModel);
 
-        public bool IsCollectionType()
```
ArcturusZhang, line 166: This method is removed because this method is never used and we have the same method in `TypeFactory` which is widely used everywhere

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            var arguments = attributeData.ConstructorArguments;
+            serializationHook = arguments[0].Value as string;
+            deserializationHook = arguments[1].Value as string;
+
+            hooks = (serializationHook, deserializationHook);
+            return serializationHook != null || deserializationHook != null;
+        }
+
+        public bool TryGetCodeGenModelAttributeValue(AttributeData attributeData, out string[]? usage, out string[]? formats)
```
lirenhe, line 86: Just curious about the best practise here, can we just return an empty array here instead of setting it as null?

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            return true;
+        }
+
+        private static string[]? ToStringArray(ImmutableArray<TypedConstant> values)
+        {
+            if (values.IsDefaultOrEmpty)
+            {
+                return null;
```
lirenhe, line 112: can we use return Array.Empty<string>() here?

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
ArcturusZhang: We should only return null when null has a meaning.
Here null has meanings for these two properties because they are patches to usage and formats, `null` means "do not patch"

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
ArcturusZhang: Explained above

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
ArcturusZhang: another reason is that this has been implemented this way therefore I do not really want to risk to break it since it is working properly.
All the things I have done is to move it here.

lirenhe: I see.   Considering this is the existing behavour, let's keep it to avoid the risk of breaking.

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
```diff
@@ -0,0 +1,121 @@
...
+                    case nameof(Azure.Core.CodeGenModelAttribute.Usage):
+                        usage = ToStringArray(namedArgument.Value.Values);
+                        break;
+                    case nameof(Azure.Core.CodeGenModelAttribute.Formats):
+                        formats = ToStringArray(namedArgument.Value.Values);
+                        break;
+                }
+            }
+
+            return true;
```
chunyu3, line 105: always return true?
Shall we return false when no usage or format got?

chunyu3 (approved):

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
ArcturusZhang: make sense. let me change

lirenhe (approved):

PR #3448, opened 2023-05-29
Fix some issues around the property type, value type and existing member in DPG

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3449

# Description

This is a follow up for the comment we have in my previous PR: https://github.com/Azure/autorest.csharp/pull/3425/files#r1203289487
I referenced the same way mgmt/HLC is doing to get value type for a property with existing member.

The changes in this PR, they actually did not change anything in behavior.
The code looks like this way:

```diff
internal Model DeserializeModel(JsonElement root)
{
	/* omit irrelavants */
	Optional<int> optionalInt = default;
	/* unchanged content */
-	return new Model(optioanlInt);
+	return new Model(optionalInt.Value);
}
// reference: the signature of ctor of Model looks like:
// internal Model(int optionalInt);
```

In the definition of `Optional<T>`, there is an implicit operator: https://github.com/Azure/autorest.csharp/blob/03db394404c19613abbe25d550faddcd4a0c1a77/src/assets/Generator.Shared/Optional.cs#L109
And our constructors will never accept `Optional<T>` as parameter, therefore the usages:
- new Model(optioanlInt);
- new Model(optioanlInt.Value);

are actually the same, in the first one, the implicit operator will automatically be invoked

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -203,28 +203,31 @@ public static TypeDeclarationOptions CreateTypeAttributes(string defaultName, st
             return new TypeDeclarationOptions(defaultName, defaultNamespace, defaultAccessibility, isAbstract, false);
         }
 
-        public static MemberDeclarationOptions CreateMemberDeclaration(string defaultName, CSharpType defaultType, string defaultAccessibility, ISymbol? existingMember, TypeFactory typeFactory)
+        public static CSharpType GetTypeFromExisting(ISymbol existingMember, CSharpType defaultType, TypeFactory typeFactory)
         {
-            if (existingMember != null)
+            var newType = existingMember switch
```
chunyu3, line 208: What if existingMember == null?  ISymbol is interface, it can be null.

It is better to check inside the function, otherwise, this function may be called in incorrect way.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -57,13 +58,16 @@ public ModelTypeProviderFields(InputModelType inputModel, TypeFactory typeFactor
             foreach (var inputModelProperty in inputModel.Properties)
             {
                 var originalFieldName = inputModelProperty.Name.ToCleanName();
-                var (originalFieldType, fieldValueType) = GetPropertyDefaultType(inputModel.Usage, inputModelProperty, typeFactory);
+                var propertyType = GetPropertyDefaultType(inputModel.Usage, inputModelProperty, typeFactory);
+                var optionalViaNullability = !inputModelProperty.IsRequired &&
```
chunyu3, line 62: What's the meaning of `optionalViaNullability`?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -184,45 +203,27 @@ private static FieldDeclaration CreateFieldFromExisting(ISymbol existingMember,
...
             _ => throw new NotSupportedException($"'{existingMember.ContainingType.Name}.{existingMember.Name}' must be either field or property.")
         };
 
-        private static (CSharpType PropertyType, CSharpType ValueType) GetPropertyDefaultType(in InputModelTypeUsage modelUsage, in InputModelProperty property, TypeFactory typeFactory)
+        private static CSharpType GetPropertyDefaultType(in InputModelTypeUsage usage, in InputModelProperty property, TypeFactory typeFactory)
         {
             var propertyType = typeFactory.CreateType(property.Type);
 
-            if (modelUsage == InputModelTypeUsage.Output ||
+            if (!usage.HasFlag(InputModelTypeUsage.Input) ||
```
chunyu3, line 220: Why change usage from `Output` to `Input` here?

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
ArcturusZhang: No, here `existingMember` is never null. See the signature, it does not have the question mark.
Please note that we enabled nullable static analysis, the compiler ensures that, here `existingMember` is not null.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
ArcturusZhang: the usage is a "flag", and actually it has two unique flags - Input and Output.
RoundTrip is a combination of input and output.

The condition `usage == Output` is the same as `!usage.HasFlag(Input)`. Iterate all the cases to find out:
| usage value | result of `== Output` | result of `!HasFlag(Input) |
| :-- | :-- | :-- |
| None | false | false |
| Input | false | false |
| Output | true | true |
| RoundTrip | false | false |

archerzz (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
ArcturusZhang: There are some comments in SchemaObjectType:
```
            // We represent property being optional by making it nullable
            // Except in the case of collection where there is a special handling
```
But I personally suppose a better name for this is `nullabilityViaOptionality`

ArcturusZhang: I will add the same comment here

ArcturusZhang: And please do not think this is a new thing - DPG previously had this, when building the serializations (on feature/v3): https://github.com/Azure/autorest.csharp/blob/c397a85ec1a0eb4d650a1d3c5900e794770a4d4e/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs#L165

The things I do here is to promote that to here when we are building it therefore we could pass this flag through instead of calculation again when we are building the serialization.

chunyu3 (approved):

lirenhe (approved):

PR #3453, opened 2023-05-31
feat(dpg): support property bag for protocol methods from swagger-based definition

### LINKED ISSUE #3428: Add a option to enable options bag for some methods
In swagger we won't have the client.tsp capabilities and since we don't want an automatic decision to convert to options bag based on threshold, so we will add a configuration to configure which methods will apply options bag.

archerzz:
# Description
- add autorest configuration property `group-parametes-method-list` which is a list of operation IDs which will have property bag in the generated protocol methods
- add property `InputOperation.GroupParameters` to indicate whether an input operation should have property bag in the generated protocol method
- refactor existing property bag codes, and add new `LowLevelPropertyBag`
- update logic to generate codes to invoke rest client methods, now property bag expansion is supported
  - make `Paramter.IsPropertyBag = true` for parameters of rest client methods when property bag is enabled
  - when generating the parameters to pass to rest client methdos, then we will generate codes to expand property bag and refer to the corresponding property- add input mode types `Azure.Core.RequestContent` and `Azure.Core.RequestContentNullable` into `TypeFactory` so that `RequestContent` can be a property of generated property bag model
- update sample generation codes to support property bag
- add test cases

resolve #3428

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -41,6 +41,8 @@ public TypeFactory(OutputLibrary library)
             InputEnumType enumType             => _library.ResolveEnum(enumType).WithNullable(inputType.IsNullable),
             // TODO -- this is a temporary solution until we refactored the type replacement to use input types instead of code model schemas
             InputModelType { Namespace: "Azure.Core.Foundations", Name: "Error" } => SystemObjectType.Create(AzureResponseErrorType, AzureResponseErrorType.Namespace!, null).Type,
+            InputModelType { Namespace: "Azure.Core", Name: "RequestContent" } => KnownParameters.RequestContent.Type,
+            InputModelType { Namespace: "Azure.Core", Name: "RequestContentNullable" } => KnownParameters.RequestContentNullable.Type,
```
ArcturusZhang, line 45: Maybe we could simplify?
```suggestion
            InputModelType { Namespace: "Azure.Core", Name: "RequestContent" or "RequestContentNullable" } => typeof(RequestContent),
```
since the two types here actually are identical

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -66,6 +66,9 @@ public static FormattableString GetIdentifiersFormattable(this IEnumerable<strin
                 _ => FormattableStringFactory.Create(GetNamesForMethodCallFormat(count, 'I'), identifiers.ToArray<object>())
             };
 
+        public static FormattableString GetParameterReferenceExpressions(this IReadOnlyCollection<Parameter> parameters)
+            => GetIdentifiersFormattable(parameters.Select(p => p.IsPropertyBag ? $"options.{p.Name.FirstCharToUpperCase()}" : p.Name), parameters.Count);
```
ArcturusZhang, line 70: since we get the name of our parameters by calling `ToVariableName`, we should not just simply call FirstCharToUpperCase, because `ToVariableName` might do something more especially we are in a common place not specific to DPG.
Could we change this to `ToCleanName` instead? This is opposite operation to `ToVariableName` I believe.

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
```diff
@@ -27,7 +27,8 @@ internal record InputOperation(
     OperationLongRunning? LongRunning,
     OperationPaging? Paging,
     bool GenerateProtocolMethod,
-    bool GenerateConvenienceMethod)
+    bool GenerateConvenienceMethod,
+    bool GroupParameters)
```
ArcturusZhang, line 31: Should we change this to something else? We already have another thing called `Grouped` or something. This might introduce ambiguity.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -86,7 +86,7 @@ public static RestClientMethod BuildRequestMethod(InputOperation operation, Para
                 BuilderHelpers.EscapeXmlDescription(operation.Description),
                 responseType,
                 request,
-                parameters,
+                (operation.GroupParameters ? parameters.Select(p => (p != KnownParameters.WaitForCompletion && p != KnownParameters.RequestContext) ? p with { IsPropertyBag = true } : p ).ToArray() : parameters),
```
ArcturusZhang, line 89: We could reomve the outside parathesis
BTW could we add some comments explaining what this line is doing?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -205,7 +205,12 @@ private static (CSharpType PropertyType, CSharpType ValueType) GetPropertyDefaul
             }
 
             var valueType = propertyType;
-            if (propertyType.IsValueType && !property.IsRequired)
+            if (propertyType.IsValueType &&
+                !property.IsRequired &&
+                !isPropertyBag) // For property bag properties, we keep the original type nullability, because in method parameters, we have two types of optional parameters
+                // "int a = 1, in? b = 3", if we will transform all optional parameters as model nullable properties, then there will be compilation error in the property bag
+                // expansion logic, e.g. "Foo(options.A)" won't match "Foo(int a = 1)", since "options.A" is of type int?
+                // Another way to resolve this issue is to change the expansion expression like "Foo(options.A ?? 1)", which is urgly.
```
ArcturusZhang, line 213: well I believe the reason here could be simplified:
We have to add nullable to the type when we are building the properties, because the "nullibility via optionality" has not been considered.
But when we are in an options bag, this "nullibility via optionality" has been done by the request builder or operation builder, therefore we should do that again here.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
archerzz: I tend to keep the current logic, because `GetType()` is widely used and we don't know whether this simplification will be right in the future.

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
archerzz: Changed to `ToCleanName`

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
archerzz: added some comments

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
archerzz: Updated comments

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
archerzz: Changed to `HasPropertyBag`

archerzz:
We've decided to pause this work.

PR #3458, opened 2023-06-02
[typespec-csharp] Float TypeSpec dependencies and move to peerDeps

mikeharder:
- Related to Azure/azure-sdk-for-net#37162, but these should be mergable in either order

## To Do
- Update scripts which test emitter in net repo to pin typespec packages


src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -34,6 +34,26 @@
...
         "dist/src/**"
     ],
+    "dependencies": {
+        "@autorest/csharp": "3.0.0-beta.20230309.1",
+        "fs-extra": "^10.1.0",
+        "js-yaml": "^4.1.0",
+        "json-serialize-refs": "^0.1.0-0",
+        "winston": "^3.8.2"
+    },
+    "peerDependencies": {
```
mikeharder, line 44: Should any of these packages be `dependencies` instead of `peerDependencies`?  The biggest difference is `peerDependencies` prevents multiple versions of a dependency being loaded side-by-side if downstream packages request incompatible ranges.  So any package we want to prevent being loaded side-by-side should be a `peerDependency`.

src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -34,6 +34,26 @@
...
+        "@azure-tools/typespec-autorest": ">=0.30.0 <1.0.0",
+        "@azure-tools/typespec-azure-core": ">=0.30.0 <1.0.0",
+        "@azure-tools/typespec-client-generator-core": ">=0.30.0 <1.0.0",
+        "@typespec/compiler": ">=0.44.0 <1.0.0",
+        "@typespec/eslint-config-typespec": ">=0.7.0 <1.0.0",
+        "@typespec/eslint-plugin": ">=0.44.0 <1.0.0",
```
timotheeguerin, line 50: those 2 should be a `devDependency`

src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -34,6 +34,26 @@
...
+        "@typespec/http": ">=0.44.0 <1.0.0",
+        "@typespec/library-linter": ">=0.44.0 <1.0.0",
```
timotheeguerin, line 52: devDependency too

src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -34,7 +34,27 @@
     "files": [
         "dist/src/**"
     ],
+    "dependencies": {
+        "@autorest/csharp": "3.0.0-beta.20230309.1",
+        "fs-extra": "^10.1.0",
+        "js-yaml": "^4.1.0",
+        "json-serialize-refs": "^0.1.0-0",
+        "winston": "^3.8.2"
```
timotheeguerin, line 42: this one might be good to remove/understand why we need it. Typespec comes with a tracing engine that will be configurable by the user https://microsoft.github.io/typespec/introduction/configuration/tracing

src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -34,7 +34,27 @@
...
+    },
+    "peerDependencies": {
+        "@azure-tools/typespec-autorest": ">=0.30.0 <1.0.0",
```
timotheeguerin, line 45: autorest should also not be a dependency, maybe a devDependency if you are using it in testing/regression tests

src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -34,7 +34,27 @@
...
+        "@azure-tools/typespec-azure-core": ">=0.30.0 <1.0.0",
+        "@azure-tools/typespec-client-generator-core": ">=0.30.0 <1.0.0",
+        "@typespec/compiler": ">=0.44.0 <1.0.0",
+        "@typespec/http": ">=0.44.0 <1.0.0",
+        "@typespec/openapi": ">=0.44.0 <1.0.0",
```
timotheeguerin, line 50: same of openapi as for autorest

timotheeguerin:
left some more comments on dependencies that should maybe not be here or be as dev but depending on the current state of use that might need a bigger change that just moving them around

src/CADL.Extension/Emitter.Csharp/package.json
```diff
@@ -34,7 +34,27 @@
     "files": [
         "dist/src/**"
     ],
+    "dependencies": {
+        "@autorest/csharp": "3.0.0-beta.20230309.1",
+        "fs-extra": "^10.1.0",
+        "js-yaml": "^4.1.0",
+        "json-serialize-refs": "^0.1.0-0",
```
mikeharder, line 41: This package is not very widely used (we are the only dependents according to npmjs), and is a pre-release version. We try to avoid such dependencies in our GA packages.

Tracking: https://github.com/Azure/autorest.csharp/issues/3459

src/CADL.Extension/Emitter.Csharp/package.json
mikeharder: Tracked by https://github.com/Azure/autorest.csharp/issues/3462

src/CADL.Extension/Emitter.Csharp/package.json
mikeharder: @m-nash: What do you think?

src/CADL.Extension/Emitter.Csharp/package.json
mikeharder: @m-nash: What do you think?

m-nash:
Kicked off the regen PR to validate in the lang repo.  Where does the pinning happen to make sure that we have repeatability?

src/CADL.Extension/Emitter.Csharp/package.json
m-nash: Our library has no need for this dependency but at some point in time it was required for some reason.  If you can remove it without breaking anything I would say lets do that.

src/CADL.Extension/Emitter.Csharp/package.json
m-nash: Our library has no need for this dependency but at some point in time it was required for some reason.  If you can remove it without breaking anything I would say lets do that.

mikeharder: Removed

src/CADL.Extension/Emitter.Csharp/package.json
mikeharder: Removed in https://github.com/Azure/autorest.csharp/pull/3491

src/TypeSpec.Extension/Emitter.Csharp/package.json
```diff
@@ -34,7 +34,31 @@
...
+        "@azure-tools/typespec-client-generator-core": ">=0.31.0 <1.0.0",
+        "@typespec/compiler": ">=0.45.0 <1.0.0",
+        "@typespec/http": ">=0.45.0 <1.0.0",
+        "@typespec/rest": ">=0.45.0 <1.0.0",
+        "@typespec/versioning": ">=0.45.0 <1.0.0"
+    },
     "devDependencies": {
+        "@azure-tools/typespec-azure-core": "0.31.0",
+        "@azure-tools/typespec-client-generator-core": "0.31.0",
+        "@typespec/compiler": "0.45.0",
```
m-nash, line 55: should this pr include a package-lock.json at the root level to pin all these dev dependencies?

mikeharder: Good point.  I updated the `package-lock.json` that already exists in the repo root, and verified that it uses the version of `@typespec/compiler` listed in `devDependencies` (`0.45.0`), rather than `latest` (`0.45.2`).  So we shouldn't need to duplicate the `@typespec` and `@azure-tools` packages in the root `package.json`.

Issue #3459, opened 2023-06-02
[typespec-csharp] Remove dependency on "json-serialize-refs"


mikeharder:
Can we remove the dependency on "json-serialize-refs"?  This package is not very widely used (we are the only dependents according to npmjs), and is a pre-release version.  We try to avoid such dependencies in our GA packages.

mikeharder:
@m-nash: I think it's important we at least consider removing this dependency before GA.  Who's the right person to do this?

m-nash:
@mikeharder I will take this.

PR #3460, opened 2023-06-02
[typespec-csharp] Pin dependency "json-serialize-refs"

mikeharder:
- Pre-release versions must be pinned since breaking changes are allowed at any time

I strongly recommend removing this dependency (#3459), but if it cannot be removed, it should definitely be pinned since it is a pre-release version, which allows breaking changes at any time.


mikeharder:
@m-nash: I think we should merge this, since it can't hurt, and we're not sure if we're going to be able to remove the dependency.

m-nash (approved):

PR #3461, opened 2023-06-02
Update shared sources

mikeharder:
- Fix pipeline "autorest.csharp - ci" in feature/v3
- Started failing after https://github.com/Azure/autorest.csharp/pull/3448


ArcturusZhang (approved):

ArcturusZhang:
Changes already included in https://github.com/Azure/autorest.csharp/pull/3441

PR #3470, opened 2023-06-06
Fix the name issue in model factory and extension class for DPG and HLC

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3471

# Description

1. This PR normalizes the usages between `Configuration.Namespace` and `_inputNamespace.Name`.
The first `Configuration.Namespace` represents the namespace configured in our configuration file (tspLocation for typespec SDKs or autorest.md for swagger SDKs), the second `_inputNamespace.Name` represents the top level `namespace` defined in their typespec file, or the `title` defined in the swagger when it is a swagger-based DPG SDK.

Since we already ensures the `Configuration.Namespace` and `Configuration.LibraryName` is never null, I removed quite a few null value fallback statements in this PR, such as 
```
var clientNamespace = Configuration.Namespace;// ?? _inputNamespace.Name; this part is removed because the Configuration.Namespace is never null
```
2. This PR changes the name of model factory and aspdotnet extension class to use the `RPName` which is the last segment in the client namespace.
3. This PR also refactors code around getting the above name and namespace.

# Checklist

# Ready to Land?


ArcturusZhang:
Hi @m-nash could you please have a review on this PR and on its corresponding [regen](https://github.com/Azure/azure-sdk-for-net/pull/36912)?
In the regen, there are some name changes in model factory - I will add customization to change their name back in the package is a GA package.
In the meantime, I also notice for some packages, the last segment of the namespace does not really have a meaning, for instance `Azure.IoT.Hub.Service` (ServiceModelFactory does not make much sense). But since the name of model factory class it very easy to change by using some customization code, I believe it is acceptable.
No mgmt RP affected in this change.

ArcturusZhang:
image
@m-nash 
Currently the name of our client is constructing from the namespace defined in typespec or title in swagger (as you could see on the above screenshot).
In DPG packages, now we only have one client, therefore we will have `XXClient`, `XXClientOptions`, `XXModelFactory` and `XXClientBuilderExtensions`.
Since this PR changes the way we name the model factory and aspdotnet extension, those names will no longer align with each other.

I am calling this out to get your attention.

ArcturusZhang:
@m-nash I have fixed the compilation issues, naming issues, etc, in the [regen PR](https://github.com/Azure/azure-sdk-for-net/pull/36912), and there are quite a few my comments explaining the reason of making the changes.
Most of the RPs impacted by this change I have reverted the breaking changes in them by adding customized code. Some of them have breaking changes because I believe the new name makes more sense and those packages are beta.

In my future iterations, I will carry the fix commit to the new regen, but I will not be able to migrate those comments around.

m-nash:
Left a few comments on the regen preview. https://github.com/Azure/azure-sdk-for-net/pull/36934.

In general we don't need to fix naming for anything that isn't GA so lets make sure we double check every library.

ArcturusZhang:
@m-nash this is the new regen: https://github.com/Azure/azure-sdk-for-net/pull/37060
In this PR
- for GA packages, I introduced customization code for them to mitigate breaking changes if any
- for non-GA packages, I fixed possible issues in their current customization code if any, but the model factory name is always the new one we generated

m-nash:
hmmm, now that I see this name here https://apiview.dev/Assemblies/Review/bf8831449b2c44e7a404911dde573209/87393b4d700c41c99db81b88dc26255c?diffRevisionId=be7c0d0176834f22a9b1254a30a35584&doc=False&diffOnly=True#Azure.IoT.Hub.Service.Models.ServiceModelFactory, we might want to consider a different algo for "FindRP".

Would it make sense to use everything after `ResourceManager` if it exists and everything after `Azure` if it doesn't?

ArcturusZhang:
> hmmm, now that I see this name here [apiview.dev/Assemblies/Review/bf8831449b2c44e7a404911dde573209/87393b4d700c41c99db81b88dc26255c?diffRevisionId=be7c0d0176834f22a9b1254a30a35584&doc=False&diffOnly=True#Azure.IoT.Hub.Service.Models.ServiceModelFactory](https://apiview.dev/Assemblies/Review/bf8831449b2c44e7a404911dde573209/87393b4d700c41c99db81b88dc26255c?diffRevisionId=be7c0d0176834f22a9b1254a30a35584&doc=False&diffOnly=True#Azure.IoT.Hub.Service.Models.ServiceModelFactory), we might want to consider a different algo for "FindRP".
> 
> Would it make sense to use everything after `ResourceManager` if it exists and everything after `Azure` if it doesn't?

@m-nash I would agree but we should have a discussion around this in our next sync up for details

ArcturusZhang:
@m-nash 
I fixed the issues in the regen again: https://github.com/Azure/azure-sdk-for-net/pull/37306
For GA libraries, I added customized code to avoid breaking change.
For non-GA libraries, I fixed their customized code to let the model factory or client extension to take the new name.

Here is the things I saw:
For some library, the new algorithm might work well. For instance, `Azure.IoT.Hub.Service` => `IoTHubServiceModelFactory`
For some library, the new algorithm creates cumbersome names. For instance, `Azure.AI.OpenAI` => `AIOpenAIModelFactory`

I do not think we could provide an algorithm that could cover all the cases - considering we could always introduce customization code to rename it, I believe the current solution might be acceptable
Well, same statement also applies well to our previous algorithm which only takes the last segment of the namespace.
The advantage of the new algorithm is that there would be a lot lower chance to have collisions on the name of model factories between different libraries

m-nash (approved):

ArcturusZhang:
@m-nash I am kind of blocked before I could merge this PR.
Apicompat is preventing us to change anything on the code gen attributes such as `CodeGenType`: https://github.com/Azure/azure-sdk-for-net/pull/37457/checks?check_run_id=14853310814
I believe we should update to make them to ignore our code gen related attributes

PR #3483, opened 2023-06-08
Make extension clients public to enable mocking for extension methods

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3493

# Description

<del>This is a simplified version of #3175 </del>

This PR enables a general way for our customer to mock the extension methods.
In general, to mock an extension method in out .Net mgmt SDK takes several unified steps of mocking:
1. Identify the type that the extension method is extending (referred as `extendee` below), follow the pattern `Mockable{RPName}{ExtendeeName}`, you will get a type which contains the real method to mock (referred as `extension class` below). For instance if the extendee is `ResourceGroupResource` in `Azure.ResourceManager.Compute`, you will get `MockableComputeResourceGroupResource`.
2. Mock the method on the `extension class` with the same parameter list as the extension method you would like to mock. For instance, to mock the extension method `public static Response DoSomething(this ResourceGroupResource resourceGroup, string someParameter, CancellationToken cancellationToken)`, you should mock the method `public virtual Response DoSomething(string someParameter, CancellationToken cancellationToken)` on the `extension class` `MockableComputeResourceGroupResource`.
3. Mock a hidden method `GetCachedClient` on the extendee to make it return the above mocked `extension class`.

By following the above three steps, your mock now should work in the same way that you usually call it.

This PR unifies the previous internal implementation to ensure we have the same pattern everywhere so that our customer gets a unified experience of mocking the extension methods.

This PR also tries to remove all the validations in the public extension class to make mocking more feasible. For instance, previously we have the code:
```csharp
public static partial class ComputeExtensions
{
	public static Response DoSomething(this ResourceGroupResource resourceGroup, string name)
	{
		Argument.ThrowIfNull(name, nameof(name));
		// implementation
	}
}
```
This will cause the mocking invocation has to pass in a non-null argument as well - but when we are mocking those instance method, since the implementation of the body has been completely replaced, we never have this issue.
Therefore this PR removes all these validations in the static class. But we still enforces those parameter to be non-null or non-empty because the same check will exist in the corresponding Mocking Extensions.

A follow up issue of this is to make a test or analyzer rule to enforce the pattern in all of our SDKs including the generated code and customized code. Because changing the internal implementation here will change the way of mocking things, changing it would be regarded as a behavioral breaking change and we should never change that once the `extension classes` are public.

# Checklist

# Ready to Land?


ArcturusZhang:
TODO:  remove the useless extra static factory methods for those extension clients.

ArcturusZhang:
We have multiple pattern in our extension method, which we might improve to simplify the experience of mocking. For instance,

1. The extension method is redirecting the method call to an extension client, for instance, if this method is extending `TenantResource`, we will have a `{RP}TenantResourceExtension` (or other name, but there is a pattern) with a method with exactly the same signature. Example:
```csharp
public static Response<T> DoSomething(this TenantResource tenantResource, string p, CancellationToken token = default)
{
	return GetTenantResourceExtension(tenantResource).DoSomething(tenantResource, p, token);
}
```
This includes the methods that return a collection.
2. The extension method is a helper method to get an instance with a name. This method is reusing the method returning the `ArmCollection`.
```csharp
[ForwardsClientCalls]
public static async Task<Response<StorageAccountResource>> GetStorageAccountAsync(this ResourceGroupResource resourceGroupResource, string accountName, StorageAccountExpand? expand = null, CancellationToken cancellationToken = default)
{
    return await resourceGroupResource.GetStorageAccounts().GetAsync(accountName, expand, cancellationToken).ConfigureAwait(false);
}
```
`resourceGroupResource.GetStorageAccounts()` is the static extension method described in type 1.
3. The extension method is extending ArmClient with a scope. This is calling a scope method or getting the collection of a scope resource.
```csharp
public static FakePolicyAssignmentCollection GetFakePolicyAssignments(this ArmClient client, ResourceIdentifier scope)
{
    return GetArmResourceExtensionClient(client, scope).GetFakePolicyAssignments();
}
```
4. The extension method is returning a resource instance from a given resource id
```csharp
public static GuestConfigurationAssignmentResource GetGuestConfigurationAssignmentResource(this ArmClient client, ResourceIdentifier id)
{
    return client.GetResourceClient(() =>
    {
        GuestConfigurationAssignmentResource.ValidateResourceId(id);
        return new GuestConfigurationAssignmentResource(client, id);
    }
    );
}
```

To mocking these methods, the customers need to do the following:
| Method type | How to mock |
| :-- | :-- |
| 1 | 1. See the `T`, get the `{RP}TExtension` type. 2. Mock the same method on `{RP}TExtension`. 3. Mock the `GetCachedClient` method on `T` |

ArcturusZhang:
Some thoughts: https://gist.github.com/ArcturusZhang/bd224f859d4895dcd9e838719f7aef39

ArcturusZhang:
TODO - more validation to move
We also need to move the scope resource type validation check into the mocking extension. Instead of checking it in the static extension method.

m-nash:
@ArcturusZhang I assume we are not merging this until we have feedback from our beta?  If so can we add the do not merge label to this PR?

ArcturusZhang:
> @ArcturusZhang I assume we are not merging this until we have feedback from our beta? If so can we add the do not merge label to this PR?

Yes, I added the label.

ArcturusZhang:
TODO -- there is a cref reference to `Resource.CreateResourceIdentifier` method. It causes compilation errors if we add overloads for it using customization code. Need to update the generator to full specify it.

Packages.Data.props
```diff
@@ -29,7 +29,7 @@
     <PackageReference Update="System.Net.ClientModel" Version="1.0.0-beta.1" />
     <PackageReference Update="Azure.Core" Version="1.35.0" />
     <PackageReference Update="Azure.Core.Experimental" Version="0.1.0-preview.18" />
-    <PackageReference Update="Azure.ResourceManager" Version="1.6.0" />
+    <PackageReference Update="Azure.ResourceManager" Version="1.8.0-alpha.20230725.1" />
```
ArthurMa1978, line 32: Pending on release 1.8.0 stable

ArthurMa1978 (approved):

Packages.Data.props
ArcturusZhang: updated

PR #3484, opened 2023-06-08
Use dev versions of TypeSpec packages

hallipr:
To test against the latest dev version of TypeSpec pacakges, we use an npm packages from the CadlRanch repo to update the root package.json and the emitter package.json to the dev package versions.   These package.json and subsequent package-lock.json changes are published as a build artifact for consumption in later jobs.  We will also conditionally ignore build breaks if we're building against dev packages.

### DIFF
--- a/eng/UseTypeSpecNextVersions.ps1
+++ b/eng/UseTypeSpecNextVersions.ps1
@@ -1,0 +1,29 @@
+$repoRoot = Resolve-Path "$PSScriptRoot\.."
+Push-Location $repoRoot
+
+Write-Host "node -v"
+node -v
+
+Write-Host "npm -v"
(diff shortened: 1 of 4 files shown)


hallipr:
@m-nash Please confirm changes

hallipr:
With the "Use next versions" checkbox disabled, the pipeline runs as normal:

Pipeline run
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2833438

Generated PR
https://github.com/Azure/azure-sdk-for-net/pull/36928

Validation build
https://dev.azure.com/azure-sdk/public/_build/results?buildId=2833627

m-nash (approved):

PR #3494, opened 2023-06-13
Change discriminator to a InputModelProperty with Input Type

ArcturusZhang:
# Description

This PR only moves the construction of the "discriminator property" on the base model with a discriminator, from the ModelTypeProviderFields class in the generator, to the emitter.

Previous behavior:
The emitter never output a property for the discriminator in tspCodeModel.json on the base model, instead we have to construct a property in the generator when building the model in `ModelTypeProviderFields` class. This does not really make sense. 
And this makes the input types cannot take a type for the discriminator as the code model schemas could do - in code model schemas, discriminators can be enums.

Current behavior:
The emitter will create a property for the discriminator, now the base model will have a property for its discriminator in the list of its properties. This aligns the behavior in code model schemas, and when we convert code model schemas to input types, we will not lose any information

BTW this PR fixes the issue that the discriminator is not included in the protocol method samples

# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -529,15 +529,32 @@ export function getInputType(
             const baseModel = getInputModelBaseType(m.baseModel);
             const properties: InputModelProperty[] = [];
 
+            const discriminator = getDiscriminator(program, m);
```
lirenhe, line 532: Please add the unit test for this change.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -39,17 +39,17 @@ public ModelTypeProviderFields(InputModelType inputModel, TypeFactory typeFactor
...
-                var originalFieldName = discriminator.ToCleanName();
-                var inputModelProperty = new InputModelProperty(discriminator, discriminator, "Discriminator", InputPrimitiveType.String, true, false, true);
-                var field = CreateField(originalFieldName, typeof(string), inputModel, inputModelProperty, false);
-                fields.Add(field);
-                fieldsToInputs[field] = inputModelProperty;
-                var parameter = Parameter.FromModelProperty(inputModelProperty, field.Name.ToVariableName(), field.Type);
-                parametersToFields[parameter.Name] = field;
-                serializationParameters.Add(parameter);
-            }
+            //if (discriminator is not null)
```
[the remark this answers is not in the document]
ArcturusZhang: I will remove this

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/modelType.test.ts
```diff
@@ -0,0 +1,114 @@
+import { TestHost } from "@typespec/compiler/testing";
+import assert, { deepStrictEqual } from "assert";
```
lirenhe, line 2: Have you sort the imports?

package.json
```diff
@@ -3,9 +3,10 @@
   "private": true,
   "description": "package.json intended for in-repo use only, package.json used for publishing is located in src/AutoRest.CSharp/package.json",
   "devDependencies": {
-    "@azure-tools/typespec-csharp": "src/TypeSpec.Extension/Emitter.Csharp",
     "@azure-tools/cadl-ranch-mockapis": "file:test/CadlRanchMockApis",
+    "@azure-tools/typespec-csharp": "src/TypeSpec.Extension/Emitter.Csharp",
     "@microsoft.azure/autorest.testserver": "3.3.24",
+    "@types/mocha": "^10.0.1",
```
chunyu3, line 9: autorest.csharp does not depend on `@types/mocha`. `Emitter.Csharp` depends on it which is already his package.json. 
We may remove this.

src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs
```diff
@@ -1124,7 +1125,9 @@ private static bool IsPropertyAssignable(ObjectTypeProperty property)
 
         private string FixKeyWords(string serializedName)
         {
-            var result = serializedName.Replace('-', '_');
+            // TODO -- this is incorrect, we should never change the property name here otherwise the serialized request content is wrong
+            // if the name changed after calling this method, we should use a dictionary instead of using anonymous object
+            var result = serializedName.Replace('-', '_').Replace(".", string.Empty);
```
chunyu3, line 1130: Each property has `name` and `serializedName`. `Name` will be the variable name in the model which should be updated to comply c# naming convention. `serializedName` is used in serialization which should not be updated. In this function, the input parameter string is not updated. That's why it does not bring error during serialization. 
Maybe we need to update the method signature `private string FixKeyWords(string name)` to avoid confusing, and we need call it for property name.

lirenhe (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -556,6 +573,7 @@ export function getInputType(
 
             // Temporary part. Derived types may not be referenced directly by any operation
             // We should be able to remove it when https://github.com/Azure/typespec-azure/issues/1733 is closed
+            // TODO -- remove this since the above issue is closed
```
chunyu3, line 576: the issue `1733` does not resolve this. In polymorphism scenario, the operation only reference the base model, and we still need to add derived model because we need the derived models. remove this TODO, suggested
```suggestion
            
```

package.json
ArcturusZhang: Sure, I will revert this change

chunyu3 (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
ArcturusZhang: What should we write here? Because by reading this comment, I saw the issue is closed and I suppose we should remove this workaround. According to your statements, we need to update this comment

src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs
ArcturusZhang: This is solving the case that serializedName has special characters. We could have a serialized name called "string", or "name.kind" which are both invalid identifier.
This is not about something is confusing, I added this todo and change the code to temporary workaround an issue I found in the samples.
To make things correct, we should never change anything in the "serialized name" as my statement here which means we should remove this method in the future.
I already opened an issue for this: https://github.com/Azure/autorest.csharp/issues/3499

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
ArcturusZhang: Well I removed this extra line

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/modelType.test.ts
ArcturusZhang: There is not really a feature to sort the imports. Since the prettier does not complain about this, I just let them be.
BTW I removed some unused import statements

PR #3503, opened 2023-06-15
Correct the date format of RequestConditions

### LINKED ISSUE #3351: Expected date format of IfModifiedSince
Actual:
https://github.com/Azure/autorest.csharp/blob/0d501998cca71449d49f04a171aa3db87c07220c/test/CadlRanchProjects/azure/core/traits/Generated/TraitsClient.cs#L169
We set "O" as a format, so the payload would be like: ""2022-08-26T14:38:00.0000000Z"

The expected behavior in cadl-ranch is ["Fri, 26 Aug 2022 14:38:00 GMT"](https://github.com/Azure/cadl-ranch/blob/0000ee363bfd27a34ec8b717172f3a0e610e61a4/packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts#L23)

pshao25:
Fix #3351


src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputParameterConverter.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+            RequestLocation requestLocation;
+            Enum.TryParse<RequestLocation>(location, ignoreCase: true, out requestLocation);
+
+            if (kind == null)
+            {
+                throw new JsonException("Parameter must have kind");
+            }
+            InputOperationParameterKind parameterKind;
+            Enum.TryParse<InputOperationParameterKind>(kind, ignoreCase: true, out parameterKind);
```
ArcturusZhang, line 94: How about we combine these two lines or just use Enum.Parse and let it panic when it cannot 
```suggestion
            Enum.TryParse<InputOperationParameterKind>(kind, ignoreCase: true, out var parameterKind);
```

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -337,7 +337,9 @@ private void BuildParameters()
                         requestConditionHeaders |= header;
                         requestConditionRequestParameter ??= operationParameter;
                         requestConditionSerializationFormat = requestConditionSerializationFormat == SerializationFormat.Default
+                            ? (operationParameter.SerializationFormat == SerializationFormat.Default
                             ? SerializationBuilder.GetSerializationFormat(operationParameter.Type)
+                            : operationParameter.SerializationFormat)
                             : requestConditionSerializationFormat;
```
ArcturusZhang, line 343: we have a double "a ? b : c" expression here, which makes this really hard to read. Could we convert this to a method?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputParameterConverter.cs
```diff
@@ -0,0 +1,122 @@
...
+                    return SerializationFormat.DateTime_RFC7231;
+                }
+
+                if (requestLocation == RequestLocation.Body)
+                {
+                    return SerializationFormat.DateTime_RFC3339;
+                }
+            }
+
+            return SerializationFormat.Default;
```
chunyu3, line 119:
```suggestion
            return SerializationBuilder.GetSerializationFormat(parameterType);
```

chunyu3: otherwise, the serialization format of other kind except `DateTime` will be lost.

lirenhe (approved):

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputParameterConverter.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+            Enum.TryParse<RequestLocation>(location, ignoreCase: true, out var requestLocation);
+
+            if (kind == null)
+            {
+                throw new JsonException("Parameter must have kind");
+            }
+            Enum.TryParse<InputOperationParameterKind>(kind, ignoreCase: true, out var parameterKind);
+
+            var parameter = new InputParameter(name, nameInRequest, description, parameterType, requestLocation, defaultValue, virtualParameter, groupBy, parameterKind, isRequired, isApiVersion, isResourceParameter, isContentType, isEndpoint, skipUrlEncoding, explode, arraySerializationDelimiter, headerCollectionPrefix, GetSerializationFormat(parameterType, requestLocation));
```
archerzz, line 95: For records with many properties, better to initialize using named parameters, like:
```suggestion
            var parameter = new InputParameter(
            Name: name,
           ...
           SerializationFormat: GetSerializationFormat(parameterType, requestLocation));
```

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputParameterConverter.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                resolver.AddReference(id, parameter);
+            }
+
+            return parameter;
+        }
+
+        private static SerializationFormat GetSerializationFormat(InputType parameterType, RequestLocation requestLocation)
+        {
+            if (parameterType is InputPrimitiveType type && type.Kind == InputTypeKind.DateTime)
```
archerzz, line 107: maybe using pattern matching like:
```suggestion
            if (parameterType is InputPrimitiveType { Kind : InputTypeKind.DateTime})
```

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -305,6 +305,13 @@ private IEnumerable<Parameter> BuildSpreadParameters(ModelTypeProvider model)
 
         private void BuildParameters()
         {
+            SerializationFormat GetSerializationFormat(InputParameter parameter)
```
archerzz, line 308: Can you explain why we need a check on `serializationFormat`?

pshao25: It is not a check. It is getting the serialization format. And it comes from Dapeng's comment that extract that logic into a function.

archerzz (approved):

chunyu3 (approved):

Issue #3466, opened 2023-06-02
Sample generator need support DFE


ArthurMa1978:
The [DFE](https://gist.github.com/JoshLove-msft/e7e85e44d6fd24c87e362a37260c9d21) is designed for DataFactory expression, the sample generator should support this.

PR #3509, opened 2023-06-19
Support generation DataFactoryElement in Sample Codegen

RodgeFu:
fixes https://github.com/Azure/autorest.csharp/issues/3466

Support generating sample code for DFE(DataFactoryElement) which may be Expression, Masked String, AKV reference or Literal value. 

Details about DataFactoryElement can be found at https://gist.github.com/JoshLove-msft/e7e85e44d6fd24c87e362a37260c9d21

Related work: Support DFE to generate Example Model properly is in the PR below in azure-sdk-tools repo:
https://github.com/Azure/azure-sdk-tools/pull/6364


ArcturusZhang (commented):
Shall we add something in the `MgmtMockAndSample` test project to verify this new thing could work properly?
You could add some new operations/models into the swagger that the test project `MgmtMockAndSample` is using, also if you added them you should add some examples files to it as well.

src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -142,7 +142,8 @@ private static CodeWriter AppendDataFactoryElementValue(this CodeWriter writer,
             }
             else
             {
-                CSharpType literlType = type.Arguments[0];
+                const int DFE_LITERAL_TYPE_ARG_INDEX = 0;
+                CSharpType literlType = type.Arguments[DFE_LITERAL_TYPE_ARG_INDEX];
```
ArcturusZhang, line 146: This is still a magic number. How about we just simply use `type.Arguments.First()` which has more clear semantic meaning.

src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -104,12 +105,50 @@ private static CodeWriter AppendFrameworkTypeValue(this CodeWriter writer, CShar
...
+            if (exampleValue.Schema is ObjectSchema os && os.Name.StartsWith(DFE_OBJECT_SCHEMA_PREFIX))
+            {
+                const string DFE_OBJECT_PROPERTY_TYPE = "type";
+                const string DFE_OBJECT_PROPERTY_VALUE = "value";
+                Dictionary<string, string> DFE_FACTORY_METHOD_MAPPING = new Dictionary<string, string>()
+                {
+                    ["Expression"] = "FromExpression",
+                    ["SecureString"] = "FromSecretString",
+                    ["AzureKeyVaultSecretReference"] = "FromKeyVaultSecretReference"
+                };
```
ArcturusZhang, line 133: Could we define this constant dictionary somewhere else like a helper for DFE types? Or at least at beginning of this class. We could either initialize it in a static constructor if it cannot initialize directly in the static field.

src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -104,12 +105,50 @@ private static CodeWriter AppendFrameworkTypeValue(this CodeWriter writer, CShar
...
+
+                string dfeType = (string)exampleValue.Properties![DFE_OBJECT_PROPERTY_TYPE].RawValue!;
+                string dfeValue = (string)exampleValue.Properties![DFE_OBJECT_PROPERTY_VALUE].RawValue!;
+
+                writer.UseNamespace(type.Namespace!);
+                writer.AppendRaw($"{type.ToString().Trim()}.{DFE_FACTORY_METHOD_MAPPING[dfeType]}(");
```
ArcturusZhang, line 139: Please use
```suggestion
                writer.AppendRaw($"{type}.{DFE_FACTORY_METHOD_MAPPING[dfeType]}(");
```
instead.
Also it would be more readable if we could first get the method name
```csharp
var dfeMethodName = DFE_FACTORY_METHOD_MAPPING[dfeType];
```
and then use it in the interpolation

src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -104,12 +105,50 @@ private static CodeWriter AppendFrameworkTypeValue(this CodeWriter writer, CShar
...
+                writer.AppendRawValue(dfeValue, typeof(string));
```
ArcturusZhang, line 140: Please use 
```suggestion
                writer.Append($"{dfeValue:L}");
```
where the format identifier `L` here means "literal". And we could combine this line into the above line together and get one single `Append` call.

src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -104,12 +105,50 @@ private static CodeWriter AppendFrameworkTypeValue(this CodeWriter writer, CShar
...
             if (exampleValue.Schema is ObjectSchema objectSchema)
                 return writer.AppendComplexFrameworkTypeValue(objectSchema, type.FrameworkType, exampleValue);
 
             return writer.AppendRawValue(exampleValue.RawValue, type.FrameworkType, exampleValue.Schema.Type);
         }
 
+        private static CodeWriter AppendDataFactoryElementValue(this CodeWriter writer, CSharpType type, ExampleValue exampleValue)
+        {
+            if (type.FrameworkType != typeof(DataFactoryElement<>))
+                throw new ArgumentException("DataFactoryElement<> is expected but got: " + type.ToString());
```
ArcturusZhang, line 120: We kind of do not need this because we kind of ensures this in our implementation.
How about we replace this guard by `Debug.Assert` which will be automatically removed in the release version of the generator.

RodgeFu: it's better to throw error as soon as possible to help on troubleshooting

ArthurMa1978 (approved):

PR #3510, opened 2023-06-19
Use ref from azure-sdk-for-net repo in sparse checkout

hallipr:
When using the sparse checkout step, we typically don't pass a commitish.  Git then uses the repo's default branch for checkout. This breaks scenarios where we want to change the checked out commit by using the resources tab when queuing the build.  Adding the `Commitish` parameter and using the repository resource's ref corrects this.

### DIFF
--- a/eng/pipelines/sdk-update.yml
+++ b/eng/pipelines/sdk-update.yml
@@ -5,9 +5,6 @@ parameters:
     type: boolean
     default: false
 
-variables:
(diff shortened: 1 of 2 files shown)


hallipr:
Fixes: https://github.com/Azure/azure-sdk-for-net/issues/37033

jsquire (approved):

PR #3533, opened 2023-06-28
Support parameters type change to request method for convenience method of paging

### LINKED ISSUE #3528: [Bug]Fail to build Generated Pageable convenience method
When there is extensible enum parameter in a pageable operation, the generated convenience method cannot compile. 
image

root cause:
In the pageable convenience method signature, the type is enum, but in the Create***Request method, the parameter type is string.

example:
https://github.com/chunyu3/azure-sdk-for-net/blob/82cd4888560c95b36c7a0d4765e032ee7607009d/sdk/loadtestservice/Azure.Developer.LoadTesting/src/Generated/LoadTestRunClient.cs#L1215

pshao25:
Fix #3528 

Currently we just concatenate formal parameter names to generate `CreateXXXRequest` and `CreateXXXNextRequest`. That fails when the parameter is not primitive. 

Solution: for every parameter in CreateXXXRequest, look for its corresponding parameter in convenience method and then do the convertion.


test/TestProjects/Pagination-TypeSpec/Pagination-TypeSpec.tsp
```diff
@@ -212,18 +212,86 @@ model TextBlockItem {
...
   @doc("A collection id may optionally be specified. Only entries in the specified (or default) collection will be returned.")
   op ListPaginationLedgerEntries is Azure.Core.Foundations.ResourceList<LedgerEntry, ListLedgerEntryInput, PagedLedgerEntries>;
 }
 
+interface AdditionalParameter {  
+  @summary("List the dimension values for the given metric dimension name.")
+  @doc("List the dimension values for the given metric dimension name.")
+  ListMetricDimensionValues is Azure.Core.ResourceList<
+    DimensionValueList, QueryParametersTrait<MetricDimensions>
+  >;
```
ArcturusZhang, line 292: What does the first generic parameter here in `ResourceList` mean? Is it the type of the item or the type of the `XXXListResult` which represents a page?

We are generating a method returning `Pageable<DimensionValueList>`, therefore, if our pageable implementation is correct, the generic parameter above means the type of the element. If so, I suggest we change the model name here, or just use string, or add some comment here to describe that this model `DimensionValueList` is the type of the element of the pageable operation to avoid future confusion

pshao25: DimensionValueList is the item type. 

Emmmm, it comes from service definition. Maybe we should ask service team to change?

chunyu3 (approved):

lirenhe (approved):

PR #3553, opened 2023-07-01
Bump Azure.Core.Expressions.DataFactory version to latest beta

JoshLove-msft:
# Description
In order to leverage the new types of Azure.Core.Expressions.DataFactory in code gen, bumping to the latest beta. This also required updating the Azure.Core dependency to the latest in order to avoid downgrade errors. It also required the following additional fixes:

- Updating PropertyReferenceTypeAttribute to support passing InternalPropertiesToInclude
- Including these internal properties in BuildProperties for SystemObjectTypes

# Checklist

# Ready to Land?


test/TestProjects/MgmtMockAndSample/src/Generated/Models/MhsmNetworkRuleSet.Serialization.cs
```diff
@@ -58,7 +59,7 @@ internal static MhsmNetworkRuleSet DeserializeMhsmNetworkRuleSet(JsonElement ele
             }
             Optional<NetworkRuleBypassOption> bypass = default;
             Optional<NetworkRuleAction> defaultAction = default;
-            Optional<IList<MhsmipRule>> ipRules = default;
+            Optional<IList<DataFactorySecretString>> ipRules = default;
```
JoshLove-msft, line 62: This (and other similar instances) should not be changing. We will need to figure out a way to avoid these false positives that are caused by DataFactorySecretString having only a single Value property.

src/AutoRest.CSharp/Mgmt/Decorator/TypeReferenceTypeChooser.cs
```diff
@@ -42,6 +42,12 @@ internal static class TypeReferenceTypeChooser
 
             foreach (System.Type replacementType in TypeReferenceTypes)
             {
+                // DataFactory reference types should only be used to replace generated types in a DataFactory namespace.
```
JoshLove-msft, line 45: This was necessary to avoid false positive replacements. If we need a more robust attribute-based solution we can investigate that as well, but this should at least unblock the DataFactory release.

test/AutoRest.TestServerLowLevel.Tests/httpInfrastructure.cs
```diff
@@ -169,50 +169,62 @@ public class HttpInfrastructureTests : TestServerLowLevelTestBase
         });
 
         [Test]
+        [Ignore("https://github.com/Azure/autorest.csharp/issues/3554")]
```
archerzz, line 172: After merge of #3570, those test cases should work now.

m-nash (approved):
As long as the final regen preview testing goes well this is ready to merge

PR #3555, opened 2023-07-02
Fix codegen so that we can generate samples for MySql properly

RodgeFu:
Fix codegen so that we can generate samples for MySql properly

### DIFF
--- a/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
+++ b/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
@@ -747,6 +747,13 @@ namespace AutoRest.CSharp.Mgmt.AutoRest
         public override CSharpType? FindTypeByName(string originalName)
         {
             _schemaOrNameToModels.TryGetValue(originalName, out TypeProvider? provider);
+
+            if (provider == null)
+            {
+                // Try to search declaration name too if no key matches. i.e. Resource Data Type will be appended a 'Data' in the name and won't be found through key
+                provider = _schemaOrNameToModels.FirstOrDefault(s => s.Value is MgmtObjectType mot && mot.Declaration.Name == originalName).Value;
+            }
+
             provider ??= ResourceSchemaMap.Values.FirstOrDefault(m => m.Type.Name == originalName);
             return provider?.Type;
         }
--- a/src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
+++ b/src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+using System.Transactions;
 using System.Xml;
 using AutoRest.CSharp.Generation.Types;
 using AutoRest.CSharp.Generation.Writers;
@@ -151,8 +152,18 @@ namespace AutoRest.CSharp.MgmtTest.Extensions
             writer.Append($"new {type}(");
             foreach (var parameter in publicCtor.GetParameters())
             {
-                // we here assume the parameter name is the same as the serialized name of the property. This is not 100% solid
-                var value = exampleValue.Properties[parameter.Name!];
+                ExampleValue value;
+                if (exampleValue.Properties.ContainsKey(parameter.Name!))
+                    // we here assume the parameter name is the same as the serialized name of the property. This is not 100% solid
+                    value = exampleValue.Properties[parameter.Name!];
+                else
+                {
+                    // if we can't find it directly, try to figure out the serailizedName form propertyMetadataDict
+                    string? serializedName = propertyMetadataDict.FirstOrDefault(p => string.Equals(p.Key, parameter.Name!, StringComparison.OrdinalIgnoreCase)).Value?.SerializedName;
+                    if (string.IsNullOrEmpty(serializedName) || !exampleValue.Properties.ContainsKey(serializedName))
+                        throw new InvalidOperationException($"Unabled to find example value for ctor parameter, type={type.FullName}, param={parameter.Name ?? "<null>"}");
+                    value = exampleValue.Properties[serializedName];
+                }
                 writer.AppendExampleValue(value, parameter.ParameterType);
                 writer.AppendRaw(",");
             }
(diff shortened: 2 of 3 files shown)


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -747,6 +747,13 @@ public override TypeProvider FindTypeProviderForSchema(Schema schema)
         public override CSharpType? FindTypeByName(string originalName)
         {
             _schemaOrNameToModels.TryGetValue(originalName, out TypeProvider? provider);
+
+            if (provider == null)
+            {
+                // Try to search declaration name too if no key matches. i.e. Resource Data Type will be appended a 'Data' in the name and won't be found through key
+                provider = _schemaOrNameToModels.FirstOrDefault(s => s.Value is MgmtObjectType mot && mot.Declaration.Name == originalName).Value;
```
ArcturusZhang, line 754: I did not get why we need this - this scenario "Resource Data Type will be appended a 'Data' in the name and won't be found through key" is handled on line 750 (before change)

src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -151,8 +152,18 @@ private static CodeWriter AppendComplexFrameworkTypeValue(this CodeWriter writer
...
+                    // we here assume the parameter name is the same as the serialized name of the property. This is not 100% solid
+                    value = exampleValue.Properties[parameter.Name!];
+                else
+                {
+                    // if we can't find it directly, try to figure out the serailizedName form propertyMetadataDict
+                    string? serializedName = propertyMetadataDict.FirstOrDefault(p => string.Equals(p.Key, parameter.Name!, StringComparison.OrdinalIgnoreCase)).Value?.SerializedName;
+                    if (string.IsNullOrEmpty(serializedName) || !exampleValue.Properties.ContainsKey(serializedName))
+                        throw new InvalidOperationException($"Unabled to find example value for ctor parameter, type={type.FullName}, param={parameter.Name ?? "<null>"}");
+                    value = exampleValue.Properties[serializedName];
+                }
```
ArcturusZhang, line 166: How about we just simplify this by using `TryGetValue`
```suggestion
                if (!exampleValue.Properties.TryGetValue(parameter.Name!, out var value))
                {
                    // if we can't find it directly, try to figure out the serailizedName form propertyMetadataDict
                    string? serializedName = propertyMetadataDict.FirstOrDefault(p => string.Equals(p.Key, parameter.Name!, StringComparison.OrdinalIgnoreCase)).Value?.SerializedName;
                    if (string.IsNullOrEmpty(serializedName) || !exampleValue.Properties.ContainsKey(serializedName))
                        throw new InvalidOperationException($"Unabled to find example value for ctor parameter, type={type.FullName}, param={parameter.Name ?? "<null>"}");
                    value = exampleValue.Properties[serializedName];
                }
```

ArcturusZhang: Also could we update the description to show in which real life case, would this KeyNotFoundException happen?
I kind of do not get why this could possibly happen and this is also the reason when I wrote this piece of code I did not add any extra checks.

Please keep in mind, if there is something wrong in the swagger example, we should not mitigate that here, instead we should open a PR in the swagger repo to fix the example json files.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
ArcturusZhang: You see two dictionaries here: `_schemaOrNameToModels` and `ResourceSchemaMap`
`ResourceSchemaMap` is a sub-dictionary from `AllSchemaMap`.
And `AllSchemaMap` is initialized [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs#L116) by this `InitializeModels` method, and this method is returning the `_schemaOrNameToModels` dictionary (there is an implicit operator taking effect during the return).
Therefore looking things up inside `_schemaOrNameToModels` is the same as looking things up in `AllSchemaMap`.

This is the reason why I wonder why we have to do this, because it is happening below.

But if there is another case of this happening, when the schema.Name is different from the name of the corresponding type provider, we should remove line 750. Now the only case of this is "ResourceData".

src/AutoRest.CSharp/MgmtTest/AutoRest/SourceCodeProject.cs
```diff
@@ -41,9 +42,11 @@ public async Task<Compilation> GetCompilationAsync()
         private static Project CreateSourceCodeProject(string sourceCodePath, string[] sharedSourceFolders)
         {
             var sourceCodeProject = CreateGeneratedCodeProject();
-            var sourceCodeGeneratedDirectory = Path.Join(sourceCodePath, GeneratedCodeWorkspace.GeneratedFolder);
+            DirectoryInfo di = new DirectoryInfo(sourceCodePath);
+            var genFolders = di.EnumerateDirectories(GeneratedCodeWorkspace.GeneratedFolder, SearchOption.AllDirectories).ToList();
```
ArcturusZhang, line 46: the `ToList` here seems redundant?

src/AutoRest.CSharp/MgmtTest/AutoRest/SourceCodeProject.cs
```diff
@@ -41,9 +42,11 @@ public async Task<Compilation> GetCompilationAsync()
...
 
-            sourceCodeProject = GeneratedCodeWorkspace.AddDirectory(sourceCodeProject, sourceCodePath, skipPredicate: sourceFile => sourceFile.StartsWith(sourceCodeGeneratedDirectory));
+            sourceCodeProject = GeneratedCodeWorkspace.AddDirectory(sourceCodeProject, sourceCodePath,
+                skipPredicate: sourceFile => genFolders.Exists(f => sourceFile.StartsWith(f.FullName)));
```
ArcturusZhang, line 49: Could we add some comments to describe the things going on here?

ArcturusZhang (approved):

src/AutoRest.CSharp/MgmtTest/AutoRest/SourceCodeProject.cs
RodgeFu: sure

src/AutoRest.CSharp/MgmtTest/AutoRest/SourceCodeProject.cs
RodgeFu: Delay init/iterate also brings higher troubleshooting/maintain cost and sometimes worse perf if used improperly, so it's better to use it only when there is a good reason.

Issue #3499, opened 2023-06-14
[Sample] The sample composer should never change the serialized name of a property


ArcturusZhang:
**Describe the issue or request**

https://github.com/Azure/autorest.csharp/blob/f8cbac6be058c5ac64e01f6d096d38ed11e21452/src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs#L1125

We should never change the serialized name of a property when building the request content in our samples.
Once we changed it even to make sure our generated sample code could compile, it makes the payload **incorrect** which is really misleading especially in a sample.

Fortunately this only happens when the property's serialized name collides with a keyword of dotnet, or it contains invalid identifier characters.

**Describe your ideas for solutions**

When building the request content, and when one above situation happens, we should switch to use a Dictionary as Request Content data source which takes string as keys and it always compiles.
Or if we want to keep consistency, we could either use Dictionary everywhere. But this might be harder to read.
<del>I am not sure if there is any attribute from System.Text.Json we could put on those name-changed property to ensure the serialization is correct.</del> unfortunately we cannot

Issue #3383, opened 2023-05-10
[Sample] Avoid var in generated samples


m-nash:
When we make this switch we need to be sure to use `Response<T>` and not T when writing the line for convenience method responses.

```c#
Response<Model> result = client.Method(...);
```

Issue #3337, opened 2023-04-30
[Doc] We need to make refactor to the xml doc generator


ArcturusZhang:
**Describe the issue or request**

Currently the code written in [LowLevelExampleComposer](https://github.com/Azure/autorest.csharp/blob/768a83a695500265d208e757d5ea0a6443ebfac2/src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs#L19) is purely concatenation of strings without any checks (rolsyn) or formatting (rolsyn) or automations (code writer) by our existing tool set. We need to refactor it so that we could remove a lot of duplicated code (like writing a type into code with proper name) and benefit from our existing tools.

PR #3558, opened 2023-07-03
Refactor the sample composer to use CodeWriter instead

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3383
Fixes https://github.com/Azure/autorest.csharp/issues/3337
Fixes https://github.com/Azure/autorest.csharp/issues/3499

# Description

Several classes like `InputClientExample`, `InputOperationExample`, etc are introduced to hold the example values for those types, preparing for the up-coming example generation.

For our current implementation, the advantage of these new types could:
1. decouple our logic when writing the samples. Previously we are writing code and in the meantime, get mock values for that type, therefore this kind of values do not have any context, and might be problematic when we have circles in the reference tree.
2. reduce the amount of duplicated code. Previously in our implementation, the samples for protocol methods and convenience methods both are mocking the values, and because of the type differences, we have to write similar code twice.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
```diff
@@ -67,5 +75,10 @@ public static async Task ExecuteAsync(GeneratedCodeWorkspace project, InputNames
                 modelFactoryFullName: modelFactoryProvider?.FullName,
                 aspExtensionClassName: library.AspDotNetExtension.FullName));
         }
+
+        private class XmlStringWriter : StringWriter
+        {
+            public override Encoding Encoding => Encoding.UTF8;
+        }
```
ArcturusZhang, line 82: Should remove

src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs
```diff
@@ -300,6 +299,32 @@ private void ComposeCodeSnippet(ConvenienceMethod convenienceMethod, string meth
...
+                case (null, not null):
+                    ComposeHandlePageableResponseCode(codeWriter, clientMethod, methodName, isAsync, useAllParameters);
+                    break;
+                case (null, null):
+                    ComposeHandleNormalResponseCode(codeWriter, clientMethod, methodName, isAsync, useAllParameters);
+                    break;
+            }
+        }
+
         internal void ComposeCodeSnippet(LowLevelClientMethod clientMethod, string methodName, bool async, bool allParameters, StringBuilder builder)
```
ArcturusZhang, line 328: These methods using StringBuilder should be removed soon

PR #3576, opened 2023-07-10
Use parameter type instead of name to check protocol method

### LINKED ISSUE #3361: [DPG] should use type of the parameter instead of parameter name to determine whether a method is protocol method
When we are dealing with the existing contract, we need to know whether the method is a protocol method, we are using the name of the parameter: https://github.com/Azure/autorest.csharp/blob/d835b0b7bffae08c1037ccc5824e928eaac55b96/src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs#L96

But we should use the type and the name of the parameter to make sure this method is protocol method (first parameter is RequestContent, second is RequestContext)

pshao25:
Fix #3361


src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
```diff
@@ -93,6 +94,9 @@ private protected List<IMethodSymbol> EnsureMethodSet()
         }
 
         private bool IsClient(INamedTypeSymbol type) => type.Name.EndsWith("Client");
-        private bool IsProtocolMethod(IMethodSymbol method) => method.Parameters.Length > 0 && method.Parameters.Last().Name == "context";
+        private bool IsProtocolMethod(IMethodSymbol method) =>
+            method.Parameters.Length > 0
+            && ((method.Parameters.Last().Type as INamedTypeSymbol)!.IsSameType(KnownParameters.RequestContext.Type)
+            || (method.Parameters.Last().Type as INamedTypeSymbol)!.IsSameType(KnownParameters.RequestContextRequired.Type));
```
ArcturusZhang, line 100: shall we have a local variable to get the last parameter type?

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Input/Source/ProtocolCompilationInput.cs
pshao25: What is the value? `method.Parameters.Last()` is an `O(1)` operation.

lirenhe (approved):

PR #3584, opened 2023-07-13
feat(common): support oasis repeatability headers

### LINKED ISSUE #1950: Support for OASIS Request Repeatability Headers

The Azure API Guidelines say that all operations should be idempotent. This is mainly an issue for POST operations. For situations where a POST needs to be made idempotent, the guidelines recommend

☑️ YOU SHOULD support repeatable requests according as defined in [OASIS Repeatable Requests Version 1.0](https://docs.oasis-open.org/odata/repeatable-requests/v1.0/repeatable-requests-v1.0.html).

Some teams have begun following this advice and including the OASIS Request Repeatability headers in their REST API.

This raises the question of how the Azure SDKs should support OASIS Repeatability headers. 

 I think we should treat these headers specially -- meaning _not_ simply expose them as optional parameters that are set by the caller.  Instead, I think the SDK should set these headers within the method for any operation that supports them.

The headers involved are `Repeatability-Request-ID`, which the SDK should set to a freshly minted UUID, and `Repeatability-First-Sent`, which should be set to the current date/time.

The csharp generator should check for the presence of these headers and produce the additional logic as described here.

**Affected Service:**
Batch

archerzz:
# Description

- remove repeatability headers from the parameters of generated methods
- in create request methods, set the value of repeatability headers
- update sample/test projects
- add test cases
- minor refactoring

resolve #1950

# Checklist

# Ready to Land?


test/TestProjects/FirstTest-TypeSpec/src/Generated/FirstTestTypeSpecClient.cs
```diff
@@ -534,18 +534,17 @@ public virtual Response FriendlyModel(RequestContent content, RequestContext con
...
         /// </list>
         /// </summary>
-        /// <param name="repeatabilityFirstSent"> The DateTime to use. </param>
         /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
         /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
         /// <returns> The response returned from the service. </returns>
-        /// <include file="Docs/FirstTestTypeSpecClient.xml" path="doc/members/member[@name='AddTimeHeaderAsync(DateTimeOffset?,RequestContext)']/*" />
-        public virtual async Task<Response> AddTimeHeaderAsync(DateTimeOffset? repeatabilityFirstSent = null, RequestContext context = null)
+        /// <include file="Docs/FirstTestTypeSpecClient.xml" path="doc/members/member[@name='AddTimeHeaderAsync(RequestContext)']/*" />
+        public virtual async Task<Response> AddTimeHeaderAsync(RequestContext context = null)
```
archerzz, line 541: One thing to note that if only one of the headers is defined, we'll also remove it from the method parameters. Reasons are:
- Those are headers, not request content or query parameters. Service should not define their own headers with the same name.
- In some cases, defining only one header makes sense. For example, if the service doesn't care about the time, then only providing `Repeatability-Request-ID` is OK.
- Here are just test projects, not real ones.

From what I know, Java generator adopts the same strategy.

archerzz:
We have some problems in existing test recordings. I've filed an issue on `Azure.Core.TestFramework`: https://github.com/Azure/azure-sdk-for-net/issues/37587

archerzz:
Regen preview PR: https://github.com/Azure/azure-sdk-for-net/pull/37584

lirenhe (commented):
There is a special header test of  repeatability in cadl-ranch.
Could you also implement that test case in this PR?

archerzz:
> There is a special header test of repeatability in cadl-ranch. Could you also implement that test case in this PR?

@lirenhe Theat case spec has already been added into our `CadlRanchProject` (e.g. `TraitClient`). I've implemented the test in https://github.com/Azure/autorest.csharp/pull/3584/files#diff-a8fcae87ba63919556add5e1f2a6c079bfc220e816aa4177845ccabc0a1e8ad9

src/AutoRest.CSharp/Common/Output/Models/Shared/Constant.cs
```diff
@@ -52,7 +52,23 @@ public static Constant NewInstanceOf(CSharpType type)
             return new Constant(NewInstanceSentinel, type);
         }
 
+        public static Constant ExpressionOf(FormattableString expression, CSharpType type) => new Constant(new Constant.Expression(expression), type);
```
chunyu3, line 55:
```suggestion
        public static Constant FromExpression(FormattableString expression, CSharpType type) => new Constant(new Constant.Expression(expression), type);
```

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -79,6 +80,18 @@ public static Parameter FromInputParameter(in InputParameter operationParameter,
                 RequestLocation: requestLocation);
         }
 
+        private static Constant? GetDefaultValue(InputParameter operationParameter, TypeFactory typeFactory) => operationParameter switch
+        {
+            { DefaultValue: not null } => BuilderHelpers.ParseConstant(operationParameter.DefaultValue.Value, typeFactory.CreateType(operationParameter.DefaultValue.Type)),
+            { NameInRequest: var nameInRequest } when nameInRequest.Equals(RequestHeader.RepeatabilityRequestId, StringComparison.OrdinalIgnoreCase) =>
```
chunyu3, line 86: If the Repeatability parameter has a default value, it will not enter this option. Maybe we need to put repeatability parameter at the top.

archerzz: > If the Repeatability parameter has a default value, it will not enter this option. Maybe we need to put repeatability parameter at the top.

I think that's a wrong usage. People should not set a default value for repeatability headers. The value **must be different** per call. Think of the following scenario:
- The first time we call `Foo()`. `Repeatability-Request-ID` is a default value (`abc`).
- The server side will remember `abc`.
- The next time we call `Foo()` with the same header value. Server will treat it as a retry, and will not do anything.

chunyu3: Yes. It is rare to set default value for repeatability. But according to the code, it may have this problem.

archerzz: > Yes. It is rare to set default value for repeatability. But according to the code, it may have this problem.

If we set the default value, the repeatability request feature will not work. See [spec 3.1.1 Repeatability-Request-ID](https://docs.oasis-open.org/odata/repeatable-requests/v1.0/cs01/repeatable-requests-v1.0-cs01.html#sec_RepeatabilityRequestID):
> The value of the Repeatability-Request-ID is an opaque string representing a client-generated, globally unique for all time, identifier for the request.

The value must be **globally unique** to mark every request. We should not set the same value for different requests.

chunyu3: What I mean is that to consider the function itself (ignoring the backend limitation of the repeatability headers ), the first option `{ DefaultValue: not null }` may meet first and then following options to handle repeatability headers may not executed.
So it may be better to put special options head of the more general one.

chunyu3 (approved):

chunyu3 (changes requested):
Please implement cadl-ranch test `special-headers\repeatability`

chunyu3 (approved):

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
m-nash: @archerzz can you open an issue in typespec repo to add a linter rule for never setting a default value for these headers?

m-nash (approved):

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
archerzz: I tend to treat this as an error in original spec. Better to generate the wrong codes so that the spec owner can find the problem. Another way to prevent it is to add a linter rule.

Issue #3537, opened 2023-06-28
CodeGen doesn't support common types in v4 & v5


ArthurMa1978:
For the new common types [v4](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v4/types.json) and [v5](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v5/types.json), some properties were specified the format, which caused the CodeGen can't indicate these types as common types.

ArthurMa1978:
Here is a sample https://github.com/Azure/azure-sdk-for-net/pull/37213 which have to use directive to mitigate

live1206:
Done

PR #3622, opened 2023-07-25
Fix format check for ResourceId

live1206:
Part of #3537

### DIFF
--- a/src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
+++ b/src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
@@ -75,7 +75,7 @@ internal static class SchemaExtensions
             switch (property.SerializedName)
             {
                 case "id":
-                    if (property.Schema.Type == AllSchemaTypes.String)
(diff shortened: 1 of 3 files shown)


src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
```diff
@@ -75,7 +75,7 @@ public static bool IsResourceModel(this Schema schema)
             switch (property.SerializedName)
             {
                 case "id":
-                    if (property.Schema.Type == AllSchemaTypes.String)
+                    if (property.Schema.Type == AllSchemaTypes.String || property.Schema.Type == AllSchemaTypes.ArmId)
```
ArthurMa1978, line 78: Please add test for this change

ArthurMa1978 (approved):

src/AutoRest.CSharp/Mgmt/Decorator/SchemaExtensions.cs
live1206: Good point, updated the existing test to verify this scenario.

PR #3633, opened 2023-08-02
fix(dpg): same type property only mock once

### LINKED ISSUE #3632: [bug] property with the same type will not be mocked in generated sample codes
## Problem

In current example composer algorithm, we will only mock value for a type for once. That's not correct. If a type is used in different path in the object hierarchy (e.g. not a loop), we should mock value for it. For example, in `Batch`, the generated sample codes could be like:

```csharp
            var task = new BatchTaskCreateParameters("<id>", "<commandLine>")
            {
                DisplayName = "<DisplayName>",
                ExitConditions = new ExitConditions()
                {
                    ExitCodes =
                    {
                        new ExitCodeMapping(1234, new ExitOptions()
                        {
                            JobAction = JobAction.None,
                            DependencyAction = DependencyAction.Satisfy,
                        })
                    },
                    ExitCodeRanges =
                    {
                        new ExitCodeRangeMapping(1234, 1234, ) // <= error!
```

Note the last line, the missing value should be `new ExitOptions()` which has been mocked in lines above.

archerzz:
# Description

- fix DFS logic to remove visited type from hashset after being traversed
- add test case

fix #3632

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Utilities/HashSetExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                HashSet = hashSet;
+                ItemToRemove = item;
+            }
+
+            public void Dispose()
+            {
+                HashSet.Remove(ItemToRemove);
+            }
+        }
```
ArcturusZhang, line 46: Looks like the static factory method used above supersedes this class, maybe we could remove it.

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Utilities/HashSetExtensions.cs
archerzz: Yes. I forgot to delete it. Was testing whether the `Disposable.Create()` could fulfill our need.

archerzz:
Preview PR: https://github.com/Azure/azure-sdk-for-net/pull/38000

As you can see, some missing mock values are added.

lirenhe (approved):

PR #3636, opened 2023-08-04
Head as boolean

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3456

- Add emitter option `head-as-boolean`

# Checklist

# Ready to Land?


test/TestProjects/FirstTest-TypeSpec/tests/Generated/Samples/Samples_FirstTestTypeSpecClient.cs
```diff
@@ -760,6 +760,50 @@ public async Task Example_StringFormat_Convenience_Async()
             var result = await client.StringFormatAsync(Guid.NewGuid(), body);
         }
 
+        [Test]
+        [Ignore("Only validating compilation of examples")]
+        public void Example_HeadAsBoolean()
```
lirenhe, line 765: Do we have any cadl-ranch test cases?  We should verify whether this API do return true/false.

chunyu3: Yes. I am working on add cadl-ranch test cases.

eng/Generate.ps1
```diff
@@ -257,6 +260,8 @@ if (!($Exclude -contains "CadlRanchProjects")) {
     }
 }
 
+# add Head-As-Boolen project
+Add-CadlRanch-TypeSpec "server/path/single" "headAsBoolean-typespec-" $cadlRanchProjectDirectory "$cadlRanchProjectDirectory/server/path/single-headAsBoolean"
```
ArcturusZhang, line 264: if we want to add new cadl-ranch cases, we just need to add the path of the case into the `testProjects.json` instead of changing the script.
If the cadl-ranch case is not ready, we should also avoid this change. Instead, you could introduce a regular `TestProject`, and open another PR to introduce the cadl ranch case when the cadl-ranch PR is merged.
In your case, since we have something tested in `FirstTest-TypeSpec`, if the cadl-ranch case is not ready, we could omit it for now by revert the changes here and add it in another PR by adding a new path into `testProjects.json`

test/TestProjects/FirstTest-TypeSpec/src/Generated/FirstTestTypeSpecClient.cs
```diff
@@ -678,6 +678,76 @@ public virtual Response StringFormat(Guid subscriptionId, RequestContent content
...
+        /// </list>
+        /// </summary>
+        /// <param name="id"> The String to use. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="id"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="id"/> is an empty string, and was expected to be non-empty. </exception>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns> The response returned from the service. </returns>
+        /// <include file="Docs/FirstTestTypeSpecClient.xml" path="doc/members/member[@name='HeadAsBooleanAsync(string,RequestContext)']/*" />
+        public virtual async Task<Response<bool>> HeadAsBooleanAsync(string id, RequestContext context = null)
```
ArcturusZhang, line 698: Is this design reviewed by anyone?
Currently, in our protocol methods, no matter protocol method returns or does not return a body, we always use `Response` as return type, we never have `Response<BinaryData>`.
In this case, it is similar - if we follow the pattern, we should return Response in protocols and return `Response<bool>` in convenience maybe?

eng/Generate.ps1
chunyu3: Yes. There is no cadl-ranch project for head operation. And we can use this existing project to test head-as-boolean as other language. After cadl-ranch add the special test, we will remove this.

test/TestProjects/FirstTest-TypeSpec/src/Generated/FirstTestTypeSpecClient.cs
chunyu3: @ArcturusZhang It is consistent with previous design. see `HeadAsBooleanTrue-LowLevel`. We don't change it.

ArcturusZhang (approved):

lirenhe (approved):

Issue #3634, opened 2023-08-04
[Defect] The usages of child models are not calculated when the base model is used in Azure core operation template


chunyu3:
When a base model is used as response in Azure Core operation template, the usages of its child models are not correct calculated.

e.g. CLU  https://github.com/Azure/azure-rest-api-specs/blob/787a5bf853a7be902041e9793fb50791d167a483/specification/cognitiveservices/Language.Conversations/main.tsp#L1075-L1079
```
op analyzeConversation is Azure.Core.StandardResourceOperations.ResourceCollectionAction<
  Conversations,
  AnalyzeConversationTask,
  AnalyzeConversationTaskResult
>;
```

PR #3642, opened 2023-08-07
Fix3634

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3634
Fix https://github.com/Azure/autorest.csharp/issues/3649
Problem:
When the return type is base model in Azure.Core operation template, the usage of the derived model is not correct. It does not contain `Output` usage.

Root cause:
When the return type is base model in Azure.Core operation template, the tsp compiler will compile it as anonymous model, so resolveUsage API will not set correct Usage for this model.
emitter will re-calculate those model's usage, but it is not propagated to its descendants models. 

Solution:
propagated usage of the base model to its descendants models in emitter.
# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -788,21 +791,24 @@ export function getUsages(
...
+            if (effectiveType.kind === "Model") {
+                if (effectiveType.templateMapper?.args) {
+                    for (const arg of effectiveType.templateMapper.args) {
+                        if (
+                            arg.kind === "Model" &&
+                            "name" in arg &&
+                            arg.name !== ""
+                        ) {
+                            affectTypes.push(
+                                getFriendlyName(program, arg) ?? arg.name
```
ArcturusZhang, line 803: is there a method in TCGC that could replace the getFriendlyName method here from the compiler?
And I asusme the TCGC common interface refactor could replace all of these, therefore I think it is OK if we just use this as a temporary solution and defer to the common interface refactor

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/usage.test.ts
```diff
@@ -253,4 +253,126 @@ describe("Test getUsages", () => {
         const usages = getUsages(sdkContext, services[0].operations);
         assert(usages.roundTrips.includes("Foo"));
     });
+
+    it("Test the usage of body polymorphism type in azure core resource operation.", async () => {
+        const program = await typeSpecCompile(
```
ArcturusZhang, line 258: Could we also add a test case that is using the enum-typed discriminator since this is one of the issue your PR trying to fix?

chunyu3: yes. I will add test for polymorphism scenario UT. Thanks @ArcturusZhang

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: Yes. TCGC will handle all the Naming and Usage of a model. Those logic will remove after we adopt TCGC common interface.

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/usage.test.ts
chunyu3: @ArcturusZhang  Add discriminator Unit Test in Emitter.

ArcturusZhang (approved):

lirenhe (approved):

Issue #3581, opened 2023-07-12
Support property with type of union of models in the response


lirenhe:
Here is the typespec defined by service team (reviewed and merged)
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/OpenAI.Inference/models/images.tsp#L85
```
model ImageGenerations {
  @doc("A timestamp when this job or item was created (in unix epochs).")
  @projectedName("csharp", "InternalCreatedSecondsAfterUnixEpoch")
  created: int64;

  #suppress "@azure-tools/typespec-autorest/union-unsupported" "openapi v2 not required"
  @doc("The images generated by the operator.")
  @projectedName("csharp", "InternalEmittedImageResponseItems")
  data: ImageLocation[] | ImagePayload[];
}

```
Confirmed this is a valid case from typespec prospective.

Currently, dotnet CodeGen will report the following error: 

net/sdk/openai/Azure.AI.OpenAI/src/Generated/ImageGenerations.Serialization.cs(34,43): error CS1001: Identifier expected [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/openai/Azure.AI.OpenAI/src/Azure.AI.OpenAI.csproj::TargetFramework=netstandard2.0]

Please check the raw file for more details.

https://github.com/azure-sdk/azure-sdk-for-net/pull/4203/files#diff-ce13784db9766dadcd61ef73512e49a91e8fd8369516241e4bd738447fee0c56

We should find a solution for supporting property with type of union of models in the response in DPG.

Issue #3557, opened 2023-07-03
`DerivedModels` in `InputModelType` is never assigned and never used properly anywhere


ArcturusZhang:
**Describe the issue or request**

We defined `DerivedModels` property on `InputModelType`, but it is never assigned anywhere.
It is used here: https://github.com/Azure/autorest.csharp/blob/bfb7beaa8a9d136e608f11ce259660561c79146d/src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs#L1225 and it is incorrect.

Issue #3356, opened 2023-05-05
[DPG] Generator confidence levels


ArcturusZhang:
We would like to introduce a new concept into our generator about the confidence levels.

We call a thing is "low confident" when the emitter or our generator has to make up a name for it.
We call a thing is "high confident" when it already has its name properly.

When the generator is about to generate the convenience method, if anything inside that convenience method is "low confident", the generator should only generate a protocol method and skip the convenience method automatically.

For instance, if we have this typespec:
```
model Foo {
   value: string | int;
}
```
This corresponds to a model with the name `Foo`, it has one required property `Value`, and the type of property `Value` is a union type, but the typespec did not specify the name of the union type (depending on how we implements union types, but either way, it needs an extra name), therefore the emitter or generator has to make up a name, therefore the type of this property should have the flag "low confident" and the convenience method using this model (directly or indirectly) should not be generated.

ArcturusZhang:
All the types we currently have:

- InputPrimitiveType: primitive type always has high confidence
- InputLiteralType: literal type may be low confident
- InputUnionType: union types always have low confidence (for now, since we did not really implement it yet)
- InputModelType: model types may be low confident
- InputEnumType: enum types may be low confident
- InputListType: list type itself always has high confidence, but the element may have low confidence, this is handled by ElementType
- InputDictionaryType: dictionary type itself always has high confidence, but the key type or value type may have low confidence, this is handled by KeyType and ValueType individually
- InputIntrinsicType: intrinsic type always has high confidence

PR #3643, opened 2023-08-07
DPG confidence level mechanism

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3557
Fixes https://github.com/Azure/autorest.csharp/issues/3356
Fixes https://github.com/Azure/autorest.csharp/issues/3581 by preventing the generator to generate convenience method for it.

Supersedes https://github.com/Azure/autorest.csharp/pull/3354

# Description

For all the types we currently have:
- `InputLiteralType`: literal types are not confident when the underlying type is numbers **with names that are not valid C# identifiers**, which causes its underlying `InputEnumType` to be not confident
- `InputUnionType`: union types are always not confident (for now, there might be some exceptions for this but we did not implemented yet)
- `InputModelType`: model types are not confident when any type of its properties is not confident
- `InputEnumType`: `InputEnumType` should be confident, except when it is used by a `InputLiteralType` (see InputLiteralType for criteria)
- `InputListType` and `InputDictionaryType`: these container types do not have its own confident levels, or it should stay the same confident level as its `ElementType` or its `ValueType`
- Other types are always confident

In this PR, when building the ConvenienceMethod inside the `OperationMethodChainBuilder` class, we will check if the operation is confident for generating convenience method by checking its parameter types and response types. If any of them is not confident, we will short circuit and hide the corresponding convenience method by making the convenience method internal.
In the meantime, a record for reasons of making convenience method internal is created and the writer will write a message in comment to avoid surprises.

This PR also changes the generated type of unions to `object`.
Previously the union types in the model will generate from its first type, for instance, if a union is `string | int32`, in the generated code, we will have `string`. This PR changes this behavior to generate them as `object`.
Since it is `object`, when serializing, it should work properly depending on what type the customer assigns. Some of them are unsupported and exceptions will be thrown.
When deserializing, our SDK has the functionality to properly deserialize it to its correct type based on the value in json. If it is primitive type, a primitive type will return. If it is an array, an array will return. If it is a complex object or dictionary, a `Dictionary<string, object>` will return.

This PR also introduces a new test projects and corresponding test cases to verify the generated content is expected to have correct accessibility.

Considering the first part of this change will make the operation with unions internal, and the post processor will automatically make the models in use of the operation internal as well, therefore the change to type of object should be acceptable.
In real life, if a service team wants to introduce SDKs with union types, the operation involved with unions will not be public. They need to do either of the following:
1. write their own method to make it public, and customize the corresponding properties and the related serializations (use `CodeGenMemberSerializationHooks` for instance)
2. specify in the client teaspoon so that the union disappears.

# Checklist

# Ready to Land?


ArcturusZhang:
@m-nash could you please have a review on this PR?
Now we are detecting the low confident content inside the generator, and also I introduced a cache so that information will not be calculated multiple times.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputLiteralTypeConverter.cs
```diff
@@ -27,14 +27,14 @@ public static InputLiteralType CreateInputLiteralType(ref Utf8JsonReader reader,
...
             var isFirstProperty = id == null && name == null;
             bool isNullable = false;
-            Object? value = null;
+            object? value = null;
             InputType? type = null;
 
             while (reader.TokenType != JsonTokenType.EndObject)
             {
                 var isKnownProperty = reader.TryReadReferenceId(ref isFirstProperty, ref id)
-                    || reader.TryReadString(nameof(InputListType.Name), ref name)
```
archerzz, line 36: Is this a typo before?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs
```diff
@@ -87,7 +87,7 @@ public static InputPrimitiveType ReadPrimitiveType(ref Utf8JsonReader reader, st
             while (reader.TokenType != JsonTokenType.EndObject)
             {
                 var isKnownProperty = reader.TryReadReferenceId(ref isFirstProperty, ref id)
-                    || reader.TryReadBoolean(nameof(InputType.IsNullable), ref isNullable)
+                    || reader.TryReadBoolean(nameof(InputPrimitiveType.IsNullable), ref isNullable)
```
archerzz, line 90: To make it more specific?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -544,6 +546,15 @@ private void WriteProtocolMethodDocumentationWithExternalXmlDoc(LowLevelClientMe
             }
         }
 
+        private void WriteConvenienceMethodOmitReason(ConvenienceMethodOmittingMessage? message)
```
archerzz, line 549:
```suggestion
        private void WriteConvenienceMethodOmitReasonIfNecessary(ConvenienceMethodOmittingMessage? message)
```

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -628,7 +639,8 @@ private static FormattableString BuildProtocolMethodSummary(MethodSignature meth
             builder.AppendLine($"<list type=\"bullet\">");
             builder.AppendLine($"<item>{Environment.NewLine}<description>{Environment.NewLine}This <see href=\"https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/ProtocolMethods.md\">protocol method</see> allows explicit creation of the request and processing of the response for advanced scenarios.{Environment.NewLine}</description>{Environment.NewLine}</item>");
 
-            if (clientMethod.ConvenienceMethod != null)
+            // we only append the relative convenience method information when the convenience method is public
+            if (clientMethod.ConvenienceMethod != null && clientMethod.ConvenienceMethod.Signature.Modifiers.HasFlag(MethodSignatureModifiers.Public))
```
archerzz, line 643:
```suggestion
            if (clientMethod.ConvenienceMethod?.Signature.Modifiers.HasFlag(MethodSignatureModifiers.Public) is true)
```

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
```diff
@@ -0,0 +1,143 @@
...
+namespace AutoRest.CSharp.Output.Models
+{
+    internal static class OperationConfidentChecker
+    {
+        private static ConcurrentDictionary<InputType, bool> _cache = new();
+
+        public static bool IsConfident(InputOperation operation)
+        {
+            // check parameters
+            foreach (var parameter in operation.Parameters)
```
archerzz, line 21:
```suggestion
            if (operation.Parmeters.Where(p => !p.IsContentType && !p.IsApiVersion && !p.IsEndpoint).Any(p => !IsConfident(p.Type)))
            {
                return false;
            }
```

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
```diff
@@ -0,0 +1,143 @@
...
+                // skip the special parameters since some particular types are still interpreted as unions (for instance content type could have union of literal types which is actually enums)
+                if (parameter.IsContentType || parameter.IsApiVersion || parameter.IsEndpoint)
+                    continue;
+
+                var isConfident = IsConfident(parameter.Type);
+                if (!isConfident)
+                    return false;
+            }
+            // check response
+            foreach (var response in operation.Responses)
```
archerzz, line 32:
```suggestion
            if (operations.Responses.Any(r => !IsConfident(r.BodyType)))
            {
                return false;
            }
```

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        private static bool IsConfident(InputType? type)
+        {
+            if (type == null)
+                return true;
+
+            if (_cache.TryGetValue(type, out var result))
+                return result;
+
+            var visitedModels = new Dictionary<InputModelType, bool?>();
```
archerzz, line 51: Why not `HashSet<InputModelType>`?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputLiteralTypeConverter.cs
ArcturusZhang: Yes

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs
ArcturusZhang: Yes and to unify

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
ArcturusZhang: This is not only for "visited models", it is also storages the result of those visited models.
We cannot get the result for one type and just put it into the global cache because sometimes non-confident thing pollutes the entire tree.
I write a test case for this
```
@discriminator("kind")
model Pet {}

model Cat extends Pet { kind: "cat"; }

model Dog extends Pet { kind: "dog"; unionProperty: string | int32; } // <- this type pollutes the entire tree and we need to backport Cat and Pet to be not confident as well
```

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
ArcturusZhang: For readability and the ability to clearly debugging this, I do not want to change it using Linq here.

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
archerzz: I see. Another comment. It looks like `IsConfident()` could be invoked many times, and some calculation is duplicated. Maybe we should change `visitedModels` as a static one?

ArcturusZhang: No. the idea here is that the global dictionary `_cache` is the final result cache. Things go into that will never be calculated again.
`visitedModels` here is a local cache to cache everything being calculated or calculated in this call.
I do not combine this local cache with global cache because the result of one model is not finalized until the tree has been go over.

I will add a read cache at the beginning of `WalkType` so that if the type has been calculated in previous calls of `IsConfident`, we will not call it again.

ArcturusZhang:
@m-nash could you take a look at this PR?

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
```diff
@@ -0,0 +1,142 @@
...
+            // the low confident part in derived types will pollute it back to the base class - if any derived class has low confident part (like union types), we will set the base as "not confident", but other derived types are not affected.
+            if (type.DiscriminatorPropertyName != null && type.DerivedModels.Count > 0)
+            {
+                foreach (var dm in type.DerivedModels)
+                {
+                    isConfident = isConfident && WalkType(dm, visitedModels);
+                }
+
+                // if one of the derived types is not confident, we should make all of them not confident
+                if (!isConfident)
```
chunyu3, line 106: the confident of a model may not affect its brother model.  The derived model may use separated in an operation as input or output.

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethodOmittingMessage.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using AutoRest.CSharp.Common.Input;
+
+namespace AutoRest.CSharp.Output.Models
+{
+    internal record ConvenienceMethodOmittingMessage
```
lirenhe, line 9: Seems this class only provide some static string message, is it possible to combine it with  ConvenienceMethodGenerationInfo class?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
```diff
@@ -148,17 +150,5 @@ private static IReadOnlyList<InputEnumTypeValue> NormalizeValues(IReadOnlyList<I
 
             return concreteValues;
         }
-
-        //private static IReadOnlyList<InputEnumTypeValue> convertToConcreteValues<EnumType, ValueType>(IReadOnlyList<InputEnumTypeValue> allowedValues) where EnumType : InputEnumTypeValue
```
m-nash, line 152: should this be removed?

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
```diff
@@ -544,6 +546,15 @@ private void WriteProtocolMethodDocumentationWithExternalXmlDoc(LowLevelClientMe
             }
         }
 
+        private void WriteConvenienceMethodOmitReasonIfNecessary(ConvenienceMethodOmittingMessage? message)
+        {
+            // TODO -- create wiki links to provide guidance here: https://github.com/Azure/autorest.csharp/issues/3624
+            if (message == null)
+                return;
+
+            _writer.Line($"// {message.Message}");
```
m-nash, line 555: I am pretty sure this is the case, but can we verify that this won't show up in docs / intellisense?  If it does move it to the first line of the method.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -93,20 +93,52 @@ public LowLevelClientMethod BuildOperationMethodChain()
...
         }
 
-        private bool ShouldGenerateConvenienceMethod()
+        private ConvenienceMethodGenerationInfo ShouldGenerateConvenienceMethod()
         {
-            return Operation.GenerateConvenienceMethod
-                && (!Operation.GenerateProtocolMethod
-                || IsConvenienceMethodMeaningful());
+            // we do not generate convenience method if the emitter does not allow it.
+            if (!Operation.GenerateConvenienceMethod)
```
m-nash, line 109: This might be a future follow up item, but I would like to get to a place where a client has a list of methods, and each method is either a protocol or a convenience method.  Versus today we have all convenience methods linked off of their protocol methods.  This won't work well if we ever omit the protocol method.

If we don't end up doing this as part of this PR please link the tracking issue in this comment for context.

m-nash:
One additional comment in the regen preview https://github.com/Azure/azure-sdk-for-net/pull/38181#discussion_r1294882249

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
ArcturusZhang: No, it affects others exactly because of the case you think of. Take this case:
```
@discriminator("kind")
model Pet {}

model Dog extends Pet { kind: "dog", union: string | int32; } // not confident

model Cat extends Pet { kind: "cat" } // confident
```
Dog is not confident, therefore Pet is not confident. **When it is not confident, we should make it internal.**
If Pet is internal, Cat must be internal. Otherwise you get this error from compiler: Inconsistent accessibility: base class 'Pet' is less accessible than class 'Cat'
Therefore we let the low confident to pollute the entire tree to ensure this. Consider this (exactly the case you said in your comment):
```
// the three models see above
op op1(@body body: Pet): void;

op op2(@body body: Cat): void;
```
The first operation `op1` is internal because `Pet` is not confident.
The second operation `op2` must be internal as well, because `Cat` is internal (see reason above), if `Cat` is internal, the operation must be internal as well.
This accessibility change does not come from "whether they are used or not", therefore this is not a work that the post processor should do. This is because the `Pet` is not confident - which is caused by the fact that `Dog` is not confident, which means the "not confident" part will pollute the whole tree.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
ArcturusZhang: I saw it is commented and never used, I just removed it.
Do we still need it in the future? I could revert this removal

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
ArcturusZhang: I believe this PR is not a good place to do this. I will open up an issue for this.

ArcturusZhang: I regard this as a refactor work item: https://github.com/Azure/autorest.csharp/issues/3665

src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs
ArcturusZhang: It does not show up in intellisense:
image
Therefore I assume it will not show up in docs neither

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
chunyu3: So it is because of `internal`.  In that case, when base is not confident, all his descendants should be non-confident and generate as internal No matter it is follow polymorphism or not.

And now we enlarge the meaning of `confident` is to the model can be generated and generated public.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -531,6 +523,7 @@ export function getInputType(
         if (!model) {
             const baseModel = getInputModelBaseType(m.baseModel);
             const properties: InputModelProperty[] = [];
+            const derivedModels: InputModelType[] = [];
```
chunyu3, line 526: `DerivedModels` is optional, can we not default set it as empty array? If no derived models, set `DerivedModels` undefined.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -556,12 +550,13 @@ export function getInputType(
...
             if (model.DiscriminatorPropertyName && m.derivedModels) {
                 for (const dm of m.derivedModels) {
-                    getInputType(
+                    const derivedModel = getInputType(
                         context,
                         getFormattedType(program, dm),
                         models,
                         enums
                     );
+                    derivedModels.push(derivedModel as InputModelType);
```
chunyu3, line 559: If you update `derivedModels` here, only polymorphism with discriminator will add the derived models, other nomal inheritance will missing.
We should add dervicedModels for base model of all kinds of inheritance.

ArcturusZhang: I fixed this and add a few new test cases

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethodOmittingMessage.cs
ArcturusZhang: The info is for internally passing information around when building those methods, this message is the final generated type, and there is an issue tracking the things we want to add into this message (wiki links, faqs, etc)

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
ArcturusZhang: this is fixed

chunyu3 (approved):

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
```diff
@@ -0,0 +1,142 @@
...
+
+            var isConfident = type.BaseModel != null ? WalkType(type.BaseModel, visitedModels) : true;
+            foreach (var property in type.Properties)
+            {
+                if (!isConfident)
+                    break;
+                isConfident = isConfident && WalkType(property.Type, visitedModels);
+            }
+
+            // the low confident part in derived types will pollute it back to the base class - if any derived class has low confident part (like union types), we will set the base as "not confident", but other derived types are not affected.
```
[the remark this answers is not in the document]
ArcturusZhang: Actually this comment is not correct any more
low confident stuff should pollute the whole tree or impact the whole tree.
I will fix this.

lirenhe (approved):

Issue #3544, opened 2023-06-29
DPG:  Handle Non-standard special headers for client-Request-Id and return-Client-request-Id


lirenhe:
Based on the discussion:  Azure.Core has a policy to append `return-client-resquest-id` and `client-request-id`. 
So Codegen will automatically add this policy to pipeline in client constructors and remove those two head parameters from method signature.

Needed for Batch service

m-nash:
Needed for Batch

PR #3656, opened 2023-08-11
set Client request id inline

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3544

- For special head `client-request-id` and `return-client-request-client-id`, we will emit it from method signature and add directly in CreateXXXMethod

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Requests/RequestHeader.cs
```diff
@@ -32,9 +32,20 @@ internal class RequestHeader
...
+        {
+            "client-request-id",
+        };
+
+        public static HashSet<string> ReturnClientRequestIdResponseHeaders = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            "return-client-request-id",
+        };
+
+        public static bool IsClientRequestIdHeader(string headerName) => ClientRequestIdHeaders.Contains(headerName) || ReturnClientRequestIdResponseHeaders.Contains(headerName);
```
archerzz, line 47: If using `.Contains()`, then we can just use one `HashSet` to host both `client-request-id` and `return-client-request-id`.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -228,6 +230,14 @@ static string AddAllowedValues(string description, ICollection<ChoiceValue> choi
 
         private static Constant? ParseConstant(RequestParameter parameter, TypeFactory typeFactory)
         {
+            if (parameter.In == HttpParameterIn.Header && RequestHeader.ClientRequestIdHeaders.Contains(parameter.Language.Default.SerializedName ?? parameter.Language.Default.Name))
```
archerzz, line 233: Is there a case when `SerializedName` is empty?

test/AutoRest.TestServerLowLevel.Tests/paging.cs
```diff
@@ -157,12 +157,12 @@ await foreach (var page in pageableAsync.AsPages())
         public Task PagingMultiple_Continuation() => Test(async (endpoint) =>
         {
             var client = new PagingClient(Key, endpoint, null);
-            var pages = client.GetMultiplePages(null, null, null, new()).AsPages().ToArray();
+            var pages = client.GetMultiplePages(null, null, new()).AsPages().ToArray();
```
lirenhe, line 160: What triggered the change in this file?

lirenhe:
I didn't see any updates in the generated code for return-client-request-id, is this expected?

test/AutoRest.TestServerLowLevel.Tests/paging.cs
chunyu3: Yes. in the paging service, it define `client-request-id` header, we need to remove it from method signature.

src/AutoRest.CSharp/Common/Output/Models/Requests/RequestHeader.cs
chunyu3: we handle `client-request-id` and `return-client-request-id` differently. (the different constant expression), so I separate them.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
chunyu3: It is rare, but possible.

chunyu3:
> I didn't see any updates in the generated code for return-client-request-id, is this expected?

It is expected. The Cadl-ranch test does not have`return-client-request-id` head parameter.

lirenhe:
> > I didn't see any updates in the generated code for return-client-request-id, is this expected?
> 
> It is expected. The Cadl-ranch test does not have`return-client-request-id` head parameter.

No a requirement for this PR but shall we update that test case? We need to ensure all cases are covered.

src/AutoRest.CSharp/Common/Output/Models/Requests/RequestHeader.cs
archerzz: You can define them as constant, and still put them into the same HashSet. See https://github.com/Azure/autorest.csharp/blob/8fd5781ccb57e422aaccd918270be3ffc6bd0d6c/src/AutoRest.CSharp/Common/Output/Models/Requests/RequestHeader.cs#L27-L34

That can also simplify the logic in `Parameter.cs` a little bit.

test/AutoRest.TestServerLowLevel.Tests/paging.cs
archerzz: It's defined here: https://github.com/Azure/autorest.testserver/blob/a949278ab15ec464bfa71da2a9e9c518be022b50/swagger/paging.json#L157-L162

src/AutoRest.CSharp/Common/Output/Models/Requests/RequestHeader.cs
chunyu3: @archerzz I prefer to separate clientRequestId and returnClientRequstID into different set. They are different. And I define ClinetRequestID a set (not a string available) because it may be customized to another head name instead of `client-request-id`. And in the future, we will remove ClientRequestIDPolicy from pipeline, and then `x-ms-client-request-id` will included and handle the same way.

chunyu3:
> > > I didn't see any updates in the generated code for return-client-request-id, is this expected?
> > 
> > 
> > It is expected. The Cadl-ranch test does not have`return-client-request-id` head parameter.
> 
> Not a requirement for this PR but shall we update that test case? We need to ensure all cases are covered.

We have discussed this when we add this cadl-ranch test and decide not not include it. I will discuss with chenjie about it again.

archerzz (approved):

lirenhe (approved):

Issue #3519, opened 2023-06-22
TypeSpec C#: @@include can't override inherited @@internal


trrwilson:
We're using the `@internal` decorator (applied in a `client.tsp` file via `@@`) to hide a long-running operation route from the public surface of our client libraries, as we'd like to wrap the presentation of it with custom code.

This works well and produces the desired effect of directing the emission of the C# method and its response model hierarchy to be `internal` instead of `public`. It also applies to the options model used for the operation -- and that's the one thing we'd still like to retain in the public surface for reuse.

We tried to use the `@include` decorator for the options model in conjunction with the `@internal` used on the route that consumes that options model, but it unfortunately doesn't appear that `@include` can override the serialization change effected by `@internal`.

Issue/request: conditionalize model emission visibility to retain `public` if `@include` is used, even if the parsing hierarchy has been made `@internal`.

Example client.tsp body (where `@internal` works, but `@include` doesn't take effect in .NET):

```typescript
// Azure-specific long-running operations should be treated as implementation details
@@internal(Azure.OpenAI.startGenerateImage);

// Some models from client-internal LROs are still desired for custom public surface
@@include(Azure.OpenAI.ImageGenerationOptions)
```

*Edit*: adding link for our original impacted .tsp:

- https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/OpenAI.Inference/client.tsp

This is intended to make the long-running operations internal while still preserving the public visibility of the options and response payload models involved (as those are then wrapped in a custom implementation).

chunyu3:
Hello @trrwilson would you please provide the typespec files you uses and we want to check how you conjunction `@include` and `@internal`? 

And do you mean that the model will be used in more than one operations, one is decorated with `@internal`, the other is not?

tadelesh:
When a model is `internal` but `include`, we should not generate it as internal model. 
I think this should be done in TCGC. When doing transitive closure detection for internal models, we should not label one model with internal if it has @include decoration.

@iscai-msft what do you think?

PR #3671, opened 2023-08-18
Support @access and @usage

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/3519. 
We are introducing the decorator `@access` and `@usage`.
Regen: https://github.com/Azure/azure-sdk-for-net/pull/38813


src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -302,6 +302,7 @@ function setUsage(
     models: Map<string, InputModelType | InputEnumType>
 ) {
     for (const [name, m] of models) {
+        if (m.Usage !== undefined && m.Usage !== Usage.None) continue;
```
archerzz, line 305: Why are some models already set  `Usage`?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -142,6 +144,13 @@ function getCSharpInputTypeKindByIntrinsicModelName(
             return InputTypeKind.Float32;
         case "float64":
             return InputTypeKind.Float64;
+        case "uri":
```
archerzz, line 147: I think this fixes other issues?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -625,6 +637,13 @@ export function getInputType(
                     enums,
                     literalTypeContext
                 );
+                if (
+                    model.Namespace === "Azure.Core.Foundations" &&
+                    model.Name === "Error" &&
+                    isInputModelType(inputType)
+                ) {
+                    inputType.Accessibility = undefined;
```
archerzz, line 645: Why `undefined?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -754,6 +773,21 @@ export function getInputType(
...
+    context: SdkContext,
+    source: Model | Enum,
+    target: InputModelType | InputEnumType
+) {
+    const sourceUsage = getUsageOverride(context, source);
+    if (sourceUsage === UsageFlags.Input) {
+        target.Usage = Usage.Input;
+    } else if (sourceUsage === UsageFlags.Output) {
+        target.Usage = Usage.Output;
+    } else if (sourceUsage === (UsageFlags.Input | UsageFlags.Output)) {
```
archerzz, line 786: Is there a `Usageflag.InputOutput` or `Usage.RoundTrip`?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -199,7 +200,9 @@ export function loadOperation(
         Summary: summary,
         Deprecated: getDeprecated(program, op),
         Description: desc,
-        Accessibility: isInternal(sdkContext, op) ? "internal" : undefined,
+        Accessibility: isInternal(sdkContext, op)
```
archerzz, line 203: Can we move `internal` into `getAccess`?

archerzz (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
pshao25: No, `UsageFlags.Input | UsageFlags.Output` is by design

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
```diff
@@ -23,10 +23,25 @@ internal class DpgOutputLibrary : OutputLibrary
...
+        public IEnumerable<ModelTypeProvider> Models
+        {
+            get
+            {
+                foreach (var key in _models.Keys)
+                {
+                    if (key is { Namespace: "Azure.Core.Foundations", Name: "Error" })
+                    {
+                        continue;
+                    }
```
ArcturusZhang, line 35: this is not a good solution.
If we have more special types to replace, we have to change two places.
What is the issue when our previous solution?

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
```diff
@@ -23,10 +23,25 @@ internal class DpgOutputLibrary : OutputLibrary
...
+                    yield return _models[key];
+                }
+            }
+        }
         public IReadOnlyList<LowLevelClient> RestClients { get; }
         public ClientOptionsTypeProvider ClientOptions { get; }
-        public IEnumerable<TypeProvider> AllModels => new List<TypeProvider>(_enums.Values).Concat(_models.Values);
+        public IEnumerable<TypeProvider> AllModels => new List<TypeProvider>(_enums.Values).Concat(Models);
+        public IEnumerable<string> AccessOverrideModels => Enums.Where(e => e.IsAccessibilityOverride).Select(e => e.Declaration.Name)
```
ArcturusZhang, line 43: Should this be `AccessOverriddenModels`?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
pshao25: Yes, to upgrade the new version of tcgc which contains getAllModels, some gaps needs to fill.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
pshao25: If I don't set it to `undefined`, it will have `Accessibility` inherited from its containing model. Then postprocessor will think "it is a model which needs to keep instead of remove even there is no place using it". But our expected behavior is to remove it.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -989,3 +1032,29 @@ export function getFormattedType(program: Program, type: Type): FormattedType {
...
+            getFormattedType(context.program, x),
+            models,
+            enums
+        ) as any;
+    const skipSubNamespaces = isGlobalNamespace(context.program, namespace);
+    navigateTypesInNamespace(
+        namespace,
+        {
+            model: (x) =>
+                x.name !== "" && !["AuthFlow", "MyFlow"].includes(x.name) && x.kind === "Model" && computeModel(x),
```
pshao25, line 1055: Here `!["AuthFlow", "MyFlow"].includes(x.name)` is a temporary workaround. Will try to move it before merge.

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
pshao25: Isn't this our previous solution? Or are you saying we modify `_models` instead of `Models`?

ArcturusZhang: Something like:
```csharp
public IEnumerable<ModelTypeProvider> Models
{
get
{
foreach (var (key, provider) in _models)
{
var type = TypeFactory.CreateType(key);
if (type is {IsFrameworkType: false, Implemention: ModelTypeProvider m} && m == provider)
yield return m;
}
}
```
The idea here is trying to know if the type corresponding to this key has been replaced or not and we should do it in a more generic way.

pshao25: This snippet will create a lot of useless class when the type is not error. I have a new solution to resolve your first concern. You could check out the latest iteration.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
pshao25: Unfortunately, we cannot do this currently. The logic for `isInternal` is a little bit different from `getAccess`. Anyway we are going to deprecate `@internal`. Will remove this after that.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
pshao25: These are the overridden ones by `@usage`.

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
ArcturusZhang: Why is this "creating" classes?
The `TypeFactory.CreateType` does not really "create" a type, it calls the `ResolveModel` method on the library in this case, which queries the `_model` dictionary using the model and returns the key.
This process never creates a class.

My initial concern is that this piece of code (both version) can only handles this "error" case, it is not generic. If we have another thing we have to replace, we need to change two places. And this is not a thing that the library should know or care about.

pshao25: So to call this `TypeFactory.Create` you have to know "If I am a model then there would be a cache". Do you think it is readable?

Anyway, what's the problem with my latest code? There are no "two places" now.

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
```diff
@@ -38,6 +38,7 @@ public EnumType(InputEnumType input, string defaultNamespace, string defaultAcce
 
             DefaultName = input.Name.ToCleanName();
             DefaultAccessibility = input.Accessibility ?? defaultAccessibility;
+            IsAccessibilityOverridden = input.Accessibility != null;
```
chunyu3, line 41: Why do we need this field in output model?  Output model will contain finally information and will not need any computation.  Can we just set the accessibility value?

pshao25: This property is used to send to post processor. If a model is access overridden, the post process should not remove it.

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
ArcturusZhang: I read your latest implementation, and there are still two places and this `Models` could only handle the replacement for errors correctly now.
This is the concern.

BTW we do not really need to consider "if this will create a new instance or not". Space consumption is the last thing we need to consider and if you notice, we are creating small object everywhere, such as when you use `typeof()` for a `CSharpType` parameter, the implicit operator always creates a new instance.

Also if you think the `TypeFactory.Create` does not have enough readability, go ahead to rename it to like `TypeFactory.GetOrCreate` to indicate it has a cache. This is not the issue we are having here.

ArcturusZhang (approved):

lirenhe (approved):

PR #3672, opened 2023-08-18
Refactor the sample composer to use `CodeWriter` instead

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3383
Fixes https://github.com/Azure/autorest.csharp/issues/3337
Fixes https://github.com/Azure/autorest.csharp/issues/3499
Fixes https://github.com/Azure/autorest.csharp/issues/3732

<del>Depends on https://github.com/Azure/autorest.csharp/pull/3711</del>

Possible follow ups:
1. Manually fix the format in collection initializers since Roslyn has a bug right now: https://github.com/dotnet/roslyn/issues/8269 (tracking https://github.com/Azure/autorest.csharp/issues/3732)
2. refactor some double iterations in the current implementation. They are temporary changes to avoid too many changes in the current regen

# Description

## How this example writer work now?

This PR separates our previous sample generator into two parts:
1. makes up mock values. This is done in the new `ExampleMockValueBuilder` class in `AutoRest.CSharp.Common.Input.Examples` namespace
3. consumes the example values (wherever it comes from) and produces C# sample code.

The biggest benefit of this separation is that in the future, if we want to consume the examples in the typespec, we just need an extra thing to deserialize those values into the structures we are having right now in this PR and the rest part introduced in this PR will write them into sample code.

## How does the example writer write sample code?

Now the sample writer will write things based on the CSharpType of the property or the parameter regardless what type of the example value is. We must consider this because we would like to let the generator always generates code that could compile properly.
Therefore in this PR, when we are writing the code, we will always check the type of the property or parameter first, and then try to write the value in a value that could compile. If the type or format of the value mismatches, the generator will always generate `default` there to ensure it could compile properly.

This whole mechanism lives in an extension class of `CodeWriter` called `CodeWriterExampleExtensions`.

## Refactors around the xml writer

Now our generator uses `XDocument` to build a xml document instead of using `StringBuilder` and putting everything manually by ourselves.
When writing the standalone xml document file, the generator will not write the code inside it because we need the sample code to be reduced and formatted by Rolsyn before we put them in the xml file.
Therefore the generator now just writes a test method name to where the sample code should be.

In the `GeneratedCodeWorkspace`, this PR adds a step after every file has been reduced and formatted to read the generated sample file, and parse it to get the implementation by the test method name we wrote before in the xml file, then replace it with the reduced and formatted code.

## Changes on the generated samples

1. When the `RequestContext` is required on the signature, we will pass `null` instead of `new RequestContext()`. Because we make it required to avoid ambiguous calls between protocol and convenience, `null` value is still allowed there.
2. Quite a few example values changed, because now they are centrally generated, therefore for the same parameter or property in protocol method and convenience method, we will always use the same value, previously they are not using the same value therefore we have changes now.
4. When we are building an anonymous object, we are now using `Dictionary<string, object>` to avoid the case that we get compilation errors when the property's serialized name has special characters or collides with C# keywords in anonymous objects (see issue #3499 for details)

## Issues in anonymous objects

Previously we always use anonymous object as the initialization in the samples for protocol methods which is nice to read for our customers. But unfortunately, as in https://github.com/Azure/autorest.csharp/issues/3499 , when the serialized name of a property is not a proper identifier in C# (which happens occasionally, such as a property named `continue` which is a C# keyword, or when a property name has hyphen in it), we no longer could construct it using the anonymous object syntax. When this happens, we have to use a dictionary.

in the implementation, when we could use the anonymous object syntax, we will use it. Only when we cannot, we switch to dictionaries.

# Checklist

# Ready to Land?


test/TestProjects/FirstTest-TypeSpec/src/Generated/Docs/FirstTestTypeSpecClient.xml
```diff
@@ -672,179 +695,216 @@ Console.WriteLine(result.GetProperty("optionalLiteralBool").ToString());
...
-    },
+bleEnum.One
+},
     FloatFixedEnum = FloatFixedEnum.One,
-    FloatFixedEnumCollection = 
+    FloatFixedEnumCollection =
 {
-        FloatFixedEnum.One
-    },
+um.One
```
ArcturusZhang, line 744: Need to fix this on the `XmlFormatter`

test/TestProjects/FirstTest-TypeSpec/src/Generated/Docs/FirstTestTypeSpecClient.xml
```diff
@@ -672,179 +695,216 @@ Console.WriteLine(result.GetProperty("optionalLiteralBool").ToString());
...
     OptionalLiteralBool = true,
-    OptionalNullableList = 
+    OptionalNullableList =
 {
-        1234
-    },
-}, BinaryData.FromString("<your binary data content>"), new Dictionary<string, BinaryData>
+1234
+},
+}, BinaryData.FromObjectAsJson(new Dictionary<string, object>()), new Dictionary<string, BinaryData>()
```
ArcturusZhang, line 726: Also need to refine this

samples/AnomalyDetector/tests/Generated/Samples/Samples_AnomalyDetectorClient.cs
```diff
@@ -89,22 +91,65 @@ public void Example_DetectUnivariateEntireSeries_AllParameters()
...
+new TimeSeriesPoint((float)123.45F)
+{
+Timestamp = DateTimeOffset.Parse("2022-05-10T14:57:31.2311892-04:00"),
+}
+})
+            {
+                Granularity = TimeGranularity.Yearly,
+                CustomInterval = 1234,
+                Period = 1234,
+                MaxAnomalyRatio = (float)123.45F,
```
ArcturusZhang, line 126: need to investiagte

samples/AnomalyDetector/src/Generated/Docs/AnomalyDetectorClient.xml
```diff
@@ -2,78 +2,102 @@
...
+AzureKeyCredential credential = new AzureKeyCredential("<key>");
+AnomalyDetectorClient client = new AnomalyDetectorClient(endpoint, credential);
+
+UnivariateDetectionOptions options = new UnivariateDetectionOptions(new List<TimeSeriesPoint>()
 {
-        Timestamp = DateTimeOffset.UtcNow,
-    }
+new TimeSeriesPoint(123.45F)
+{
+Timestamp = DateTimeOffset.Parse("2022-05-10T14:57:31.2311892-04:00"),
```
chunyu3, line 28: Can we format this sample code? It is not easy to read.

src/AutoRest.CSharp/Common/Input/InputTypes/Examples/InputOperationExample.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace AutoRest.CSharp.Common.Input.Examples
+{
+    // TODO -- currently we do not need the responses. In the case that in the future we need to handle the responses in examples, we need to add a new class InputExampleResponse and a new property here
+    internal record InputOperationExample(InputOperation Operation, IReadOnlyList<InputParameterExample> Parameters);
```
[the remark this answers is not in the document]
ArcturusZhang: we could add them when we need it. This is the reason I leave a TODO here.

samples/AnomalyDetector/src/Generated/Docs/AnomalyDetectorClient.xml
ArcturusZhang: Of course the sample code is formatted. You see the bad format here because of a bug in Rolsyn: https://github.com/dotnet/roslyn/issues/8269 (which I leave TODO items in our code as well [here](https://github.com/Azure/autorest.csharp/blob/d92e59263cca015f9a250c95c3a3715d17e67fc4/src/AutoRest.CSharp/LowLevel/Extensions/CodeWriterExampleExtensions.cs#L235))

We could do a manual fix - we are already doing some in `XmlFormatter` class if you notice.
It is complicated to do that, and it only happens in collection initializations, therefore I would like to leave it to another PR

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -107,9 +110,9 @@ public IEnumerable<LowLevelClientMethod> CustomMethods()
         }
 
 
-        public static IEnumerable<LowLevelClientMethod> BuildMethods(TypeFactory typeFactory, IEnumerable<InputOperation> operations, ClientFields fields, string namespaceName, string clientName, SourceInputModel? sourceInputModel)
+        public static IEnumerable<LowLevelClientMethod> BuildMethods(LowLevelClient? client, TypeFactory typeFactory, IEnumerable<InputOperation> operations, ClientFields fields, string namespaceName, string clientName, SourceInputModel? sourceInputModel, IReadOnlyDictionary<string, InputClientExample>? clientParameterExamples)
```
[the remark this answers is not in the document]
ArcturusZhang: Good idea and we have one parameter less for this method which is good.

src/AutoRest.CSharp/Common/Input/InputTypes/Examples/InputOperationExample.cs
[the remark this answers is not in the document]
chunyu3: In previous sample/doc generation, we will show the result in sample see https://github.com/Azure/autorest.csharp/blob/e373769d6923262334156659e19554b62f89009b/test/TestProjects/FirstTest-TypeSpec/tests/Generated/Samples/Samples_FirstTestTypeSpecClient.cs#L52-L53

I think it is better to keep this.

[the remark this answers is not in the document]
ArcturusZhang: This part is still there in the generated samples.
This is something different.
When do we need the response **VALUES** here? Please note the above code never have a response value.
When we want to generate "Test code" with assertions in them. For example, if we want to generate code like this:
```csharp
// call the method above and get a response back
Assert.AreEqual("valueInResponse", response.GetProperty("name"));
```
We do not have this plan. Therefore I add it in a TODO.

samples/AnomalyDetector/src/Generated/Docs/AnomalyDetectorClient.xml
chunyu3: When will Rolsyn bug will be resolved? Just a little concern about the format, if we generate the sample in that format, customer may complain it.

ArcturusZhang: I do not know therefore I created the item above.
This issue is not introduced by this refactor which is reason I have no intention to fix it here.

chunyu3 (approved):

samples/AnomalyDetector/src/Generated/Docs/AnomalyDetectorClient.xml
ArcturusZhang: it is simpler to fix than I thought - I have fixed this issue in this PR.

lirenhe (approved):

PR #3674, opened 2023-08-18
Move shared source from Generator.Shared to Azure.Core.Shared

annelo-msft:
This change moves all shared source files except the `CodeGen*Attribute` files in Generator.Shared into Azure.Core.Shared.  They are now mastered in azure-sdk-for-net, with this PR: https://github.com/Azure/azure-sdk-for-net/pull/38286


Directory.Build.targets
```diff
@@ -18,19 +18,51 @@
     <Compile Include="$(AzureCoreSharedSources)AsyncLockWithValue.cs" LinkBase="Shared/Core" />
     <Compile Include="$(AzureCoreSharedSources)AzureKeyCredentialPolicy.cs" LinkBase="Shared/Core" />
     <Compile Include="$(AzureCoreSharedSources)ClientDiagnostics.cs" LinkBase="Shared/Core" />
+    <Compile Include="$(AzureCoreSharedSources)ChangeTrackingDictionary.cs" LinkBase="Shared/Core" />
```
m-nash, line 21: You will need to update [DownloadShareSource](https://github.com/Azure/autorest.csharp/blob/feature/v3/eng/DownloadSharedSource.ps1#L35-L39) since we have a inclusion list of these files and this is checked at ci time to see that all of these files are in sync with the upstream.

Since you are using an alpha which won't be in main yet temporarily change the branch [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/eng/DownloadSharedSource.ps1#L22) and [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/eng/DownloadSharedSource.ps1#L23) so it can find all the new files in the new location.

Simply replace `git checkout` with `git checkout [branch]`

m-nash:
We need to change this somehow https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/build/CodeGeneration.targets#L89.

We need to be aware of people using this outside of azure-sdk-for-net.  I don't know how we can bundle these files such that they will not be broken but this definitely a test we will want to do / verify.

m-nash:
Need to update [DownloadShareSource](https://github.com/Azure/autorest.csharp/blob/feature/v3/eng/DownloadSharedSource.ps1) to include all the new files that are moving to Azure.Core.

Since you are using an alpha version there are [2 spots here](https://github.com/Azure/autorest.csharp/blob/feature/v3/eng/DownloadSharedSource.ps1#L22-L23) where you need to change `git checkout` to `git checkout [your branch]` so that the new files can be found.

annelo-msft:
> We need to change this somehow https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/build/CodeGeneration.targets#L89.

Can we leave this on a first iteration?  It will be an empty folder, but that should not be breaking.

m-nash (approved):

Issue #2456, opened 2022-07-21
Rename configuration `rename-rules` to `acronym-mapping`


ArcturusZhang:
The name `rename-rules` looks similar to `rename-mapping` and is not quite fully described what this configuration could do.
We would like to rename it to reflect its real purpose.

RodgeFu:
PR: 
https://github.com/Azure/autorest.csharp/pull/3676
https://github.com/Azure/azure-sdk-for-net/pull/38294

PR #3676, opened 2023-08-19
Rename configuration rename-rules to acronym-mapping

RodgeFu:
Fixes https://github.com/Azure/autorest.csharp/issues/2456


ArcturusZhang (commented):
I suggest we do it in two PRs in the generator:
1. open a generator PR (**the first PR**) to make the generator to accept both `rename-mapping` and `acronym-mapping` as keys (`acronym-mapping` should have higher priority)
2. the regen of the first PR should be no change.
3. we could merge the first PR.
4. open a PR (**the SDK PR**) in the azure-sdk-for-net repo to rename all config to `acronym-mapping`. There should also be no code change when doing this. Merge this PR.
5. open another generator PR (**the second PR**) to make the generator only accept `acronym-mapping`. The regen of this PR should also contain no change.

I personally prefer we should do **the second PR** later - maybe months later, because there might be ongoing PR waiting to merge into the SDK if our generator no longer supports `rename-mapping`, they will not get a notice, and the generator might just work fine.

Also we need to change the mgmt template as well to use the new config name in **the SDK PR**: https://github.com/Azure/azure-sdk-for-net/blob/858b7ba9da51891ccb77b7aa88d94a44b9bffeb2/eng/templates/Azure.ResourceManager.Template/src/autorest.md?plain=1#L26

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
```diff
@@ -265,7 +265,7 @@ internal static MgmtConfiguration GetConfiguration(IPluginCommunication autoRest
                 operationPositions: autoRest.GetValue<JsonElement?>("operation-positions").GetAwaiter().GetResult(),
                 requestPathToSingletonResource: autoRest.GetValue<JsonElement?>("request-path-to-singleton-resource").GetAwaiter().GetResult(),
                 overrideOperationName: autoRest.GetValue<JsonElement?>("override-operation-name").GetAwaiter().GetResult(),
-                renameRules: autoRest.GetValue<JsonElement?>("rename-rules").GetAwaiter().GetResult(),
+                renameRules: GetAcronymMappingConfig(autoRest),
```
ArcturusZhang, line 268: Could we rename the name of the argument here as well?
```suggestion
                acronymMapping: GetAcronymMappingConfig(autoRest),
```

docs/mgmt/polishing.md
```diff
@@ -18,7 +18,7 @@
 A mechanism of renaming is introduced in the generator to unify the casing of some words.
```
ArcturusZhang, line 18: We should also change the name of this section

ArcturusZhang:
```suggestion
A mechanism of acronym mapping is introduced in the generator to unify the casing of acronyms.
```

ArcturusZhang (approved):
Approved with two comments

ArthurMa1978 (approved):

Issue #3653, opened 2023-08-09
`{}` generates invalid source


heaths:
**Describe the issue or request**

In TypeSpec, you can declare a plain object using only `{}` but the C# generator emits erroneous source named simply _.cs_ i.e., just an extension - no file name:

```C#
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

// <auto-generated/>

#nullable disable

namespace Azure.AI.Language.Conversations
{
    /// <summary> Initializes a new instance of. </summary>
    internal ()
{
}
}
}
```

See [this comment](https://github.com/Azure/azure-sdk-for-net/pull/37820#issuecomment-1669139641) for more context.

**Describe your ideas for solutions**

This should be supported and, perhaps, either generate a property-less model with stable name based on the property, or `BinaryData` if we end up using `{}` to signify any valid JSON object. See [our internal Teams channel discussion](https://teams.microsoft.com/l/message/19:906c1efbbec54dc8949ac736633e6bdf@thread.skype/1690838455083?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1690838455083&teamName=Azure%20SDK&channelName=TypeSpec%20(Cadl)%20Discussion%20%F0%9F%90%AE&createdTime=1690838455083) for more details.

heaths:
One thought about how this could conceptually work is to use the same naming convention generation that you use for, say, inline enums. E.g. (though, perhaps not exact):

```
model Foo {
  bar?: {};
}
```

You'd generate something like you would if that were an enum e.g., `FooBar` with no properties.

As for use cases, this could be a placeholder. Maybe we don't want to support this, but in that case I think you'd want to err immediate rather than err downstream when trying to compile source. Easier to diagnose if a TypeSpec author generating an SDK would get a message saying "we don't support {}" or whatever.

chunyu3:
This issue is one scenario of anonymous model. There is an issue https://github.com/Azure/autorest.csharp/issues/3623 to track all those.

lirenhe:
There is the discussion related to this topic in https://github.com/Azure/typespec-azure/issues/3389

Issue #3623, opened 2023-07-25
Handle Anonymous models


pshao25:
Initial problem comes from https://github.com/Azure/autorest.csharp/issues/3592#issuecomment-1649195720. User wants to define invalid spec like 
```
op postClusterScale(): {
  @body body: ClusterScale;
};
```
Two errors here:
1. The return type is an anonymous type.
2. There is a body decorator in return type. Cannot believe typespec compiler even allows it pass!

We should properly handle the anonymous type.

**Scenario 1: return type is an anonymous body type**
```
op postClusterScale(): {
  @body property: ClusterScale;
};
```
Actual: Successfully generate the model `ClusterScale` with **internal** constructor. Interesting...
Expected.

**Scenario 2: return type is an anonymous normal type**
```
op postClusterScale(): {
  property: ClusterScale;
};
```
Actual: models with empty name
![image](https://github.com/Azure/autorest.csharp/assets/97225342/9df6440a-0e7e-48c2-9ad1-c55bd6a8c1e0)
Expected:
1. Fail because it is anonymous.
**2. Make it `low confident`.** issue to tyspec

**Scenario 3: return type is an empty return type**
```
op postClusterScale(): {}
```
Actual: models with empty name
![image](https://github.com/Azure/autorest.csharp/assets/97225342/9df6440a-0e7e-48c2-9ad1-c55bd6a8c1e0)
Expected:
Regard `{}` as a `object` => Record<unknown>.

**Scenario 4: Input type is an anonymous type**
```
model Input {
  @path
  id: string;
  detail: {
    info: string;
    type: string;
  }
};
op operation(...Input): void;
```
Actual: models with empty name
![image](https://github.com/Azure/autorest.csharp/assets/97225342/97bc3532-27be-492c-a7d3-4cae55ed8aff)
Expected: Fail or make it `low confident`.

**Scenario 5: Input type is an alias with anonymous type**
```
alias Input = {
  @path
  id: string;
  @body
  detail: {
    info: string;
    type: string;
  }
};
op operation(...Input): void;
```
Actual: models with empty name
![image](https://github.com/Azure/autorest.csharp/assets/97225342/97bc3532-27be-492c-a7d3-4cae55ed8aff)
Expected: It should be the same as below. (But we cannot generate below spec, which means we have another bug with spread scenario.)
```
alias DetailAlias = {
    info: string;
    type: string;
};
alias Input = {
  @path
  id: string;
  @body
  detail: DetailAlias;
};
[...]

PR #3681, opened 2023-08-21
Handle anonymous model

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3653
Fixes https://github.com/Azure/autorest.csharp/issues/3623

# Description

Update the confident level mechanism. Now when we find any anonymous model in an operation, we will mark it as "not confident" and will not generate the corresponding convenience method (instead of making it internal).

Now the confident level is an enumeration indicating an increasing level of the confidence:
```csharp
public enum ConvenienceMethodConfidentLevel
{
	Confident = 0, // default level
	Internal = 1, // the case of union, number-literal, etc
	Remove = 2 // anonymous models
}
```

In order to make everything working properly, we must assign a name to all models, otherwise a class without name will be generated and it cannot compile which nullifies all our mechanism. This PR takes the id of the model as its name - which is a number. Since we will remove the corresponding convenience method, the post processor will work properly and remove every involved models including the anonymous model.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -87,7 +87,12 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
 
         private static InputModelType CreateInputModelTypeInstance(string? id, string? name, string? ns, string? accessibility, string? deprecated, string? description, string? usageString, string? discriminatorValue, string? discriminatorPropertyName, InputModelType? baseModel, List<InputModelProperty> properties, List<InputModelType> derivedModels, bool isNullable, ReferenceResolver resolver)
         {
-            name = name ?? throw new JsonException("Model must have name");
+            if (name == null)
+                throw new JsonException("Model must have name");
+
+            if (string.IsNullOrEmpty(name))
```
archerzz, line 93:
```suggestion
            if (name == "")
```

null is checked above

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
```diff
@@ -1,51 +1,61 @@
...
 
 namespace AutoRest.CSharp.Output.Models
 {
     internal static class OperationConfidentChecker
     {
-        private static ConcurrentDictionary<InputType, bool> _cache = new();
+        private static ConcurrentDictionary<InputType, ConvenienceMethodConfidentLevel> _cache = new();
 
-        public static bool IsConfident(InputOperation operation)
+        public static ConvenienceMethodConfidentLevel GetConfidentLevel(InputOperation operation, TypeFactory typeFactory)
```
archerzz, line 18: Better to use noun here, and in similar case like `OperationConfidenceCheck`.

```suggestion
        public static ConvenienceMethodConfidenceLevel GetConfidentLevel(InputOperation operation, TypeFactory typeFactory)
```

archerzz (commented):
Looks good to me.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
ArcturusZhang: well I change it to name.Length == 0 since it is the best practice - actually for readability I would rather keep `IsNullOrEmpty` because it is basically doing the same check if you check the source of it.
https://stackoverflow.com/questions/7872633/most-performant-way-of-checking-empty-strings-in-c-sharp

src/AutoRest.CSharp/LowLevel/Output/OperationConfidentChecker.cs
```diff
@@ -1,51 +1,61 @@
...
         {
+            var confidentLevel = ConvenienceMethodConfidentLevel.Confident;
             // check parameters
             // skip the special parameters since some particular types are still interpreted as unions (for instance content type could have union of literal types which is actually enums)
             foreach (var parameter in operation.Parameters.Where(p => !p.IsContentType && !p.IsApiVersion && !p.IsEndpoint))
             {
-                if (!IsConfident(parameter.Type))
-                    return false;
+                var level = GetConfidentLevel(parameter.Type, typeFactory);
```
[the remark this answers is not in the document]
ArcturusZhang: I introduced a `UpdateConfidenceLevel(ref ConvenienceMethodConfidenceLevel current, ConvenienceMethodConfidenceLevel level)` to do this comparison

archerzz (approved):

src/AutoRest.CSharp/LowLevel/Output/OperationConfidenceChecker.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        private static ConvenienceMethodConfidenceLevel WalkUnionType(InputUnionType unionType)
+        {
+            // the union types are always not confident.
+            // if any exceptions happen in the future, we could handle it here.
+            return ConvenienceMethodConfidenceLevel.Internal;
+        }
+
+        private static bool IsUnreasonableName(string name) => name.StartsWith('_');
```
chunyu3, line 194: The anonymous model name will be updated to `id`. so this function will not filter out anonymous model and set it as low confident. So where do you handle anonymous model?

And the model name which is not valid, such as keyword, the name will perfix `_`, so all those model are marked as low confident?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -60,13 +61,21 @@ public ModelTypeProvider(InputModelType inputModel, string defaultNamespace, Sou
...
             _derivedTypes = derivedTypes;
             _defaultDerivedType = defaultDerivedType ?? (inputModel.IsUnknownDiscriminatorModel ? this : null);
         }
 
+        private static string GetValidIdentifier(string name)
+        {
+            if (SyntaxFacts.IsValidIdentifier(name))
+                return name;
+
+            return $"_{name}"; // prepend a underscore to make sure the name is valid identifier
```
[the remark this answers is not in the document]
ArcturusZhang: There are multiple rules:
1. If it is a keyword, prepend `@`
2. If the leading character is a digit, prepend `_`.

In my case, when a model is anonymous, I will use `id` as its name, which is a number, therefore I just implement the `_` case here.
As I claimed above in comment, this is only a workaround.

src/AutoRest.CSharp/LowLevel/Output/OperationConfidenceChecker.cs
ArcturusZhang: > So where do you handle anonymous model?
I did it in the `TypeSpecInputModelTypeConverter`

Why I did not just use `inputModel.Name` and check if it is valid identifier?
I would like to make sure this could work with customization code, therefore I choose to check the `modelTypeProvider.Declaration.Name` instead of checking `InputModel.Name`.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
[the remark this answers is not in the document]
chunyu3: @ArcturusZhang Thanks for clarifying. But if this method only handle the anonymous model which is updated to digital, can we add more comments about this method, and change the method to specify this, otherwise, the method may consider to suffix `_` for all invalid identifier.

```
if (char.IsDigit(name[0]))
{
    return $"_{name}";
} 

```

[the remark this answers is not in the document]
ArcturusZhang: This is fixed.
I change this method to make it only change the name when the leading character is a digit. In other cases, the name does not change until we decide about the capitalization issue.

src/AutoRest.CSharp/LowLevel/Output/OperationConfidenceChecker.cs
ArcturusZhang: This method is now removed.
Instead, for model case, since anonymous model related things will be removed completely - there is no point to consider customizations (well in the meantime, they could always disable the post processor, and see those models, and introduce customization code, but since it relies on internally generated indices, it is not recommended)
for literal case, the check is:
first check if the name in input enum value is not valid identifier, then we compare the declaration name is the same as input enum value to ensure it has not been customized. If both match, we mark it as low confident.

chunyu3 (approved):

lirenhe (approved):

Issue #3596, opened 2023-07-17
[Feature] Overload function to allow method signature change for Model Factory - parameter


ArthurMa1978:
Swagger may introduce non-breaking changes by adding / removing optional parameters, which currently is breaking changes for .Net SDK.
CodeGen should generate overload function to make it non-breaking change automatically instead of handwriting manual code now.

ArcturusZhang:
Reopen since last time the PR resolved this has been reverted because of some accidents.

PR #3686, opened 2023-08-24
Mitigate breaking change for optional parameter change

live1206:
Fixes https://github.com/Azure/autorest.csharp/issues/3596.

This PR introduces `SignatureType`, it holds three portions of code:
- current
- custom
- baseline contract

It provides 
- OverloadingMethods: for now represents the missing methods in baseline contract due to additional optional parameters in the new contract, this is causing breaking changes


src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
```diff
@@ -102,4 +105,44 @@ private MethodSignature MakeSync()
...
+                && x.Parameters.SequenceEqual(y.Parameters, new ParameterComparer());
+            return result;
+        }
+
+        public int GetHashCode([DisallowNull] MethodSignature obj)
+        {
+            return HashCode.Combine(obj.Name, obj.ReturnType);
+        }
+
+        private bool CompareType(CSharpType? x, CSharpType? y)
```
archerzz, line 134: Can this method be moved to `CSharType`?

src/AutoRest.CSharp/Common/Output/Models/MethodSignatureBase.cs
```diff
@@ -3,15 +3,91 @@
...
+        {
+            if (typeSymbol is null)
+            {
+                // TODO: handle missing type
+                yield break;
+            }
+            var methods = typeSymbol!.GetMembers().OfType<IMethodSymbol>();
+            foreach (var method in methods)
+            {
+                var typeFactory = new TypeFactory(MgmtContext.Library);
```
archerzz, line 29: Need to avoid repetitively creating `TypeFactory`

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -17,25 +18,27 @@
...
 {
-    internal sealed class ModelFactoryTypeProvider : TypeProvider
+    internal sealed class ModelFactoryTypeProvider : SignatureTypeProvider
     {
         protected override string DefaultName { get; }
         protected override string DefaultAccessibility { get; }
 
         private IEnumerable<MethodSignature>? _methods;
-        public IEnumerable<MethodSignature> Methods => _methods ??= Models.Select(CreateMethod);
+        public override IEnumerable<MethodSignature> Methods => _methods ??= Models.Select(CreateMethod).ToList();
```
archerzz, line 32: remove `ToList()`? since the return type is `IEnumerable`

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureTypeProvider.cs
```diff
@@ -0,0 +1,188 @@
...
+        protected SignatureTypeProvider? PreviousContract => _previousContract ??= CreateFromCompilation(_sourceInputModel?.PreviousContract);
+
+        // TODO: store the implementation of missing methods along with declaration
+        public virtual IEnumerable<MethodSignature> Methods { get; }
+
+        protected override string DefaultName { get; }
+
+        protected override string DefaultAccessibility { get; }
+
+        public void WriteMissingOverloadMethod(CodeWriter writer, MethodSignature currentMethodToCall, MethodSignature previousMethodToAdd, IList<MethodParameter> missingParameters)
```
archerzz, line 50: No sure if we should put a write method in a `TypeProvider`.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -141,7 +141,7 @@ public static CodeWriter WriteFieldDeclarations(this CodeWriter writer, IEnumera
             return writer.Line();
         }
 
-        public static IDisposable WriteMethodDeclaration(this CodeWriter writer, MethodSignatureBase methodBase, params string[] disabledWarnings)
+        public static IDisposable WriteMethodDeclaration(this CodeWriter writer, MethodSignatureBase methodBase, bool ignoreOptional = false, params string[] disabledWarnings)
```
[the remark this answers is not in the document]
live1206: Parameters is `readonly` in `MethodSignature`

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
live1206: [EqualsTo ](https://github.com/Azure/autorest.csharp/blob/2ea857bbb2a719d7a1790dd501c856ba38de32eb/src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs#L96) has a different implementation with its own purpose.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: It is referenced multiple times.
Needs to materialize it here, it throws while visit it the second time.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureTypeProvider.cs
live1206: The purpose it to reuse this logic in multiple writers, not only `ModelFactoryWriter`

src/AutoRest.CSharp/Common/Output/Models/MethodSignatureBase.cs
live1206: Exposed `TypeFactory` in `MgmtContext` instead.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureTypeProvider.cs
live1206: Moved to `BreakingChangeWriter`

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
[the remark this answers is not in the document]
live1206: removed by make a copy of `MethodSignature` with required parameters.

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
live1206: Moved it to `TypeExtensions.cs`.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -446,5 +448,55 @@ public static bool RequiresToList(CSharpType from, CSharpType to)
 
             return to.FrameworkType == typeof(IReadOnlyList<>) || to.FrameworkType == typeof(IList<>);
         }
+
+        public CSharpType? GetCsharpType(ITypeSymbol symbol)
```
ArcturusZhang, line 452: should be `GetCSharpType`

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -26,25 +26,34 @@ internal sealed class ModelFactoryTypeProvider : TypeProvider
         protected override string DefaultName { get; }
         protected override string DefaultAccessibility { get; }
 
-        private IEnumerable<MethodSignature>? _methods;
-        public IEnumerable<MethodSignature> Methods => _methods ??= Models.Select(CreateMethod);
+        private IReadOnlyList<MethodSignature>? _methods;
+        public IReadOnlyList<MethodSignature> Methods => _methods ??= Models!.Select(CreateMethod).ToList();
 
-        public IEnumerable<SerializableObjectType> Models { get; }
+        public IEnumerable<SerializableObjectType>? Models { get; }
```
ArcturusZhang, line 32: it is not really good to make a collection nullable because an empty list of models look like means just the same as null.

I saw we have a new ctor for this type provider, if the ctor must initialize this property, we could initialize it as `Enumerable.Empty<SerializableObjectType>()`

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
```diff
@@ -102,4 +109,30 @@ private MethodSignature MakeSync()
             };
         }
     }
+
+    internal class MethodSignatureComparer : IEqualityComparer<MethodSignature>
```
ArcturusZhang, line 113: I believe it would be better if we could add a static field in MethodSignature for this with a more descriptive name, and make this class private which would make things more readable, such as
```csharp
internal record MethodSignature(/* the properties */)
{
    public static IEquality<MethodSignature> ParameterAndReturnTypeEqualityComparer = new MethodSignatureComparer();
}

private class MethodSignatureComparer : IEqualityComparer<MethodSignature>
{}
```
So that in other places, we could use `MethodSignature.ParameterAndReturnTypeEqualityComparer` (or other descriptive names) to get the comparer.

We might could do the same for the ParameterComparer below.

src/AutoRest.CSharp/Mgmt/Decorator/TypeExtensions.cs
```diff
@@ -40,6 +40,20 @@ public static bool EqualsByName(this CSharpType left, CSharpType right)
...
+            {
+                return true;
+            }
+
+            if (left is null || right is null)
+            {
+                return false;
+            }
+            return left.GetType().Equals(right.GetType());
+        }
```
ArcturusZhang, line 55: is this method expected? 
Please note `left.GetType()` will always returns you `typeof(CSharpType)`
If you want to check if the underlying type of these two CSharpType instances, you should do:
`left.FrameworkType.Equals(right.FrameworkType)`
Also `CSharpType` class has an implementation of `Equals` to compare whether two types are the same type, and some other variant of this method such as `EqualsIgnoreNullable` etc.

What is the purpose of this method?

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -40,6 +48,216 @@ protected TypeProvider(string defaultNamespace, SourceInputModel? sourceInputMod
...
+                {
+                    continue;
+                }
+
+                if (!CurrentContainAllPreviousParameters(previousMethod, item))
+                {
+                    continue;
+                }
+
+                if (!previousMethod.ReturnType.EqualsBySystemType(item.ReturnType))
```
ArcturusZhang, line 230: based on the context here, I believe we should use `Equals` here because we are trying the see if these return types are the same or not, and `CSharpType.Equals` is written for that purpose.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtContext.cs
```diff
@@ -16,6 +17,8 @@ internal static class MgmtContext
 
         public static MgmtOutputLibrary Library => Context.Library;
 
+        public static TypeFactory TypeFactory => Context.TypeFactory;
```
ArcturusZhang, line 20: I am OK with this change, but the DpgLibrary just has a public property `TypeFactory` for the type factory it is using.
Mgmt could follow the same pattern and do not have to change this.
But that might be a big irrelavant change, it is fine here.

src/AutoRest.CSharp/Mgmt/Decorator/TypeExtensions.cs
live1206: `typeof(CSharpType)` just get the type info, not the other details, which is taken into consideration in `CSharpType.Equals` method. Found that [TypeExtension.EqualsByName](https://github.com/Azure/autorest.csharp/blob/63b56fd85c988fe795e1ed7829fbad1f09a35c18/src/AutoRest.CSharp/Mgmt/Decorator/TypeExtensions.cs#L26) serves the purpose, I just need to update it to nullable.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtContext.cs
live1206: Since we already have created an instance of TypeFactory in [context](https://github.com/Azure/autorest.csharp/blob/63b56fd85c988fe795e1ed7829fbad1f09a35c18/src/AutoRest.CSharp/Common/Output/Models/Types/BuildContextOfT.cs#L46), why create another instance [here](https://github.com/Azure/autorest.csharp/blob/63b56fd85c988fe795e1ed7829fbad1f09a35c18/src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs#L33) again?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: The new constructor is removed, this can be reverted.

src/AutoRest.CSharp/Common/Output/Models/MethodSignatureBase.cs
```diff
@@ -3,8 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using AutoRest.CSharp.Generation.Types;
+using AutoRest.CSharp.Mgmt.AutoRest;
 using AutoRest.CSharp.Output.Models.Shared;
+using AutoRest.CSharp.Output.Models.Types;
 using AutoRest.CSharp.Utilities;
+using Microsoft.CodeAnalysis;
```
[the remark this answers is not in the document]
live1206: reverted

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -446,5 +448,55 @@ public static bool RequiresToList(CSharpType from, CSharpType to)
...
+            if (namedTypeSymbol == null)
+            {
+                throw new InvalidCastException($"Unexpected type {symbol}");
+            }
+
+            if (namedTypeSymbol.TypeArguments.Any())
+            {
+                wrapers.Push(namedTypeSymbol.Name);
+                return UnWrapType(namedTypeSymbol.TypeArguments[0], wrapers);
+            }
```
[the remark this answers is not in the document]
live1206: Handled multiple arguments and merged it to the original `TryCreateType` method, but unit testing for INamedTypeSymbol to CSharpType is not an easy task, the original method does not have UT either. This could be a new task requiring more time.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: Commented below.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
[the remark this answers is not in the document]
live1206: Besides, we have the existing test projects to guard this logic, found an issue of it from `TypesSchemaMapping`

src/AutoRest.CSharp/Common/Generation/Writers/BreakingChangeWriter.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using AutoRest.CSharp.Output.Models;
+using AutoRest.CSharp.Output.Models.Shared;
+using MethodParameter = AutoRest.CSharp.Output.Models.Shared.Parameter;
+
+namespace AutoRest.CSharp.Generation.Writers
+{
+    internal static class BreakingChangeWriter
```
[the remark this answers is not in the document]
live1206: This extension method could be used for other purpose, not only for breaking change, renamed it to `WriteOverloadMethod` and moved it to the existing `CodeWriterExtensions.cs`

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
live1206: Sounds good to me, updated

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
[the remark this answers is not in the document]
live1206: This logic is not needed for ModelFactory until dealing with Resource, postpone the change to later PR instead.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtContext.cs
ArcturusZhang: Dpg and HLC is not using BuildContext.
Mgmt is the only target that is using the BuildContext.
We have a plan to unify them, therefore in the near future, Mgmt will no longer use the BuildContext as well.

live1206: OK, thanks for the info. Sounds like we can leave the refactoring till then.

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -22,10 +24,13 @@ internal class CSharpGen : IPlugin
...
             ValidateConfiguration();
-
             Directory.CreateDirectory(Configuration.OutputFolder);
             var project = await GeneratedCodeWorkspace.Create(Configuration.AbsoluteProjectFolder, Configuration.OutputFolder, Configuration.SharedSourceFolders);
-            var sourceInputModel = new SourceInputModel(await project.GetCompilationAsync());
+
+            var baselineContract = await LoadBaselineContract();
+            var sourceInputModel = baselineContract is not null
+                ? new SourceInputModel(await project.GetCompilationAsync(), previousContract: baselineContract)
+                : new SourceInputModel(await project.GetCompilationAsync());
```
ArcturusZhang, line 33: could this just simply be
```suggestion
            var sourceInputModel = new SourceInputModel(await project.GetCompilationAsync(), previousContract: baselineContract);
```

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
```diff
@@ -98,5 +105,31 @@ private MethodSignature MakeSync()
...
+                    return true;
+                }
+
+                if (x is null || y is null)
+                {
+                    return false;
+                }
+
+                var result = x.Name == x.Name
+                    && x.ReturnType.EqualsByName(y.ReturnType)
```
ArcturusZhang, line 124: could this be `EqualsIgnoreNullable` instead of `EqualsByName`?
This `EqualsByName` is kind of a workaround for some special situation in MPG which should be removed - or promoted into the class of CSharpType if it is really needed somewhere

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -29,19 +32,28 @@ internal sealed class ModelFactoryTypeProvider : TypeProvider
         protected override string DefaultName { get; }
         protected override string DefaultAccessibility { get; }
 
-        private IEnumerable<Method>? _methods;
-        public IEnumerable<Method> Methods => _methods ??= Models.Select(CreateMethod);
+        // TODO: remove this intermediate state once we generate it before output types
+        private IReadOnlyList<Method>? _methods;
+        private IReadOnlyList<Method> ShouldNotBeUsedForOutPut([CallerMemberName] string caller = "")
+        {
+            Debug.Assert(caller == nameof(OutputMethods) || caller == nameof(SignatureType), $"This method should not be used for output. Caller: {caller}");
```
ArcturusZhang, line 39: is it really necessary to add this kind of guard on a private method?
If this is internal I would agree that we need a guard to avoid abuse of the method.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -18,6 +19,8 @@
 using AutoRest.CSharp.Output.Models.Serialization.Xml;
 using AutoRest.CSharp.Output.Models.Shared;
 using AutoRest.CSharp.Utilities;
+using Azure.Core.Expressions.DataFactory;
+using Azure.ResourceManager.Models;
```
ArcturusZhang, line 23: this might need a clean up

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,247 @@
...
+    internal class SignatureType
+    {
+        private readonly string _defaultNamespace;
+        private readonly string _defaultName;
+        private readonly SignatureType? _customization;
+        private readonly SignatureType? _previousContract;
+
+        // Missing means the method with the same name is missing from the current contract
+        // Updated means the method with the same name is updated in the current contract, and the list contains the previous method and current methods including overload ones
+        private readonly (IReadOnlyList<MethodSignature> Missing, IReadOnlyList<(List<MethodSignature> Current, MethodSignature Previous)> Updated)? _methodChangeset;
```
ArcturusZhang, line 25: since this is quite often used inside this class, could we make a struct or record to represent the type here?
It could make it more readable.

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -4,9 +4,15 @@
...
 using System.Diagnostics;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using AutoRest.CSharp.Generation.Types;
 using AutoRest.CSharp.Input.Source;
+using AutoRest.CSharp.Mgmt.AutoRest;
+using AutoRest.CSharp.Mgmt.Decorator;
 using AutoRest.CSharp.Output.Builders;
+using AutoRest.CSharp.Output.Models.Shared;
```
ArcturusZhang, line 15: this needs a clean up

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
```diff
@@ -3,22 +3,29 @@
...
 {
     internal record MethodSignature(string Name, FormattableString? Summary, FormattableString? Description, MethodSignatureModifiers Modifiers, CSharpType? ReturnType, FormattableString? ReturnDescription, IReadOnlyList<Parameter> Parameters, IReadOnlyList<CSharpAttribute>? Attributes = null, IReadOnlyList<CSharpType>? GenericArguments = null, IReadOnlyDictionary<CSharpType, FormattableString>? GenericParameterConstraints = null, CSharpType? ExplicitInterface = null, string? NonDocumentComment = null)
         : MethodSignatureBase(Name, Summary, Description, NonDocumentComment, Modifiers, Parameters, Attributes ?? Array.Empty<CSharpAttribute>())
     {
+        public static IEqualityComparer<MethodSignature> ParameterAndReturnTypeEqualityComparer = new MethodSignatureParameterAndReturnTypeEqualityComparer();
+
         public MethodSignature WithAsync(bool isAsync) => isAsync ? MakeAsync() : MakeSync();
 
+        public MethodSignature DisableOptionalParameters()
+            => this with { Parameters = Parameters.Select(p => p.ToRequired()).ToList() };
```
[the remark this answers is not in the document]
live1206: sure

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
live1206: updated

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
live1206: Don't remember why I used `EqualsByName` in the first place, `Equals` works for now.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: sounds good

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
live1206: Updated

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: Updated

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
live1206: Updated

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: Had a long discussion with @m-nash, we would like to avoid the misusage of this within this class even.

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
ArcturusZhang: There are other places that are using EqualsByName as well (other comparer defined in this PR), be sure to change them all

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
ArcturusZhang: OK - we might need to add some document on this method to explicitly document it.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,248 @@
...
+using AutoRest.CSharp.Mgmt.AutoRest;
+using AutoRest.CSharp.Output.Models.Shared;
+using System;
+using System.Linq;
+using AutoRest.CSharp.Mgmt.Decorator;
+using System.Diagnostics.CodeAnalysis;
+
+namespace AutoRest.CSharp.Output.Models.Types
+{
+    internal class SignatureType
```
ArcturusZhang, line 16: The name of this type is kind of hard to understand, could we add some xml doc to explain the purpose?
Maybe we need to come up with a new name for this type

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -258,6 +276,31 @@ public bool Equals(Parameter? x, Parameter? y)
...
+                {
+                    return true;
+                }
+
+                if (x is null || y is null)
+                {
+                    return false;
+                }
+
+                var result = x.Type.EqualsByName(y.Type) && x.Name == y.Name;
```
live1206, line 294: This needs to be `EqualsByName`

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,247 @@
...
+                {
+                    continue;
+                }
+
+                if (!CurrentContainAllPreviousParameters(previousMethod, item))
+                {
+                    continue;
+                }
+
+                if (!previousMethod.ReturnType.EqualsByName(item.ReturnType))
```
live1206, line 88: This needs to be `EqualsByName`

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
live1206: There are other 2 places using `EqualsByName` can't be replaced by `Equals`, because the types having different implementations, which is compared in [Equals](https://github.com/Azure/autorest.csharp/blob/178d7ba7b164d702a2f66468e7e38ef8c0f10168/src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs#L118)

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: Added documentation

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: Added

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
ArcturusZhang: Ah I see.
Could you move the EqualsByName to the `CSharpType` class? It is now an extension method.
Double confirm: this method also checks if they have the same namespace as well as the name?

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,254 @@
...
+    ///     - custom
+    ///     - baseline contract
+    ///     current union custom compare with baseline contract outputs the changeset, we can apply different rules with it.
+    /// </summary>
+    internal class SignatureType
+    {
+        private readonly string _defaultNamespace;
+        private readonly string _defaultName;
+        private readonly SignatureType? _customization;
+        private readonly SignatureType? _baselineContract;
```
ArcturusZhang, line 28: I am confused why we have to make this type recursive?
For instance, do we ever need the `OverloadMethods` property on this `_customization` field?
It feels like we do not need everything on SignatureType for customization - then maybe we should not choose `SignatureType` as the type of `customization`.

ArcturusZhang: For example, I only see `Methods` property is used on `_customization`.
Does it mean this `_customization` should just be `_customizationMethods` with a type of `IReadOnlyList<MethodSignature>`?
Same comment applies to `_baselineContract` as well

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,254 @@
...
+                if (TryGetPreviousMethodWithLessOptionalParameters(current, previous, out var currentMethodToCall, out var missingParameters))
+                {
+                    overloadMethods.Add(new OverloadMethodSignature(currentMethodToCall, previous.DisableOptionalParameters(), missingParameters, previous.Description, true));
+                }
+            }
+            return overloadMethods;
+        }
+
+        private HashSet<MethodSignature>? _methodsToSkip;
+        public HashSet<MethodSignature> MethodsToSkip => _methodsToSkip ??= EnsureMethodsToSkip();
```
ArcturusZhang, line 71: maybe `ImmutableHashSet` is better here since we do not have a `IReadOnlySet`

ArcturusZhang: Oh I was wrong - we have a `IReadOnlySet<T>` interface and here we should use `IReadOnlySet<T>` as the type

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: Yes, only `Methods` is in use for now. We will have `Properties` later, so there will be `_customMethods` and `_customProperties`, it grows every time we add a new member.
It would be more convenient to hold them in the same type for custom recursively.

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
live1206: It doesn't, is namespace part of type name here?

live1206: No, it's not, I will add namespace to the method.

live1206: Updated

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: Updated

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: Updated

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -257,5 +257,42 @@ internal static CSharpType FromSystemType(BuildContext context, Type type, IEnum
...
+                return false;
+
+            if (Arguments.Length != other.Arguments.Length)
+                return false;
+
+            for (int i = 0; i < Arguments.Length; i++)
+            {
+                if (Arguments[i].Name != other.Arguments[i].Name)
+                    return false;
+            }
```
ArcturusZhang, line 293: should we here recursively check `EqualsByName` on all the arguments? Here we only check their names.

live1206: nice catch, updated.

PR #3698, opened 2023-08-29
fix(common): random error due to repetitive loading configuration

### LINKED ISSUE #3697: Random concurrent issue when accessing configuration.IntrinsicTypesToTreatEmptyStringAsNull
Almost all of us have met this random issue: when generating code, there will be such duplicate item in a HashSet, which shouldn't.
image

Very likely that two threads are updating the same `configuration.IntrinsicTypesToTreatEmptyStringAsNull` which is a global static instance (which means not thread safe).

Personally I met another issue once on my local machine, when accessing `configuration.IntrinsicTypesToTreatEmptyStringAsNull` at https://github.com/Azure/autorest.csharp/blob/f04ade5c151eda6c61ea401ae6516926391332a7/src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs#L451, with error
image

Error message shows two threads are modifying the configuration.

**Some observation but might not be the root cause:**
Which makes `IntrinsicTypesToTreatEmptyStringAsNull` special is https://github.com/Azure/autorest.csharp/blob/f04ade5c151eda6c61ea401ae6516926391332a7/src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs#L161

archerzz:
# Description

We have mulitple `IPlugin` which could load configuration multiple times. For example, in `src/AutoRest.CSharp/readme.md`, we load plugins `csharpgen` and `csharpproj`. The bad thing is that plugin invocation is async, so loading configuration could be concurrent. And unfortunately the configuration loading is not concurrency safe. For example, `_intrinsicTypesToTreatEmptyStringAsNull.UnionWith(additionalIntrinsicTypesToTreatEmptyStringAsNull);` is not thread-safe.

To fix that, we should make sure the configuration is loaded only once. This commit uses a flag `IsInitialized` and a lock `InitLock` to make sure the configuration initialization is executed only once and exclusively.

fix #3697

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
+++ b/src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
@@ -17,6 +17,10 @@ namespace AutoRest.CSharp.Input
     {
         internal static readonly string ProjectFolderDefault = "../";
 
+        private static readonly object InitLock = new object();
+
+        private static bool IsInitialized = false;
+
         public static class Options
         {
             public const string OutputFolder = "output-folder";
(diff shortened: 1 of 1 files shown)


pshao25:
1. Lock on `modelsToTreatEmptyStringAsNull` is enough, we don't have to lock all the initialize procedure.
2. I personally don't like to introduce a lock. Maybe we could 1) try to revisit the configuration why there must be both `modelsToTreatEmptyStringAsNull` and `additionalIntrinsicTypesToTreatEmptyStringAsNull`, 2) revisit the design to reduce the times of initialization.

archerzz:
> 1. Lock on `modelsToTreatEmptyStringAsNull` is enough, we don't have to lock all the initialize procedure.
> 2. I personally don't like to introduce a lock. Maybe we could 1) try to revisit the configuration why there must be both `modelsToTreatEmptyStringAsNull` and `additionalIntrinsicTypesToTreatEmptyStringAsNull`, 2) revisit the design to reduce the times of initialization.

1. The whole `Configuration` instance is a contended data. All properties initialization is not thread-safe, according to the .NET memory model. We cannot only fix what we met.
2. There are not many threads access this area. Probably only 2 and the number will not increase, since we're moving away from swagger. The cost of lock is acceptable. Reducing the time of initialization does not fix the problem, it just reduces the probability.

archerzz:
preview PR: https://github.com/Azure/azure-sdk-for-net/pull/38424

archerzz:
I didn't expect that `Configuration` can be loaded multiple times in tests, e.g. `OutputLibraryTestBase`. So, restricting `Configuration` to load only once will break test cases. We need to refactor how we access `Configuration` which I filed an issue #3699 to track.

For this pr, the focus is to provide an urgent fix.

archerzz:
close this, in favor of #3703

PR #3701, opened 2023-08-30
Exploring what partial shared classes look like

annelo-msft:
Work in progress


m-nash:
Prefer we close in favor of this solution here https://github.com/Azure/autorest.csharp/pull/3434/commits/8d3e7c211e741ee4d517a2daf96883d856cb0934#diff-c3ec2274d241a0d6bc4ae6c34e926e310082c290372840042c96bdb94e71cf1e

PR #3703, opened 2023-08-31
fix(configuration): `Configuration` is not thread-safe

### LINKED ISSUE #3697: Random concurrent issue when accessing configuration.IntrinsicTypesToTreatEmptyStringAsNull

archerzz:
# Description

- add a new record `CSharpProjConfiguration` which hosts the minimal set of properties used by `CSharpProj` plugin
- refactor `CSharpProj` plugin to read configuration from `CSharpProjConfiguration` instead of `Configuration`
- refactor configuration initialization logic to support load `CSharpProjConfiguration` independently
- run OutputLibraryTest sequentially to prevent potential concurrency issue because each `OutputLibraryTestBase` test will load different `Configuration`

fix #3697

# Checklist

# Ready to Land?


archerzz:
Preview PR: https://github.com/Azure/azure-sdk-for-net/pull/38443

There is no code change.

pshao25:
The idea looks good. Could we have an e2e test to check configuration.init only called once?

archerzz:
> The idea looks good. Could we have an e2e test to check configuration.init only called once?

The problem is that it's triggered from autorest. I can do the test offline with some changes which cannot be applied in the official codes.

pshao25 (approved):

test/AutoRest.TestServer.Tests/Mgmt/OutputLibrary/OutputLibraryTestBase.cs
```diff
@@ -22,6 +22,7 @@
 
 namespace AutoRest.TestServer.Tests.Mgmt.OutputLibrary
```
lirenhe, line 23: I would like to see more tests on CSharpProjConfiguration

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -596,5 +599,57 @@ private static bool ReadOption(JsonElement root, string option)
...
+                IsMgmtTestProject: MgmtTestConfiguration.HasMgmtTestConfiguration(autoRest),
+                LibraryName: GetLibraryNameOption(autoRest, defaultLibraryName),
+                Namespace: GetNamespaceOption(autoRest, defaultNamespace),
+                SkipCSProjPackageReference: GetSkipCSProjPackageReferenceOption(autoRest),
+                RelativeProjectFolder: relativeProjectFolder,
+                Generation1ConvenienceClient: GetGeneration1ConvenienceClientOption(autoRest)
+            );
+        }
+
+        private static string GetOutputFolderOption(IPluginCommunication autoRest) => TrimFileSuffix(GetRequiredOption<string>(autoRest, Options.OutputFolder));
```
lirenhe, line 635: Could you help to explain why we need those internal functions?

src/AutoRest.CSharp/Common/Output/Models/Types/BuildContext.cs
```diff
@@ -9,20 +9,25 @@ namespace AutoRest.CSharp.Output.Models.Types
...
         }
 
         public OutputLibrary? BaseLibrary { get; protected set; }
 
         public CodeModel CodeModel { get; }
         public SchemaUsageProvider SchemaUsageProvider { get; }
         public string DefaultName => CodeModel.Language.Default.Name;
-        public string DefaultNamespace => Configuration.Namespace;
-        public string DefaultLibraryName => Configuration.LibraryName;
+        public string DefaultNamespace { get; }
```
chunyu3, line 29: It will decouple the `DefaultName` and `DefaultNamespace` in BuildContext from the Configuration. When the value in Configuration is changed, the BuildContext will not aware it.

Need to investigate if the changes in Configuration is needed to reflect the BuildContext or not.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -596,5 +599,57 @@ private static bool ReadOption(JsonElement root, string option)
...
+        private static string? GetProjectFolderOption(IPluginCommunication autoRest) => autoRest.GetValue<string?>(Options.ProjectFolder).GetAwaiter().GetResult();
+        private static bool GetAzureArmOption(IPluginCommunication autoRest) => GetOptionBoolValue(autoRest, Options.AzureArm);
+        private static bool GetSkipCSProjPackageReferenceOption(IPluginCommunication autoRest) => GetOptionBoolValue(autoRest, Options.SkipCSProjPackageReference);
+        private static bool GetGeneration1ConvenienceClientOption(IPluginCommunication autoRest) => GetOptionBoolValue(autoRest, Options.Generation1ConvenienceClient);
+        private static string GetLibraryNameOption(IPluginCommunication autoRest, string defaultLibraryName) => autoRest.GetValue<string?>(Options.LibraryName).GetAwaiter().GetResult() ?? defaultLibraryName;
+        private static string GetNamespaceOption(IPluginCommunication autoRest, string defaultNamespace) => autoRest.GetValue<string?>(Options.Namespace).GetAwaiter().GetResult() ?? defaultNamespace;
     }
+
+    // configuration properties neede by CSharpProj
+    public record CSharpProjConfiguration(
```
chunyu3, line 645: Shall we separate the `CSharpProjConfiguration` from Configuration. It may not comply c# style, a public struct/class need to be in a .cs file with the same name.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
archerzz: It's a refactor to extract common codes. See the changes on other places in `Configuration`.

src/AutoRest.CSharp/Common/Output/Models/Types/BuildContext.cs
archerzz: The preview PR doesn't include any code change. I think we're fine here.

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -62,7 +62,7 @@ public static async Task RunAsync(CommandLineOptions options)
                 workspace = await new CSharpGen().ExecuteAsync(codeModel);
                 if (options.IsNewProject)
                 {
-                    new CSharpProj().Execute(Configuration.Namespace, outputPath, (yaml.Contains("x-ms-format: dfe-", StringComparison.Ordinal)));
+                    new CSharpProj().Execute(Configuration.Namespace, outputPath, (yaml.Contains("x-ms-format: dfe-", StringComparison.Ordinal)), Configuration.ToCSharpProjConfiguration());
```
ArcturusZhang, line 65: I assume since part of your PR is making the test cases running sequentially, the concurrent issue is resolved.
Therefore the changes related with the new `CSharpProjConfiguration` is purely refactor and not resolving any issue, is this correct?
The idea of making the configuration class static is "to avoid the fact that we have to passing an instance around everywhere", which is exactly the thing we are doing right here. See the PR here: https://github.com/Azure/autorest.csharp/pull/2005
What is the advantage of having this new record of configuration?
Or if we have to have it, we could let the `Configuration` static class to hold it internally and redirect its values in static properties.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -596,5 +599,57 @@ private static bool ReadOption(JsonElement root, string option)
...
+                AbsoluteProjectFolder: absoluteProjectFolder,
+                AzureArm: GetAzureArmOption(autoRest),
+                IsMgmtTestProject: MgmtTestConfiguration.HasMgmtTestConfiguration(autoRest),
+                LibraryName: GetLibraryNameOption(autoRest, defaultLibraryName),
+                Namespace: GetNamespaceOption(autoRest, defaultNamespace),
+                SkipCSProjPackageReference: GetSkipCSProjPackageReferenceOption(autoRest),
+                RelativeProjectFolder: relativeProjectFolder,
+                Generation1ConvenienceClient: GetGeneration1ConvenienceClientOption(autoRest)
+            );
+        }
```
ArcturusZhang, line 633: What is the difference between this method and the method `Initialize` above?

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
```diff
@@ -596,5 +599,57 @@ private static bool ReadOption(JsonElement root, string option)
...
+        (
+            AbsoluteProjectFolder: AbsoluteProjectFolder,
+            AzureArm: AzureArm,
+            IsMgmtTestProject: MgmtTestConfiguration is not null,
+            LibraryName: LibraryName,
+            Namespace: Namespace,
+            SkipCSProjPackageReference: SkipCSProjPackageReference,
+            RelativeProjectFolder: RelativeProjectFolder,
+            Generation1ConvenienceClient: Generation1ConvenienceClient
+        );
```
ArcturusZhang, line 614: Considering the comment I left above, if we have to have this new record for "a part of" configurations, we should have it the other way around. Something like:
```csharp
public static class Configuration
{
	private static CSharpProjConfiguration _csharpProjConfig;
	public static string AbsoluteProjectFolder => _csharpProjConfig?.AbsoluteProjectFolder ?? throw new InvalidOperationException("The configuration has not been initialized");
}
```

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
pshao25: > Therefore the changes related with the new CSharpProjConfiguration is purely refactor and not resolving any issue, is this correct?

Not correct. See the PR title. It is resolving the problem when two threads initialize configuration together. So I'm afraid your below solution is not going to work.

test/AutoRest.TestServer.Tests/Mgmt/OutputLibrary/OutputLibraryTestBase.cs
archerzz: added

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
archerzz: done

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
ArcturusZhang: In the actual executions, our generator never runs in parallel (they should be running in different process). And we are making the tests to run in sequence. Therefore the issue should be gone.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
archerzz: In `StandaloneRunner`, we can directly convert ` Configuration` into `CSharpConfiguration`, instead of initializing twice.

src/AutoRest.CSharp/Common/AutoRest/Plugins/Configuration.cs
archerzz: `CSharpProjConfiguration` doesn't need to be a singleton, because it is used by `CSharpProj` which is a small plugin which doesn't have many references to the configuration.

chunyu3 (approved):

ArcturusZhang (approved):

PR #3711, opened 2023-09-04
refactor the serialization format so that the logic for it is no longer scattered everywhere

ArcturusZhang:
# Description

The `SerializationFormat` is not provided by the emitter and in our generator, we need it in serialization.

Previously this is calculated when we build the information for serializations, therefore in properties, we could never get it. But when we are writing the samples, we need this information.

This PR unifies the calculation of `SerializationFormat` for properties and parameters and make sure this information will be stored in the properties or parameters.

The sample generation needs this `SerializationFormat` information, therefore I made this PR to provide it.

# Checklist

# Ready to Land?


test/CadlRanchProjects/encode/bytes/src/Generated/Models/Base64urlArrayBytesProperty.cs
```diff
@@ -35,29 +35,15 @@ internal Base64urlArrayBytesProperty(IList<BinaryData> value)
...
-        /// </item>
-        /// <item>
-        /// <term>BinaryData.FromObjectAsJson(new { key = "value" })</term>
-        /// <description>Creates a payload of { "key": "value" }.</description>
-        /// </item>
-        /// <item>
-        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
-        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// <term>BinaryData.FromBytes(new byte[] { 1, 2, 3 })</term>
+        /// <description>Creates a payload of "AQID".</description>
```
ArcturusZhang, line 46: This change is expected.
The item of this list has an encode of base 64 string. Therefore our help message should write this information just like we did in the normal property here: https://github.com/ArcturusZhang/autorest.csharp/blob/2ff52249838b64f59584cbfa07e9ac5563522429/test/CadlRanchProjects/encode/bytes/src/Generated/Models/Base64BytesProperty.cs#L29

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
```diff
@@ -45,7 +45,75 @@ internal record InputParameter(
...
+            arraySerializationDelimiter,
+            headerCollectionPrefix)
+    {
+        _serializationFormat = serializationFormat;
+    }
+
+    private SerializationFormat? _serializationFormat;
+    public SerializationFormat SerializationFormat => _serializationFormat ??= GetSerializationFormat(Type, Location);
+
+    private static SerializationFormat GetSerializationFormat(InputType parameterType, RequestLocation requestLocation)
```
chunyu3, line 96: When a InputParameter is converted from codeModel(modelerFour), the _serializationFormat is not null, this function will not be called. We may need the strategy for parameter from swagger, also.

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
```diff
@@ -45,7 +45,75 @@ internal record InputParameter(
...
+            isApiVersion,
+            isResourceParameter,
+            isContentType,
+            isEndpoint,
+            skipUrlEncoding,
+            explode,
+            arraySerializationDelimiter,
+            headerCollectionPrefix)
+    {
+        _serializationFormat = serializationFormat;
```
chunyu3, line 90: Shall we set the  _serializationFormat here? Since we will re-calculate it when use the property, we will not need to initialize it.

This internal constructor is not needed. When convert an inputParameter from modelerFour, we just leave the `_serializationFormat` null, and re-calculate it when use.

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
```diff
@@ -45,7 +45,75 @@ internal record InputParameter(
...
+        bool isRequired,
+        bool isApiVersion,
+        bool isResourceParameter,
+        bool isContentType,
+        bool isEndpoint,
+        bool skipUrlEncoding,
+        bool explode,
+        string? arraySerializationDelimiter,
+        string? headerCollectionPrefix,
+        SerializationFormat serializationFormat) : this(
```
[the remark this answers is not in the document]
ArcturusZhang: No. If we define the property here, we cannot have the getter implementation as we are having now.
This ctor is a workaround especially for the code model converter when we are converting code model types to input types.
I do not want to change much on that side therefore I introduced this as a seaming layer to keep everything unchanged for that part.

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
ArcturusZhang: This is not the purpose. See the comment below

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
ArcturusZhang: Ideally, we should have everything working in the same way, which is: calculate the serialization format when we first access it based on its type.
The reason we have to have this is that in the code model converter, it is calculating the serialization format in the converter and passed it in. I do not want to change behavior there therefore I introduced this to make sure at least on the DPG side everything is working fine (nothing changed in the generated code proves it).
I do not want to dig into that part too much because this is a dependency for the sample refactor.

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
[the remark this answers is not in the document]
ArcturusZhang: In addition, I hope the definition of our records could follow this rule:
If a property is not deserialized from the tspCodeModel, we should not put it into the record definition to emphasize.
In this case, serialization format is not deserialized from tspCodeModel, therefore I would like to define it separately.

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
ArcturusZhang: I have a try and updated the code. Now code model converter no longer needs to pass in the serialization format calculated elsewhere. Now everything is calculated in `InputParameter` and nothing changed.
For more details if you are interested, you could check above hidden comment.

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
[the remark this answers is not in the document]
ArcturusZhang: **UPDATE** I updated the `CodeModelConverter` class now we no longer need it to pass in a `SerializationFormat` value to the parameter.

chunyu3 (approved):

ArthurMa1978 (approved):

Issue #3630, opened 2023-08-01
Sample generator should not throw exception when hit example errors


ArthurMa1978:
For example, a property is required but not given a value in the example, the sample generator can handle this by providing a default value and adding a comment that says this property needs a value, and mentions the example in swagger should be fixed.

ArcturusZhang:
FYI the related implementation could be found here: https://github.com/Azure/autorest.csharp/blob/0220ddc134d4e799b838c7e9e6d8783aaf45abd8/src/AutoRest.CSharp/MgmtTest/Models/MockTestCase.cs#L105

PR #3728, opened 2023-09-09
Use default value instead of throwing exception in samplegen for method parameter without example value

RodgeFu:
Fixes https://github.com/Azure/autorest.csharp/issues/3630

# Checklist

# Ready to Land?


src/AutoRest.CSharp/MgmtTest/Models/MockTestCase.cs
ArcturusZhang, line 1: Could we add a case in the MgmtMockAndSample test project to see the change we made here?

src/AutoRest.CSharp/MgmtTest/Models/MockTestCase.cs
```diff
@@ -100,9 +101,23 @@ private IEnumerable<Parameter> GetAllPossibleParameters()
...
-                    // if this is a required parameter and we did not find the corresponding parameter in the examples, we throw an exception.
+                    // if this is a required parameter and we did not find the corresponding parameter in the examples
                     if (parameter.DefaultValue == null)
-                        throw new InvalidOperationException($"Cannot find an example value for required parameter `{parameter.Name}` in example {Name}");
+                    {
+                        // this parameter is not from body which means its type should be primary which means "default" keyword should be able to handle
+                        // the default value (also string because we disabled nullable in generated code)
+                        var warning = $"No value is provided for {parameter.Name} in example '{this.Name}'. Please consider adding a proper example value for it in swagger";
+                        AutoRestLogger.Warning(warning).Wait();
+                        var pv = new ExampleParameterValue(parameter, $"default; /* Warning: {warning}*/");
```
ArcturusZhang, line 111: The ExampleParameterValue could be written inline or standalone depending on the context it is used. We could not know that here because we do not have the context because this method is only providing the data.
For instance, if we have this value, and when this parameter is used in a method, we will write code like this:
```csharp
var response = client.DoSomething(default; /* Warning: {warning}*/);
```
which certainly will not compile. This might work when the parameter is not inline, but this will definitely break when the arameter is not inline.

If we just want to fix the issue, we could just create a `ExampleParameterValue` with the default value of this type in it. Something like:
```csharp
object? value = parameter.Type is {IsFramewrokType: true} && parameter.Type.IsValueType ?
    Activator.CreateInstance(parameter.Type.FrameworkType) : null;
exampleValue = new ExampleValue(parameter.Type, value);
}
```
then everything should be fine to use for the other part of the code. But the issue of this solution is that the rest part of the code never know it should add a comment to warn that something wrong happens.

src/AutoRest.CSharp/MgmtTest/Models/MockTestCase.cs
RodgeFu: added

src/AutoRest.CSharp/MgmtTest/Models/MockTestCase.cs
RodgeFu: update to use "default /* Warning: {warning}*/" which supports inline scenario

test/TestProjects/MgmtMockAndSample/specification/mockSwagger/examples/createVault.json
```diff
@@ -1,7 +1,6 @@
 {
   "parameters": {
     "resourceGroupName": "sample-resource-group",
-    "vaultName": "sample-vault",
```
ArcturusZhang, line 4: How about we duplicate this example, and remove this value in the new example, so that we could leave some comment on the name of the example to avoid future confusions

ArcturusZhang (approved):

test/TestProjects/MgmtMockAndSample/specification/mockSwagger/examples/createVault.json
RodgeFu: updated

ArthurMa1978 (approved):

Issue #3502, opened 2023-06-15
The location of endpoint paramter changes in the client constructor


chunyu3:
The location of endpoint parameter changes according to whether it has default value or not.
 
e.g.

- endpoint has default value

public ParametersLowlevelClient(AzureKeyCredential credential, Uri endpoint, ParametersLowlevelClientOptions options)

- endpoint has no default value
public ParametersLowlevelClient(Uri endpoint, AzureKeyCredential credential, ParametersLowlevelClientOptions options)

It may not consistent among sdks.

Shall we place endpoint as the first parameter always?

chunyu3:
As discussed, we will put `endpoint` parameter the first one, before credential and options.

PR #3737, opened 2023-09-14
Endpoint order

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3502

- order the parameter, and put endpoint the first parameter.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -156,13 +156,13 @@ private IEnumerable<ConstructorSignature> BuildPrimaryConstructors(IReadOnlyList
...
             {
-                yield return CreatePrimaryConstructor(requiredParameters.Concat(optionalToRequired).ToArray());
+                yield return CreatePrimaryConstructor(requiredParameters.Concat(optionalToRequired).OrderBy(parameter => parameter.Name != "endpoint").ToArray());
             }
             else
             {
                 foreach (var credentialField in Fields.CredentialFields)
                 {
-                    yield return CreatePrimaryConstructor(requiredParameters.Append(CreateCredentialParameter(credentialField!.Type)).Concat(optionalToRequired).ToArray());
+                    yield return CreatePrimaryConstructor(requiredParameters.Append(CreateCredentialParameter(credentialField!.Type)).Concat(optionalToRequired).OrderBy(parameter => parameter.Name != "endpoint").ToArray());
```
ArcturusZhang, line 165: this reads so much like a "patch" to our code.
Could we come up with a solution that does not look like a patch?

I heard the root cause of this is that in swagger, the endpoint is always optional. Is it possible we could make it required in the `CodeModelConverter` or anywhere else and then we do not need to do anything here and fix the issue?

chunyu3: hello @ArcturusZhang The root cause is the endpoint is optional or it has default value. We also need to honor the default value and use it to call the secondary constructor in primary constructor. In that way, the endpoint parameter should maintain optional with default value. We cannot just convert it as required, and ignore the default value.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -174,32 +180,45 @@ private IEnumerable<ConstructorSignature> BuildSecondaryConstructors(IReadOnlyLi
...
+                var allarguments = arguments.Concat(optionalParametersArguments);
+                yield return CreateSecondaryConstructor(requiredParameters, allarguments.ToArray());
             }
             else
             {
                 foreach (var credentialField in Fields.CredentialFields)
                 {
-                    yield return CreateSecondaryConstructor(requiredParameters.Append(CreateCredentialParameter(credentialField!.Type)).ToArray(), optionalParametersArguments);
+                    var credentialParameter = CreateCredentialParameter(credentialField!.Type);
+                    var allarguments = arguments.Concat(new List<FormattableString>() { $"{credentialParameter.Name}" }).Concat(optionalParametersArguments);
```
lirenhe, line 211: Should it be 'allArguments'?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -156,13 +156,19 @@ private IEnumerable<ConstructorSignature> BuildPrimaryConstructors(IReadOnlyList
 
             if (Fields.CredentialFields.Count == 0)
             {
-                yield return CreatePrimaryConstructor(requiredParameters.Concat(optionalToRequired).ToArray());
+                /* order the constructor paramters as (endpoint, requiredParameters, OptionalParamters).
+                 * use the OrderBy to put endpoint as the first parameter.
+                 * */
+                yield return CreatePrimaryConstructor(requiredParameters.Concat(optionalToRequired).OrderBy(parameter => parameter.Name != "endpoint").ToArray());
```
lirenhe, line 162: Do we need to care about the case when comparing 'endpoint'?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
chunyu3: resolved.

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
chunyu3: `endpoint` is current hardcoded. but it is better to compare case-insensitive. updated. Thanks,

ArcturusZhang (approved):

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -156,13 +156,19 @@ private IEnumerable<ConstructorSignature> BuildPrimaryConstructors(IReadOnlyList
 
             if (Fields.CredentialFields.Count == 0)
             {
-                yield return CreatePrimaryConstructor(requiredParameters.Concat(optionalToRequired).ToArray());
+                /* order the constructor paramters as (endpoint, requiredParameters, OptionalParamters).
+                 * use the OrderBy to put endpoint as the first parameter.
+                 * */
+                yield return CreatePrimaryConstructor(requiredParameters.Concat(optionalToRequired).OrderBy(parameter => !parameter.Name.Equals("endpoint", StringComparison.OrdinalIgnoreCase)).ToArray());
```
pshao25, line 162: We have a lot of places directly use "endpoint". I would suggest make it a constant variable.

pshao25 (approved):

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
ArcturusZhang: actually we could reference the `KnownParameters` where we have a parameter created for endpoint

chunyu3: Yes. Maybe we need to file a MQ task to update `endpoint` to `KnowParameters.Endpoint.Name`. And after migrate mgp to InputModel, we will update this together.

lirenhe (approved):

PR #3739, opened 2023-09-14
Handle breaking change of additional optional parameters

live1206:
This PR introduces SignatureTypeProvider, it holds three portions of code:
- current
- custom
- previous contract

It provides 
- OverloadingMethods: for now represents the missing methods in previous contract due to additional optional parameters in the new contract, this is causing breaking changes


live1206:
Merged the change into #3686

Issue #3744, opened 2023-09-18
The class name for the `cref` should include the full namespace


ArthurMa1978:
The document explains the derived classes of a base class parameter in the model factory, like this:

```
        /// <param name="providerSpecificDetails">
        /// The migration provider custom settings.
        /// Please note <see cref="MigrationProviderSpecificSettings"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
        /// The available derived classes include <see cref="VMwareCbtMigrationDetails"/>.
        /// </param>
```

However, if `VMwareCbtMigrationDetails` changes in a new release, a new overload method will be added in the model factory class, which will cause a build error:

```
error CS0419: Ambiguous reference in cref attribute
```

This is because the `cref` here is mistaken for the method name in the model factory instead of the class name. 
To fix this, we need to add the namespace.:

```
<see cref="Models.VMwareCbtMigrationDetails>
```

PR #3747, opened 2023-09-18
fix the class name for cref element to include full namespace

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3744

# Description

Previously, in some cases when our generator wants to write a type into `cref`, it will not write its full name.
The discriminator extra descriptions have this issue. The comment there is written like:
```csharp
/// please note that <see cref="Foo"/> is the base class
```
This comment is in the class of `Foo` as well as the entry for `Foo` class in the model factory.

It is OK to have this in the `Foo` class file, but it is actually incorrect in the model factory because `cref` is actually pointing to the method `Foo` defined in the model factory.

A worse case happens when our model factory must have customization to mitigate a breaking change. When this happens, we have two methods with `Foo` as their names (they are overloads to each other), but the cref now will not compile any more, because the compiler is complaining the ambiguity.

This PR changes the type of `description` in multiple places from `string` to `FormattableString`, so that the type information could be properly stored in the `FormattableString` and handled eventually and correctly by `CodeWriter`. Otherwise despite the description is built from a formattable string, it evaluates its value and invokes the `ToString` method on CSharpType which has issues of new lines etc.

In the changes this PR made, we always write full name in crefs, such as
```
/// Please <see cref="global::NamespaceOfFoo.Foo"/>
```
and let Roslyn to trim off whatever it assumes necessary to trim off. There might be some additional namespace added in the test projects and regen, but since Roslyn thinks it is necessary to keep, it is OK to keep them - they do not really harm anything anyway.

# Checklist

# Ready to Land?


ArthurMa1978 (approved):

live1206:
Should we handle the Description in [MethodSignature](https://github.com/Azure/autorest.csharp/blob/712159bdc66d48d01a6bb7b5302df3600f964b79/src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs#L15C75-L15C86) as well? It could be types with same name in different namespaces.

ArcturusZhang:
> Should we handle the Description in [MethodSignature](https://github.com/Azure/autorest.csharp/blob/712159bdc66d48d01a6bb7b5302df3600f964b79/src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs#L15C75-L15C86) as well? It could be types with same name in different namespaces.

I will update that as well.

ArcturusZhang:
> Should we handle the Description in [MethodSignature](https://github.com/Azure/autorest.csharp/blob/712159bdc66d48d01a6bb7b5302df3600f964b79/src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs#L15C75-L15C86) as well? It could be types with same name in different namespaces.

well I tried to update the methodsignature.Summary and description property to FormattableString, and finding more of similar cases.
This might make this PR big and out of its initial purpose.
I will create an issue to track this possible refactor item.

live1206:
> > Should we handle the Description in [MethodSignature](https://github.com/Azure/autorest.csharp/blob/712159bdc66d48d01a6bb7b5302df3600f964b79/src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs#L15C75-L15C86) as well? It could be types with same name in different namespaces.
> 
> well I tried to update the methodsignature.Summary and description property to FormattableString, and finding more of similar cases. This might make this PR big and out of its initial purpose. I created an issue to track this possible refactor item: #3750

I was expecting more similar cases, we can limit this PR to just fix it for parameter.

Issue #3754, opened 2023-09-19
Support conditional request header in convenience method


chunyu3:
Currently we will convert etag to MatchCondition and RequestCondition in protocol method, but in convenience method it is not.

PR #3755, opened 2023-09-19
on board cadl-ranch conditional request test

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3754

- onboard cadl-ranch conditional request test project
- convert to Etag for if-match/if-non-match parameter in convenience method

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -590,7 +590,7 @@ private void AddParameter(string name, InputParameter inputParameter, CSharpType
                 return;
             }
 
-            var convenienceMethodParameter = BuildParameter(inputParameter);
+            var convenienceMethodParameter = BuildParameter(inputParameter, frameworkParameterType);
```
lirenhe, line 593: why do we introduce this change?  I didn't find the place we use this change in the PR.

chunyu3: It is to resolve issue https://github.com/Azure/autorest.csharp/issues/3754. In the cadl-ranch test, the operation is simple which only has one head parameter, and it is not need to generate convenience method. So it is not reflect the fix there.

test/CadlRanchProjects/special-headers/conditional-request/src/Generated/ConditionalRequestClient.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// </description>
+        /// </item>
+        /// </list>
+        /// </summary>
+        /// <param name="ifMatch"> The request should only proceed if an entity matches this string. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns> The response returned from the service. </returns>
+        /// <include file="Docs/ConditionalRequestClient.xml" path="doc/members/member[@name='PostIfMatch(ETag?,RequestContext)']/*" />
+        public virtual Response PostIfMatch(ETag? ifMatch = null, RequestContext context = null)
```
ArcturusZhang, line 98: I am curious why we do not generate it using `RequestCondition`?
I assume when we have if-match, etc, we will encapsulate them in a `RequestCondition` model previously, or am I remember this wrong?

chunyu3: hello @ArcturusZhang There are three scenarios according to the existence of four conditional request header (if-match, if-not-match, ifModifiedSince, ifUnmodifiedSince):

1. If only has one of if-match/if-not-match header, it will use `Etag`
2. If has if-match and if-not-match, it will use `MatchCondition`
3. Others use `RequestCondition`

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
pshao25: But I don't see the convenience generated. Shall we add one that could generate convenience in cadl-ranch?

test/CadlRanchProjects/special-headers/conditional-request/src/Generated/ConditionalRequestClient.cs
pshao25: If so, why there is only `Etag`. Seems the coverage in cadl-ranch is not enough?

ArcturusZhang: i see, thank you!
If we could add a test project to handle the other combinations, it would be the best. Please ignore this if we already have one.

chunyu3: We already has a test project `CollapseRequestCondition-LowLevel` which is from swagger to test all those scenarios, but there is no such project from typespec.
But the combination/transformation is in autorest.csharp codegen side, it should be the same for swagger and typespec.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
chunyu3: actually, the issue that convenience method does not recognize the framework type as protocol method does is a common defect, and it is not related to Conditional Reuqest. Conditional Request is just a case which has the problem.

lirenhe (approved):

pshao25 (approved):

PR #3760, opened 2023-09-20
Expression infrastructure + serialization logic

AlexanderSher:


src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -163,12 +214,16 @@ internal static CSharpType FromSystemType(Type type, string defaultNamespace, So
...
+            if (!IsFrameworkType)
+                return false;
 
-            return IsFrameworkType ? new CSharpType(FrameworkType, false, Arguments) : new CSharpType(Implementation, Namespace, Name, IsValueType, IsEnum, false, Arguments);
+            return FrameworkType.Equals(typeof(IList<>)) ||
+                FrameworkType.Equals(typeof(IEnumerable<>)) ||
+                FrameworkType == typeof(IReadOnlyList<>) ||
+                FrameworkType.Equals(typeof(IDictionary<,>)) ||
+                FrameworkType == typeof(IReadOnlyDictionary<,>);
         }
```
ArcturusZhang, line 227: since we have another `IsCollectionType(CSharpType)` method in `TypeFactory` as well and it is widely used in a lot of places, do we really need this new one?

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,912 @@
...
+
+                if (format == SerializationFormat.Duration_Seconds_Float)
+                {
+                    return TimeSpanExpression.FromSeconds(element.GetDouble());
+                }
+
+                return element.GetTimeSpan(format.ToFormatSpecifier());
+            }
+
+            throw new NotSupportedException($"Framework type {frameworkType} is not supported");
```
ArcturusZhang, line 881: There are a few RPs failing in the regen. I did not go through all of them, but one of them ([log](https://dev.azure.com/azure-sdk/590cfd2a-581c-4dcb-a12e-6568ce786175/_apis/build/builds/3101210/logs/266)) is complaining this:
Exception: Framework type Azure.ResourceManager.Resources.Models.ExtendedLocationType is not supported
Considering the existence of our customization code, there are possibilities that we could put any framework type in the model, and we should not throw here.
Here is the model: https://github.com/Azure/azure-sdk-for-net/blob/5d5acaabce95ab28faee3c1eb6d164bf7b7d9442/sdk/resourcemanager/Azure.ResourceManager/src/Resources/Generated/Models/ExtendedLocationType.cs#L14
it is an extensible enum.
How do we previously handle this type?

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
```diff
@@ -56,6 +54,9 @@ private void WriteObjectSerialization(CodeWriter writer, SerializableObjectType
 
             using (writer.Namespace(declaration.Namespace))
             {
+                //[TODO] Azure.Core is not needed to compile, but previous version of writer has been adding it.
+                writer.UseNamespace("Azure.Core");
```
ArcturusZhang, line 58: we keep this in order to reduce the amount of diff in the regen?

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: Well there are 4 errors in total.
2 of them are reporting `ExtendedLocationType`
the other 2 are reporting `RegionType` which is also an extensible enum from `Azure.ResourceManager` assembly.
They should share the same root cause.

src/AutoRest.CSharp/Common/Output/Models/Statements/AssignValueStatement.cs
ArcturusZhang, line 1: since we have quite a few types of `statement` and `expression`, could we put them together in one directory so that they are easy to locate.
as well as those `KnowXXX`
Another consideration on this is that I believe in C# AST, statements are special expressions that returns nothing. Therefore maybe we could put them into a big namespace such as
```
- Expressions
    - Statements
    - ValueExpressions
```

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
ArcturusZhang: well in the regen, I saw that we are actually adding `Azure.Core` namespace in quite a few test projects.
Maybe we could just remove this.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
AlexanderSher: No, and it is not used anywhere. Will remove.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
AlexanderSher: Yes. These types have custom deserialization methods. I need to merge the pre-build check from consolidation (or maybe relax the condition for type representation in expression tree).

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
AlexanderSher: There are much more files where `Azure.Core` will be removed. I will remove this line in separate commit to this PR once everything else is finalized.

src/AutoRest.CSharp/Common/Output/Models/Statements/AssignValueStatement.cs
AlexanderSher: Sure, I can do that.

AlexanderSher: Done

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
AlexanderSher: Removed

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
AlexanderSher: Fixed

ArcturusZhang (approved):

AlexanderSher:
There is one unexpected change in azure-sdk-for-net repo that I need to investigate before merge:

image

AlexanderSher:
> There is one unexpected change in azure-sdk-for-net repo that I need to investigate before merge:
> 
> image

Ok, this one is fixed + couple tests are added to cover this case.

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
```diff
@@ -58,15 +58,19 @@ public EnumType(InputEnumType input, string defaultNamespace, string defaultAcce
             ValueType = typeFactory.CreateType(input.EnumValueType);
             IsStringValueType = ValueType.Equals(typeof(string));
             IsIntValueType = ValueType.Equals(typeof(int)) || ValueType.Equals(typeof(long));
+            IsFloatValueType = ValueType.Equals(typeof(float));
```
ArcturusZhang, line 61: do we need to include `double` here? Could we ever have double value enum?
But even if we never have double value enum, it would not harm if we add it i assume

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
```diff
@@ -58,15 +58,19 @@ public EnumType(InputEnumType input, string defaultNamespace, string defaultAcce
...
+            IsNumericValueType = IsIntValueType || IsFloatValueType;
             SerializationMethodName = IsStringValueType && IsExtensible ? "ToString" : $"ToSerial{ValueType.Name.FirstCharToUpperCase()}";
```
ArcturusZhang, line 63: I assume this is a potential refactor here to let it have a `IEnumerable<Method>` here instead of just having the serialization method name.

src/AutoRest.CSharp/Common/Output/Models/Types/SerializableObjectType.cs
```diff
@@ -22,6 +23,8 @@ protected SerializableObjectType(string defaultNamespace, SourceInputModel? sour
         {
         }
 
+        public INamedTypeSymbol? GetExistingType() => ExistingType;
```
ArcturusZhang, line 26: why we need this method wehn we have the public property `ExistingType`?

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
AlexanderSher: Yes. I have it in consolidation PR where it is done for both `ModelTypeProvider` and `MgmtObjectType`.

src/AutoRest.CSharp/Common/Output/Models/Types/SerializableObjectType.cs
AlexanderSher: `ExistingType` is protected. This method will be become unused when serialization method creation is moved into `ModelTypeProvider`, which is done in consolidation PR.

src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonPropertySerialization.cs
```diff
@@ -1,40 +1,35 @@
...
 using AutoRest.CSharp.Generation.Types;
-using AutoRest.CSharp.Output.Models.Types;
 
 namespace AutoRest.CSharp.Output.Models.Serialization.Json
 {
     internal class JsonPropertySerialization : PropertySerialization
     {
-        public JsonPropertySerialization(string parameterName, string propertyName, string serializedName, CSharpType propertyType, CSharpType? valueType, JsonSerialization valueSerialization, bool isRequired, bool shouldSkipSerialization, bool shouldSkipDeserialization, bool optionalViaNullability, string? serializationValueHook = null, string? deserializationValueHook = null)
-            : base(propertyName, serializedName, propertyType, valueType, isRequired, shouldSkipSerialization, shouldSkipDeserialization)
+        public JsonPropertySerialization(string parameterName, ValueExpression value, string serializedName, CSharpType valueType, CSharpType? serializedType, JsonSerialization valueSerialization, bool isRequired, bool shouldSkipSerialization, bool shouldSkipDeserialization, string? customSerializationMethodName = null, string? customDeserializationMethodName = null)
```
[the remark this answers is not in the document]
AlexanderSher: Good point. This is a problem of the `ObjectTypeProperty.ValueType` property name, because it is not a type of the property, but a type we use in serialization/ungrouping, but I don't want to rename it cause I don't know how it is used in MPG. 

I have updated the `PropertySerialization` to use `TypedValueExpression` instead of `ValueExpression` and removed separate `ValueType` property.

src/AutoRest.CSharp/Common/Output/Models/Types/EnumType.cs
AlexanderSher: According to the code in `TypeSpecInputEnumTypeConverter`, it can be only `int`, `float` or `string`. But yeah, I can add it.

m-nash (approved):

PR #3761, opened 2023-09-21
Validate operation client parameters

live1206:
There are parameters default as `client` in operations, which is wrong in the first place.
Generator does not throw on this, but azure-sdk-for-net code failed to build.
It would be better to fail earlier during generation.


src/AutoRest.CSharp/Mgmt/Models/MgmtRestClientBuilder.cs
```diff
@@ -44,11 +40,28 @@ public MgmtRestClientBuilder(OperationGroup operationGroup)
...
                 .Where(p => p.Implementation == ImplementationLocation.Client)
-                .Distinct(new ParameterCompareer());
+                .Distinct(new ParameterCompareer()).ToList();
+            ValidateMgmtOperationParameters(parameters);
+            return parameters;
+        }
+
+        private static void ValidateMgmtOperationParameters(IReadOnlyList<RequestParameter> parameters)
+        {
+            var set = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "$host", "apiVersion" };
```
ArcturusZhang, line 55: I suggest we use `Origin` here to check whether they are `synthesized/host` or `synthesized/api-version`.
In our new input types which are supposed to replace modeler four schemas, we have a few flags on a parameter to indicate whether a parameter is one of the special ones (api-version, endpoint/host, content-type, etc).
There is a converter from schema parameters to input parameters, where we are using the `Origin` as source of truth: https://github.com/Azure/autorest.csharp/blob/176fba6b20dc3a05daab5998780d08514010f628/src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs#L136
This could help if we use the same thing here because eventually, all these things need to be converted to the input parameters.

live1206: Thans for the info, updated

ArcturusZhang (approved):

src/AutoRest.CSharp/Mgmt/Models/MgmtRestClientBuilder.cs
```diff
@@ -44,11 +40,28 @@ public MgmtRestClientBuilder(OperationGroup operationGroup)
...
+        private static void ValidateMgmtOperationParameters(IReadOnlyList<RequestParameter> parameters)
+        {
+            var set = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "modelerfour:synthesized/host", "modelerfour:synthesized/api-version" };
+            foreach (var parameter in parameters)
+            {
+                var parameterName = parameter.Origin ?? string.Empty;
+                if (!set.Contains(parameterName))
+                {
+                    throw new InvalidOperationException($"{parameterName} should be method parameter for operation");
+                }
```
ArcturusZhang, line 62: we are using the origin of the parameter as its name in the error message here.
Also, would it help if we could include the information of the corresponding operation of this parameter?

live1206: updated

src/AutoRest.CSharp/Mgmt/Models/MgmtRestClientBuilder.cs
live1206: @ArcturusZhang Check [this](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3111871&view=logs&j=145668b3-025a-5324-2d33-5b69e74f926a&t=d2bbf9b3-8667-59db-9211-fb3fd1ece838&l=7750) out, the request parameter `ApiVersion` has empty origin...

live1206: Had a closer look at it. 
This is because in generic resource, ApiVersion is defined as method parameter, but generator update all `apiVersion` to client parameter. It ends up into a client ApiVersion with empty origin.

live1206: Will be resolved in https://github.com/Azure/autorest.csharp/pull/3791

live1206: resolved

live1206:
Verified in https://github.com/Azure/azure-sdk-for-net/pull/39149/

ArthurMa1978 (approved):

PR #3764, opened 2023-09-22
add a [DebuggerDisplay] attribute to show necessary information on the debugger

ArcturusZhang:
# Description

with the attribute, when debugging, we could see some information for the type provider:
image

before this change, the debugger is only showing the type which usually is useless.
image

The attribute is now applied to:
- TypeProvider
- `Languages` which contains the name of a schema in code model
image

# Checklist

# Ready to Land?


chunyu3 (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -10,6 +11,7 @@
 
 namespace AutoRest.CSharp.Output.Models.Types
 {
+    [DebuggerDisplay("Name: {Declaration.Name}, Namespace: {Declaration.Namespace}")]
```
pshao25, line 14: Looks good to me. 

Just curious, will this advance the evaluating of `Declaration`? For example, we use `Declaration` in step 10, so it will be calculated in step 10. But if in step 5 I expand the debug view to see the debugger information, `Declaration` is calculated in step 5.

live1206, line 14: what is the difference between `DefaultName` and `Declaration.Name`?
same for `DefaultNamespace` and `Declaration.Namespace`.

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
```diff
@@ -16,6 +15,9 @@
 // ReSharper disable once CheckNamespace
 namespace AutoRest.CSharp.Input
 {
+    [DebuggerDisplay("Languages(Name: {Default.Name})")]
```
live1206, line 18: Not quite sure about the usage here since we can't flatten all the nested properties into debugger. We might print the important ones, but as you pointed out [here](https://github.com/Azure/autorest.csharp/pull/3761#discussion_r1333818562), the source of truth is `RequestParameter.Origin`, should we print that instead?

live1206: never mind, it does make sense to print name here since it matches with swagger definition better.

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
ArcturusZhang: DefaultName and DefaultNamespace are the values the schemas/input would like it to have.
The Declaration.Name and Declaration.Namespace are the values that the generator decides to write into the code which takes other factors into consideration such as the customization code.

ArcturusZhang: Yes, debugging might change the internal state of the object.
This is the same as you check those properties in the debugger view - which changes the state of the instance as well.
Debuggers have been optimized to avoid such cases for instance sometimes you might notice you cannot expand the values of an `IEnumerable<T>` instance because it would change the state of the instance.

Here, I believe it is safe to have them here because the Declaration of a TypeProvider is immutable and if we calculate it early, it would not cause any bad side effects

live1206 (approved):

lirenhe (approved):

Issue #3659, opened 2023-08-14
[Feature] Test auto generation


chunyu3:
Generate Test:

- Align with .NET test framework
- operation-level test, write test for an operation
-  can compile

Issue #3371, opened 2023-05-09
Deferred loading needs to be introduced to the DPG target


ArcturusZhang:
**Describe the issue or request**

Currently the information inside the `DpgOutputLibrary` is not guaranteed to be loaded when we could access them, this introduces troubles: https://github.com/Azure/autorest.csharp/blob/547671905492511ffe91b0e2f1f603c30389d7f7/src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs#L290

**Describe your ideas for solutions**

Moving everything in DpgOutputLibraryBuilder into the DpgOutputLibrary and put them in a way of "deferred loading" should solve this problem. The deferred loading mechanism could guarantee the property is fully initialized when we access it somewhere else.

lirenhe:
Please help to confirm whether this is still an issue for dotnet codegen

PR #3785, opened 2023-09-27
Generate test scaffolding

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3872
Fixes https://github.com/Azure/autorest.csharp/issues/3659

<del>Depends on https://github.com/Azure/autorest.csharp/pull/3775</del>

# Description

This PR generates test scaffolding for convenience of service onboarding.

Features:
1. A `TestBase` class and a `TestEnvironment` class will be generated.
2. Factory methods will be generated for all root level clients in the `TestBase` class to instrument the client and the corresponding client options.
3. The body of the generated test cases look exactly the same as the generated samples, except for the responses which will not be generated in test cases.
4. All the generated test cases will have `Ignore` attribute, because those generated test cases are impossible to pass, and they will fail the pipelines in this repo.

New configurations and changes to existing configurations:
1. `generate-sample-project`: default to `true`. If turned off, the samples will no longer generate (as well as the sample xml file in doc). We should only use this when there are issues in sample generation.
2. `generate-test-project`: now the default value of this configuration is false.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/Output/Tests/DpgClientTestProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+using AutoRest.CSharp.Input.Source;
+using AutoRest.CSharp.LowLevel.Output.Samples;
+using AutoRest.CSharp.Output.Models;
+using AutoRest.CSharp.Output.Models.Shared;
+using AutoRest.CSharp.Output.Samples.Models;
+using NUnit.Framework;
+
+namespace AutoRest.CSharp.LowLevel.Output.Tests
+{
+    internal class DpgClientTestProvider : DpgClientSampleProvider
```
chunyu3, line 21: In my opinion, the test provider and the sample provider does not has the inheritance relationship. instead, the test provider and sample provider has many common stuffs. So maybe it is better to define a base class to contain the common stuff, and let test provider and sample provider inheritance the base class.

src/assets/TestFramework/RecordedTestBase.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using NUnit.Framework;
+
+namespace Azure.Core.TestFramework
+{
+    [TestFixture(true)] // this is only mocking the RecordedTestBase to make sure the test could properly run and skip
+    public abstract class RecordedTestBase<TEnvironment> where TEnvironment : TestEnvironment, new()
```
chunyu3, line 9: Is this class similar as `ManagementRecordedTestBase`? in the future we will consolidate as one for both mgmt and dpg?

eng/Generate.ps1
```diff
@@ -134,7 +134,7 @@ function Add-Directory ([string]$testName, [string]$directory, [boolean]$forTest
     $readmeConfigurationPath = Join-Path $directory "readme.md"
     $testArguments = $null
     if (Test-Path $readmeConfigurationPath) {
-        $testArguments = "--require=$readmeConfigurationPath --clear-output-folder=true"
+        $testArguments = "--require=$readmeConfigurationPath --clear-output-folder=true --generate-test-scaffolding=true"
```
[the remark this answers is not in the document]
ArcturusZhang: I add this flag for a few reasons:
1. Generation of the test cases is slightly different from generation of samples. For instance, the test cases need to instrument the client and client options, therefore we have to use a ctor with a client options, this is different from samples (in which we always use the simplest ctor). Therefore generation of the test cases have the possibility that it cannot compile. This flag helps our customer to mitigate the issue if present.
2. Our customer should never modify the generated test cases because they are generated. The build analyze pipeline on azure-sdk-for-net repo validates if they get modified. Therefore these code snippets are only references for them to write their own test cases. They could copy them and then change them. If they copy them and finish their own test based on the code we generate, they no longer need the generated test cases any more.

Here is the work flow I imagine for this feature:
1. They generate the code with this flag on, and get the test cases. (the flag is default off)
2. They copy the test cases and modify them to finalize the test cases.
3. They generate the code with this flag off, the generated test cases are gone.

src/AutoRest.CSharp/LowLevel/Output/Tests/DpgClientTestProvider.cs
ArcturusZhang: I do not have any ideas of how to name that common base type. Any suggestions?

samples/AnomalyDetector/tests/Generated/Tests/AnomalyDetectorClientTests.cs
```diff
@@ -0,0 +1,740 @@
...
+namespace AnomalyDetector.Tests
+{
+    public partial class AnomalyDetectorClientTests : AnomalyDetectorTestBase
+    {
+        public AnomalyDetectorClientTests(bool isAsync) : base(isAsync)
+        {
+        }
+
+        [Test]
+        [Ignore("Only validating compilation of test scaffoldings")]
```
pshao25, line 26: Does customers see the same `Ignore`?

src/assets/TestFramework/RecordedTestBase.cs
ArcturusZhang: ManagementRecordedTestBase inherits this.

samples/AnomalyDetector/tests/Generated/Tests/AnomalyDetectorClientTests.cs
ArcturusZhang: Yes. We must have this Ignore. Otherwise our pipeline in this repo will try to run the test and fail

samples/AnomalyDetector/tests/Generated/Tests/AnomalyDetectorClientTests.cs
```diff
@@ -0,0 +1,740 @@
...
+        public async Task DetectUnivariateEntireSeries_ShortVersion()
```
pshao25, line 27: What is the difference between these tests and ones in samples?

ArcturusZhang: Same but without response handling

samples/AnomalyDetector/tests/Generated/Tests/AnomalyDetectorClientTests.cs
pshao25: Then it is better to be like `[Ignore("Remove this line to start writing tests")]`?

Directory.Build.props
```diff
@@ -6,6 +6,7 @@
     <IsTestProject Condition="$(MSBuildProjectName.EndsWith('.Tests'))">true</IsTestProject>
     <CentralPackagesFile>$(MSBuildThisFileDirectory)Packages.Data.props</CentralPackagesFile>
     <AzureCoreSharedSources>$(MSBuildThisFileDirectory)/src/assets/Azure.Core.Shared/</AzureCoreSharedSources>
+    <FakeAzureCoreTestFrameworkSources>$(MSBuildThisFileDirectory)/src/assets/TestFramework/</FakeAzureCoreTestFrameworkSources>
```
pshao25, line 9: Can we just name `AzureCoreTestFrameworkSources`?

samples/AnomalyDetector/tests/Generated/Tests/AnomalyDetectorClientTests.cs
pshao25: Why there is no response handling? Shouldn't there are some `Assert`s?

eng/Generate.ps1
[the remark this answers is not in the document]
chunyu3: Yes, customer should never modify the generated code, and we put the generated test code under test\Generated. And customer will add more test under `test` directory. Current they can copy them to `test` directory and modify them. The final goal is to auto-gen test which is completed and workable by default. Current we can set this flag false by default because the quality is not good enough. In future, we will change it to `true` by default ( or maybe remove this flag). 
 How about change it to `--generate-test` without `scaffolding`?

Directory.Build.props
ArcturusZhang: Because they are fake, they are only here to make sure our code could compile.
TestFramework is not a library, we cannot depend on them, therefore we create files to mimic their interfaces.

samples/AnomalyDetector/tests/Generated/Tests/AnomalyDetectorClientTests.cs
ArcturusZhang: We do not have the "expected values" to assert for the response values.
I could consider to add some assertions on the status code.

eng/Generate.ps1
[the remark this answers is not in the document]
ArcturusZhang: OK I could rename them

samples/AnomalyDetector/tests/Generated/Tests/AnomalyDetectorClientTests.cs
ArcturusZhang: I changed the message to `Please remove the Ignore attribute to let the test method run`

eng/Generate.ps1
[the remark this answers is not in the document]
ArcturusZhang: This flag is renamed to `--generate-tests` to align with one of our existing build target in mgmt (`/t:GenerateTests`)

src/AutoRest.CSharp/LowLevel/Output/Tests/DpgClientTestProvider.cs
ArcturusZhang: Thank you!
Well as for the renaming and moving to common, we could do it later when we find it is necessary

ArcturusZhang: Done

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
```diff
@@ -69,6 +68,31 @@ public static async Task ExecuteAsync(GeneratedCodeWorkspace project, InputNames
                 project.AddGeneratedFile($"{modelFactoryProvider.Type.Name}.cs", modelFactoryWriter.ToString());
             }
 
+            if (Configuration.GenerateTests)
+            {
+                // write test base and test env
+                var testBaseWriter = new ExpressionTypeProviderWriter(library.DpgTestBase);
+                testBaseWriter.Write();
+                project.AddGeneratedTestFile($"../../tests/Generated/Tests/{library.DpgTestBase.Type.Name}.cs", testBaseWriter.ToString());
```
AlexanderSher, line 76: I would prefer not to override `ToString` method but pass `CodeWriter` instance as parameter to `ExpressionTypeProviderWriter` constuctor or `Write` method.

src/AutoRest.CSharp/LowLevel/Output/ExpressionTypeProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        protected virtual IEnumerable<string> BuildUsings()
+        {
+            yield break;
+        }
+
+        public virtual CSharpType? Inherits => null;
+
+        private IEnumerable<CSharpType>? _implements;
+        public virtual IEnumerable<CSharpType> Implements => _implements ??= BuildImplements();
```
AlexanderSher, line 30: Do we need any of these methods to be lazy in any of the derived types?

src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs
```diff
@@ -334,6 +334,61 @@ private IEnumerable<Parameter> GetSubClientFactoryMethodParameters()
             return candidates.OrderBy(c => c.Parameters.Count).FirstOrDefault();
         }
 
+        internal MethodSignatureBase? GetEffectiveCtorWithClientOptions()
```
AlexanderSher, line 337: Can we unify it with method above?

ArcturusZhang: Yes we could - I was just considering the original copy of this method (`GetEffectiveCtor`) is already a workaround which will be removed in the future therefore to avoid any accident I just duplicated it.
I will unify this with the method above.

src/AutoRest.CSharp/LowLevel/Output/ExpressionTypeProvider.cs
ArcturusZhang: I think some of them could and some of them have to be lazy loaded.
For instance, `Inherits` could almost be determined when we construct this type provider, as well as Implements.
But others like the members of this type, could not be determined when we constructed this type, just like how we did the `BuildProperties` in models.
Let us take the `Resource` as an example, the methods of it is very likely to involve other resources or collections that is being constructed.
I am not 100% sure until we get there to make other types like this, therefore I could change them now since they really could be fully determined in the ctor here for the test case type provider. But if some cross reference issues occur in the future (most likely to happen in mgmt plane code), we might need to make them lazy loaded.

Also Michael is kind of a fan of lazy loading them - we even have an issue for that to convert dpg things to lazy loaded here:  https://github.com/Azure/autorest.csharp/issues/3371
This issue was created quite a time ago that he found in some circumstances when debugging things, the variables in the ctor is not really fully initialized.

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
ArcturusZhang: Emmm any reason for that?
this is kind of a pattern in mgmt writers already introduced quite a time ago.
I could change it back, but we might need to align this

src/AutoRest.CSharp/LowLevel/Output/ExpressionTypeProvider.cs
ArcturusZhang: Also any suggestions for the name of this type? I only leave this name as a temporary placeholder

src/AutoRest.CSharp/LowLevel/Output/Tests/DpgClientTestProvider.cs
chunyu3: Thanks. in the `DPGClientTestProvider` there are remaining some variable/type suffixed with `Sample`. We can refact this later, current it is ok. resolve this.

chunyu3 (approved):

eng/testProjects.json
```diff
@@ -44,6 +44,7 @@
         "parameters/collection-format",
         "parameters/spread",
         "parameters/body-optionality",
+        "client/structure/default",
```
ArcturusZhang, line 47: we have this test case, but we did not update this file.
Because of this, this test case will not regenerate when we run the script `/eng/Generate.ps1` also it will not be checked on our pipeline.
Now we introduce this back

test/CadlRanchProjects/client/structure/default/tspconfig.yaml
```diff
@@ -1,3 +1,4 @@
 options:
   "@azure-tools/typespec-csharp":
-    namespace: Client.Structure.Service.Default
\ No newline at end of file
+    namespace: Client.Structure.Service.Default
+    generate-sample-project: false
```
ArcturusZhang, line 4: because of [this issue](https://github.com/Azure/autorest.csharp/issues/4151), I have to disable sample generation for this project.
[This PR](https://github.com/Azure/autorest.csharp/pull/4082) might fix the issue for this test case and we could reenable when it merges.

src/AutoRest.CSharp/LowLevel/AutoRest/LowLevelTarget.cs
ArcturusZhang: Updated.

src/AutoRest.CSharp/LowLevel/Output/ExpressionTypeProvider.cs
ArcturusZhang: Updated

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -162,14 +162,18 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
                     options["branded"] === true
                         ? undefined
                         : options["branded"],
-                generateTestProject:
-                    options["generateTestProject"] === true
+                "generate-sample-project":
```
lirenhe, line 165: could you help to explain why the logic of generate-sample-project & generate-test-project are different?

test/TestProjects/Accessibility-LowLevel-NoAuth/src/Generated/Configuration.json
```diff
@@ -7,5 +7,6 @@
     "..\\..\\..\\..\\..\\artifacts\\bin\\AutoRest.CSharp\\Debug\\net7.0\\Azure.Core.Shared"
   ],
   "public-clients": true,
-  "skip-csproj-packagereference": true
+  "skip-csproj-packagereference": true,
+  "generate-test-project": true
```
lirenhe, line 11: Can we enable some projects with  "generate-sample-project": true

ArcturusZhang: its default value is `true` therefore we always have sample projects.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
ArcturusZhang: because default value of generate-sample-project is true, but generate-test-project is false.

lirenhe (approved):

PR #3790, opened 2023-09-28
Add using to RequestContent in convenience method

### LINKED ISSUE #3788: Convenience methods calling protocol methods should Dispose RequestContent?
Feel free to double check me, but I believe we should be calling Dispose on RequestContent instances we create in the convenience layer of DPG libraries.

Here is my thinking:

- RequestContent is IDisposable
- For protocol methods, the caller owns the RequestContent and is responsible for disposing it
    - Do/should we show this in generated samples?
- For convenience methods, we create the RequestContent in the model's serialization routine in the convenience method
    - Since we created the RequestContent, we're responsible for calling Dispose on it
- It doesn't look like we're putting the value returned from ToRequestContent in `using` statements:
    - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs#L155
    - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/src/Generated/OpenAIClient.cs#L51

If we agree with the above, the action items here would be:
- [ ] Show RequestContent in protocol method generated samples with `using` statement
- [ ] Add `using` statement to RequestContent created in convenience methods that call protocol methods

pshao25:
Fix #3788 

Regen: https://github.com/Azure/azure-sdk-for-net/pull/39142


ArcturusZhang (approved):
LGTM

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
```diff
@@ -21,17 +21,20 @@ internal record ConvenienceMethod(MethodSignature Signature, IReadOnlyList<Proto
...
+            var (parameterValues, contentInfo, spreadVariable) = PrepareConvenienceMethodParameters(contextVariable);
 
-            var converter = EnsureConvenienceBodyConverter(spreadVariable, contextVariable);
+            var converter = EnsureConvenienceBodyConverter(spreadVariable, contextVariable, contentInfo);
 
             return (parameterValues, converter);
         }
 
-        private (IReadOnlyList<FormattableString> ParameterValues, CodeWriterDeclaration? SpreadVariable) PrepareConvenienceMethodParameters(CodeWriterDeclaration contextVariable)
+        private record ContentInfo(CodeWriterDeclaration ContentVariable, FormattableString ContentValue);
```
chunyu3, line 31: `ContentInfo` is a little confusing and we may lose the meaning of this struct. Maybe we can name it as `RequestContentParamterInfo`

lirenhe (approved):

PR #3820, opened 2023-10-12
Change to node16 module resolution

pshao25:
According to https://microsoft.github.io/typespec/extending-typespec/basics#3-install-and-initialize-typescript, use node16 module resolution in emitter.


pshao25:
Already in feature/v3

Issue #3822, opened 2023-10-12
upgrade typespec library to 0.49.0


chunyu3:
upgrade typespec library to 0.49.0

PR #3823, opened 2023-10-12
onboard special-words cadl-ranch project

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3822
Fix https://github.com/Azure/autorest.csharp/issues/3845
Fix https://github.com/Azure/autorest.csharp/issues/3846

- Upgrade typespec libraries to 0.49.0
- Handle ambiguous property name when it is same as the type name. suffix `Value` on the property name
- Handle the operation group name empty issue: when the operation name is `Operations`, don't remove this as a normal client suffix.
- onboard special-words project

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Properties/launchSettings.json
```diff
@@ -628,10 +628,6 @@
       "commandName": "Project",
       "commandLineArgs": "--standalone $(SolutionDir)\\test\\CadlRanchProjects\\azure\\client-generator-core\\access\\src\\Generated -n"
     },
-    "typespec-azure/client-generator-core/internal": {
-      "commandName": "Project",
-      "commandLineArgs": "--standalone $(SolutionDir)\\test\\CadlRanchProjects\\azure\\client-generator-core\\internal\\src\\Generated -n"
```
lirenhe, line 633: Could you help to explain why we remove this test project?

chunyu3: it is removed from cadl-ranch

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -45,7 +46,7 @@ protected TypeProvider(string defaultNamespace, SourceInputModel? sourceInputMod
         private TypeDeclarationOptions BuildType()
         {
             return BuilderHelpers.CreateTypeAttributes(
-                DefaultName,
+                DefaultName.ToCleanName(),
```
ArcturusZhang, line 49: This is wrong, please call ToCleanName in ModelTypeProvider:
https://github.com/Azure/autorest.csharp/blob/8671af7f3cb128ec8e9f18926833c3e15949d473/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs#L74

m-nash:
Can we rename this to adopt special words since upgrading to 49 was already completed?

chunyu3:
> Can we rename this to adopt special words since upgrading to 49 was already completed?

Thanks @m-nash . updated

test/CadlRanchProjects/special-words/tspconfig.yaml
```diff
@@ -0,0 +1,4 @@
+options:
+  "@azure-tools/typespec-csharp":
+    use-overloads-between-protocol-and-convenience: false
+    model-namespace: false
```
chunyu3, line 4: Exclude the model namespace because there is a client named as `Models` which will cause conflict when the model namespace is applied.
Current we suppress this by excluding the model-namespace, and filed an issue https://github.com/Azure/autorest.csharp/issues/3847 to resolve in future.

lirenhe (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -52,7 +52,7 @@ public ModelTypeProvider(InputModelType inputModel, string defaultNamespace, Sou
             _typeFactory = typeFactory!;
             _inputModel = inputModel;
             _sourceInputModel = sourceInputModel;
-            DefaultName = GetValidIdentifier(inputModel.Name); // TODO -- this is only a workaround only to solve the anonymous model names, in other cases, the name is unchanged.
+            DefaultName = inputModel.Name.ToCleanName(); // TODO -- this is only a workaround only to solve the anonymous model names, in other cases, the name is unchanged.
```
ArcturusZhang, line 55: we could remove the TODO since your change resolves it.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -39,8 +39,20 @@ public ModelTypeProviderFields(InputModelType inputModel, TypeFactory typeFactor
 
             var visitedMembers = new HashSet<ISymbol>(SymbolEqualityComparer.Default);
 
-            foreach (var inputModelProperty in inputModel.Properties)
+            foreach (var property in inputModel.Properties)
             {
+                var inputModelProperty = property;
+                var name = BuilderHelpers.DisambiguateName(inputModel.Name.ToCleanName(), property.Name);
```
ArcturusZhang, line 45: here the `DisambiguateName` method append `Value` to the property name when it collides with the type name, which is OK.
If you read the code carefully, you will find we have code from L56-L60, and I quote here:
```csharp
var originalFieldName = inputModelProperty.Name.ToCleanName();
if (originalFieldName == inputModel.Name)
{
    originalFieldName = $"{originalFieldName}Property";
}
```
this is trying to append the word `Property` to this name if it is the same as the model - we need to verify if this is working. And since we choose the word `Value` now, we need to remove this piece of code.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -39,8 +39,20 @@ public ModelTypeProviderFields(InputModelType inputModel, TypeFactory typeFactor
...
+                if (property.Name != name) // the previous name is ambiguous, need to update the name
+                {
+                    inputModelProperty = new InputModelProperty(name,
+                        property.SerializedName,
+                        property.Description,
+                        property.Type,
+                        property.IsRequired,
+                        property.IsReadOnly,
+                        property.IsDiscriminator);
+                }
```
ArcturusZhang, line 55: By reading the code below, it is unnecessary to create a new InputModelProperty here.
This new property is duplicating everything except for the name.
But code below are using the local variable `originalFieldName` for the name of the property.
Therefore we just need to assign the new name to `originalFieldName` and everything should work fine.

chunyu3: But it will affected the serialize and deserialize. We only update the fieldName in the Model struct, but we need to use origin name for serialize/deserialize, that why we need to create the new InputModelProperty

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
chunyu3: This code snippet is added recently after I generate the PR. So I am not aware it. Thanks @ArcturusZhang  to inform this. And I tried, change `originalFieldName` is enough.

chunyu3: But we still need to handle other Name ambiguate as DisambiguateName handled, so I remove this logic and use DisambiguateName function.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
chunyu3: Thanks. `originalFieldName` is enough. I remove this code snippet.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
chunyu3: removed. @ArcturusZhang

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -41,11 +41,7 @@ public ModelTypeProviderFields(InputModelType inputModel, TypeFactory typeFactor
 
             foreach (var inputModelProperty in inputModel.Properties)
             {
-                var originalFieldName = inputModelProperty.Name.ToCleanName();
-                if (originalFieldName == inputModel.Name)
-                {
-                    originalFieldName = $"{originalFieldName}Property";
-                }
+                var originalFieldName = BuilderHelpers.DisambiguateName(inputModel.Name.ToCleanName(), inputModelProperty.Name.ToCleanName(), "Property");
```
ArcturusZhang, line 44: This still has issues.
Imagine this scenario:
We have this typespec:
```
model Foo
{
foo?: string;
}
```
which has the conflict between model name and property name, and we should generate:
```csharp
public partial class Foo
{
public string FooProperty {get;}
}
```
this has no issue.
But if the customer writes a customization file like this:
```csharp
[CodeGenModel("Foo")]
public partial class Bar
{}
```
which changes the generated name of this model to `Bar`. Now there is no conflicts between the model name and the property name, and we should generate
```csharp
public partial class Bar
{
public string Foo {get;}
}
```
But the implementation here will not know it.

ArcturusZhang: The solution of this, could be:
we need the ctor of this `ModelTypeProviderFields` to accept an extra parameter `CSharpType type`, and pass in the `Type` property of `ModelTypeProvider` in when it is constructing the `ModelTypeProviderFields` instance [here](https://github.com/chunyu3/autorest.csharp/blob/70bd982d529d382e602cd661cedcab4a35a37cf9/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs#L100)
Then here we could do:
```suggestion
                var originalFieldName = BuilderHelpers.DisambiguateName(_type, inputModelProperty.Name.ToCleanName(), "Property");
```
Also please revert the parameter type change in `DisambiguateName` method because it is not needed any more.

We could always have the `Type` property fully initialized before we initialize the `Properties` which calls `BuildProperties` which calls the `EnsureFields` which initialize the `ModelTypeProviderFields` instance. Therefore I assume this change will cause a circular invocation and lead to stack overflow.

chunyu3: Thanks @ArcturusZhang to bring out customization. To consider customization, thing becomes complex, We may need to consider the the name of the property is customized.
```
public partial class Foo
{
    [CodeGenMember("Foo")]
    public string RenamedFoo { get; } 
}
```

and more complex senarios: such as combine Model-Renaming and Property-Renaming together.

ArcturusZhang: this case is already considered in this type

chunyu3: resolved.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
```diff
@@ -29,7 +29,7 @@ internal sealed class ModelTypeProviderFields : IObjectTypeFields<InputModelProp
         public IReadOnlyList<Parameter> SerializationParameters { get; }
         public int Count => _fields.Count;
 
-        public ModelTypeProviderFields(InputModelType inputModel, TypeFactory typeFactory, ModelTypeMapping? sourceTypeMapping)
+        public ModelTypeProviderFields(InputModelType inputModel, CSharpType type, TypeFactory typeFactory, ModelTypeMapping? sourceTypeMapping)
```
ArcturusZhang, line 32: could we kind of specify the `type` parameter to make it more clear? Something like `enclosingType`, or `modelType`

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs
chunyu3: updated

Issue #2833, opened 2022-11-03
Proper use of AdditionalProperties


tracyboehrer:
What is the proper way to handle AdditionalProperties.  Nothing I've tried in OpenAPI actually works as expected.

When I use:

```
additionalProperties: true
```

or

```
additionalProperties:
   type: object
```

autorest will add

```
        /// <summary> Additional Properties. </summary>
        public IDictionary<string, object> AdditionalProperties { get; }
```

To the class.  But this doesn't work with System.Text.Json.JsonSerializer apparently.

We've stopped using the OpenAPI notation in the definition, and just manually include

```
        [JsonExtensionData]
        public IDictionary<string, JsonElement> AdditionalProperties { get; set; } = new Dictionary<string, JsonElement>();
```

ArcturusZhang:
Hi @tracyboehrer we are developing a new feature that enables public serialization/deserialization of the generated models.
For more details and samples of how to use that please refer to: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/System.ClientModel/README.md

PR #3838, opened 2023-10-14
Update serialization code to use new signatures

ArcturusZhang:
Supersedes https://github.com/Azure/autorest.csharp/pull/3434

Fixes https://github.com/Azure/autorest.csharp/issues/3353
Fixes https://github.com/Azure/autorest.csharp/issues/2351
Fixes https://github.com/Azure/azure-sdk-for-net/issues/16977
Fixes https://github.com/Azure/autorest.csharp/issues/2833

# Description

This feature literally changes every model, therefore this PR will have a lot of files changed. Details of features included in this PR are listed below.

In summary, now a model's serialization class will look like this:
```csharp
public partial class Foo : IUtf8JsonSerializable, IJsonModel<Foo>
{
    void IUtf8JsonSerializable(Utf8JsonWriter writer) => ((IJsonModel<Model>)this).Write(writer, ModelReaderWriterOptions.DefaultWireOptions);

    void IJsonModel<Foo>.Write(Utf8JsonWriter writer, ModelReaderWriterOptions options)
	{
		/* previous serialization implementation with some minor changes around Json and Wire format and raw data field */
	}

    Foo IJsonModel<Foo>.Create(ref Utf8JsonReader reader, ModelReaderWriterOptions options)
	{
		/* validates the format and call DeserializeFoo method to return a model as result */
	}

	internal static Foo DeserializeFoo(JsonElement element, ModelReaderWriterOptions options = null)
	{
		/* previous deserialization implementation with some minor changes around Json and Wire format and raw data field*/
	}

	BinaryData IPersistableModel<Foo>.Write(ModelReaderWriterOptions options)
    {
		/* call ModelReaderWriter to write the model into a Binrary data */
        return ModelReaderWriter.Write(this, options);
    }

	Foo IPersistableModel<Foo>.Create(BinaryData data, ModelReaderWriterOptions options)
	{
		/* validates the format and call DeserializeFoo method to return a model as result */
	}

	string IPersistableModel<Foo>.GetFormatFromOptions(ModelReaderWriterOptions options) => "J"; // return json when the model ONLY supports Json, return Xml when the model supports Xml

	/* ToRequestContent and FromResponse method unchanged. */
}
```

For serialization partial class

- [x] All model have to implement `IJsonModel<T>` where T is itself
	- [x] If the model is struct, it needs to implement `IJsonModel<object>` as well, otherwise compilation will fail
	- [x] If the model is the "unknown" model in a discriminated set, it needs to implement `IJsonModel<BaseType>`
- [x] A few new implementations have to be added to implement the `IJsonModel<T>` interface.
- [x] The main logic of serializing lives in the explicit implementation of `IJsonModel<T>`, and `IUtf8JsonSerializable` just calls the implementation
- [x] The `DeserializeXXX` method remains mostly unchanged, a new optional parameter `ModelReaderWriterOptions options` is added.
- [x] When the model is abstract, it should have an attribute `[DeserializationProxy(typeof(<the unknown internal type of it>))]`
- [x] Write readonly properties into JSON when `options.Format == ModelReaderWriterFormat.Json`
- [x] Write the private "additional properties" into JSON when `options.Format == ModelReaderWriterFormat.Json`.
(question) should we change this to `options.Format != ModelReaderWriterFormat.Wire`?
- [x] Read the private "additional properties" from JSON when `options.Format == ModelReaderWriterFormat.Json`
- [x] every model needs to implement `IUtf8JsonSerializable` (this might change in the future)

We leave the `ToRequestContent` method and `FromResponse` method unchanged, until the corresponding APIs in Azure.Core get into its next stable release.

For the model itself

- [x] Every model should have a new private "AdditionallProperties" property: 
```csharp
private Dictionary<string, BinaryData> _serializedAdditionalRawData; // when this model does not have any derived classes
protected internal Dictionary<string, BinaryData> _serializedAdditionalRawData; // when this model does not have any derived classes
```
- [x] This private/internal property automatically becomes AdditionalProperties when the model gets the flag on. Therefore the `_serializedAdditionalRawData` change does not happen on the models that already have additional properties. 
Since `_serializedAdditionalRawData` is just a private/protected internal version of AdditionalProperties, we will reuse the `AdditionalProperties` functionalities to implement it.
- [x] The internal ctor for deserialization will always accept the additional properties property as a property, no matter it is private or public.
- [x] The model factory entries do not add the private additional properties property as a parameter, but the model factory will include the public additional properties as a parameter. This is the same as previous behavior of model factory, no API changes of the existing entries are expected in model factory.
- [x] Every model now should have a new internal empty ctor if it does not have public empty ctor (for deserialization).
- [x] All models now will have both initialization ctor and serialization ctor (the internal ctor)
- [x] Model factory now includes those models without a deserializer previously, meaning new methods will be added to the model factory.

# Checklist

# Ready to Land?


eng/DownloadSharedSource.ps1
```diff
@@ -36,7 +36,8 @@ $files = @('AsyncLockWithValue.cs', 'ClientDiagnostics.cs', 'DiagnosticScope.cs'
     'OperationInternalBase.cs', 'OperationInternal.cs', 'OperationInternalOfT.cs', 'TaskExtensions.cs', 'Argument.cs', 'Multipart/MultipartFormDataContent.cs',
     'Multipart/MultipartContent.cs', 'AzureKeyCredentialPolicy.cs', 'AppContextSwitchHelper.cs',
     'OperationPoller.cs', 'FixedDelayWithNoJitterStrategy.cs', 'SequentialDelayStrategy.cs',
-    'ForwardsClientCallsAttribute.cs', 'AsyncLockWithValue.cs', 'VoidValue.cs', 'AzureResourceProviderNamespaceAttribute.cs',
+    'ForwardsClientCallsAttribute.cs', 'VoidValue.cs', 'AzureResourceProviderNamespaceAttribute.cs',
```
ArcturusZhang, line 39: 'AsyncLockWithValue.cs' is removed because we already have this file on the first line of this list

src/AutoRest.CSharp/Common/Output/Expressions/KnownValueExpressions/XAttributeExpression.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Xml.Linq;
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
+
+namespace AutoRest.CSharp.Common.Output.Expressions.KnownValueExpressions
+{
+    internal sealed record XAttributeExpression(ValueExpression Untyped) : TypedValueExpression<XAttribute>(Untyped)
```
ArcturusZhang, line 9: this is not a new class. This is an existing record previously defined in the same file as another record.
This change just extract this to make it to have its own file

src/AutoRest.CSharp/Common/Output/Expressions/Snippets.New.cs
```diff
@@ -50,7 +50,7 @@ public static RequestContextExpression RequestContext(CancellationTokenExpressio
 
             public static ValueExpression RequestFailedException(BaseResponseExpression response) => Instance(Configuration.ApiTypes.RequestFailedExceptionType, response);
 
-            public static ResourceIdentifierExpression ResourceIdentifier(ValueExpression resourceData) => new(Instance(typeof(ResourceIdentifier), new MemberExpression(resourceData, "Id")));
+            public static ResourceIdentifierExpression ResourceIdentifier(ValueExpression idString) => new(Instance(typeof(ResourceIdentifier), idString));
```
ArcturusZhang, line 53: This method is only used in a scenario that we are meant to create an expression of `new ResourceIdentifier(idString)`, the previous implemenation is wrong and misleading

src/AutoRest.CSharp/Common/Output/Models/Serialization/PropertySerialization.cs
```diff
@@ -28,23 +28,16 @@ internal abstract class PropertySerialization
...
-        public bool ShouldSkipSerialization { get; }
-        public bool ShouldSkipDeserialization { get; }
+        public bool ShouldExcludeInWireSerialization { get; }
 
-        protected PropertySerialization(string parameterName, TypedValueExpression value, string serializedName, CSharpType? serializedType, bool isRequired, bool shouldSkipSerialization) :
-            this(parameterName, value, serializedName, serializedType, isRequired, shouldSkipSerialization, false)
-        {
-        }
-
-        protected PropertySerialization(string parameterName, TypedValueExpression value, string serializedName, CSharpType? serializedType, bool isRequired, bool shouldSkipSerialization, bool shouldSkipDeserialization)
```
ArcturusZhang, line 39: we remove this ctor and the property `ShouldSkipDeserialization` because there is never a scenario that we are assigning this property to `true`. It is really not needed anywhere.

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectTypeProperty.cs
```diff
@@ -221,10 +223,5 @@ private static FormattableString CreateExtraPropertyDiscriminatorSummary(CSharpT
             }
             return updatedDescription ?? $"";
         }
-
-        public override string ToString()
-        {
-            return $"ObjectTypeProperty {{Name: {Declaration.Name}, Type: {Declaration.Type}}}";
-        }
```
ArcturusZhang, line 228: this method was introduced by me a long time ago to let the debugger show more information about this type. We never need to call this in our code.
Therefore in this PR, I replaced this method with `DebuggerDisplay`

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -144,8 +144,7 @@ await foreach (var file in project.GetGeneratedFilesAsync())
                 {
                     // Ignore any errors while trying to output crash information
                 }
-                await AutoRestLogger.Fatal($"Internal error in AutoRest.CSharp{ErrorHelpers.FileIssueText}\nException: {e.Message}\n{e.StackTrace}");
-                return false;
+                throw;
```
ArcturusZhang, line 147: it turns out this line is not working properly and the stack traces are swallowed somehow.
it is safe to throw the exception here because we have a try catch and print the stack traces clause in the place where calls this Execute method.

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
```diff
@@ -305,6 +305,7 @@ private string GetSrcCsproj()
 
   <ItemGroup>
     <PackageReference Include=""Azure.Core"" />
+    <PackageReference Include=""System.Net.ClientModel"" />
```
ArcturusZhang, line 308: we will change to an official version of this package when it is ready

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
```diff
@@ -322,7 +323,7 @@ private string GetSrcCsproj()
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include=""System.Net.ClientModel"" Version=""1.0.0-beta.1"" />
+    <PackageReference Include=""System.Net.ClientModel"" Version=""1.0.0-alpha.20231102.5"" />
```
ArcturusZhang, line 326: we will change to an official version of this package when it is ready

samples/AnomalyDetector/src/Generated/AnomalyDetectorClient.cs
```diff
@@ -1182,7 +1182,7 @@ public virtual Pageable<BinaryData> GetMultivariateModels(int? skip, int? maxCou
         {
             HttpMessage FirstPageRequest(int? pageSizeHint) => CreateGetMultivariateModelsRequest(skip, maxCount, context);
             HttpMessage NextPageRequest(int? pageSizeHint, string nextLink) => CreateGetMultivariateModelsNextPageRequest(nextLink, skip, maxCount, context);
-            return GeneratorPageableHelpers.CreatePageable(FirstPageRequest, NextPageRequest, e => BinaryData.FromString(e.GetRawText()), ClientDiagnostics, _pipeline, "AnomalyDetectorClient.GetMultivariateModels", "models", "nextLink", context);
+            return GeneratorPageableHelpers.CreatePageable(FirstPageRequest, NextPageRequest, (e, o) => BinaryData.FromString(e.GetRawText()), ClientDiagnostics, _pipeline, "AnomalyDetectorClient.GetMultivariateModels", "models", "nextLink", context);
```
pshao25, line 1185: We don't pass the parameter `o` to that implementation body?

ArcturusZhang: No. And the implementation here does not have a place to accept it

src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
```diff
@@ -240,16 +240,16 @@ private static FormattableString GetValueFactory(CSharpType? pageItemType)
             {
                 // When `JsonElement` provides access to its UTF8 buffer, change this code to create `BinaryData` from it.
                 // See also GeneratorPageableHelpers.ParseResponseForBinaryData
-                return $"e => {BinaryDataType}.{nameof(BinaryData.FromString)}(e.{nameof(JsonElement.GetRawText)}())";
+                return $"(e, o) => {BinaryDataType}.{nameof(BinaryData.FromString)}(e.{nameof(JsonElement.GetRawText)}())";
```
chunyu3, line 243: Why add a parameter `o` in the lambda which is not used in the body?

chunyu3:
LGTM, just a simple comment. I approved it. 
And now, we will always generate serialize/deserialize method for models ignore its usage?

chunyu3 (approved):

src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
ArcturusZhang: This is the delegate that we previously pass in the `DeserializeModel` method.
This PR changes the signature of `DeserializeModel` to take an extra optional parameter `ModelReaderWriterOptions` (corresponding to `o` here).
This method always falls back to `wire` when the options argument is absent, and this situation is exactly the wire scenario.

I change this only because I do not want to change the majority case where we pass in the `DeserializeModel` as the delegate.

src/AutoRest.CSharp/Common/Output/Models/Serialization/PropertySerialization.cs
```diff
@@ -33,23 +33,16 @@ internal abstract class PropertySerialization
         public CSharpType? SerializedType { get; }
 
         public bool IsRequired { get; }
-        public bool ShouldSkipSerialization { get; }
-        public bool ShouldSkipDeserialization { get; }
+        public bool ShouldExcludeInWireSerialization { get; }
```
live1206, line 36: This name is confusing to me. "Should Exclude InWire serialization?" Could you elaborate it further?

src/AutoRest.CSharp/Common/Output/Models/Types/SerializableObjectType.cs
```diff
@@ -40,9 +37,28 @@ protected SerializableObjectType(string defaultNamespace, SourceInputModel? sour
...
+                return false;
+
+            if (Inherits != null && Inherits is not { IsFrameworkType: false, Implementation: SystemObjectType })
+                return false;
+
+            return true;
+        }
+
+        protected readonly string _privateAdditionalPropertiesPropertyDescription = "Keeps track of any properties unknown to the library.";
+        protected readonly string _privateAdditionalPropertiesPropertyName = "_serializedAdditionalRawData";
```
live1206, line 61: These can be const

src/AutoRest.CSharp/LowLevel/Output/DpgClient.cs
```diff
@@ -21,7 +22,7 @@
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal class LowLevelClient : TypeProvider
+    internal class DpgClient : TypeProvider
```
live1206, line 25: feels like the renaming can be a separate PR

src/AutoRest.CSharp/Common/Decorator/DefaultDerivedSchema.cs
```diff
@@ -4,15 +4,13 @@
...
-using System.Text;
 using AutoRest.CSharp.Input;
-using AutoRest.CSharp.Output.Models.Types;
 
 namespace AutoRest.CSharp.Common.Decorator
 {
     internal static class DefaultDerivedSchema
     {
-        public const string DefaultDerivedExtension = "x-ms-autorest-defaultDerivedSchema";
+        private const string _defaultDerivedExtension = "x-ms-autorest-defaultDerivedSchema";
```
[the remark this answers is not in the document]
ArcturusZhang: thank you, I will revert this

src/AutoRest.CSharp/Common/Output/Models/Serialization/PropertySerialization.cs
ArcturusZhang: some background:
the serialization options has a `format` property, which accepts following format: wire (W), json (J) and xml (X)
This usually categorizes into wire and not wire.
Our previous system of serialization is always wire-serialization which serialize a thing and use the serialized result in the payload, that is why it is called "wire"
This property here simply mean "should exclude in wire serialization" where wire means the thing above.
When should a property be excluded in wire serialization? Usually when the property is readonly. We should not use the word `IsReadOnly` because only property could be readonly.

src/AutoRest.CSharp/Common/Input/ApiTypes.cs
```diff
@@ -12,6 +12,7 @@
 using AutoRest.CSharp.Output.Models;
 using AutoRest.CSharp.Output.Models.Requests;
 using AutoRest.CSharp.Output.Models.Shared;
+using Azure.Core;
```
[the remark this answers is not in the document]
ArcturusZhang: Sure, I will revert those changes

src/AutoRest.CSharp/Common/Output/Expressions/KnownValueExpressions/SerializableObjectTypeExpression.cs
```diff
@@ -19,8 +19,17 @@ public static MemberExpression DeserializeDelegate(SerializableObjectType serial
...
 
-        public static SerializableObjectTypeExpression Deserialize(SerializableObjectType serializableObjectType, ValueExpression element)
-            => new(serializableObjectType, new InvokeStaticMethodExpression(serializableObjectType.Type, $"Deserialize{serializableObjectType.Declaration.Name}", new[] { element }));
+        public static SerializableObjectTypeExpression Deserialize(SerializableObjectType model, ValueExpression element, ValueExpression? options = null)
+        {
+            var arguments = options == null ? new[] { element } : new[] { element, options };
+            return new(model, new InvokeStaticMethodExpression(model.Type, $"Deserialize{model.Declaration.Name}", arguments));
+        }
+
+        public static SerializableObjectTypeExpression DeserializeXml(SerializableObjectType model, ValueExpression xElement, ValueExpression? options = null)
```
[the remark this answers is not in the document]
ArcturusZhang: this is removed

src/AutoRest.CSharp/LowLevel/Output/DpgClient.cs
ArcturusZhang: sure

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
```diff
@@ -78,15 +81,33 @@ private void WriteObjectSchema(CodeWriter writer, ObjectType schema)
...
+                    continue;
+
                 WriteProperty(writer, property);
 
                 if (property.FlattenedProperty != null)
                     WriteProperty(writer, property.FlattenedProperty);
             }
         }
 
+        // TODO -- this is workaround
```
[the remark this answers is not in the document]
ArcturusZhang: we need to consolidate the way we treat FieldDeclaration and properties first. I will update the comment.
Here I say it is a workaround because this thing is now declared as a property (because this is the only way to include it into the ctor now without massive changes) but written as a field.

src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
ArcturusZhang: I changed this to the other way around - leaving this unchanged, and where we call `Foo.DeserializeFoo` we now have:
```
e => Foo.DeserializeFoo(e)
```
because the signature of `Foo.DeserializeFoo` has changed from
```
static Foo DeserializeFoo(JsonElement element)
```
to
```
static Foo DeserializeFoo(JsonElement element, ModelReaderWriterOptions options = null)
```

archerzz (approved):

ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Output/Builders/XmlSerializationMethodsBuilder.cs
```diff
@@ -215,11 +255,20 @@ private static IEnumerable<MethodBodyStatement> BuildDeserializeBody(XElementExp
...
+                if (parameterValues.TryGetValue(parameter.Name, out var argument))
+                    arguments.Add(argument);
+                else
+                {
+                    // this must be the raw data property
+                    arguments.Add(new PositionalParameterReference(parameter.Name, Null));
+                }
+            }
+
+            yield return Return(New.Instance(objectSerialization.Type, arguments.ToArray()));
```
live1206, line 271: have we considered to make an overload to take in `IReadOnlyList<>` to void `ToArray` of making another copy?

src/AutoRest.CSharp/Common/Output/Expressions/ValueExpressions/FormattableStringExpression.cs
```diff
@@ -15,18 +15,19 @@ namespace AutoRest.CSharp.Common.Output.Expressions.ValueExpressions
     /// </summary>
     internal sealed record FormattableStringExpression : ValueExpression
     {
-        public FormattableStringExpression(string format, IReadOnlyList<ValueExpression> args)
+        public FormattableStringExpression(string format, params ValueExpression[] args)
```
live1206, line 18: shall we keep the `IReadOnlyList` overload?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -177,6 +187,13 @@ private Method CreateMethod(SerializableObjectType model)
                     continue;
                 }
 
+                if (ctorParameter.IsRawData)
+                {
+                    // we do not want to include the raw data as a parameter of the model factory entry method, therefore here we skip the parameter, and use empty dictionary as argument
```
live1206, line 192: we are putting `null` not empty dictionary here, right?

src/AutoRest.CSharp/Common/Output/Expressions/ValueExpressions/FormattableStringExpression.cs
ArcturusZhang: Sure.
I changed this because the `IReadOnlyList` overload is quite unconvenient. But we could keep it

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -888,14 +1173,14 @@ public static ValueExpression GetFrameworkTypeValueExpression(Type frameworkType
                 return element.GetTimeSpan(format.ToFormatSpecifier());
             }
 
-            throw new NotSupportedException($"Framework type {frameworkType} is not supported");
+            throw new NotSupportedException($"Framework type {frameworkType} is not supported, please add `CodeGenMemberSerializationHooks` to specify the serialization of this type with the customized property");
```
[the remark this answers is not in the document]
ArcturusZhang: This could only happen when the generator itself will not generate it, but the customer changes the type of a property to that framework type. Therefore I added some information for our customer to fix it.
This is a fact that it is impossible that we support all framework type, we could consider support some, if those show up very frequently

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -392,9 +678,9 @@ private static IEnumerable<MethodBodyStatement> WriteObjectInitialization(JsonEl
                 yield return Declare(dictionary, New.Instance(objAdditionalProperties.Type));
                 yield return new ForeachStatement("property", element.EnumerateObject(), out var property)
                 {
-                    DeserializeIntoObjectProperties(serialization.Properties, objAdditionalProperties.ValueSerialization, new JsonPropertyExpression(property), new DictionaryExpression(objAdditionalProperties.Type.Arguments[0], objAdditionalProperties.Type.Arguments[1], dictionary), propertyVariables, shouldTreatEmptyStringAsNull)
+                    DeserializeIntoObjectProperties(serialization.Properties, objAdditionalProperties, new JsonPropertyExpression(property), new DictionaryExpression(objAdditionalProperties.Type.Arguments[0], objAdditionalProperties.Type.Arguments[1], dictionary), options, propertyVariables, shouldTreatEmptyStringAsNull).ToArray()
                 };
-                yield return new AssignValueStatement(propertyVariables[objAdditionalProperties], dictionary);
+                yield return Assign(propertyVariables[objAdditionalProperties], dictionary);
```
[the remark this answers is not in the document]
ArcturusZhang: well I would like to make the refactor in a separate PR. This does not originate from me therefore I do not know if there is a reason for it. It does look weird any way.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriter.cs
```diff
@@ -175,6 +176,9 @@ public CodeWriter Append(FormattableString formattableString)
                     case CodeWriterDeclaration declaration:
                         Append(declaration);
                         break;
+                    case ValueExpression expression:
+                        this.WriteValueExpression(expression);
```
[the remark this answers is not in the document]
ArcturusZhang: the `WriteValueExpression` is an extension method of `CodeWriter` therefore we have to have the `this.` here.

I believe we could consider to change it to an instance method - it just is as important as the other methods of writing things since we are using expressions more and more
But we could put it into a separate PR, this PR is complicated enough

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -22,76 +28,327 @@
...
+
+                if (xml != null)
+                {
+                    /*  using MemoryStream stream = new MemoryStream();
+                        using XmlWriter writer = XmlWriter.Create(stream);
+                        ((IXmlSerializable)this).Write(writer, null);
+                        writer.Flush();
+                        if (stream.Position > int.MaxValue)
+                        {
+                            return BinaryData.FromStream(stream);
```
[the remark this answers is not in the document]
ArcturusZhang: OK - is there a workaround for this?
I actually copied this from Michael's change.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
ArcturusZhang: Yes we did this intentionally.
I discussed this with Michael. Kind of it is an internal only property, customer will not have the chance to touch it, therefore if we could do the null check carefully, it is OK that we assign null.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriter.cs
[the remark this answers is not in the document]
live1206: OK, I see

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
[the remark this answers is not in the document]
live1206: OK, let's do is in a separate PR

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: What's the benefit using null instead of empty dictionary here? we don't need to check null afterwards during serialization generation.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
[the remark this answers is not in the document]
live1206: I don't think the position can exceed `int.MaxValue`, for [`MemoryStream`](https://github.com/dotnet/runtime/blob/4dffd80c4d77c27e772a0be26e8036af77fbb26e/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L25) the capacity is `int`, same as `position`. I think we can simply remove this check and use `FromStream` directly.
And `MemoryStream` is used to hold the buffer in [BinaryData](https://github.com/dotnet/runtime/blob/7fcbc81f2e367944d57815cf6bfa2bad3f6b7754/src/libraries/System.Memory.Data/src/System/BinaryData.cs#L329C45-L329C45) as well.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
ArcturusZhang: > What's the benefit using null instead of empty dictionary here? we don't need to check null afterwards during serialization generation if empty dictionary is always in place.

less allocation

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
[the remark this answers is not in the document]
ArcturusZhang: since the implementation of position in memory stream  will never exceed `int.MaxValue`, we could remove the if branch and only keep the else part.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -22,76 +28,325 @@
...
+                }
+
+                if (xml != null)
+                {
+                    /*  using MemoryStream stream = new MemoryStream();
+                        using XmlWriter writer = XmlWriter.Create(stream);
+                        ((IXmlSerializable)this).Write(writer, null);
+                        writer.Flush();
+                        // in the implementation of MemoryStream, `stream.Position` could never exceed `int.MaxValue`, therefore this if is redundant, we just need to keep the else branch
+                        //if (stream.Position > int.MaxValue)
```
live1206, line 206: Update the comment as well

live1206 (approved):

live1206 (approved):

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: well I intentionally kept them and changed those to comments (with more comments)

live1206 (approved):

PR #3843, opened 2023-10-16
Create an sdk pr even if generation fails

hallipr:


m-nash (approved):

eng/pipelines/sdk-update.yml
```diff
@@ -199,8 +199,15 @@ stages:
           Matrix: $[dependencies.Generate_Job_Matrix.outputs['generate_job_matrix.matrix']]
           AutorestCSharpVersion: $(AutorestCSharpVersion)
           BranchName: $(BranchName)
+          ContinueOnError: true
```
ArcturusZhang, line 202: Is it possible that we could make the step to fail when there is errors, but we still could run the following step `Create_PR`?

The reason I ask this is that in practice, we might get some random glitch of failure, and if the step fails, we could only rerun the step and work this around, such as failure like this:
image
And to work issues like this around, rerunning the whole thing might not solve it (every PR has the possibility to have this issue when a bunch of jobs running together), and when every step is green, there is no button to rerun the failed steps:
image
And when the steps fails, we could:
image

PR #3857, opened 2023-10-22
Turn `dfe-object` into `DataFactoryElement<object>` instead of `DataFactoryElement<BinaryData>`

DavidKarlas:
# Description
To be honest I'm not sure what BinaryData is supposed to represent, but ADF SetVariable accepts any JSON value either `123` or `"some String"`, so it seems better to use `DataFactoryElement<object>`, because `DataFactoryElement<BinaryData>` throws exception when it sees raw `123`... This depends on https://github.com/Azure/azure-sdk-for-net/pull/39436 being merged first...

# Checklist

   - Well I want someone with better understanding of things to look into this and say I talk nonsense or that, yes this should be merged...

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
@@ -287,7 +287,7 @@ namespace AutoRest.CSharp.Generation.Types
             XMsFormat.DataFactoryElementOfDateTime => typeof(DataFactoryElement<DateTimeOffset>),
             XMsFormat.DataFactoryElementOfDuration => typeof(DataFactoryElement<TimeSpan>),
             XMsFormat.DataFactoryElementOfUri => typeof(DataFactoryElement<Uri>),
-            XMsFormat.DataFactoryElementOfObject => typeof(DataFactoryElement<BinaryData>),
+            XMsFormat.DataFactoryElementOfObject => typeof(DataFactoryElement<object>),
             XMsFormat.DataFactoryElementOfListOfString => typeof(DataFactoryElement<IList<string>>),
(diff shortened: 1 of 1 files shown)


ArcturusZhang:
It is not quite a good idea that we use `object` in the generated libraries. If the `object` type gives the users of our libraries an illusion that the SDK could handle any type - even if it is a type created by them, but actually it is not, the generated libraries could only handle a very limited set of types. When this kind of requirement occurs, we use `BinaryData` because it is a "data holder" which handles everything.
We might need to change to another perspective if we are meeting problems when using the BinaryData type

DavidKarlas:
I'm fine with leaving BinaryData, but in that case it needs to be able to parse `123` or `"someString"` as input, and not always expect json object, maybe I'm missunderstanding something here...
Notice that dfe-object is very specific, in sense that it is not expected of ADF or SDK to understand this value, since this is parameter set by user/developer to be passed hence they can set any object and ADF passes it as is without any processing...

ArcturusZhang:
integers, strings, they are all valid json elements, therefore you could use `BinaryData.FromObjectAsJson` to convert them to BinaryData instance (which might not be intuitive). Please note if you want literal string, you need to include the quote inside your strings, such as
```
var data = BinaryData.FromObjectAsJson("\"something\"");
```
I suppose, invocations like this:
```
var data = new BinaryData(123);
```
might also work for integers (but you need to verify, I am not quite sure).
There are some `ToByteArray` or something like this APIs for strings in the `Encoding` class which could let you convert a string to bytes, and that is also OK to pass into the ctor of `BinaryData`.

The point is that `BinaryData` has the ability to represent everything, and it is really not a good idea for an SDK to export `object` for input, because the library itself cannot really guarantee it could handle all the types properly.

DavidKarlas:
I believe proper fix is here: https://github.com/Azure/azure-sdk-for-net/pull/39467

Issue #3858, opened 2023-10-24
[defect] method parameter does not honor @proejctedName


chunyu3:
When apply projected name for parameter, codegen does not recognize this projected name.

PR #3859, opened 2023-10-24
honor projected name for parameter

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3858

- honor @projectedName for operation parameter
- add one testcases for special-words

# Checklist

# Ready to Land?

### DIFF
--- a/src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
+++ b/src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
@@ -64,7 +64,7 @@ import {
     getFormattedType,
     getInputType
 } from "./model.js";
-import { capitalize } from "./utils.js";
+import { capitalize, getProjectedNameForCsharp } from "./utils.js";
 
 export function loadOperation(
     context: EmitContext<NetEmitterOptions>,
@@ -259,7 +259,7 @@ export function loadOperation(
                     : InputOperationParameterKind.Client
                 : InputOperationParameterKind.Method;
         return {
-            Name: param.name,
+            Name: getProjectedNameForCsharp(sdkContext, param) ?? param.name,
             NameInRequest: name,
             Description: getDoc(program, param),
             Type: inputType,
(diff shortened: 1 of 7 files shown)


src/TypeSpec.Extension/Emitter.Csharp/src/lib/utils.ts
```diff
@@ -52,3 +55,14 @@ export function getModelName(context: SdkContext, m: Model): string {
 
     return name;
 }
+
+export function getProjectedNameForCsharp(
```
ArcturusZhang, line 59: based on the purpose of this function, I think we should name it as `getNameForCSharp`

ArcturusZhang (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -259,7 +259,7 @@ export function loadOperation(
                     : InputOperationParameterKind.Client
                 : InputOperationParameterKind.Method;
         return {
-            Name: param.name,
+            Name: getProjectedNameForCsharp(sdkContext, param) ?? param.name,
```
lirenhe, line 262: As this is inside c# emitter, do we need to include csharp in function name?

lirenhe: I am wondering whether we could also put ' ?? param.name' inside function body and just call it getTypeName()

chunyu3: Different Type (Operation, Model, ModelProperty) may have different name logic, some may have friendlyName, or some may get the Name to combine the template args. And it is in a order, such as if projectedName is undefined, and try to get FriendlyName, if also undefined, then it origin name.
So we cannot combine all those in `getTypeName` and we need to just handle projectedName and return undefined if it is not defined.

lirenhe (approved):

PR #3874, opened 2023-10-27
refactor `CreateResourceIdentifier` method in MPG

ArcturusZhang:
# Description

This PR refactors the code to use expressions to write the `CreateResourceIdentifier` method.
To do this, a `FormattableStringExpression` is introduced, using this new expression you could just use a `FormattableString` instance built from other expressions and it will just turns out to be the way you constructed it
```csharp
var writer = new CodeWriter();
var a = new VariableReference(typeof(bool), "a");
var b = new VariableReference(typeof(bool), "b");
var expression = And(a, b);
writer.WriteValueExpression(new FormattableStringExpression("the result is {0}", new[] {expression}));
// this will write:
// $"the result is {a && b}"
```

One side effect of this change is that now the `CreateResourceIdentifier` method will have some extra parameter documentation.

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
@@ -10,6 +10,7 @@ using AutoRest.CSharp.Common.Output.Expressions.Statements;
 using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
 using AutoRest.CSharp.Common.Output.Models;
 using AutoRest.CSharp.Generation.Types;
+using AutoRest.CSharp.Utilities;
 using Azure.ResourceManager.Models;
 using SwitchExpression = AutoRest.CSharp.Common.Output.Expressions.ValueExpressions.SwitchExpression;
 
@@ -590,6 +591,27 @@ namespace AutoRest.CSharp.Generation.Writers
                     break;
                 case StringLiteralExpression(var literal, false):
                     writer.Literal(literal);
+                    break;
+                case FormattableStringExpression(var formattable):
+                    writer.AppendRaw("$\"");
+                    var argumentCount = 0;
+                    foreach ((var span, bool isLiteral) in StringExtensions.GetPathParts(formattable.Format))
+                    {
+                        if (isLiteral)
+                        {
+                            writer.AppendRaw(span.ToString());
+                            continue;
+                        }
+
+                        var arg = formattable.GetArgument(argumentCount);
+                        argumentCount++;
+                        // append the argument. The argument must be something assignable to ValueExpression here
+                        var a = (ValueExpression)arg!;
+                        writer.AppendRaw("{");
+                        writer.WriteValueExpression(a);
+                        writer.AppendRaw("}");
+                    }
+                    writer.AppendRaw("\"");
                     break;
             }
 
--- a/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
@@ -592,10 +592,10 @@ namespace AutoRest.CSharp.Generation.Writers
                 case StringLiteralExpression(var literal, false):
                     writer.Literal(literal);
                     break;
-                case FormattableStringExpression(var formattable):
+                case FormattableStringExpression(var format, var args):
                     writer.AppendRaw("$\"");
                     var argumentCount = 0;
-                    foreach ((var span, bool isLiteral) in StringExtensions.GetPathParts(formattable.Format))
+                    foreach ((var span, bool isLiteral) in StringExtensions.GetPathParts(format))
                     {
                         if (isLiteral)
                         {
@@ -603,12 +603,11 @@ namespace AutoRest.CSharp.Generation.Writers
                             continue;
                         }
 
-                        var arg = formattable.GetArgument(argumentCount);
+                        var arg = args[argumentCount];
                         argumentCount++;
-                        // append the argument. The argument must be something assignable to ValueExpression here
-                        var a = (ValueExpression)arg!;
+                        // append the argument
                         writer.AppendRaw("{");
-                        writer.WriteValueExpression(a);
+                        writer.WriteValueExpression(arg);
                         writer.AppendRaw("}");
                     }
                     writer.AppendRaw("\"");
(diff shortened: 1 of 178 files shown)


src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
```diff
@@ -591,6 +592,27 @@ public static void WriteValueExpression(this CodeWriter writer, ValueExpression
...
+                    var argumentCount = 0;
+                    foreach ((var span, bool isLiteral) in StringExtensions.GetPathParts(formattable.Format))
+                    {
+                        if (isLiteral)
+                        {
+                            writer.AppendRaw(span.ToString());
+                            continue;
+                        }
+
+                        var arg = formattable.GetArgument(argumentCount);
```
AlexanderSher, line 606: This assumes that numbers in format are sequential, e.g.: `$"/subscriptions/{0}/resourceGroups/{1}/providers/Microsoft.Compute/exactMatchModel1s/{2}"`
Not sure if that is critical or not.

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -451,31 +458,64 @@ protected virtual FormattableString CreateDescription()
...
-        private static Type? ToFormatTypeByName(string? name) => name switch
-        {
-            "location" => typeof(AzureLocation),
-            _ => null
-        };
+        private static CSharpType GetReferenceType(Reference reference)
+            => reference.Name switch
+            {
+                "location" when reference.Type.EqualsIgnoreNullable(typeof(string)) => typeof(AzureLocation),
+                _ => reference.Type
```
AlexanderSher, line 465: Do we really expect any type here, or those are limited to strings?

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
ArcturusZhang: Would it be sequential part of the rule when the formattable string is built?
For instance, `$"{foo}{bar}"` should be built into something like `string.Format("{0}{1}", foo, bar)`
Since this is happening internally, there is really no point that it could happen not in a sequence

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
ArcturusZhang: Well the old code is matching parameter called "location" with any type.
But since these are path parameters, there is really a very limited set of types it could choose from  - integers and strings basically.
It works perfectly fine if I remove the constraint but I just think that AzureLocation, this type by design only takes in a `string` as input, therefore I added the constraint.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
ArcturusZhang: Or do you think that it is only reasonable the `FormattableStringExpression` should define like this:
```csharp
public record FormattableStringExpression(string Format, param ValueExpression[] args) : ValueExpression
```
the format here is just a string containing `{0}` like `some{0}else{1}`
and we might need to do some validation such as validating the number of `{x}` should be the same as the args
This could avoid the case that we might be passing in a arbitrary FormattableString into this expression with non-expression arguments (such as real string values) and we could only get the error when we write it.

src/AutoRest.CSharp/Common/Output/Expressions/Snippets.New.cs
```diff
@@ -50,7 +50,7 @@ public static RequestContextExpression RequestContext(CancellationTokenExpressio
 
             public static ValueExpression RequestFailedException(BaseResponseExpression response) => Instance(Configuration.ApiTypes.RequestFailedExceptionType, response);
 
-            public static ResourceIdentifierExpression ResourceIdentifier(ValueExpression resourceData) => new(Instance(typeof(ResourceIdentifier), new MemberExpression(resourceData, "Id")));
+            public static ResourceIdentifierExpression ResourceIdentifier(ValueExpression id) => new(Instance(typeof(ResourceIdentifier), id));
```
ArcturusZhang, line 53: This method is only used in a scenario of xml serialization to serialize the ID in resource data that we are meant to create an expression of `new ResourceIdentifier(idString)`, the previous implemenation is wrong and misleading

ArcturusZhang: Here we could just make it simply since it is only calling the ctor of resource identifier

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
```diff
@@ -46,7 +46,7 @@ public class MgmtDebugConfiguration
                 SkipCodeGen = Configuration.DeserializeBoolean(skipCodeGen, false);
                 GenerateReport = Configuration.DeserializeBoolean(generateReport, true);
                 ReportOnly = Configuration.DeserializeBoolean(reportOnly, false);
-                ReportFormat = reportFormat?.GetString() ?? "yaml";
+                ReportFormat = Configuration.IsValidJsonElement(reportFormat) ? reportFormat?.GetString() ?? "yaml" : "yaml";
```
archerzz, line 49: Shall we throw an exception if the format is invalid?

archerzz (approved):

ArthurMa1978 (approved):

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtConfiguration.cs
ArcturusZhang: I do not think so - this is not a code gen related feature, it is only generating the reports. Not everyone is using it therefore making it throw any exception does not make sense.
BTW here IsValidJsonElement is not checking if the format is invalid as it is implied. It is checking if the kind of this element is null or invalid, this is not always caused by an invalid input. When omitting the format, we will get a `default` JsonElement for this thing, and default always has a `invalid` kind

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
ArcturusZhang: I changed the signature of FormattableStringExpression to the "format and args" way. There is a validation running on in debug mode.

PR #3875, opened 2023-10-28
Adding ReadOnlyMemory support

m-nash:
# Description
Addresses this comment https://github.com/Azure/azure-sdk-for-net/pull/39305#discussion_r1374076919

Adds support for ReadOnlyMemory<T>

For swagger you can use `"x-ms-embedding-vectory": true`
For tsp you will need to customize the property type, but the generator should handle serialization / initialization for you.  https://github.com/Azure/typespec-azure/issues/3781

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
@@ -75,6 +75,9 @@ namespace AutoRest.CSharp.Generation.Writers
 
         public static FormattableString? GetParameterInitializer(this CSharpType parameterType, Constant? defaultValue)
         {
+            if (parameterType.IsValueType)
+                return null;
+
             if (TypeFactory.IsCollectionType(parameterType) && (defaultValue == null || TypeFactory.IsCollectionType(defaultValue.Value.Type)))
             {
                 defaultValue = Constant.NewInstanceOf(TypeFactory.GetImplementationType(parameterType).WithNullable(false));
--- a/src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
+++ b/src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
@@ -76,7 +76,9 @@ namespace AutoRest.CSharp.Generation.Writers
         public static FormattableString? GetParameterInitializer(this CSharpType parameterType, Constant? defaultValue)
         {
             if (parameterType.IsValueType)
+            {
                 return null;
+            }
 
             if (TypeFactory.IsCollectionType(parameterType) && (defaultValue == null || TypeFactory.IsCollectionType(defaultValue.Value.Type)))
             {
--- a/src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
+++ b/src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
@@ -117,6 +117,8 @@ namespace AutoRest.CSharp.Input
         public string? Namespace => TryGetValue("x-namespace", out object? value) ? value?.ToString() : null;
         public string? Usage => TryGetValue("x-csharp-usage", out object? value) ? value?.ToString() : null;
 
+        public bool IsEmbeddingsVector => TryGetValue("x-ms-embedding-vector", out var value) && Convert.ToBoolean(value);
+
         public string[] Formats
         {
             get
--- a/src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
+++ b/src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
@@ -113,7 +113,7 @@ namespace AutoRest.CSharp.Common.Output.Builders
                 utf8JsonWriter.WritePropertyName(serialization.SerializedName),
                 serialization.CustomSerializationMethodName is {} serializationMethodName
                     ? InvokeCustomSerializationMethod(serializationMethodName, utf8JsonWriter)
-                    : SerializeExpression(utf8JsonWriter, serialization.ValueSerialization, serialization.EnumerableValue) // EnumerableValue is equal to Value if its not an array
+                    : SerializeExpression(utf8JsonWriter, serialization.ValueSerialization, serialization.EnumerableValue ?? serialization.Value)
             };
         }
 
--- a/src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
+++ b/src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
@@ -480,13 +480,18 @@ namespace AutoRest.CSharp.Common.Output.Builders
             // Reading a property value
             if (jsonPropertySerialization.ValueSerialization is not null)
             {
-                return new[]
+                List<MethodBodyStatement> statements = new List<MethodBodyStatement>
                 {
                     CreatePropertyNullCheckStatement(jsonPropertySerialization, jsonProperty, propertyVariables, shouldTreatEmptyStringAsNull),
-                    DeserializeValue(jsonPropertySerialization.ValueSerialization, jsonProperty.Value, out var value),
-                    Assign(propertyVariables[jsonPropertySerialization], value),
-                    Continue
+                    DeserializeValue(jsonPropertySerialization.ValueSerialization, jsonProperty.Value, out var value)
                 };
+
+                AssignValueStatement assignStatement = TypeFactory.IsReadOnlyMemory(jsonPropertySerialization.SerializedType!)
+                    ? Assign(propertyVariables[jsonPropertySerialization], New.Instance(jsonPropertySerialization.SerializedType!, value))
+                    : Assign(propertyVariables[jsonPropertySerialization], value);
+                statements.Add(assignStatement);
+                statements.Add(Continue);
+                return statements;
             }
 
             // Reading a nested object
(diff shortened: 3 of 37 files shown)


src/AutoRest.CSharp/Common/Output/Expressions/Statements/UnaryOperatorStatement.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
+
+namespace AutoRest.CSharp.Common.Output.Expressions.Statements
+{
+    internal record UnaryOperatorStatement(ValueExpression Value) : DeclarationStatement;
```
ArcturusZhang, line 8: I suppose we should make this statement taking the same parameter as `UnaryOperatorStatement` just like we did for InvokeStaticMethodStatement and expression.
We could add an implicit or explicit operator to cast from the expression to statement
Or we should rename this type to something like `ValueExpressionStatement` to represent the cases that we are using a value expression as a statement, since this class now is not take a `UnaryOperatorExpression` therefore it is very likely that this statement will not write as a unary operator thing, it depends on what we passed in.
If we do this change, it is also very reasonable that we could introduce an implicit or explicit operator for this from `ValueExpression` because there is a lot of expression could be standalone statements

test/TestProjects/Customizations-TypeSpec/src/Customizations/ModelWithCustomizedProperties.cs
```diff
@@ -75,5 +75,29 @@ public partial class ModelWithCustomizedProperties
...
+        public ReadOnlyMemory<float> VectorReadOnly { get; }
+
+        /// <summary> Property type changed to ReadOnlyMemory&lt;float&gt;. </summary>
+        public ReadOnlyMemory<float>? VectorOptionalReadOnly { get; }
+
+        /// <summary> Property type changed to ReadOnlyMemory&lt;float&gt;. </summary>
+        public ReadOnlyMemory<float>? VectorNullableReadOnly { get; }
+
+        /// <summary> Property type changed to ReadOnlyMemory&lt;float&gt;. </summary>
+        public ReadOnlyMemory<float>? VectorOptionalNullableReadOnly { get; }
```
ArcturusZhang, line 101: I assume in the (near) future, typespec might support this so that we do not have to customize this to get `ReadOnlyMemory`, right?

src/AutoRest.CSharp/Common/Output/Expressions/Statements/UnaryOperatorStatement.cs
AlexanderSher: > I suppose we should make this statement taking the same parameter as UnaryOperatorStatement just like we did for InvokeStaticMethodStatement and expression.

Agreed. Only unary operators that change the variable can be statements, so I would prefer to have `UnaryOperatorStatement(string Operator, ValueExpression Operand)`

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
```diff
@@ -117,6 +117,8 @@ internal partial class RecordOfStringAndAny
         public string? Namespace => TryGetValue("x-namespace", out object? value) ? value?.ToString() : null;
         public string? Usage => TryGetValue("x-csharp-usage", out object? value) ? value?.ToString() : null;
 
+        public bool IsEmbeddingsVector => TryGetValue("x-ms-embedding-vector", out var value) && Convert.ToBoolean(value);
```
AlexanderSher, line 120: Is it swagger-only?

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -87,7 +87,7 @@ public static IEnumerable<MethodBodyStatement> WriteProperties(Utf8JsonWriterExp
                 }
                 else if (property.SerializedType is { IsNullable: true })
                 {
-                    var checkPropertyIsInitialized = TypeFactory.IsCollectionType(property.SerializedType) && property.IsRequired
+                    var checkPropertyIsInitialized = (TypeFactory.IsCollectionType(property.SerializedType) && !TypeFactory.IsReadOnlyMemory(property.SerializedType)) && property.IsRequired
```
AlexanderSher, line 90: Nit: braces aren't needed.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -113,7 +113,7 @@ private static MethodBodyStatement WriteProperty(Utf8JsonWriterExpression utf8Js
                 utf8JsonWriter.WritePropertyName(serialization.SerializedName),
                 serialization.CustomSerializationMethodName is {} serializationMethodName
                     ? InvokeCustomSerializationMethod(serializationMethodName, utf8JsonWriter)
-                    : SerializeExpression(utf8JsonWriter, serialization.ValueSerialization, serialization.Value)
+                    : SerializeExpression(utf8JsonWriter, serialization.ValueSerialization, serialization.EnumerableValue) // EnumerableValue is equal to Value if its not an array
```
AlexanderSher, line 116: This is a potential bug prone. I would rather change `SerializeExpression` itself, specifically `JsonArraySerialization array` branch:
```cs
JsonArraySerialization array => SerializeArray(utf8JsonWriter, array, GetEnumerableExpression(array.ImplementationType, expression))
```
where
```
private static GetEnumerableExpression(CSharpType arrayImplementationType, ValueExpression expression)
{
    if (expression is TypedValueExpression typed && TypeFactory.IsReadOnlyMemory(typed))
    {
        expression = expression.NullableStructValue(typed.Type).Property(nameof(ReadOnlyMemory<byte>.Span));
    }
    return new EnumerableExpression(TypeFactory.GetElementType(arrayImplementationType), expression))
}

Comment (AlexanderSher) [reply src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs:116]: In consolidation branch, `SerializeExpression` uses `TypedValueExpression`, so I'll remove the casting during merge.

File: src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs [hunk 6]
```diff
@@ -480,13 +480,18 @@ private static MethodBodyStatement DeserializeIntoObjectProperty(JsonPropertySer
             // Reading a property value
             if (jsonPropertySerialization.ValueSerialization is not null)
             {
-                return new[]
+                List<MethodBodyStatement> statements = new List<MethodBodyStatement>
                 {
                     CreatePropertyNullCheckStatement(jsonPropertySerialization, jsonProperty, propertyVariables, shouldTreatEmptyStringAsNull),
-                    DeserializeValue(jsonPropertySerialization.ValueSerialization, jsonProperty.Value, out var value),
-                    Assign(propertyVariables[jsonPropertySerialization], value),
```
AlexanderSher, line 487: Can we call `New.Instance(jsonPropertySerialization.SerializedType, value)` in the `DeserializeValue` and keep this method as is.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -544,7 +549,7 @@ private static MethodBodyStatement CreatePropertyNullCheckStatement(JsonProperty
                 };
             }
 
-            if (!jsonPropertySerialization.IsRequired &&
+            if ((!jsonPropertySerialization.IsRequired || TypeFactory.IsReadOnlyMemory(serializedType!)) && // even if ReadOnlyMemory is required we leave the list empty if the payload doesn't have it
```
AlexanderSher, line 552: `serializedType` is not guaranteed to be not null here. We should check for null explicitly.

test/TestProjects/Customizations-TypeSpec/src/Customizations/ModelWithCustomizedProperties.cs
m-nash: Yes https://github.com/Azure/typespec-azure/issues/3781

src/AutoRest.CSharp/Common/Input/CodeModelPartials.cs
m-nash: For now yes until this is resolved https://github.com/Azure/typespec-azure/issues/3781

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
m-nash: I don't believe so because the assignment we call is conditional

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -75,6 +75,9 @@ public static FormattableString GetIdentifiersFormattable(this IEnumerable<strin
 
         public static FormattableString? GetParameterInitializer(this CSharpType parameterType, Constant? defaultValue)
         {
+            if (parameterType.IsValueType)
+                return null;
```
jsquire, line 79:
```suggestion
            if (parameterType.IsValueType)
            {
                return null;
            }
```

Looks like the other single-line `if` statements in this file are using brackets.   Plus, we're not monsters.

jsquire (approved):

Issue #3773, opened 2023-09-22
SDK for service with @useAuth(ApiKeyAuth<ApiKeyLocation.header, "Ocp-Apim-Subscription-Key">) doesn't compile.


rhurey:
**Describe the issue or request**
Have typespec using

``` TypeScript
@useAuth(ApiKeyAuth<ApiKeyLocation.header, "Ocp-Apim-Subscription-Key">)
@useAuth(BearerAuth)
@server(
    "{endpoint}/computervision",
    "Analyze an image to generate a description of its content.",
    {
      @doc("""
  Supported Cognitive Services endpoints (protocol and hostname, for example:
  https://<resource-name>.cognitiveservices.azure.com).
  """)
      endpoint: url,
    }
  )
```

And the generated SDK for it doesn't build because it tries to use an AzureKeyCredentialPolicy that is marked internal:
``` C#
_pipeline = HttpPipelineBuilder.Build(options, Array.Empty<HttpPipelinePolicy>(), new HttpPipelinePolicy[] { new AzureKeyCredentialPolicy(_keyCredential, AuthorizationHeader) }, new ResponseClassifier());
```
            
**Describe your ideas for solutions**
Looking at other generated code in the repro, it appears that the .csproj file needs to have a link to the AzureKeyCredential file added:
``` xml
<Compile Include="$(AzureCoreSharedSources)AzureKeyCredentialPolicy.cs" LinkBase="Shared" />
```

(Reference source: https://github.com/Azure/azure-sdk-for-net/blob/b1b123c1dcb1ba0fbe206fcbd3a1f8280d4410ce/sdk/anomalydetector/Azure.AI.AnomalyDetector/src/Azure.AI.AnomalyDetector.csproj#L21C21-L21C21)

Trivially worked around by manually adding the reference to the project, but took time to track down.

chunyu3:
The same issue as https://github.com/Azure/autorest.csharp/issues/3569

Issue #3569, opened 2023-07-07
Scaffolding should add shared source when azure key credential is used


m-nash:
If a spec uses azure key credential we need to add a shared source line to the csproj file in order to have the constructor compile.

`    <Compile Include="$(AzureCoreSharedSources)AzureKeyCredentialPolicy.cs" LinkBase="Shared/Core" />`

PR #3883, opened 2023-10-31
add AzureKeyCredentialPolicy.cs into project file

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/3773
Fix https://github.com/Azure/autorest.csharp/issues/3569
- Add AzureKeyCredentialPolicy.cs into project file(.csproj) when the sdk need azure key authentication.

# Checklist

# Ready to Land?


test/CadlRanchProjects/authentication/api-key/src/Authentication.ApiKey.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <Compile Include="$(AzureCoreSharedSources)AzureResourceProviderNamespaceAttribute.cs" LinkBase="Shared/Core" />
+    <Compile Include="$(AzureCoreSharedSources)AzureKeyCredentialPolicy.cs" LinkBase="Shared/Core" />
```
lirenhe, line 13: what is the impact if this line was not added?  Seems we didn't get any problem of the compile error.

chunyu3: Yes. It used to be added in `Directory.Build.targets` for autorest.csharp local test projects. so those test projects can compile successfully in autorest.csharp. But for those outside projects, without this, build will fail if the project has azure key authentication. To resolve this compile failure, we will add this to each sdk project file, and remove it from `Directory.Build.targets`.

pshao25:
I thought the solution is supposed to be: only when azure key credential is used, we add it to project file.

This solution looks a workaround to me.

chunyu3:
> only when azure key credential is used, we add it to project file.
> 
> This solution looks a workaround to me.

Yes. @pshao25 we only add `AzurekeyCredentialPolicy` to project file when azure key credential is used. See  in the code we will check the actual authentication used, then decide to add this or not.

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
```diff
@@ -16,14 +16,16 @@ internal class NewProjectScaffolding
         private string _testDirectory;
         private string _serviceDirectory;
         private bool _isAzureSdk;
+        private bool _NeedAzureKeyAuth;
```
pshao25, line 19:
```suggestion
        private bool _needAzureKeyAuth;
```

pshao25 (approved):

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
```diff
@@ -292,7 +294,13 @@ private string GetSrcCsproj()
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include=""$(AzureCoreSharedSources)AzureResourceProviderNamespaceAttribute.cs"" LinkBase=""Shared/Core"" />
+    <Compile Include=""$(AzureCoreSharedSources)AzureResourceProviderNamespaceAttribute.cs"" LinkBase=""Shared/Core"" />";
+            if (_needAzureKeyAuth)
```
archerzz, line 298: You actually can define `srcBrandedCsprojContent` as `@$"xxxx"`, e.g. multiple line formattable string. So there the codes can be like:
```suggestion
           { (_needAzureKeyAuth ? @"....." : @"....")}
```

archerzz (commented):
Can we have a test case which requires `AzureKeyCredentialPolicy.cs` to compile?

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
chunyu3: hello @archerzz  I have thought about this. but it cannot work for some encoding issue. I remember. So we need to maintain as it is.

chunyu3:
> Can we have a test case which requires `AzureKeyCredentialPolicy.cs` to compile?

Actually there are some projects in autorest.csharp which need `AzureKeyCredentialPolicy.cs`, so in this PR, we update the project files for them. No need to add more.

ArcturusZhang (approved):

archerzz (approved):

lirenhe (commented):
Could you provide the link of the Regen PR?  Thanks

lirenhe (approved):

PR #3892, opened 2023-10-31
Support changes in preview pipeline template

hallipr:
- remove npm scripts in favor of common powershell scripts eng/scripts/(initialize|build|test).ps1
- remove common chsred code script in favor of direct invocations where appropriate
- remove duplicate typescript generation code


eng/scripts/test.ps1
```diff
@@ -0,0 +1,74 @@
+#Requires -Version 7.0
```
m-nash, line 1: The name of this file made me think this was some kind of temporary test script.  Can we call it run-tests or something like this so it sounds more purposeful?

eng/pipelines/preview.yml
```diff
@@ -24,11 +32,11 @@ pool:
 extends:
   template: eng/pipelines/templates/stages/archetype-autorest-preview.yml@azure-sdk-tools
   parameters:
-    RegenerationJobCount: 20
+    RegenerationJobCount: 10
```
m-nash, line 35: What motivated this change?

m-nash (approved):

eng/pipelines/preview.yml
hallipr: There was no considerable change in runtime when fanning out so wide.

Compare 20 legs:
image

to 10:
image

hallipr: Until we can use a weighted grouping, were stuck with some leg taking about 20 minutes.

eng/scripts/test.ps1
hallipr: I renamed the scripts to:
- Initialize-Repository.ps1
- Build-Packages.ps1
- Test-Packages.ps1

PR #3895, opened 2023-11-01
Fix paging operation return type issue

### LINKED ISSUE #3860: Pagination operation is not showing correct return type
```

@doc("The result of listing the Pools in an Account.")
@pagedResult
model BatchPoolListResult {
    @doc("The list of Pools.")
    @items
    value?: BatchPool[];

    #suppress "@azure-tools/typespec-azure-core/casing-style" "The names of Property types must use camelCase"
    @doc("The URL to get the next set of results.")
    @nextLink
    `odata.nextLink`?: string;
}

@doc("A Pool in the Azure Batch service.")
model BatchPool {
    @doc("A string that uniquely identifies the Pool within the Account. The ID can contain any combination of alphanumeric characters including hyphens and underscores, and cannot contain more than 64 characters. The ID is case-preserving and case-insensitive (that is, you may not have two IDs within an Account that differ only by case).")
    @visibility("read")
    id?: string;

    @doc("The display name for the Pool. The display name need not be unique and can contain any Unicode characters up to a maximum length of 1024.")
    @visibility("read")
    displayName?: string;

    @doc("The URL of the Pool.")
    @visibility("read")
    url?: string;
}

#suppress "@azure-tools/typespec-azure-core/casing-style" "The names of Model types must use camelCase"
@doc("Client related response header values")
model BatchRequestHeaders {
    @doc("""
    An OData $filter clause. For more information on constructing this filter, see
    https://docs.microsoft.com/en-us/rest/api/batchservice/odata-filters-in-batch#list-pools.
    """)
    @query
    $filter?: string;

    @doc("An OData $select clause.")
    @query({
        format: "csv",
    })
    $select?: string[];

    @doc("An OData $expand clause.")
    @query({
        format: "csv",
    })
    $expand?: string[];
}

#suppress "@azure-tools/typespec-azure-core/casing-style" "The names of Model types must use camelCase"
@doc("Client related response header values")
model BatchResponseHeaders {
    @doc("The client-request-id provided by the client during the request. This will be returned only if the return-client-request-id parameter was set to true.")
    @header
    `client-request-id`?: string;

    @doc("A unique identifier for the request that was made to the Batch service. If a request is consistently failing and you have verified that the request is properly formulated, you may use this value to report the error to Microsoft. In your report, include the value of this request ID, the approximate time that the request was made, the Batch Account against which the request was made, and the region that Account resides in.")
    @header
    `request-id`?: string;

    @doc("The ETag HTTP response header. This is an opaque string. You can use it to detect whether the resource has changed between requests. In particular, you can pass the ETag to one of the If-Modified-Since, If-Unmodified-Since, If-Match or If-None-Match headers.")
    @header
    @projectedName("python", "etag")
    ETag?: string;

    @doc("The time at which the resource was last modified.")
    @header
    @encode(DateTimeKnownEncoding.rfc7231)
    `Last-Modified`?: utcDateTime;
}

@doc("An error response received from the Azure Batch service.")
@error
model BatchError {
    @doc("An identifier for the error. Codes are invariant and are intended to be consumed programmatically.")
    code: string;
}

interface Pools {
  @doc("Lists all of the Pools in the specified Account.")
  @route("/pools")
  @get
  @convenientAPI(true)
  listPools is RpcOperation<
     BatchRequestHeaders,
     BatchPoolListResult,
     BatchResponseHeaders,
     BatchError
  >;
}
```
Actual:
```C#
public virtual async Task<Response<BatchPoolListResult>> GetPoolsAsync(string filter = null, IEnumerable<string> select = null, IEnumerable<string> expand = null, CancellationToken cancellationToken = default)
```
Expected:
```C#
public virtual async AsyncPageable<BatchPool> GetPoolsAsync(string filter = null, IEnumerable<string> select = null, IEnumerable<string> expand = null, CancellationToken cancellationToken = default)
```

pshao25:
Fix #3860


ArcturusZhang (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -163,15 +163,16 @@ export function loadOperation(
     for (const res of operation.responses) {
         const body = res.responses[0]?.body;
         if (body?.type) {
+            const bodyType = getEffectiveSchemaType(sdkContext, body.type);
```
chunyu3, line 166: Just curious, why `getEffectiveSchemaType` will resolve this issue?

pshao25: customer appends some header properties to the body, which makes the body model itself become an anonymous model and lose all the decorators related to paging.

chunyu3 (approved):

test/TestProjects/Pagination-TypeSpec/tests/Generated/Samples/Samples_PaginationClient.cs
```diff
@@ -625,5 +625,121 @@ await foreach (TextBlockItem item in client.GetTextBlocklistItemsAsync("<blockli
...
+        [Ignore("Only validating compilation of examples")]
+        public void Example_GetPools_ShortVersion()
+        {
+            Uri endpoint = new Uri("<https://my-service.azure.com>");
+            TokenCredential credential = new DefaultAzureCredential();
+            PaginationClient client = new PaginationClient(endpoint, credential);
+
+            foreach (BinaryData item in client.GetPools(null, null, null, null))
+            {
+                JsonElement result = JsonDocument.Parse(item.ToStream()).RootElement;
```
lirenhe, line 639: Shall we use 'using' here?  @ArcturusZhang

lirenhe (approved):

Issue #3706, opened 2023-09-01
Our generator creates client libraries which violate AZC0006 and AZC0007


m-nash:
We should update our generator so that these two analyzer rules are not violated for azure sdks

chunyu3:
Hello @m-nash  Does this occur in MPG, DPG or both?

pshao25:
**Current HLC behavior:**
One constructor with optional `options` like
```C#
public AClient(Uri endpoint, AClientOptions options = null) {}
```

**Current DPG behavior:**
Two constructors, one with all parameters and the other with only required parameters. All the parameters will be changed to required.
```C#
public AClient(Uri endpoint): this(endpoint, new AClientOptions()){}
public AClient(Uri endpoint, AClientOptions options){}
```
When there is only one optional parameter (e.g. this case only have one optional `AClientOptions`), this scenario actually not violates analyzer.

However, when there are more than one optional parameter, it violates analyzer, like
```C#
public AClient(AzureKeyCredential credential): this(credential, new Uri("http://localhost:3000"), new AClientOptions()){}
public AClient(AzureKeyCredential credential, Uri endpoint, AClientOptions options){}
```

pshao25:
So, we will put all the optional parameters except endpoint into options 

1. `public AClient(string required, string optional ="optional", AClientOptions options = null)` becomes
```C#
public AClient(string required): this(required, new AClientOptions()){}
public AClient(string required, AClientOptions options)
{
  _optional = options.Optional; // Client parameter
}

public partial class ArrayClientOptions : ClientOptions
{
    internal string Optional { get;}

    /// <summary> Initializes new instance of ArrayClientOptions. </summary>
    public ArrayClientOptions(ServiceVersion version = LatestVersion, string optional = "optional")
    {
        Version = version switch
        {
            ServiceVersion.V1_0_0 => "1.0.0",
            _ => throw new NotSupportedException()
        };

        Optional = optional;
    }
}
```

2. `public AClient(string required, Uri endpoint = new Uri("http://localhost:3000"), AClientOptions options = null)` becomes
```C#
public AClient(string required): this(new Uri("http://localhost:3000/"), required, new AClientOptions()){}
public AClient(Uri endpoint, string required, AClientOptions options){}
```

tg-msft:
I'd clarify

> So, we will put all the optional parameters except endpoint into options

slightly to 

- We'll default to adding optional parameters in `ClientOptions`
[...]

PR #3910, opened 2023-11-07
Put all optional client parameters into client options

pshao25:
1. Change constructor appearance to https://github.com/Azure/autorest.csharp/issues/3706#issuecomment-1752374728
2. Fix the problem causing from M4, i.e., when there is no endpoint set in swagger, M4 would still claim it is an optional one with default "", which is expected as a required one.

Now constructor will only have required parameters, all the optional parameters except endpoint will go to client options.

Regen: https://github.com/Azure/azure-sdk-for-net/pull/40292


src/AutoRest.CSharp/Common/Output/Models/Types/ClientOptionsTypeProvider.cs
```diff
@@ -16,6 +17,7 @@ internal sealed class ClientOptionsTypeProvider : TypeProvider
 
         public FormattableString Description { get; }
         public IReadOnlyList<ApiVersion>? ApiVersions { get; }
+        public IList<Parameter>? AdditionalProperties { get; private set; }
```
ArcturusZhang, line 20: Is there any special meaning of this being null? An empty list should work perfectly.
Also we should not make this mutable - it is unsafe that the state of this provider could change. If it could change, how could we know it is the final state when we write it?

pshao25: I hope it could be immutable as well. However, when we built `LowLevelClient` we need to pass `clientOptions`, and the optional parameters needed by `clientOptions` comes from `LowLevelClient`. We create `clientOptions` way earlier than we create `LowLevelClient`.

ArcturusZhang: This is not a proper reason.
Something like this should work:
```
var additionalProperties = new List<Parameter>();
var options = new ClientOptionsTypeProvider(additionalProperties);

// build the client

// populate the array after we build the client
```
Or change this property to be lazy loaded so that when some one calls this, everything will be calculated and will never change.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -675,6 +675,8 @@ private IEnumerable<LowLevelClient> CreateClients(IEnumerable<ClientInfo> client
 
                 subClients.AddRange(CreateClients(clientInfo.Children, typeFactory, clientOptions, client));
 
+                clientOptions.AddAdditionalProperty(client.GetOptionalParametersInOptions());
```
ArcturusZhang, line 678: For instance, one of the solution of comment https://github.com/Azure/autorest.csharp/pull/3910/files#r1392203322 could be:
1. add a parameter in ctor of ClientOptionsTypeProvider to accept a list for the additional properties.
2. change the type of AdditionalProperties property ClientOptionsTypeProvider to `IReadOnlyList<Parameter>`
3. creates the ClientOptionsTypeProvider instance using an empty list.
4. pass the list in this method as well
5. add the parameter into the list here.

ArcturusZhang: this is exactly the same solution as the `SubClients` list we have in this method.

src/AutoRest.CSharp/Common/Output/Models/Types/ClientOptionsTypeProvider.cs
pshao25: There is circle dependency. You cannot know additionalProperties before client is built.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
pshao25: Same question "how could we know it is the final state when we write it?"

pshao25: And as I said "We create clientOptions way earlier than we create LowLevelClient". For this solution, we will pass a new list all the way from the place creating clientoptions to the place creating client.

src/AutoRest.CSharp/Common/Output/Models/Types/ClientOptionsTypeProvider.cs
```diff
@@ -16,6 +17,7 @@ internal sealed class ClientOptionsTypeProvider : TypeProvider
 
         public FormattableString Description { get; }
         public IReadOnlyList<ApiVersion>? ApiVersions { get; }
+        public IReadOnlyList<Parameter>? AdditionalProperties { get; init; }
```
ArcturusZhang, line 20: looks like in your code, we will always assign an instance to this property.
Also it really does not have a difference when it is null and it is an empty list - it represents there is no additional properties.
Therefore we could make it `IReadOnlyList<Parameter>` (remove the question mark)

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -20,6 +20,7 @@ internal record Parameter(string Name, FormattableString? Description, CSharpTyp
     {
         public CSharpAttribute[] Attributes { get; init; } = Array.Empty<CSharpAttribute>();
         public bool IsOptionalInSignature => DefaultValue != null;
+        public bool IsEndpoint => Type.Equals(typeof(Uri)) && Name == KnownParameters.Endpoint.Name;
```
ArcturusZhang, line 23: this does not look like correct - we have a `IsEndpoint` flag in `InputParameter`, and I think if we have this, this should have the same value as that flag.
https://github.com/Azure/autorest.csharp/blob/2755f061b8f45ce0017a3b61c949eba06d1ff613/src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs#L24

If there is inconsistency when it comes to the swagger inputs, we might need to refine the converter

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -82,6 +83,7 @@ public static Parameter FromInputParameter(in InputParameter operationParameter,
 
         private static Constant? GetDefaultValue(InputParameter operationParameter, TypeFactory typeFactory) => operationParameter switch
         {
+            { Name: "$host", DefaultValue.Value: ""} => (Constant?)null, // In M4, when endpoint is not set, its default value is an empty string instead of null
```
ArcturusZhang, line 86: should it be 
```suggestion
            { IsEndpoint: true, DefaultValue.Value: "" } => (Constant?)null, // In M4, when endpoint is not set, its default value is an empty string instead of null
```
and if in the modelerfour case, the corresponding parameter does not have `IsEndpoint: true` when it is actually endpoint (or host), we should fix there

src/AutoRest.CSharp/Common/Output/Models/Types/ClientOptionsTypeProvider.cs
pshao25: > we will always assign an instance to this property.

No. Only DPG will do that, HLC not.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
pshao25: I can do this. But my concern is it is just a workaround for M4's bug (we might not want to fix M4), so I just want to limit this condition to swagger, not want to impact typespec. If we use `IsEndpoint`, typespec will be included to this change.

src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
```diff
@@ -50,6 +51,21 @@ public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvid
...
+                    {
+                        if (clientOptions.ApiVersions is not null)
+                        {
+                            writer.Line();
+                        }
+
+                        foreach (var property in clientOptions.AdditionalProperties)
+                        {
+                            writer.WriteXmlDocumentationSummary(property.Description);
+                            writer.Line($"public {property.Type} {property.Name.FirstCharToUpperCase()} {{ get; set; }}");
```
archerzz, line 65: Shall we use other normalization method like `property.Name.ToCleanName()`?

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -16,7 +16,7 @@
 
 namespace AutoRest.CSharp.Output.Models.Shared
 {
-    internal record Parameter(string Name, FormattableString? Description, CSharpType Type, Constant? DefaultValue, ValidationType Validation, FormattableString? Initializer, bool IsApiVersionParameter = false, bool IsResourceIdentifier = false, bool SkipUrlEncoding = false, RequestLocation RequestLocation = RequestLocation.None, SerializationFormat SerializationFormat = SerializationFormat.Default, bool IsPropertyBag = false)
+    internal record Parameter(string Name, FormattableString? Description, CSharpType Type, Constant? DefaultValue, ValidationType Validation, FormattableString? Initializer, bool IsApiVersionParameter = false, bool IsEndpoint = false, bool IsResourceIdentifier = false, bool SkipUrlEncoding = false, RequestLocation RequestLocation = RequestLocation.None, SerializationFormat SerializationFormat = SerializationFormat.Default, bool IsPropertyBag = false)
```
archerzz, line 19: Not sure if we should insert a property v.s. append a property.

src/AutoRest.CSharp/Common/Output/Models/Types/ClientOptionsTypeProvider.cs
archerzz: Why not just return empty collection, so that we can remove `?` here? If `null` doesn't have extra meaning here, I'd like to recommend to remove `?`.

src/AutoRest.CSharp/LowLevel/Generation/DpgClientWriter.cs
```diff
@@ -259,6 +259,10 @@ private void WritePrimaryPublicConstructor(ConstructorSignature signature)
                         {
                             _writer.Line($"{field.Name:I} = {clientOptionsParameter.Name:I}.Version;");
                         }
+                        else if (parameter.IsOptionalInSignature && !parameter.IsEndpoint && !Configuration.KeepOptionalClientParametersInConstructor)
```
archerzz, line 262:
```suggestion
                        else if (parameter is { IsOptionalInSignature: true, IsEndpoint: false} && !Configuration.KeepOptionalClientParametersInConstructor)
```

src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
```diff
@@ -50,6 +51,21 @@ public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvid
...
                     }
+
+                    if (clientOptions.AdditionalProperties != null && !Configuration.KeepOptionalClientParametersInConstructor)
+                    {
+                        if (clientOptions.ApiVersions is not null)
+                        {
+                            writer.Line();
+                        }
+
+                        foreach (var property in clientOptions.AdditionalProperties)
```
ArcturusZhang, line 62: this is kind of misleading because the type of this `property` variable is actually `Parameter`.
Could we normalize it to be `var parameter in clientOptions.AdditionalParameters`?

src/AutoRest.CSharp/Common/Output/Models/Types/ClientOptionsTypeProvider.cs
ArcturusZhang: empty collection here does not change anything for HLC

test/AutoRest.TestServerLowLevel.Tests/url-path-items.cs
```diff
@@ -12,55 +12,68 @@ public class UrlPathItemsTests : TestServerLowLevelTestBase
     {
         [Test]
         public Task UrlPathItemGetAll() => TestStatus(async (host) =>
-            await new PathItemsClient(globalStringPath: "globalStringPath",
-                    globalStringQuery: "globalStringQuery",
+        {
+            var options = new AutoRestUrlTestServiceClientOptions();
+            options.GlobalStringQuery = "globalStringQuery";
```
ArcturusZhang, line 17: I personally prefer the way of object initializer which makes this more compact

src/AutoRest.CSharp/Common/Output/Models/Types/ClientOptionsTypeProvider.cs
ArcturusZhang: If this is the issue:

> if we remove the question mark, we will have to assign it in the ctor and in HLC it does not have anything to assign

then we should
1. add a new ctor that accepts an `IReadOnlyList<Parameter>` to assign value to this property
2. the old ctor calls `this` with `Array.Empty<Parameter>` for this property

in this way, HLC is unchanged as well

src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
pshao25: Not sure, `ToCleanName` is to complicated, I'm not sure if there is any side effects. So I actually prefer to use `FirstCharToUpperCase` first and we could change later is there are some special cases.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
pshao25: Just want to put it near `IsApiVersionParameter`, any problem?

src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
pshao25: I don't have strong opinion. If you insist, I could change.

src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
```diff
@@ -50,6 +51,21 @@ public static void WriteClientOptions(CodeWriter writer, ClientOptionsTypeProvid
...
+                    {
+                        if (clientOptions.ApiVersions is not null)
+                        {
+                            writer.Line();
+                        }
+
+                        foreach (var parameter in clientOptions.AdditionalParameters)
+                        {
+                            writer.WriteXmlDocumentationSummary(parameter.Description);
+                            writer.Line($"public {parameter.Type} {parameter.Name.FirstCharToUpperCase()} {{ get; set; }}");
```
ArcturusZhang, line 65:
```suggestion
                            writer.Line($"public {parameter.Type} {parameter.Name.ToCleanName()} {{ get; set; }}");
```
well I would suggest we construct something like `Property` but we currently do not have a class to define a property. We only have `ObjectTypeProperty`. This could be future work to refactor.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
ArcturusZhang: fair enough

src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
archerzz: In `ModelWriter`, we use this: https://github.com/Azure/autorest.csharp/blob/d472a8294859384089e456a8d8d80a18e328a479/src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs#L104

It's from `Declaration.ActualName`: https://github.com/Azure/autorest.csharp/blob/d472a8294859384089e456a8d8d80a18e328a479/src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs#L14

which comes from `CodeWriterDeclaration.SetActualName`: https://github.com/Azure/autorest.csharp/blob/d472a8294859384089e456a8d8d80a18e328a479/src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterDeclaration.cs#L22

There are a few ways to set `ActualName`. One path comes from `ToCleanName()`:

https://github.com/Azure/autorest.csharp/blob/d472a8294859384089e456a8d8d80a18e328a479/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs#L57

https://github.com/Azure/autorest.csharp/blob/d472a8294859384089e456a8d8d80a18e328a479/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs#L44

archerzz (approved):

src/AutoRest.CSharp/Common/Generation/Writers/ClientOptionsWriter.cs
pshao25: I created an issue https://github.com/Azure/autorest.csharp/issues/3940, since this PR changes API so it is better to get merged first.

archerzz (approved):

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
chunyu3: If it is a workaround for swagger, it may be good to add this logic in CodeModelConvert where convert CodeModel to InputModel.

test/AutoRest.TestServerLowLevel.Tests/url-path-items.cs
```diff
@@ -12,55 +12,76 @@ public class UrlPathItemsTests : TestServerLowLevelTestBase
     {
         [Test]
         public Task UrlPathItemGetAll() => TestStatus(async (host) =>
-            await new PathItemsClient(globalStringPath: "globalStringPath",
-                    globalStringQuery: "globalStringQuery",
+        {
+            var options = new AutoRestUrlTestServiceClientOptions()
```
chunyu3, line 16: If `globalStringQuery` is optional, eventually we can create a client without this parameter. But now, it seems that we already need to set this parameter in ClientOptions, that is not correct.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
ArcturusZhang: > If it is a workaround for swagger, it may be good to add this logic in CodeModelConvert where convert CodeModel to InputModel.

well this is my point as well - but if there is difficulties doing this, I am fine with the current solution because in typespec I believe `$host` is never a valid identifier name therefore it should be fine.

pshao25: I would prefer to add workaround
1. M4 is in maintaining mode.
2. It's better to get this PR in early because it changes API.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -673,7 +678,9 @@ private IEnumerable<LowLevelClient> CreateClients(IEnumerable<ClientInfo> client
                     SubClients = subClients
                 };
 
-                subClients.AddRange(CreateClients(clientInfo.Children, typeFactory, clientOptions, client));
+                subClients.AddRange(CreateClients(clientInfo.Children, typeFactory, clientOptions, client, parametersInClientOptions));
+                // parametersInClientOptions is assigned to ClientOptionsTypeProvider.AdditionalProperties before, which makes sure AdditionalProperties is readonly and won't change after ClientOptionsTypeProvider is built.
+                parametersInClientOptions.AddRange(client.GetOptionalParametersInOptions());
```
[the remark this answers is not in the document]
pshao25: Yes, if there is, then it will go to ClientOptions.

test/AutoRest.TestServerLowLevel.Tests/url-path-items.cs
pshao25: You want test `var options = new AutoRestUrlTestServiceClientOptions()`? 

This is not an e2e test, so above statement just shows we could create a class without initializing the optional properties.

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,56 @@
+## Service Client Constrctors
+
+Scope: This document only applies to DPG services.
```
ArcturusZhang, line 3:
```suggestion
Scope: This document only applies to DPG libraries.
```

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,56 @@
...
+
+There will be two **public** client constructors in a generated service client. One primary public constructor with all the user defined client parameters plus `ClientOptions`, and one secondary public constructor with all the user defined **required** client parameters. The secondary public constructor calls the primary public constructor underneath.
```
ArcturusZhang, line 5:
```suggestion
There will be two **public** client constructors in a generated service client. One primary public constructor with all the user defined client parameters plus a `ClientOptions` parameter, and one secondary public constructor with all the user defined **required** client parameters without the `ClientOptions` parameter. The secondary public constructor calls the primary public constructor underneath with default values of those optional parameters.
```

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,56 @@
...
+
+For example, user defines required parameters `Uri endpoint` and `AzureKeyCredential credential`, then the generated public constructors are:
+
+```C#
+public ServiceClient(Uri endpoint, AzureKeyCredential credential) : this(endpoint, credential, new ServiceClientOptions()) {}
+
+public ServiceClient(Uri endpoint, AzureKeyCredential credential, ServiceClientOptions options) {}
+```
+
+Normally, all the client parameters are required. But if user defines some additional optional client parameters, all the optional client parameters will go to `ClientOptions`, except `endpoint`.
```
ArcturusZhang, line 15:
```suggestion
Usually, all the client parameters are required. But if user defines some additional optional client parameters, all the optional client parameters will go to `ClientOptions`, except `endpoint`.
```

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,56 @@
...
+```C#
+public partial class ServiceClientOptions : ClientOptions
+{
+    ...
+
+    public string Optional { get; set; }
+}
+```
+
+The signature of the public constructors will keep the same, but its implementation will become to
```
ArcturusZhang, line 28:
```suggestion
The signature of the public constructors will not change, but its implementation will change to
```

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,56 @@
...
+```C#
+public ServiceClient(Uri endpoint, AzureKeyCredential credential) : this(endpoint, credential, new ServiceClientOptions()) {}
+
+public ServiceClient(Uri endpoint, AzureKeyCredential credential, ServiceClientOptions options)
+{
+    ...
```
ArcturusZhang, line 34: let us change this `...` to a comment like `/* omitting other irrelevant assignments*/

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,56 @@
...
+    _optional = options.Optional;
+}
+```
+
+If the optional client parameters user defines contains `endpoint`, it will still be in the client constructor, not going to `ClientOptions`. For example, user defines an optional parameter `string optional`, an optional `endpoint` with default value `"http://localhost:3000"`, and a required parameter `AzureKeyCredential credential`, the generated public constructors are:
```
ArcturusZhang, line 39:
```suggestion
If the user defined optional client parameters contain `endpoint`, it will still be in the client constructor, not be going to `ClientOptions`. For example, if user defines an optional parameter `string optional`, an optional `endpoint` with default value `"http://localhost:3000"`, and a required parameter `AzureKeyCredential credential`, the generated public constructors are:
```

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,56 @@
...
+
+```C#
+public ServiceClient(AzureKeyCredential credential) : this(new Uri("http://localhost:3000"), credential, new ServiceClientOptions()) {}
+
+public ServiceClient(Uri endpoint, AzureKeyCredential credential, ServiceClientOptions options)
+{
+    ...
```
ArcturusZhang, line 46: same here

docs/features/service_client_constructors.md
pshao25: This section has some problem. I adjust a bit.

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,68 @@
+## Service Client Constrctors
+
+Scope: This document only applies to DPG libraries.
+
+There will be two **public** client constructors in a generated service client:
+
+**One primary public constructor with:**
+- `endpoint` parameter (no matter it is required or optional).
```
ArcturusZhang, line 8:
```suggestion
- An `endpoint` parameter (no matter it is required or optional).
```

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,68 @@
...
+- All the user defined **required** client parameters.
+- A `ClientOptions` parameter.
+
+**One secondary public constructor with:**
+- All the user defined **required** client parameters (without the `ClientOptions` parameter and **optional** paramters).
+
+The secondary public constructor calls the primary public constructor underneath with default values of `endpoint` (if any) and `ClientOptions`.
+
+For example, user defines required parameters `Uri endpoint` and `AzureKeyCredential credential`, then the generated public constructors are:
```
ArcturusZhang, line 17:
```suggestion
For example, required parameters `Uri endpoint` and `AzureKeyCredential credential` are defined, then the generated public constructors are:
```

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,68 @@
...
+
+```C#
+public ServiceClient(Uri endpoint, AzureKeyCredential credential) : this(endpoint, credential, new ServiceClientOptions()) {}
+
+public ServiceClient(Uri endpoint, AzureKeyCredential credential, ServiceClientOptions options) {}
+```
+
+Usually, all the client parameters are required. But if user defines some additional optional client parameters, all the optional client parameters will go to `ClientOptions`, except `endpoint`.
```
ArcturusZhang, line 25:
```suggestion
Usually, all the client parameters are required. But if some additional optional client parameters are defined, all the optional client parameters will go to `ClientOptions`, except for the `endpoint` parameter.
```

docs/features/service_client_constructors.md
```diff
@@ -0,0 +1,68 @@
...
+
+For example, user defines an optional parameter `string optional`. The `ClientOptions` will have a new property as
```
ArcturusZhang, line 27:
```suggestion
For example, if an optional parameter `string optional` is defined, the `ClientOptions` will have a new property as
```

ArcturusZhang (approved):
We should not use the word user because it is kind of ambiguous - it could be the user of the generator or the user of the generator.
We could use passive phrase to avoid that

chunyu3 (approved):

lirenhe (approved):

PR #3930, opened 2023-11-14
Fix the name of the types in default parameter descriptions

AlexanderSher:


src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -97,7 +97,7 @@ public static Parameter FromInputParameter(in InputParameter operationParameter,
         public static FormattableString CreateDescription(InputParameter operationParameter, CSharpType type, IEnumerable<string>? values, Constant? defaultValue = null)
         {
             FormattableString description = string.IsNullOrWhiteSpace(operationParameter.Description)
-                ? (FormattableString)$"The {operationParameter.Type.Name} to use."
+                ? (FormattableString)$"The {TypeFactory.GetInputType(type).ToStringForDocs()} to use."
```
ArcturusZhang, line 100: Could we assume that the type here is **exactly** the type of the parameter which we are creating descriptions on?
I believe we should ensure that - and then we do not need the `TypeFactory.GetInputType` here because not every time when we create the description, it represents the "input" type.

samples/Azure.ResourceManager.Sample/Generated/Extensions/MockableSampleSubscriptionResource.cs
```diff
@@ -110,9 +110,9 @@ public virtual VirtualMachineExtensionImageCollection GetVirtualMachineExtension
         /// </list>
         /// </summary>
         /// <param name="location"> The name of a supported Azure region. </param>
-        /// <param name="publisherName"> The String to use. </param>
-        /// <param name="type"> The String to use. </param>
-        /// <param name="version"> The String to use. </param>
+        /// <param name="publisherName"> The string to use. </param>
```
m-nash, line 113: I think we should use crefs here so that when it links to a non primitive you have a nice reference.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
m-nash: Same question why didn't we just use `this.Type`?

src/AutoRest.CSharp/Mgmt/Output/Models/MgmtPropertyBag.cs
```diff
@@ -43,8 +44,8 @@ protected override TypeProvider EnsurePackModel()
             foreach (var parameter in _paramsToKeep)
             {
                 var inputParameter = _operation.Parameters.FirstOrDefault(p => string.Equals(p.Name, parameter.Name, StringComparison.OrdinalIgnoreCase));
-                var description = parameter.Description ?? $"The {parameter.Name}";
-                var property = new InputModelProperty(parameter.Name, parameter.Name, description.ToString(), inputParameter!.Type, parameter.DefaultValue == null, false, false)
+                var description = inputParameter is not null && !string.IsNullOrEmpty(inputParameter.Description) && parameter.Description is not null ? parameter.Description.ToString() : $"The {parameter.Name}";
```
archerzz, line 47: I think `inputParameter` is not null here, otherwise code gen wil fail on next line, e.g. `inputParameter!.Type`
```suggestion
                var description = !string.IsNullOrEmpty(inputParameter!.Description) && parameter.Description is not null ? parameter.Description.ToString() : $"The {parameter.Name}";
```

archerzz (approved):

src/AutoRest.CSharp/Mgmt/Output/Models/MgmtPropertyBag.cs
AlexanderSher: Fixed

m-nash (approved):

PR #3931, opened 2023-11-15
Remove automatic added apiversion if the spec is not defining it in operation

### LINKED ISSUE #3856: api-version parameter gets added even in not declared.
I'm trying to generate a client from TypeSpec and I none of my operations declare an `api-version` parameter, but when I generate the C# client, the parameter always gets added (I'm using `@versioned` mostly for book keeping, and if I remove it the emitter errors out).

The following is a small repro I created:
```typescript
import "@typespec/rest";
import "@typespec/http";
import "@typespec/versioning";

using TypeSpec.Http;
using TypeSpec.Versioning;

@service({
  title: "<placeholder>",
})
@server(
  "{endpoint}",
  "<placeholder>",
  {
    endpoint: string,
  }
)
@versioned(APIVersion)
namespace Example;


enum APIVersion {  
  v20231001Preview: "2023-10-01-preview",
}

interface API {
    @route("/foo")
    @post
    op foo(bar: string, baz: safeint): string;
}
```
This results in a client with the following create request member function:
```csharp
internal HttpMessage CreateFooRequest(RequestContent content, RequestContext context)
{
    var message = _pipeline.CreateMessage(context, ResponseClassifier200);
    var request = message.Request;
    request.Method = RequestMethod.Post;
    var uri = new RawRequestUriBuilder();
    uri.Reset(_endpoint);
    uri.AppendPath("/foo", false);
    uri.AppendQuery("api-version", _apiVersion, true);
    request.Uri = uri;
    request.Content = content;
    return message;
}
```

pshao25:
Fix #3856, fix https://github.com/Azure/autorest.csharp/issues/3939

Behavior change:
1. The `ServiceVersion` in `ClientOptions`: from "contain both versions from `@versioned` and `@service.version`" to "only contain versions from `@versioned`". In practice, we cannot write both versions in `@versioned` and `@service.version` in tsp. So in some test cases like `Authoring-TypeSpec`, you would see that I remove the version from `@service.version` to `@versioned` in tsp and the generated `ClientOptions` keeps the same.
2. The `api-version` in generated method: from "if no api-version defined in operation level than still create one" to "if no api-version defined in operation level than not magically add it". `FirstTest-TypeSpec` is an example that no `api-version` is generated in the operation.
3. Remove hard-coded logic for non-branding, which being said the branding behavior is the same as the non-branding behavior.

Ongoing cadl-ranch tests: https://github.com/Azure/cadl-ranch/pull/463
SDK change: NO.


src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -86,59 +86,51 @@ export function createModelForService(
...
+
+    const apiVersions: Set<string> | undefined = new Set<string>();
+    let defaultApiVersion: string | undefined = undefined;
     const versions = getVersions(program, service.type)[1]?.getVersions();
     if (versions) {
         for (const ver of versions) {
             apiVersions.add(ver.value);
         }
-        version = versions[versions.length - 1].value; //default version
+        defaultApiVersion = versions[versions.length - 1].value;
```
chunyu3, line 98: Better to check the count of versions. in case, service defined @versioned, but the enum is empty. Add the check to avoid crash.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -197,30 +189,39 @@ export function createModelForService(
...
+                    const apiVersionInOperation =
+                        op.Parameters[apiVersionIndex];
+                    if (!apiVersionInOperation.DefaultValue?.Value) {
+                        apiVersionInOperation.DefaultValue =
+                            apiVersionParam.DefaultValue;
+                    }
+                    /**
+                     * replace to the global apiVersion parameter if the apiVersion defined in the operation is the same as the global service apiVersion parameter.
+                     * Three checkpoints:
+                     * the parameter is query parameter,
```
lirenhe, line 209: What if api-version is defined in path instead of query parameter?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -86,59 +86,51 @@ export function createModelForService(
...
+    let defaultApiVersion: string | undefined = undefined;
     const versions = getVersions(program, service.type)[1]?.getVersions();
     if (versions) {
         for (const ver of versions) {
             apiVersions.add(ver.value);
         }
-        version = versions[versions.length - 1].value; //default version
+        defaultApiVersion = versions[versions.length - 1].value;
     }
+    const apiVersionParam: InputParameter | undefined = defaultApiVersion
```
lirenhe, line 100: Is it possible to add some unit test to verify the behavior of api-version in different cases?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
pshao25: If the enum is empty, compiler will throw error.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -86,59 +86,51 @@ export function createModelForService(
         context,
         "@azure-tools/typespec-csharp"
     );
-    const title = service.title;
     const serviceNamespaceType = service.type;
-    const apiVersions: Set<string> = new Set<string>();
-    let version = service.version;
-    if (version && version !== mockApiVersion) {
```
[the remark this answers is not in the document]
pshao25: This logic here is to exclude the version in 
```
@service(
    {
        title: "Media Types Service",
        version: "2021-03-25"
    }
)
```
Now we are not considering this version anymore. We only consider the versions in `@versioned(Versions)`

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
pshao25: This is how we define api-version as path parameter:
```
@server(
  "{Endpoint}/anomalydetector/{ApiVersion}",
  "",
  {
    Endpoint: string,
    @path
    ApiVersion: APIVersion,
  }
)
```
Then we don't need to define apiversion in operation level.

Directory.Build.props
```diff
@@ -15,6 +15,8 @@
     <NoWarn Condition="'$(IsTestGenerationSrcProject)' == 'true'">$(NoWarn);AZC0012</NoWarn>
     <!-- TODO we should figure out how to make our generator create the right thing for these two warnings https://github.com/Azure/autorest.csharp/issues/3706 -->
     <NoWarn Condition="'$(IsTestGenerationSrcProject)' == 'true'">$(NoWarn);AZC0006;AZC0007</NoWarn>
+    <!-- TODO new project need this-->
+    <NoWarn Condition="'$(IsTestGenerationSrcProject)' == 'true'">$(NoWarn);AZC0008</NoWarn>
```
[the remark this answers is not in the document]
pshao25: I opened up https://github.com/Azure/autorest.csharp/issues/4075 to track.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -43,7 +43,7 @@ internal class RestClientBuilder
         public RestClientBuilder(IEnumerable<InputParameter> clientParameters, TypeFactory typeFactory)
         {
             _typeFactory = typeFactory;
-            _parameters = clientParameters.ToDictionary(p => p.Name, BuildConstructorParameter);
+            _parameters = clientParameters.DistinctBy(p => p.Name).ToDictionary(p => p.Name, BuildConstructorParameter);
         }
```
pshao25, line 47: For versioned service, if one has query apiversion and the other has path apiversion, there will be two apiversion parameters.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -195,31 +169,18 @@ export function createModelForService(
...
-                        InputOperationParameterKind.Client &&
-                    apiVersionInOperation.Location === apiVersionParam.Location
-                ) {
-                    op.Parameters[apiVersionIndex] = apiVersionParam;
+                        defaultApiVersionConstant;
                 }
             } else {
-                op.Parameters.push(apiVersionParam);
+                apiVersionInOperation.Kind = InputOperationParameterKind.Method;
+                apiVersionInOperation.IsApiVersion = false;
```
chunyu3, line 183: Why set the `IsApiVersion` false? The parameter is api-version.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
chunyu3: I don't think typespec compiler will throw error when the enum is empty.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
pshao25: It is unversioned scenario, so it is not.

chunyu3: The parameter is api-version parameter, but it is not get the value from clientOptions. Maybe in autorest.csharp codegen, it will use `IsApiVersion` flag to do some special handling.

chunyu3 (approved):

lirenhe (approved):

PR #3935, opened 2023-11-15
Validates null for extension method parameter

ArcturusZhang:
# Description

The code in our extension looks like this
```csharp
private static MockableMgmtConstantsArmClient GetMockableMgmtConstantsArmClient(ArmClient client)
{
    return client.GetCachedClient(client0 => new MockableMgmtConstantsArmClient(client0));
}
```
This code throws NullReferenceException when client is null, which surprises our customer, for instance somehow they write code like this:
```csharp
var resourceGroup = /* some code to get it */
var vmCollection = resourceGroup.GetVirtualMachines(); // this will throw null reference exception
```

Since these operations require to have a resourceGroup, we should validate those are not null to throw meaningful exceptions.

# Checklist

# Ready to Land?


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "7.0.101",
+    "version": "8.0.100",
```
ArcturusZhang, line 3: will revert this - just updated the visual studio and the dotnet sdk 7 was removed.

test/TestProjects/MgmtScopeResource/Generated/Extensions/MgmtScopeResourceExtensions.cs
```diff
@@ -23,31 +23,43 @@ public static partial class MgmtScopeResourceExtensions
     {
         private static MockableMgmtScopeResourceArmClient GetMockableMgmtScopeResourceArmClient(ArmClient client)
         {
+            Argument.AssertNotNull(client, nameof(client));
```
ArcturusZhang, line 26: these validations are added.
The invocation will fail anywhere if the client is null - it will throw `NullReferenceException` which might confuses our customers.

test/TestProjects/MgmtScopeResource/Generated/Extensions/MgmtScopeResourceExtensions.cs
```diff
@@ -346,6 +366,7 @@ public static FakePolicyAssignmentResource GetFakePolicyAssignmentResource(this
         /// </summary>
         /// <param name="client"> The <see cref="ArmClient" /> instance the method will execute against. </param>
         /// <param name="id"> The resource ID of the resource to get. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="client"/> is null. </exception>
```
ArcturusZhang, line 369: the documents are updated accordingly since the exception might be thrown when this parameter is null.

src/AutoRest.CSharp/Common/Input/ApiTypes.cs
```diff
@@ -53,10 +53,6 @@ public FormattableString GetHttpPipelineCreateMessageFormat(bool withContext)
 
         public abstract Type ClientOptionsType { get; }
 
-        public abstract Type ArgumentType { get; }
-        public string AssertNotNullOrEmptyName => nameof(ClientUtilities.AssertNotNullOrEmpty);
-        public string AssertNotNullName => nameof(ClientUtilities.AssertNotNull);
```
ArcturusZhang, line 58: These are implemented in the snippets

ArthurMa1978 (approved):

archerzz (approved):

src/AutoRest.CSharp/Common/Output/Expressions/ExtensibleSnippets.ArgumentSnippets.cs
```diff
@@ -0,0 +1,30 @@
...
+using AutoRest.CSharp.Common.Output.Expressions.Statements;
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
+using AutoRest.CSharp.Output.Models.Shared;
+using static AutoRest.CSharp.Common.Output.Models.Snippets;
+
+namespace AutoRest.CSharp.Common.Output.Expressions
+{
+    internal abstract partial class ExtensibleSnippets
+    {
+        internal abstract class ArgumentSnippets
```
live1206, line 13: Why is this an inner class instead of its own file?

ArcturusZhang: This does not have particular reason, just a way Alexander chose to have them well organized.
It just works the same if we choose to have those individual types and put them inside ExtensibleSnippets namespace.
No one should use the class name directly - the expression `Snippets.Extensible.Argument` returns you an singleton of this class.

live1206 (approved):

PR #3937, opened 2023-11-16
Fix formatting in docs

AlexanderSher:
Unifies writing of types in docs, removes `StringBuilder` usage in the middle


src/AutoRest.CSharp/LowLevel/Generation/DpgClientWriter.cs
```diff
@@ -603,19 +603,19 @@ private void WriteSampleRefsIfNecessary(MethodSignature methodSignature, bool is
                 return;
 
             var sampleProvider = _library.GetSampleForClient(_client);
-            // do not write this part when there is no sample provider
+            // do not write this part when there is no sample providerPara
```
ArcturusZhang, line 606: typo? maybe we need to revert this change

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -484,7 +484,7 @@ protected virtual Method BuildCreateResourceIdentifierMethod()
             var signature = new MethodSignature(
                 Name: "CreateResourceIdentifier",
                 null,
-                Description: $"Generate the resource identifier of a <see cref=\"{Type.ToStringForDocs()}\"/> instance.",
```
ArcturusZhang, line 487: I assume this method `CSharpType.ToStringForDocs` is no longer needed after this change? Could we remove it?

src/AutoRest.CSharp/LowLevel/Generation/DpgClientWriter.cs
AlexanderSher: yep

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
AlexanderSher: Done

src/AutoRest.CSharp/Common/Generation/Writers/DocumentationWriterExtensions.cs
```diff
@@ -68,6 +70,48 @@ public static CodeWriter WriteXmlDocumentationReturns(this CodeWriter writer, Fo
...
+            {
+                AppendTypeWithShortNames(parameter.Type, sb);
+                sb.Append(",");
+            }
+
+            sb.Remove(sb.Length - 1, 1);
+            sb.Append(")");
+
+            memberId = sb.ToString();
+            return writer.LineRaw($"/// <include file=\"{filename}\" path=\"doc/members/member[@name='{memberId}']/*\" />");
```
ArcturusZhang, line 90: does this count as "writing documentations"? should we have a check if the writer is in the mode of writing documents?

ArcturusZhang (approved):
LGTM

src/AutoRest.CSharp/Common/Generation/Writers/DocumentationWriterExtensions.cs
AlexanderSher: No, we lie here to the code writer on purpose. `include` doesn't follow our type formatting rules, hence a separate method that doesn't call `AppendXmlDocumentation` and the string is fully constructed and added via `LineRaw`.

test/TestProjects/Models-TypeSpec/src/Generated/ModelsTypeSpecClient.cs
```diff
@@ -1022,7 +1022,7 @@ public virtual Response<OutputModel> FixedFloatEnum(FixedFloatEnum input, Cancel
         /// </item>
         /// <item>
         /// <description>
-        /// Please try the simpler <see cref="FixedFloatEnumAsync(FixedFloatEnum,CancellationToken)"/> convenience overload with strongly typed models first.
+        /// Please try the simpler <see cref="FixedFloatEnumAsync(Models.FixedFloatEnum,CancellationToken)"/> convenience overload with strongly typed models first.
```
AlexanderSher, line 1025: `Models` namespace is needed here because there is method with the same name - `FixedFloatEnum`.

archerzz (approved):

m-nash (changes requested):
This is a conflict with https://github.com/Azure/autorest.csharp/pull/3891.  Lets wait for this one to go in first.

AlexanderSher:
https://github.com/Azure/azure-sdk-for-net/pull/40091

AlexanderSher:
https://github.com/Azure/azure-sdk-for-net/pull/40126

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -102,7 +102,7 @@ public CSharpType(TypeProvider implementation, string ns, string name, bool isVa
         public bool IsLiteral { get; init; }
         public Constant? Literal { get; init; }
         public bool IsUnion { get; }
-        public CSharpType[] UnionItemTypes { get; } = Array.Empty<CSharpType>();
+        public IReadOnlyList<CSharpType> UnionItemTypes { get; } = Array.Empty<CSharpType>();
```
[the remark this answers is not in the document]
AlexanderSher: We have mutable `CSharpType[] Arguments` as a relic of the past when arrays have been used everywhere. Now we use read-only types for public properties that shouldn't be changed. Ideally, `Arguments` should be `IReadOnlyList<CSharpType>` as well, cause changing the list of generic types would cause an error

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -152,6 +152,11 @@ private bool ArgumentsEquals(Type[] genericArguments)
 
         public override int GetHashCode() => HashCode.Combine(_implementation, _type, ((System.Collections.IStructuralEquatable)Arguments).GetHashCode(EqualityComparer<CSharpType>.Default));
 
+        public CSharpType GetGenericTypeDefinition()
```
m-nash, line 155: Can you add a unit test which does what we discussed in chat?

CSharpType.Arguments should do what System.Type.GetGenericArguments() does
csharpType.GetGenericTypeDefinition().Arguments should do what type.GetGenericTypeDefinition().GetGenericArguments()

Then validate this with `new CSharpType(typeof(List<>), typeof(string))` and `new CSharpType(typeof<List<string>>)`

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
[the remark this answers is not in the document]
m-nash: Can we add a work item to fix `Arguments` in that case we can tackle in a separate PR.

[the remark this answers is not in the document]
AlexanderSher: https://github.com/Azure/autorest.csharp/issues/3958

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
AlexanderSher: Done

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -152,6 +152,11 @@ private bool ArgumentsEquals(Type[] genericArguments)
...
+            => _type is null
+                ? new(_implementation!, Namespace, Name, IsValueType, IsEnum, IsNullable, Array.Empty<CSharpType>())
+                : new(_type, IsNullable, _type.GetGenericArguments().Select(p => new CSharpType(p)).ToArray());
```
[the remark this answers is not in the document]
AlexanderSher: Fixed per discussion offline

m-nash (approved):
Very nice change thank you for this clean up!

PR #3941, opened 2023-11-16
Upgrade autorest.csharp to net7.0

AlexanderSher:


AlexanderSher:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3268492&view=logs&j=110ffb37-f6b6-5d2e-4482-cee28f4a1c07&t=c1e6de2a-7bc7-5f8d-32b9-8690377ad59f

m-nash (changes requested):
Can you add some information to the description around what makes us want this change.  What impacts it will have on external users.  What mitigations they should take if they don't want these changes, what mitigations they should take if they do.

Also since this is a CLI tool what version rules do you think we should take here around this type of breaking change?  Should this be a major version bump?

ArcturusZhang:
i am curious that why when there is net8.0 released and we update to net 7.0? could we just directly bump to net8.0?

readme.md
```diff
@@ -57,7 +57,7 @@ ___
 ## Prerequisites
 
 - [NodeJS (18.x.x)](https://nodejs.org/en/) 
-- [.NET Core SDK (5.0.100)](https://dotnet.microsoft.com/download/dotnet-core/5.0)
+- [.NET Core SDK (7.0.404)](https://dotnet.microsoft.com/en-us/download/dotnet/7.0)
```
archerzz, line 60:
```suggestion
- [.NET Core SDK (7.0.404 or above)](https://dotnet.microsoft.com/en-us/download/dotnet/7.0)
```

archerzz:
> i am curious why when there is net8.0 released and we update to net 7.0? could we just directly bump to net8.0?

@ArcturusZhang `azure-sdk-for-net` adopts up to `net7.0`. So better to align with that.

m-nash:
Making this change here https://github.com/Azure/autorest.csharp/pull/4055 so closing this PR.

PR #3943, opened 2023-11-16
Update to tsp 50

m-nash:
# Description

# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/package.json
```diff
@@ -50,18 +62,23 @@
...
-        "@typespec/http": "0.49.0",
-        "@typespec/library-linter": "0.49.0",
-        "@typespec/rest": "0.49.0",
-        "@typespec/versioning": "0.49.0",
         "fs-extra": "^10.1.0",
         "js-yaml": "^4.1.0",
         "json-serialize-refs": "0.1.0-0",
         "winston": "^3.8.2"
+    },
+    "peerDependencies": {
```
mikeharder, line 70: Does typespec-csharp actually depend on all these?  typespec-python has a much smaller list:

```
    "peerDependencies": {
        "@azure-tools/typespec-azure-core": ">=0.36.0 <1.0.0",
        "@azure-tools/typespec-client-generator-core": ">=0.37.0-dev.3 <1.0.0",
        "@typespec/compiler": ">=0.50.0 <1.0.0",
        "@typespec/http": ">=0.50.0 <1.0.0",
        "@typespec/rest": ">=0.50.0 <1.0.0",
        "@typespec/versioning": ">=0.50.0 <1.0.0"
    },
```

https://github.com/Azure/autorest.python/blob/881659f763f4ee98d47fcf25a8082ca804d4a289/packages/typespec-python/package.json#L44

src/TypeSpec.Extension/Emitter.Csharp/package.json
```diff
@@ -50,18 +62,23 @@
...
+        "@azure-tools/typespec-azure-resource-manager": ">=0.35.0 <1.0.0",
```
mikeharder, line 71:
```suggestion
        "@azure-tools/typespec-azure-resource-manager": ">=0.36.0 <1.0.0",
```

I think you'd want to set the min version to the current latest, unless you intentionally want to support version N-1 of these deps.

package.json
```diff
@@ -7,7 +7,17 @@
     "@azure-tools/cadl-ranch-mockapis": "file:test/CadlRanchMockApis",
     "@microsoft.azure/autorest.testserver": "3.3.48",
     "autorest": "3.6.3",
-    "@typespec/openapi": "^0.49.0",
-    "@typespec/openapi3": "^0.49.0"
+    "@azure-tools/typespec-azure-resource-manager": "0.36.0",
+    "@azure-tools/typespec-client-generator-core": "0.36.0",
+    "@azure-tools/typespec-azure-core": "0.36.0",
```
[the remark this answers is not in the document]
mikeharder: Tests in this repo import these packages, for example here:

https://github.com/Azure/autorest.csharp/blob/b5a406804e7659a9597ef1988c7dca5519c8c68e/test/TestProjects/FirstTest-TypeSpec/FirstTest-TypeSpec.tsp#L3

When calling `tsp compile` on these specs, these packages must be loadable by `npm`.  These packages are dependencies in `emitter.csharp/package.json`, so this may have worked previously (by accident), but `npm` doesn't guarantee that transitive dependencies are installed at the top level of `node_modules`.  The only way to guarantee these packages are loadable, is to have them in a `package.json` in a folder **above** the specs.

One other solution I just thought of, would be to move the TypeSpec tests under the `emitter.csharp` folder.  This should also work, if you run `tsp compile .` from the spec folder.

src/TypeSpec.Extension/Emitter.Csharp/package.json
m-nash: I think n-1 is best since we don't know if all emitters have updated and we want to be compatible

src/TypeSpec.Extension/Emitter.Csharp/package.json
m-nash: I am just adding everything here because I don't want to have to deal with it any longer.

package.json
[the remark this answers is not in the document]
m-nash: This is because its the only way to force a package to be installed at the root level

src/TypeSpec.Extension/Emitter.Csharp/package.json
mikeharder: But this means you also can only use features in (n-1), and to verify this, it would be best to set your devDep to (n-1).

mikeharder: I pushed an update to use latest as min peerDep, which matches autorest-python.  I'm pretty sure this is what you want, so you can use the latest features in your dependencies.

mikeharder:
@m-nash: I believe this PR is breaking the JS unit tests for typespec-csharp: https://github.com/Azure/typespec-azure/pull/3718

```
  5 passing (3s)
  32 failing

  1) Test encode duration
       encode iso8601 for duration query parameter :
     Error: No emitter name found in program
      at parseEmitterName (node_modules/@azure-tools/typespec-client-generator-core/dist/src/internal-utils.js:3:15)
      at createSdkContext (node_modules/@azure-tools/typespec-client-generator-core/dist/src/decorators.js:217:22)
      at Context.<anonymous> (src/TypeSpec.Extension/Emitter.Csharp/dist/test/Unit/encode.test.js:21:28)
```

jsquire (approved):

PR #3952, opened 2023-11-17
fix: properly ref custom model in generated code

jorgerangel-msft:
# Description

# Checklist

# Ready to Land?


test/TestProjects/Customizations-TypeSpec/src/Generated/CustomizationsInTspClient.cs
```diff
@@ -152,6 +152,114 @@ public virtual Response RoundTrip(RequestContent content, RequestContext context
             }
         }
 
+        /// <summary> Sample operation with request body of a renamed model. </summary>
+        /// <param name="input"> The RenamedModel to use. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="input"/> is null. </exception>
+        /// <include file="Docs/CustomizationsInTspClient.xml" path="doc/members/member[@name='FooAsync(RenamedModel,CancellationToken)']/*" />
+        public virtual async Task<Response<RenamedModel>> FooAsync(RenamedModel input, CancellationToken cancellationToken = default)
```
ArcturusZhang, line 160: I did not really see a case in our test project that the model type has been renamed by code gen attributes, and the parameter name changed accordingly per the description of this PR.
For instance, this one, the type of `input` has been renamed, but the parameter name `input` here is not changed (in its tsp the name of this parameter is also `input`). Should this change accordingly?

Or this PR is only to change the description when the description is absent? this is the only thing I notice changed in this PR

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -16,7 +16,7 @@
 
 namespace AutoRest.CSharp.Output.Models.Shared
 {
-    internal record Parameter(string Name, FormattableString? Description, CSharpType Type, Constant? DefaultValue, ValidationType Validation, FormattableString? Initializer, bool IsApiVersionParameter = false, bool IsResourceIdentifier = false, bool SkipUrlEncoding = false, RequestLocation RequestLocation = RequestLocation.None, SerializationFormat SerializationFormat = SerializationFormat.Default, bool IsPropertyBag = false)
+    internal record Parameter(string Name, FormattableString? Description, CSharpType Type, Constant? DefaultValue, ValidationType Validation, FormattableString? Initializer, bool IsApiVersionParameter = false, bool IsResourceIdentifier = false, bool SkipUrlEncoding = false, RequestLocation RequestLocation = RequestLocation.None, SerializationFormat SerializationFormat = SerializationFormat.Default, bool IsPropertyBag = false, string? DefaultName = null)
```
ArcturusZhang, line 19: well I do not think it is a good name here `DefaultName`. Should it be `NameInRequest` or `WireName` or something similar?
Looks like this is only to fix the issue in sample generator that if we renamed the parameter to be a new name instead of the name in its spec, the sample generator will not be able to find which parameter to use.

Theoretically we should need this name in request when building the rest operation methods (for instance when the parameter is a query), but we did that by querying the `NameInRequest` property in `InputParameter` instead of referencing `Parameter`.
But I believe it might be nice we have that here, and the sample generator could query parameters using this property `NameInRequest` to avoid the issues caused by renamings

jorgerangel-msft: sure, I can rename this property.

You are correct, this is to handle generating the samples correctly when a sample uses a custom model. It seems the lookup to find the example value for a parameter of model type uses the "original" model name vs the custom model. Though I'm not sure I entirely follow the last part of you comment, are you suggesting I change how the example values are looked up in the sample generator ?

test/CadlRanchProjects/special-words/src/Generated/And.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace SpecialWords
 {
-    /// <summary> The and. </summary>
+    /// <summary> The And. </summary>
```
m-nash, line 13: Undo this change and open an issue to track if we want to do this as well as should this be a cref.

test/TestProjects/Customizations-TypeSpec/Customizations-TypeSpec.tsp
```diff
@@ -188,3 +188,9 @@ model RootModel {
 @get
 @convenientAPI(true)
 op roundTrip(@body input: RootModel): RootModel;
+
+@route("/foo")
```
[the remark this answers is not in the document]
jorgerangel-msft: Sure. I've generated two baseline scenarios, one with a named operation parameter that references the model to rename, and one which uses the spread operator again the model to rename on the operation parameter.

1. [Named Parameter](https://github.com/jorgerangel-msft/autorest.csharp/commit/83cc9dac365248605d0c528875bd2cf8b0e4d81f). This generates correctly now without my changes. The default param description wasn't generated correctly before, but looks like this was fixed [recently](https://github.com/Azure/autorest.csharp/pull/3930/files#diff-1d8080793123f1f633e4fcb80852199dacf38a51b2fd96f50e256756507f0429).
2. [Spread Parameter](https://github.com/Azure/autorest.csharp/commit/052b0a5859f0e41942eb6586af4b2134253909c2). Here the variable names are still generated using the model name before the renaming.

Though in both scenarios, the correct model type is used so my memory must've been fuzzy on that part.

test/CadlRanchProjects/special-words/src/Generated/And.cs
jorgerangel-msft: Sure, I opened an item for this https://github.com/Azure/autorest.csharp/issues/3984

test/TestProjects/Customizations-TypeSpec/src/Generated/CustomizationsInTspClient.cs
m-nash: @ArcturusZhang do you consider that @projectedName and CodeGen custom name are a different intent from the customer?  I was thinking of these two things as two different ways to do exactly the same thing.

If the spec was swagger would you still think that changing the name with CodeGen shouldn't change this?

ArcturusZhang: Yes because the purpose of `CodeGenModel` is only to change the model on the generated code side, and projectedName is a customization on the tsp side and tsp implies something inside it.
The reason our customer has questions on this case, is that they are using the automatically generated parameter name which comes from the name of the type of the parameter.
it is hard to believe if they define something like this:
```
op foo(body: OldName): void;
```
and they introduce a CodeGenModel to change the name of the model, and they will expect the name of the  parameter should change accordingly.

Therefore my point is that it only makes sense we change the parameter name only when the customer is writing this:
```
op foo(...OldName): void;
```
because this expression implies the name of the parameter should be the same as the type name, and they would have the expectation of the generated result (even after they made a customization to the model name)

jorgerangel-msft: @ArcturusZhang - this change _shouldn't_ change the parameter name if the user supplied a named parameter in their tsp (I can double check to make sure). There has been other changes/fixes since I first opened this PR, so this PR now simply fixes the case where a user provides the model via the spread operator, like your example:

```
op foo(...OldName): void;
```

I have added a test scenario in this PR that shows that. Looks like the references in the summary descriptions were fixed in another change, as I noted [here](https://github.com/Azure/autorest.csharp/pull/3952#discussion_r1408345318). Either way, I'm open to whatever we think is best for the customer, whether that's telling them to use `@projectedName` or we add this change in the generator. I was not aware of `@projectedName` 😄

ArcturusZhang: Yeah I believe the projectedName decorator should be the correct and most convenient solution for our customer.
But in the meantime, I am OK with the changes here

m-nash: I think we are missing the fact that the generator works with more than just tsp.  I understand that if you are using tsp then @projectedName is the best solution but what if you aren't?  I believe any rename capability we have should work completely, and right now renaming via CodeGen doesn't work completely.  We should either fix it or remove the capability.

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -115,6 +115,39 @@ public static FormattableString CreateDescription(InputParameter operationParame
...
 
+        /// <summary>
+        /// This method constructs the variable name for an input parameter. If the input parameter type is an input model type,
+        /// and the input parameter name is the same as the input parameter type name, the variable name is constructed using the supplied CSharpType name. Otherwise,
+        /// it will use the input parameter name by default.
+        /// </summary>
+        /// <param name="param">The input parameter.</param>
+        /// <param name="type">The constructed CSharpType for the input parameter.</param>
+        /// <returns>A string representing the variable name for the input parameter.</returns>
+        public static string ConstructParameterVariableName(InputParameter param, CSharpType type)
```
m-nash, line 126: Does this need to be public?

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -115,6 +115,39 @@ public static FormattableString CreateDescription(InputParameter operationParame
...
+
+                    if (paramName.Equals(paramInputTypeName))
+                    {
+                        variableName = !string.IsNullOrEmpty(type.Name) ? type.Name.ToVariableName() : variableName;
+                    }
+
+                }
+            }
+
+            return variableName;
```
m-nash, line 148: previously if InputParamter was null we would throw, now we silently return an empty string.  I believe we should throw since this is an invalid scenario you can't have and empty name for a parameter

Issue #3853, opened 2023-10-18
TypeFactory can't resolve anonymous types for method return


AlexanderSher:
In `FirstTest-TypeSpec` test, `FirstTestTypeSpecClient.ReturnsAnonymousModel` method returns `Response<object>`:
https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/FirstTest-TypeSpec/src/Generated/FirstTestTypeSpecClient.cs#L824

It expected to return `Response<ReturnsAnonymousModel_101>`

ArcturusZhang:
This has been fixed in [this PR](https://github.com/Azure/autorest.csharp/pull/4091)

Issue #3876, opened 2023-10-28
Fix duplicates in the list of derived models or remove the list


AlexanderSher:
C# TypeSpec emitter produces duplicates when fills in derived models list.

Here is an example of the same model presented twice.
- as a item in the DerivedModels array: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/ConfidentLevels-TypeSpec/src/Generated/tspCodeModel.json#L719-L745
- as a separate model: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/ConfidentLevels-TypeSpec/src/Generated/tspCodeModel.json#L793-L819
Both point to the same base model id ref85.

Possible solutions:
- Fix emitter output.
- Remove `DerivedTypes` from protocol, assign them during deserialization or in a separate visitor.
- Remove `DerivedTypes` completely.

chunyu3:
When there is a `BaseModel` and a derived model `DerivedModel`, in the emitted json, `BaseModel` has a reference to `DerivedModel` and `DerivedModel` also has a reference to `BaseModel`. If both use the `$ref`, there will be a circle, and it will throw exception when deserailize this json, so we duplicate the derived model to break the circle.

The possible solution is remove one reference, either from `BaseModel` or `DerivedModel` from emitted json, and then need to calculate the relationship in codegen runtime.

AlexanderSher:
Fixed in https://github.com/Azure/autorest.csharp/pull/3967

PR #3953, opened 2023-11-17
Introduce visitors for input namespaces

AlexanderSher:
Provides mechanism to change `InputNamespace` 
Also fixes https://github.com/Azure/autorest.csharp/issues/3853


src/AutoRest.CSharp/Common/Input/InputTypes/Examples/InputClientExample.cs
```diff
@@ -5,5 +5,10 @@
 
 namespace AutoRest.CSharp.Common.Input.Examples
 {
-    internal record InputClientExample(InputClient Client, IReadOnlyList<InputParameterExample> ClientParameters);
+    /// <summary>
+    /// Set of client parameter sample values of a specific type
+    /// </summary>
+    /// <param name="Key">Type of a sample. E.g.: short version sample, all parameters sample, etc.</param>
```
ArcturusZhang, line 11: I believe the `key` here should be `key of the example`?
Just like what we have in swaggers:
```
"/request" : {
    "get" : {
	"x-ms-examples": {
	    "the key of example": "file path of the example"
	}
    }
}
```
maybe we could refer that concept here so that it would be more comprehensive

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Input/MergeDerivedModelsVisitor.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        public static InputNamespace Visit(InputNamespace rootNamespace)
+        {
+            if (rootNamespace.Models.All(static m => m.BaseModel is not {} bm || bm.DerivedModels.Contains(m)))
+            {
+                return rootNamespace;
+            }
+
+            var modelsToReplace = rootNamespace.Models
```
archerzz, line 27: A model inherits a base model, but not in `baseModel.DerivedModels`. This looks like an issue in TSP compiler or our emitter. Shall we fix it in compiler or emitter?

AlexanderSher: https://github.com/Azure/autorest.csharp/issues/3876 - we need to decide what to do with it. However, since correct model graph is critical for consolidation work, I'll keep the workaround until 3876 is resolved.

src/AutoRest.CSharp/Common/Input/InputTypes/Examples/InputClientExample.cs
AlexanderSher: Done

AlexanderSher:
Changes in azure-sdk-for-net: https://github.com/Azure/azure-sdk-for-net/pull/40125
Methods in factories are reordered because of the ordering in visitors.

src/AutoRest.CSharp/LowLevel/Output/DpgClient.cs
```diff
@@ -21,7 +21,7 @@
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal class LowLevelClient : TypeProvider
+    internal class DpgClient : TypeProvider
```
m-nash, line 24: Why the rename here?

src/AutoRest.CSharp/Common/Input/InputNamespaceVisitor.cs
```diff
@@ -0,0 +1,287 @@
...
+        //   1. Sort models by their base models, so that models that have no base models are going first
+        //   2. Iterates over models in the said order and calls `VisitModel` for each of them. Resulted models still reference source models at that point, so all references are invalid
+        //   3. From the "intermediate" models collected on the previous step, create target models with correct `BaseModel` values and empty lists for DerivedModels, CompositionModels and Properties.
+        //      Target model is saved into `typesCache` and added to the returned list.
+        //      If `VisitModel` implementation has returned the same intermediate model for more than one source model, target model will be added only once
+        //   4. After all target models are created, lists for DerivedModels and CompositionModels are filled.
+        //      Also, each property is visited in `VisitModelProperty`, and if InputModelProperty.Type is a model, it must updated it to the target model from `typesCache`.
+        private IReadOnlyList<InputModelType> VisitModels(Dictionary<InputType, InputType> typesCache, IReadOnlyList<InputModelType> source)
```
m-nash, line 52: This seems overly complicated to avoid having a setter and instead restricting who can call that setter.  We should discuss our overall strategy here going forward.

PR #3954, opened 2023-11-20
introduce property declaration

ArcturusZhang:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace AutoRest.CSharp.Output.Models
 {
-    internal record FieldDeclaration(FormattableString? Description, FieldModifiers Modifiers, CSharpType Type, CSharpType ValueType, CodeWriterDeclaration Declaration, FormattableString? DefaultValue, bool IsRequired, SerializationFormat SerializationFormat, bool IsField = false, bool WriteAsProperty = false, bool OptionalViaNullability = false, FieldModifiers? GetterModifiers = null, FieldModifiers? SetterModifiers = null, SourcePropertySerializationMapping? SerializationMapping = null)
+    internal record FieldDeclaration(FormattableString? Description, FieldModifiers Modifiers, CSharpType Type, CSharpType ValueType, CodeWriterDeclaration Declaration, FormattableString? InitializationValue, bool IsRequired, SerializationFormat SerializationFormat, bool IsField = false, bool WriteAsProperty = false, bool OptionalViaNullability = false, FieldModifiers? GetterModifiers = null, FieldModifiers? SetterModifiers = null, SourcePropertySerializationMapping? SerializationMapping = null)
```
ArcturusZhang, line 12: default value is a little bit confusing because here is not a spec of something, it is not really a default value - more like an initialization value for a field which the class use to initialize the corresponding instance.
I changed the name here to align with the name in `ObjectTypeProperty`.

ArcturusZhang: The property `writeAsProperty` is confusing as well.
If the member is a field we should construct an instance of FieldDeclaration, and if the member is a property, we should construct an instance of corresponding to property.
We should really do not have a thing which is "field declaration", but writes as a property.

src/AutoRest.CSharp/Common/Output/Models/PropertyDeclaration.cs
```diff
@@ -0,0 +1,29 @@
...
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
+using AutoRest.CSharp.Generation.Types;
+using AutoRest.CSharp.Generation.Writers;
+using AutoRest.CSharp.Output.Models;
+using Azure.Core;
+
+namespace AutoRest.CSharp.Common.Output.Models
+{
+    [DebuggerDisplay("{GetDebuggerDisplay(),nq}")]
+    internal record PropertyDeclaration(FormattableString? Description, MethodSignatureModifiers Modifiers, CSharpType PropertyType, CodeWriterDeclaration Declaration, PropertyAccessorMethod Get, PropertyAccessorMethod? Set = null, ValueExpression? InitializationValue = null)
```
AlexanderSher, line 15: One design option I would consider here is to add something like `PropertyBody`, which can be `GetOnlyPropertyBody`, `GetSetPropertyBody`, `ConstantBody`, etc.

ArcturusZhang: We still will keep the `PropertyDeclaration` here right? This comment is for the `PropertyAccessorMethod Get, PropertyAccessorMethod? Set = null` part?

I choose this because I am trying to mimic the same way in C# reflection and syntax tree.
And the get and set might have different modifiers, it might be cumbersome to have property body which contains two things.

AlexanderSher: > We still will keep the PropertyDeclaration here right?

Yes

> I choose this because I am trying to mimic the same way in C# reflection and syntax tree.

Reflection represents metadata rather than the way we want to generate the code. Roslyn semantic representation of properties has been created at the times of C# 4.0, and properties have seriously evolved since then, and they have lots of mutually-exclusive conditions that I would prefer to enforce on the level of the type (e.g. initialization value can't co-exist with getter body). What I meant is something like this:

```cs
internal record PropertyDeclaration(FormattableString? Description, MethodSignatureModifiers Modifiers, CSharpType PropertyType, CodeWriterDeclaration Declaration, PropertyBody Body);

internal record PropertyBody();

// Writes `MyProperty { get; set; }` or `MyProperty { get; set; } = InitializationValue;`
internal record TrivialPropertyBody(bool HasSetter, MethodSignatureModifiers SetterModifiers, ConstantExpression? InitializationValue) : PropertyBody;

// Writes `MyProperty { get; }` when GetterBody is null and `MyProperty { get {...} }` when it is present
internal record MethodPropertyBody(MethodBody GetterBody, MethodBody? SetterBody, MethodSignatureModifiers SetterModifiers) : PropertyBody;

// Writes `MyProperty => [[ValueExpression]]
internal record ExpressionPropertyBody(ValueExpression GetterBodyExpression) : PropertyBody;
```

ArcturusZhang: I will update accordingly.

src/AutoRest.CSharp/Common/Output/Expressions/ValueExpressions/TypedValueExpression.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using AutoRest.CSharp.Common.Output.Models;
```
live1206, line 5: can be removed

ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs
```diff
@@ -23,13 +24,24 @@ public FieldDeclaration(FieldModifiers modifiers, CSharpType type, string name,
                   writeAsProperty: writeAsProperty)
         { }
 
-        public FieldDeclaration(FieldModifiers modifiers, CSharpType type, string name, FormattableString? defaultValue, SerializationFormat serializationFormat, bool writeAsProperty = false)
+        public FieldDeclaration(FormattableString description, FieldModifiers modifiers, CSharpType type, string name, ValueExpression? initializationValue = null)
+            : this(Description: description,
```
live1206, line 28: these explicit naming seems redundant, most of them are not optional

src/AutoRest.CSharp/Common/Output/Models/Method.cs
```diff
@@ -10,8 +10,8 @@ namespace AutoRest.CSharp.Common.Output.Models
     internal class Method
     {
         public MethodSignatureBase Signature { get; }
-        public MethodBodyStatement? Body { get; }
-        public ValueExpression? BodyExpression { get; }
+        public MethodBodyStatement? Body { get; protected init; }
```
live1206, line 13: what is this change for? don't see any derived types of `Method`

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -135,7 +134,52 @@ protected internal Resource(OperationSet operationSet, IEnumerable<Operation> op
...
+                FieldModifiers.Public | FieldModifiers.Static | FieldModifiers.ReadOnly,
+                typeof(ResourceType),
+                "ResourceType",
+                Literal(ResourceType.ToString()));
+        }
+
+        private IReadOnlyList<PropertyDeclaration>? _properties;
+        public IReadOnlyList<PropertyDeclaration> Properties => _properties ??= BuildProperties().ToArray();
+
+        protected virtual IEnumerable<PropertyDeclaration> BuildProperties()
```
live1206, line 152: change return type to `IReadOnlyList<T>`? So that, there is no need of `ToArray`

live1206: Never mind, the signature can't be changed.

src/AutoRest.CSharp/Common/Output/Models/FieldDeclaration.cs
ArcturusZhang: well I intentional add them to make those parameters more clear - especially when a method consumes a lot of parameters, and you might have a few arguments with string types. It is very confusing to have somehting like this:
```
DoSomething("a", "b", "c")
```
or
```
DoSomething("a", null, null)
```
it is very likely to go wrong when we add a new required parameter (when we have to) or add a new optional parameter, if it happens to have the same type, the compiler never shows an error for you

src/AutoRest.CSharp/Common/Output/Models/Method.cs
ArcturusZhang: In a previous commit we have. But now I have reverted that. I will change this back

live1206 (approved):

PR #3957, opened 2023-11-20
Create RequestContent with different content type

### LINKED ISSUE #3793: Content-type: Application/octet-stream being base64 encoded
This started in 0.2.0-beta.20230926.2

We have a typespec operation described as:
``` tyepscript
@doc("Performs a single Image Analysis operation")
@actionSeparator("/")
@route("imageanalysis:analyze")
@post
@sharedRoute
@action("analyze")
op analyzeFromStream is Azure.Core.RpcOperation<
  {
    ...SharedAnalyzeQuery,

    @doc("The format of the HTTP payload.")
    @header()
    contentType: "application/octet-stream";

    @doc("The image to be analyzed")
    @body
    imageContent: bytes;
  },
  ImageAnalysisResult
>;
```
Starting in 20230926.2 instead of sending the data in imageContent as raw bytes, it's being base64 encoded

Looks like 
``` C#
RequestContentHelper.FromObject(imageContent),
```
was injected when calling the protocol method when it wasn't needed.

If the content type is a byte stream, don't run it through a json encoder...

pshao25:
Fix #3793


chunyu3 (commented):
We can calculate the mediatype during inputModel json serialization

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
```diff
@@ -45,7 +45,7 @@ private record RequestContentParameterInfo(CodeWriterDeclaration ContentVariable
                 {
                     if (converter.ConvenienceSpread == null)
                     {
-                        var parameter = convenienceParameter.GetConversionFormattable(protocolParameter.Type);
+                        var parameter = convenienceParameter.GetConversionFormattable(protocolParameter.Type, RequestMediaTypes?.FirstOrDefault());
```
chunyu3, line 48: This is common function to get convenience parameter for all kinds parameter, path, query, head and body. I think RequestMediaType is only affected the body parameter. Need to make sure it will break other kind of parameter.

test/CadlRanchProjects/encode/bytes/src/Generated/RequestBody.cs
```diff
@@ -161,7 +161,7 @@ public virtual async Task<Response> OctetStreamAsync(BinaryData value, Cancellat
             Argument.AssertNotNull(value, nameof(value));
 
             RequestContext context = FromCancellationToken(cancellationToken);
-            using RequestContent content = RequestContentHelper.FromObject(value);
+            using RequestContent content = value;
```
chunyu3, line 164: We don't need this line. If can implicitly cast to RequestConent, just pass the parameter in the protocol method call.

pshao25: We can't. We need `using`.

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -95,7 +95,7 @@ public static FormattableString GetIdentifiersFormattable(this IEnumerable<strin
             return conversion == null ? constantFormattable : $"{constantFormattable}{conversion}";
         }
 
-        public static FormattableString GetConversionFormattable(this Parameter parameter, CSharpType toType)
+        public static FormattableString GetConversionFormattable(this Parameter parameter, CSharpType toType, string? contentType)
```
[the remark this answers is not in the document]
pshao25: I think you are saying `bodyMediaType` which has only one value for an operation, but content type we could have several ones.

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
pshao25: Why it would break others? Could you think of a scenario, since all tests are passed and no SDK change.

chunyu3 (approved):

lirenhe (approved):

PR #3966, opened 2023-11-22
only include the public types in discriminator comments

ArcturusZhang:
# Description

We will add a comment to the base model of a discriminated set to state that the user of our libraries should use a derived class, such as:
```
        /// Please note <see cref="TokenFilter"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
        /// The available derived classes include <see cref="AsciiFoldingTokenFilter"/>, <see cref="CjkBigramTokenFilter"/>, <see cref="CommonGramTokenFilter"/> and <see cref="DictionaryDecompounderTokenFilter"/>
```
This is a list our SDK user will read and follow, therefore we should only include public types. In real life, internal types might also be included.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
```diff
@@ -96,7 +96,10 @@ public virtual FormattableString CreateExtraDescriptionWithDiscriminator()
                 List<FormattableString> childrenList = new List<FormattableString>();
                 foreach (var implementation in Discriminator.Implementations)
                 {
-                    childrenList.Add($"{implementation.Type.Implementation.Type:C}");
+                    // when the base type is public and the implementation type is not public, we skip it
+                    if (Type.IsPublic && !implementation.Type.IsPublic)
```
live1206, line 100: If `Type.IsPublic` is false, should it be added to the comment? The current code will do so.

archerzz (approved):
So some existing SDK docs will be changed to delete some internal sub-classes prompts?

ArthurMa1978 (approved):

ArcturusZhang:
> So some existing SDK docs will be changed to delete some internal sub-classes prompts?

The regen contains a few cases of this - all of its derived classes are changed to internal by customization code. I am not sure if it is intentional or not, but I assume, since the xml doc is always public, we should not expose the references of the internal classes in xml doc

src/AutoRest.CSharp/Common/Output/Models/Types/ObjectType.cs
ArcturusZhang: we really do not care if that is the case. The base is not public then the derived classes must not be public otherwise the code will not compile. Therefore in that case, everything is not public, it really does not matter what we put as comments on those, and I prefer having less code changed in our generated libraries.

PR #3969, opened 2023-11-24
Refine type replacing in MPG

ArcturusZhang:
Partially fix https://github.com/Azure/autorest.csharp/issues/3970

# Description

We should never wrap the "replaced types" into our type providers other than `SystemObjectType`.

For instance, if we replaced a model `XXXOperationStatus` with the `OperationStatus` in resourcemanager, in the generator's outputlibrary, `XXXOperationStatus` is modeled in type of `MgmtObjectType`, we should not wrap the replaced `OperationStatus` in resourcemanager any more, instead we should wrap it into `SystemObjectType` because the new model IS a system object type.

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
+++ b/src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
@@ -129,7 +129,7 @@ namespace AutoRest.CSharp.AutoRest.Plugins
                 await AutoRestLogger.Fatal(e.ErrorText);
                 return false;
             }
-            catch (Exception e)
+            catch (Exception)
             {
                 try
                 {
@@ -144,8 +144,7 @@ namespace AutoRest.CSharp.AutoRest.Plugins
                 {
                     // Ignore any errors while trying to output crash information
                 }
-                await AutoRestLogger.Fatal($"Internal error in AutoRest.CSharp{ErrorHelpers.FileIssueText}\nException: {e.Message}\n{e.StackTrace}");
-                return false;
+                throw;
             }
 
             return true;
--- a/src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
+++ b/src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
@@ -11,10 +11,10 @@ using System.Reflection;
 using System.Text;
 using AutoRest.CSharp.Common.Utilities;
 using AutoRest.CSharp.Generation.Types;
+using AutoRest.CSharp.Generation.Writers;
 using AutoRest.CSharp.Input;
 using AutoRest.CSharp.Input.Source;
 using AutoRest.CSharp.Mgmt.Decorator;
-using AutoRest.CSharp.Output.Models.Requests;
 using AutoRest.CSharp.Output.Models.Shared;
 using AutoRest.CSharp.Utilities;
 using Azure.Core;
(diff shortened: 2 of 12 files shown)


test/TestProjects/MgmtReferenceTypes/Generated/Models/ErrorResponse.Serialization.cs
```diff
@@ -22,7 +22,7 @@ void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
             if (Optional.IsDefined(Error))
             {
                 writer.WritePropertyName("error"u8);
-                writer.WriteObjectValue(Error);
+                JsonSerializer.Serialize(writer, Error);
```
ArcturusZhang, line 25: The new code is somehow correct in a way but not entirely correct.

The type of `Error` property here is `Azure.ResponseError`. It does not implement the internal interface `IUtf8JsonSerializable` therefore the `writer.WriteObjectValue` extension method will just throw: https://github.com/Azure/autorest.csharp/blob/91b51d23bbd026b4169c1823080d0688818c192f/src/assets/Azure.Core.Shared/Utf8JsonWriterExtensions.cs#L73
In the meantime, the type `Azure.ResponseError` has a `JsonConverter` attribute on it, therefore `JsonSerializer.Serialize` is the correct method to call here.
On the other hand, why this is not entirely correct? Because the `JsonConverter` only implements the `Read` part, `Write` method of it just throws, therefore the invocation here also just throws.
It is not entirely correct, but it is closer.

ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/SystemObjectType.cs
```diff
@@ -127,17 +130,33 @@ private ObjectTypeConstructor BuildConstructor(ConstructorInfo ctor, ObjectTypeC
         }
 
         protected override ObjectTypeConstructor BuildInitializationConstructor()
-            => BuildConstructor(GetCtor(_type, ReferenceClassFinder.InitializationCtorAttributeName), GetBaseObjectType()?.InitializationConstructor);
+        {
+            if (_backingObjectType is not null)
```
archerzz, line 134: Can we provide a sub-class of `SystemObjectType`, like `ReplacingSystemObjectType`, and override all those methods?

ArcturusZhang: OK this makes sense - but in the meantime I think this whole replacement story needs a thorough redesign and discussion in the consolidation. Wei might take care of this.
I will introduce an derived class of SystemObjectType for this purpose.

ArcturusZhang: A private derived class of  SystemObjectType is introduced for this purpose.

ArcturusZhang: it turns out that we cannot do this.
We cannot have a standalone class to take a `backingObjectType` because in a fairly complicated case, we somehow get into a situation of cyclic invocation which leads to stack overflow.

The reason I am adding this backing object type is because we need the descriptions on those properties and ctors, but it turns out there is no way that we could get those xml docs using reflection. (For instance, [this](https://stackoverflow.com/questions/51666875/get-xml-documentation-comments#:~:text=You%20can%27t%20get%20the%20XML%20documentation%20from%20reflection%2C,you%20can%20open%20and%20read%20from%20your%20application.))
Therefore what I have to do now is that we just use the backing object type's property to get its description if there is any

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -144,8 +144,7 @@ await foreach (var file in project.GetGeneratedFilesAsync())
                 {
                     // Ignore any errors while trying to output crash information
                 }
-                await AutoRestLogger.Fatal($"Internal error in AutoRest.CSharp{ErrorHelpers.FileIssueText}\nException: {e.Message}\n{e.StackTrace}");
-                return false;
+                throw;
```
ArcturusZhang, line 147: previously the exception stack trace somehow gets swallowed completely.
Throwing this out and this exception will be caught again where this is invoked and we will get full stack traces there.

archerzz: Is the context information sufficient? If not, we could `throw new XXXExceptio()` with more details.

archerzz (approved):

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
ArcturusZhang: we actually do not want extra information here. If we wrap it, it will add an additional layer to the stack trace which is quite unnecessary

PR #3971, opened 2023-11-26
remove the model factory if it does not contain any methods

ArcturusZhang:
# Description

# Checklist

# Ready to Land?


archerzz (approved):

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
```diff
@@ -179,8 +179,16 @@ private async Task<Project> RemoveMethodsFromModelFactoryAsync(Project project,
             return project;
 
         var root = await modelFactoryGeneratedDocument.GetSyntaxRootAsync();
-        root = root?.RemoveNodes(nodesToRemove, SyntaxRemoveOptions.KeepNoTrivia);
-        modelFactoryGeneratedDocument = modelFactoryGeneratedDocument.WithSyntaxRoot(root!);
+        Debug.Assert(root is not null);
```
ArthurMa1978, line 182: Under what kind of circumstances would it reach this assert error?

live1206 (approved):

ArthurMa1978 (approved):

test/CadlRanchProjects/authentication/http/custom/src/Generated/AuthenticationHttpCustomModelFactory.cs
```diff
@@ -1,16 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
-// <auto-generated/>
-
-#nullable disable
```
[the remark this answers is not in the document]
ArcturusZhang: Yes. You could see the changes in regen.
We have more cases when we have public serialization because we have more entries in model factory with the feature on.

src/AutoRest.CSharp/Common/Output/PostProcessing/PostProcessor.cs
ArcturusZhang: that depends on how Roslyn behaves, I am not 100 percent sure but when the code in `modelFactoryGeneratedDocument` cannot compile (maybe it needs to be non-compilable in a specific way), it will return   null here.
In most cases with compilable code, it will never be null.

PR #3976, opened 2023-11-27
refine model builder and `SystemObjectType`

ArcturusZhang:
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtTarget.cs
```diff
@@ -270,18 +267,6 @@ private static void WriteExtensionFile(GeneratedCodeWorkspace project, MgmtClien
             AddGeneratedFile(project, $"Extensions/{extensionWriter.FileName}.cs", extensionWriter.ToString());
         }
 
-        private static bool ShouldSkipModelGeneration(TypeProvider model)
```
ArcturusZhang, line 273: we must have this method because previously the replaced models are represented by their respective type provider, for instance, if the model gets replaced and before the replacement, it is a `ResourceData`, we will construct a new ResourceData for it after the replacement.
But now, we always construct a `SystemObjectType` for them, therefore we no longer need a check whether to skip them or not.

live1206 (approved):

src/AutoRest.CSharp/Common/Utilities/AutoRestLogger.cs
```diff
@@ -15,20 +14,16 @@ public static void Initialize(IPluginCommunication autoRest)
...
 
-        public static bool IsInitialized => _autoRest != null;
-
         public static async Task Warning(string message)
         {
-            if (!IsInitialized)
-                throw new InvalidOperationException("AutoRestLogger.Warning is called before initialized");
-            await _autoRest!.Warning(message);
+            if (_autoRest is not null)
+                await _autoRest.Warning(message);
```
archerzz, line 20: So, we changed the behavior from throwing exception to just hide it if not initialized? Why?

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -471,35 +469,8 @@ private bool ShouldGenerateChildrenForType(Type armCoreType)
...
-                    {
-                        var propTemp = mgmtObjectType.Properties;
-                    }
-                }
-            }
-
-            return models;
-        }
+        private IEnumerable<TypeProvider>? _models;
+        public IEnumerable<TypeProvider> Models => _models ??= SchemaMap.Values.Where(m => m is not SystemObjectType);
```
archerzz, line 473: Looks like we delete some useless loop?

archerzz (approved):

src/AutoRest.CSharp/Common/Utilities/AutoRestLogger.cs
ArcturusZhang: We have two different execution procedure:
the `CSharpGen` which consumes the autorest instance here
the `StandaloneRunner` which does not have an autorest instance.

The issue this is trying to fix is that:
our MPG generator almost always follows the autorest path therefore it has an autorest instance to write the logs with one exception:
when we are debugging the test projects in `autorest.csharp` repo in VS. When we are doing this, the generator is actually started in the standalone way, and it does not have an autorest instance, and these exceptions will be thrown which is anonying.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
ArcturusZhang: which loop are you referring to?

src/AutoRest.CSharp/Common/Utilities/AutoRestLogger.cs
ArcturusZhang: we could consider to expand the scope of this logger to provide logging functionality in all paths - we might put that as a future work since this class has the name of `AutoRestLogger`

ArthurMa1978 (approved):

Issue #3958, opened 2023-11-20
[MQ] Change type of the `CSharpType.Arguments` to `IReadOnlyList<CSharpType>`


AlexanderSher:
The content of the `Arguments` property shouldn't be changed after `CSharpType` is instantiated.

PR #4004, opened 2023-12-04
change the type `CSharpType.Arguments` from `T[]` to `IReadOnlyList<T>`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3958

# Description

This PR contains the following changes:
1. Change the type of `CSharpType.Arguments` property from `T[]` to `IReadOnlyList<T>` to ensure this type is immutable.
2. Introduce new ctors `CSharpType(Type type, IReadOnlyList<CSharpType> arguments)` and `CSharpType(Type type, bool isNullable, IReadOnlyList<CSharpType> arguments)`.
3. Remove ctor `CSharpType(Type type, Type? serializeAs)`.
4. Refactor the parts related with union types and literal types.
5. Remove the `TryCast` method on `CSharpType` because we now could use the pattern matching syntax to get similar and more flexible functionality

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -60,21 +65,6 @@ public CSharpType(Type type, bool isNullable, params CSharpType[] arguments)
             IsPublic = type.IsPublic && arguments.All(t => t.IsPublic);
         }
 
-        /// <summary>
-        /// Constructs a CSharpType that represents a union type.
-        /// </summary>
-        /// <param name="type">The type to convert.</param>
-        /// <param name="unionItemTypes">The list of union item types.</param>
-        /// <param name="isNullable">Flag used to determine if a type is nullable.</param>
-        public CSharpType(Type type, CSharpType[] unionItemTypes, bool isNullable) : this(
```
ArcturusZhang, line 69: This ctor is removed.
The reason why it is removed is:
1. this ctor is only for constructing a union type, where the `type` parameter here actually should only accept `BinaryData`, but does not really reflect on anywhere semantically.
2. this ctor is very alike to the other ctor constructing a type from a generic type with generic parameters, the only difference between them is one of them has `isNullable` on 2nd place the other has it on 3rd place, which is very confusing and easy to misuse.

This ctor is replaced by a static method `FromUnion` just like we have `FromLiteral`.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -99,12 +89,12 @@ public CSharpType(TypeProvider implementation, string ns, string name, bool isVa
...
         public bool IsValueType { get; }
         public bool IsEnum { get; }
-        public bool IsLiteral { get; init; }
-        public Constant? Literal { get; init; }
-        public bool IsUnion { get; }
-        public IReadOnlyList<CSharpType> UnionItemTypes { get; } = Array.Empty<CSharpType>();
+        public bool IsLiteral => Literal is not null;
+        public Constant? Literal { get; private init; }
+        public bool IsUnion => UnionItemTypes.Count > 0;
+        public IReadOnlyList<CSharpType> UnionItemTypes { get; private init; } = Array.Empty<CSharpType>();
```
ArcturusZhang, line 95: `IsLiteral` and `Literal` actually stand for the same thing, they should not be able to set separately because it is very likely to misuse.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -99,12 +89,12 @@ public CSharpType(TypeProvider implementation, string ns, string name, bool isVa
...
         public bool IsPublic { get; }
-        public CSharpType[] Arguments { get; } = Array.Empty<CSharpType>();
+        public IReadOnlyList<CSharpType> Arguments { get; } = Array.Empty<CSharpType>();
```
ArcturusZhang, line 97: we change this to IReadOnlyList so that it will not be able to change after this instance is constructed.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -41,13 +40,19 @@ internal class CSharpType
         }
 
         public CSharpType(Type type, params CSharpType[] arguments) : this(type, false, arguments)
+        { }
+
+        public CSharpType(Type type, IReadOnlyList<CSharpType> arguments) : this(type, false, arguments)
+        { }
+
+        public CSharpType(Type type, bool isNullable, params CSharpType[] arguments) : this(type, isNullable, arguments as IReadOnlyList<CSharpType>)
```
ArcturusZhang, line 48: this ` as IReadOnlyList<CSharpType>` here is required because without it this ctor is actually calling itself and leads to a compilation error.

ArcturusZhang:
@AlexanderSher @jorgerangel-msft could you have a review on this PR?

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
archerzz: We still need to keep the constructor on line 43?

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -80,7 +81,7 @@ protected override IEnumerable<ObjectTypeProperty> BuildProperties()
 
         private static bool IsSinglePropertyObject(ObjectTypeProperty property)
         {
-            if (!property.Declaration.Type.TryCast<ObjectType>(out var objType))
+            if (property.Declaration.Type is not { IsFrameworkType: false, Implementation: ObjectType objType })
```
archerzz, line 84: There are a lot of check statement like `is { IsFrameowkrType: false, Implementation: {type}}`. Can we provide a `IsCustomType{type}` method to replace `TryCast`?

ArcturusZhang: that is the TryCast with a new name. The purpose of this change is to remove it because we are using two different approaches of getting it, one is `TryCast` the other way is this is pattern matching. And currently `is` way has a overwhelming number therefore I am removing the method to get consistency.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -41,13 +40,19 @@ internal class CSharpType
         }
 
         public CSharpType(Type type, params CSharpType[] arguments) : this(type, false, arguments)
```
live1206, line 42: `arguments` should be cast to `IReadOnlyList`

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -274,7 +274,7 @@ private ReturnTypeChain BuildReturnTypes()
                     throw new InvalidOperationException($"Method {Operation.Name} has to have a return value");
                 }
 
-                if (responseType.TryCast<ModelTypeProvider>(out var modelType))
+                if (responseType is { IsFrameworkType: false, Implementation: ModelTypeProvider modelType })
```
live1206, line 277: Is `IsFrameworkType: false` needed? If `Implementation` is set, should `IsFrameworkType` always be false?

ArcturusZhang: Implementation will throw exceptions when IsFrameworkType is false.
C# compiler compiles this into a sequence of if statements, therefore if we do not have IsFrameworkType: false, it throws exceptions.

ArcturusZhang: Something like this:
image

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
ArcturusZhang: That `params` ctor is used in a lot of places, especially when there is only one generic argument, such as:
```
new CSharpType(typeof(IList<>), typeof(string))
```
if we remove this, we have to use it like this:
```
new CSharpType(typeof(IList<>), new CSharpType[] {typeof(string)})
```
which is cumbersome and inconvenient.

I do not like the solution of adding a ctor for the case we have one generic parameter, which is another confusion with the other ctor `CSharpType(Type type, Type? serializeAs)`

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -33,21 +32,20 @@ internal class CSharpType
         {
         }
 
-        public CSharpType(Type type, Type? serializeAs) : this(
-            type.IsGenericType ? type.GetGenericTypeDefinition() : type,
-            type.IsGenericType ? type.GetGenericArguments().Select(p => new CSharpType(p)).ToArray() : Array.Empty<CSharpType>())
-        {
-            SerializeAs = serializeAs;
-        }
```
ArcturusZhang, line 41: I removed this ctor because 
1. It is rarely used.
2. It is a pit hole when you want to construct a generic type, such as:
```csharp
new CSharpType(typeof(IList<>), typeof(string))
```
you are actually invoking this ctor, instead of the one with generic arguments.

I removed this ctor and add an `init` accessor on `SerializeAs` property, therefore if you want to construct a csharptype with SerializeAs property, you just need to do this:
```
new CSharpType(typeof(YourType))
{
    SerializeAs = typeof(string)
};
```

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
live1206: Right. And why do we throw while accessing `Implemenation` instead of making it nullable?

ArcturusZhang: This CSharpType is a wrap and a union of `System.Type` and `TypeProvider`
It might benefit and make things easier to debug when things go wrong by throwing an exception instead of making it nullable - because they are so different.
And making it nullable leads to a result of we are using `!` everywhere or we have to check if it is not null, therefore it really does not validate anything.
In the situation now, if we are 100% sure, this cannot be a framework type, we could just use `type.Implementation` and do not need to add the nullable forgiving operator

live1206: OK. This then ends up with checking `IsFrameworkType: false` everywhere we want to check the type of `Implementation`.

live1206: And if we use `!` to bypass compiler check, it still throws at run time if it is null, which is basically the same as the throwing now.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -237,25 +222,28 @@ internal static CSharpType FromLiteral(CSharpType type, object literalValue)
...
+        /// <returns></returns>
+        public static CSharpType FromUnion(IReadOnlyList<CSharpType> unionItemTypes, bool isNullable)
         {
-            provider = null;
-            if (this.IsFrameworkType)
-                return false;
-
-            provider = this.Implementation as T;
-            return provider != null;
+            return new CSharpType(typeof(BinaryData), isNullable)
```
jorgerangel-msft, line 240: 👍

jorgerangel-msft:
> @AlexanderSher @jorgerangel-msft could you have a review? This is ready to review now.

The changes related to union types and literals look fine to me.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
ArcturusZhang: Yeah both ways have their shortage. But it is introduced a long time ago therefore it is not in the scope of changing it.
This is similar to the pattern when you have a union in typescript - you will have to assert it is the type before you could use any function or property on it (as far as I could remember)

live1206 (approved):

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
archerzz: If it's a common pattern, I'd prefer a utility method with better name, which provides better readability.

ArcturusZhang: well then I could revert this change.
I remove this for alignment because there are two different usage: call TryCast or use `is`
`is` is more widely used, therefore I would like to remove this.

lirenhe (approved):

Issue #4016, opened 2023-12-08
[defect]Failed to deserialize tspCodeModel.json for recursive inheritance model


chunyu3:
Codegen failed to deserialize recursive inheritance model.

e.g.
```
@doc("extension")
model Extension extends Element {
  level: int8;
}

@doc("element")
model Element {
  extension?: Extension[];
}

@route("/recursiveExtension")
@doc("doSomething for glossary")
@put
@convenientAPI(true)
op recursiveExtension(@body input: Extension): void;
```

Issue #3974, opened 2023-11-27
[defect] infinite loop call in FindMatchesRecursively


chunyu3:
When create anonymous model name, the infinite loop call occur in method `FindMatchesRecursively` https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs#L216

Sample tsp:
[TempTypeSpecFiles.zip](https://github.com/Azure/autorest.csharp/files/13470426/TempTypeSpecFiles.zip)

asaflevi-ms:
@chunyu3 
Do you have an ETA for the fix?
We are blocked and our public preview is planned for next month.

chunyu3:
hello @asaflevi-ms We are investigating this. Have not found the root cause yet.

asaflevi-ms:
CC: @catalinaperalta

catalinaperalta:
Hi @chunyu3, any updates on this issue? Could you link me to the line in the specification that is causing the issue please?

chunyu3:
> Could you link me to the model in the specification that is causing the issue please

Hello @catalinaperalta I have found the root cause and resolve PR (https://github.com/Azure/autorest.csharp/pull/4019) is ready and will be merged soon, then a new autorest.csharp will be released to include the fix.

PR #4019, opened 2023-12-08
Handle Recursive inheritance and model circle reference

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/4016
Fix https://github.com/Azure/autorest.csharp/issues/3974
Fix https://github.com/Azure/autorest.csharp/issues/4029

- remove duplicated model schema in tspCodeModel.json
- Add visistedModel to skip visisted model in FindMatchesRecursively() to avoid infinite loop when there is a circle reference in the model.

# Checklist

# Ready to Land?


pshao25:
Do we have tests?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecReferenceHandler.cs
```diff
@@ -24,9 +24,9 @@ private class TypeSpecReferenceResolver : ReferenceResolver
 
             public override void AddReference(string referenceId, object value)
             {
-                if (!_referenceIdToObjectMap.TryAdd(referenceId, value))
+                if (!_referenceIdToObjectMap.ContainsKey(referenceId))
                 {
-                    throw new JsonException();
+                    _referenceIdToObjectMap.Add(referenceId, value);
```
ArcturusZhang, line 29: this scenario "adding a reference when the same Id existed" should never happen, therefore I believe we should revert this change.
We should make sure that definition of a model should only happen once.

chunyu3: No. we cannot make the model.derievedModels only once otherwise we will meet circle loop issue of the json deserialization issue.

chunyu3:
> Do we have tests?

hello @pshao25 we will have test in latest cadl-ranch. After upgraded cadl-ranch, I will onboard the test.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecReferenceHandler.cs
ArcturusZhang: Let us add a TODO here to emphasize this is a workaround.

The solution of this issue should be in the emitter to ensure it will not output the derived type twice and we should NEVER get a duplicated key here

ArcturusZhang (approved):
Approved if this is only a workaround and there is follow up to fix the real issue.

ArcturusZhang:
> > Do we have tests?
> 
> hello @pshao25 we will have test in latest cadl-ranch. After upgraded cadl-ranch, I will onboard the test.

I suggest we onboard the test case now, and when the cadl ranch case is onboarded we could have another PR to remove it.

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -6,9 +6,14 @@
...
+    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, bool IsNullable)
         : InputType(Name, IsNullable)
     {
+        public InputModelType(string name, string? ns, string? accessibility, string? deprecated, string? description, InputModelTypeUsage usage, IReadOnlyList<InputModelProperty> properties, InputModelType? baseModel, IReadOnlyList<InputModelType> derivedModels, string? discriminatorValue, string? discriminatorPropertyName, bool isNullable)
+            : this(name, ns, accessibility, deprecated, description, usage, properties, derivedModels, discriminatorValue, discriminatorPropertyName, isNullable)
+        {
+            BaseModel = baseModel;
+        }
```
ArcturusZhang, line 16: these changes are unacceptable.
According to our discussion yesterday, these classes (this one and the corresponding converter/deserializer) do not need to be changed, it is already possible to handle the case we are currently having.
The only issue is in the emitter.
If we would like to introduce a temporary workaround, I prefer the previous way which is simpler and straightforward to revert.

test/TestProjects/FirstTest-TypeSpec/src/Generated/tspCodeModel.json
```diff
@@ -748,19 +748,146 @@
...
+      "IsNullable": false
+     },
+     "IsRequired": false,
+     "IsReadOnly": false
+    }
+   ]
+  },
+  {
+   "$ref": "105"
+  },
```
ArcturusZhang, line 801: this is exactly the result we would like to see in the emitter, and in our yesterday's discussion, it does not require a change in the converter to work properly.

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
ArcturusZhang: And I have a try on my machine, by checking out to your branch, revert all the changes in CodeModelConverter.cs, InputModelType.cs, and TypeSpecInputModelTypeConverter.cs, then regenerate FirstTest-TypeSpec, it works fine and generated code does not have a change.
Is there anything I am missing that we really require these changes?

ArcturusZhang: I put the changes in this PR: https://github.com/Azure/autorest.csharp/pull/4038

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -3,12 +3,19 @@
...
```
ArcturusZhang, line 18: We do not have to change the first line to add a private setter to BaseModel property, we could do something like this:
image
this might reduce the amount of changes in this PR

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -19,6 +26,13 @@ internal record InputModelType(string Name, string? Namespace, string? Accessibi
         public bool IsPropertyBag { get; init; } = false;
 
         public bool IsAnonymousModel { get; init; } = false;
+        public InputModelType? BaseModel { get; private set; }
+        internal void SetBaseModel(InputModelType? baseModel, [CallerFilePath] string filepath = "", [CallerMemberName] string caller = "")
+        {
+            Debug.Assert(filepath.EndsWith($"{nameof(TypeSpecInputModelTypeConverter)}.cs"));
```
ArcturusZhang, line 32: Could we add a message here as well?

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -19,6 +26,13 @@ internal record InputModelType(string Name, string? Namespace, string? Accessibi
         public bool IsPropertyBag { get; init; } = false;
 
         public bool IsAnonymousModel { get; init; } = false;
+        public InputModelType? BaseModel { get; private set; }
+        internal void SetBaseModel(InputModelType? baseModel, [CallerFilePath] string filepath = "", [CallerMemberName] string caller = "")
```
ArcturusZhang, line 30: Could we add some xml document here to explain the reasoning of doing this?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -51,17 +51,27 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
...
-                    || reader.TryReadWithConverter(nameof(InputModelType.BaseModel), options, ref baseModel);
+                    || reader.TryReadString(nameof(InputModelType.DiscriminatorValue), ref discriminatorValue);
 
                 if (isKnownProperty)
                 {
                     continue;
                 }
-
-                if (reader.GetString() == nameof(InputModelType.Properties))
+                if (reader.GetString() == nameof(InputModelType.BaseModel))
```
ArcturusZhang, line 60: we should add some comments here to emphasize the last two properties in the JSON must be `BaseModel` and `Properties` and must in this order

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
chunyu3: added.

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
chunyu3: added.

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
chunyu3: added.

lirenhe (approved):

PR #4026, opened 2023-12-08
Mitigate breaking change of optional parameter change for ModelFactory

live1206:
Fixes https://github.com/Azure/autorest.csharp/issues/3596.

This PR introduces `SignatureType`, it holds three portions of code:
- current
- custom
- baseline contract

It provides 
- OverloadingMethods: for now represents the missing methods in baseline contract due to additional optional parameters in the new contract, this is causing breaking changes

This feature is only enabled for Mgmt.
- For DPG, there is no GA version yet, `ApiCompatVersion` is not set in .csproj
- For HLC, there are quite some hand-written/custom code, which is not supported by `TypeFactory.TryCreateType` since `CsharpType` does not represent them properly for now.


src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -54,6 +56,40 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(CodeModel codeModel)
             return project;
         }
 
+        private async Task<CSharpCompilation?> LoadBaselineContract()
+        {
+            if (!Configuration.AzureArm)
```
ArcturusZhang, line 61: let us add a TODO item here since this is temporary.

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -54,6 +56,40 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(CodeModel codeModel)
             return project;
         }
 
+        private async Task<CSharpCompilation?> LoadBaselineContract()
```
ArcturusZhang, line 59: I think maybe we could put this method inside the `SourceInputModel` class so that we do not have to manually enable this for all targets (for instance we have to add a method call on those consuming tsp)

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -253,6 +253,18 @@ public static GeneratedCodeWorkspace CreateExistingCodeProject(string outputDire
             return new GeneratedCodeWorkspace(project);
         }
 
+        public static GeneratedCodeWorkspace CreatePreviousContractFromDll(string xmlDocumentationpath, string dllPath)
```
ArcturusZhang, line 256: It does not really make sense that we return `GeneratedCodeWorkspace` here, because we are reading the contract which is not generated (now) at all.
Considering the usage of this method in other places is always followed by `GetCompilationAsync`, how about we change this method to return a `Compilation` or a `CSharpCompilation` if you need something specified to this derived type?

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -432,7 +433,10 @@ public bool TryCreateType(ITypeSymbol symbol, Func<System.Type, bool> validator,
 
             if (existingType != null && validator(existingType))
             {
-                var arguments = namedTypeSymbol.TypeArguments.Select(a => CreateType(a)).ToArray();
+                if (!TryPopulateArguments(namedTypeSymbol.TypeArguments, out var arguments))
```
ArcturusZhang, line 436: I believe the `validator` delegate should also pass through here.

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
```diff
@@ -3,22 +3,35 @@
...
 namespace AutoRest.CSharp.Output.Models
 {
     internal record MethodSignature(string Name, FormattableString? Summary, FormattableString? Description, MethodSignatureModifiers Modifiers, CSharpType? ReturnType, FormattableString? ReturnDescription, IReadOnlyList<Parameter> Parameters, IReadOnlyList<CSharpAttribute>? Attributes = null, IReadOnlyList<CSharpType>? GenericArguments = null, IReadOnlyDictionary<CSharpType, FormattableString>? GenericParameterConstraints = null, CSharpType? ExplicitInterface = null, string? NonDocumentComment = null)
         : MethodSignatureBase(Name, Summary, Description, NonDocumentComment, Modifiers, Parameters, Attributes ?? Array.Empty<CSharpAttribute>())
     {
+        public static IEqualityComparer<MethodSignature> ParameterAndReturnTypeEqualityComparer = new MethodSignatureParameterAndReturnTypeEqualityComparer();
+
         public MethodSignature WithAsync(bool isAsync) => isAsync ? MakeAsync() : MakeSync();
 
+        public MethodSignature DisableOptionalParameters()
```
ArcturusZhang, line 26: I think `WithAllParametersRequired` is a better name here

src/AutoRest.CSharp/Common/Output/Models/OverloadMethodSignature.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public OverloadMethodSignature(MethodSignature methodSignature, MethodSignature previousMethodSignature, IReadOnlyList<Parameter> missingParameters, FormattableString? description, bool isHiddenFromUser)
+        {
+            MethodSignature = methodSignature;
+            PreviousMethodSignature = previousMethodSignature;
+            MissingParameters = missingParameters;
+            Description = description;
+            IsHiddenFromUser = isHiddenFromUser;
+        }
+    }
```
ArcturusZhang, line 26: We could simply define a record for this

ArcturusZhang: and it would be better if we could have a `[DebuggerDisplay(...)]` attribute to customize the debugging output to show some useful information

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -29,26 +31,39 @@ internal sealed class ModelFactoryTypeProvider : TypeProvider
...
-        public IEnumerable<SerializableObjectType> Models { get; }
+        // This method should only be called from OutputMethods as intermediate state.
+        private IReadOnlyList<Method> ShouldNotBeUsedForOutput([CallerMemberName] string caller = "")
+        {
+            Debug.Assert(caller == nameof(OutputMethods) || caller == nameof(SignatureType), $"This method should not be used for output. Caller: {caller}");
+            return _methods ??= _models.Select(CreateMethod).ToList();
+        }
+
+        private IReadOnlyList<Method>? _outputMethods;
+        public IReadOnlyList<Method> OutputMethods
```
ArcturusZhang, line 45: Could we just use `Methods` as the name here? The `Methods` property on `TypeProvider` just means the methods a writer should write.
In the future we will have others such like `Properties`, `Constructors`, etc...

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
live1206: Added a comment, not a TODO, think we are far away from enabling it for HLC.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
live1206: Sounds good, we can just return `CSharpCompilation`.

src/AutoRest.CSharp/Common/Output/Models/MethodSignature.cs
live1206: sure

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: They are not the same thing.
`Methods` represents the methods generated at this run, it might contain duplicated methods from custom code.
`outputMethods` represents the final methods consumed by the writer, which filter out the existing methods from custom code.
There is a TODO to remove this intermediate result once we move this logic before constructing output types.

src/AutoRest.CSharp/Common/Output/Models/OverloadMethodSignature.cs
live1206: This can be a record, but I am not sure what should be set for `DebuggerDisplay`, it contains complex information of current generated method and previous GA contract method.

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
live1206: Updated

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: Not following here.

live1206: See what you mean now, passed along the validator

src/AutoRest.CSharp/Common/Output/Models/OverloadMethodSignature.cs
ArcturusZhang: Yeah it is a better to have.
If it is a class, when debugging, the debugger only shows the type name of this instance, which literally contains no information at all.
If we change this to a record, it will have a default `ToString` implementation to contain all its properties, which is better when debugging but it is usually too much.
Therefore I suggest we add a DebuggerDisplay to only show those information we find useful to show.
For instance. we could have:
OverloadMethodSignature({MethodName}, Missing: {MissingParameters})

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
ArcturusZhang: I am saying the same thing - we could keep the most frequently used property to a simpler name (such as `Methods`), and for the "methods generated at this run", we could name it as `GeneratedMethods`. The writer should not use this.

Ultimately, our writer should looks like this:
```
public void Write()
{
foreach (var method in Methods)
{
// write the method
}

foreach (var property in Property)
{
// write the property
}
}
```
It does not change the concept if we name them as `OutputMethods` and `OutputProperties`, just make it simpler.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
```diff
@@ -253,6 +253,18 @@ public static GeneratedCodeWorkspace CreateExistingCodeProject(string outputDire
...
+        public static async Task<CSharpCompilation> CreatePreviousContractFromDll(string xmlDocumentationpath, string dllPath)
+        {
+            var workspace = new AdhocWorkspace();
+            Project project = workspace.AddProject("PreviousContract", LanguageNames.CSharp);
+            project = project
+                .AddMetadataReferences(AssemblyMetadataReferences)
+                .WithCompilationOptions(new CSharpCompilationOptions(
+                    OutputKind.DynamicallyLinkedLibrary, nullableContextOptions: NullableContextOptions.Disable));
+            project = project.AddMetadataReference(MetadataReference.CreateFromFile(dllPath, documentation: XmlDocumentationProvider.CreateFromFile(xmlDocumentationpath)));
+            return await (new GeneratedCodeWorkspace(project)).GetCompilationAsync();
```
ArcturusZhang, line 265: we do not need to call `new GeneratedCodeWorkspace` here, actually the `GetCompilationAsync` just calls the `GetCompilationAsync` method on the project, therefore we could just do:
```suggestion
            return await project.GetCompilationAsync();
```
And do we really need properties only on `CSharpCompilation` instead of `Compilation`? Usually `Compilation` just works.

src/AutoRest.CSharp/Common/Input/Source/SourceInputModel.cs
```diff
@@ -127,5 +137,40 @@ private bool TryGetName(ISymbol symbol, [NotNullWhen(true)] out string? name)
 
             return name != null;
         }
+
+        private async Task<CSharpCompilation?> LoadBaselineContract()
```
ArcturusZhang, line 141: since the type that is accepting the return value of this method has the type of `Compilation`, we do not really need to make the return type `CSharpCompilation` 
(ref https://github.com/Azure/autorest.csharp/pull/4026/files#r1423306866)

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
live1206: Right

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: To clarify this further, since `SignatureType` is used recursively for custom, previous GA contract and current generated codes.
So, `SignatureType.Methods` represents methods of all three above, we can't just rename it to `GeneratedMethods`.
And there is a TODO to refine this once we can have the proper set in place before we construct, adding the TODO to `OutputMethods` as well.

live1206: Never mind, think your intension is just to have Methods as output for the writer. 
Since the intermediate methods is private, we can just rename `OutputMethods` to `Methods`.

src/AutoRest.CSharp/Common/Output/Models/OverloadMethodSignature.cs
live1206: Don't think it's straightforward to put information to `DebuggerDisplay`, we would still need to look into the object since `MissingParameters` is a list of objects.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -414,61 +415,119 @@ public bool TryCreateType(ITypeSymbol symbol, [NotNullWhen(true)] out CSharpType
...
+            return symbol switch
             {
-                throw new InvalidCastException($"Unexpected type {symbol}");
-            }
+                IArrayTypeSymbol arrayTypeSymbol => TryCreateTypeForIArrayTypeSymbol(arrayTypeSymbol, validator, out type),
+                INamedTypeSymbol namedTypeSymbol => TryCreateTypeForINamedTypeSymbol(namedTypeSymbol, validator, out type),
+
+                // We can only handle IArrayTypeSymbol of framework type and INamedTypeSymbol for now since CSharpType can't represent other types such as IArrayTypeSymbol of user types
+                // Instead of throwing an exception, wihch causes more side effects, we just return false and let the caller handle it.
+                _ => false
```
ArcturusZhang, line 425: maybe we could throw here as a emphasizer for us to update the code to handle more cases because returning false here for an unhandled case might be hiding issues.
But I could go either way here

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -88,6 +114,20 @@ private static string GetDefaultNamespace()
 
         public HashSet<MethodInfo> ExistingModelFactoryMethods { get; }
 
+        private SignatureType? _signatureType;
+        public override SignatureType? SignatureType
+        {
+            get
+            {
+                // This can only be used for Mgmt now, because there are custom/hand-written code in HLC can't be loaded into CsharpType such as generic methods
+                if (!Configuration.AzureArm)
```
ArcturusZhang, line 123: since we potentially need to support DPG, we may need a TODO item here, and the check could be
`if (!Configuration.Generate1ConvenientClient)` which means "not HLC"

test/AutoRest.TestServer.Tests/Mgmt/TestProjects/TestProjectTests.cs
```diff
@@ -593,6 +593,8 @@ public IEnumerable<Type> FindAllCollections()
             }
         }
 
+        protected Type FindModelFactory() => MyTypes().Single(IsModelFactory);
```
ArcturusZhang, line 596: I believe we need to make this nullable and use `SingleOrDefault` here because there are occasions that a project does not have a model factory (for example, for a very small project, and all models inside it happens to meet the condition of omitting in model factory)

live1206: OK then

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
ArcturusZhang: did we get this resolved?

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -261,5 +261,52 @@ public static CSharpType FromUnion(IReadOnlyList<CSharpType> unionItemTypes, boo
...
+
+        public bool TryCast<T>([MaybeNullWhen(false)] out T provider) where T : TypeProvider
+        {
+            provider = null;
+            if (this.IsFrameworkType)
+                return false;
+
+            provider = this.Implementation as T;
+            return provider != null;
+        }
```
ArcturusZhang, line 273: well last time I removed this method because we could always use a is expression to get the same result:
```
if (type is {IsFrameworkType: false, Implementation: ObjectType objectType})
```
which is the same as
```
if (type.TryCast<ObjectType>(out var objectType))
```

I do not have objections on introducing this back personally

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -752,5 +753,34 @@ static void WritePropertyAccessorModifiers(CodeWriter writer, MethodSignatureMod
                     .AppendRawIf("private ", modifiers.HasFlag(MethodSignatureModifiers.Private));
             }
         }
+
+        public static void WriteOverloadMethod(this CodeWriter writer, OverloadMethodSignature overloadMethod)
```
ArcturusZhang, line 757: I would like to suggest we do not need to introduce this method.
Instead, we should build a proper `Method` instance (with a body), and call `writer.WriteMethod(overloadMethod)` where overloadMethod has a type of `Method`
All the calculations could happen when we build the `Method` instance.

src/AutoRest.CSharp/Common/Generation/Writers/ModelFactoryWriter.cs
```diff
@@ -32,6 +33,15 @@ public void Write()
                         _writer.WriteMethodDocumentation(method.Signature);
                         _writer.WriteMethod(method);
                     }
+
+                    if (This.SignatureType is not null)
+                    {
+                        foreach (OverloadMethodSignature overloadMethod in This.SignatureType.OverloadMethods)
```
ArcturusZhang, line 39: As I said in this comment: https://github.com/Azure/autorest.csharp/pull/4026/files#r1446822634
I believe we should change the type of `OverloadMethods` to `IReadOnlyList<Method>` as its name suggests, "Methods" instead of signatures.

ArcturusZhang: In principal, we always want the writers to be stupid, which just writes everything we give to it. Therefore it would be better if the `Methods` on `ModelFactoryTypeProvider` could contain all the method we would like the writer to write, including the generated methods, and the overload methods (with its body included, the writer should not calculate the body, determining which parameter to pass along which parameter to pass a null).

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
live1206: This is not in use, introduced during the merge, I will remove it.

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
live1206: forgot to change the return type, will update it.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
live1206: removed

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
live1206: updated

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
live1206: updated

src/AutoRest.CSharp/Common/Generation/Writers/ModelFactoryWriter.cs
live1206: updated

src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs
ArcturusZhang: well my point is that we do not need to construct a `GeneratedCodeWorkspace`.
Just `await project.GetCompilationAsync()` could work as expected.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,318 @@
...
+                        parameters.Add(methodParameter);
+                    }
+                }
+
+                // Since we don't have the ability to create the missing types, if any parameter type is missing we can't continue to generate overload methods
+                if (isParameterTypeMissing)
+                {
+                    continue;
+                }
+                result.Add(new MethodSignature(method.Name, null, $"{description}", MapModifiers(method), returnType, null, parameters, IsRawSummayText: true));
```
ArcturusZhang, line 259: here is a typo: 
```suggestion
                result.Add(new MethodSignature(method.Name, null, $"{description}", MapModifiers(method), returnType, null, parameters, IsRawSummaryText: true));
```

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -150,6 +151,12 @@ public static IDisposable WriteMethodDeclaration(this CodeWriter writer, MethodS
 
         private static IDisposable WriteMethodDeclarationNoScope(this CodeWriter writer, MethodSignatureBase methodBase, params string[] disabledWarnings)
         {
+            // CSharpAtribute does not accept non-string arguments
+            if (methodBase.IsHiddenFromUser)
+            {
+                writer.Line($"[{typeof(EditorBrowsableAttribute)}({typeof(EditorBrowsableState)}.{nameof(EditorBrowsableState.Never)})]");
+            }
```
ArcturusZhang, line 158: OK if this is the reason that we add this flag, we need to change the definition of `CSharpAttribute` to make sure it could accept anything as parameters.
I believe `ValueExpression` should be a good idea. We do not really have expression when we introduce the attribute stuff

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -274,6 +281,11 @@ private static IDisposable WriteMethodDeclarationNoScope(this CodeWriter writer,
 
         public static CodeWriter WriteMethodDocumentation(this CodeWriter writer, MethodSignatureBase methodBase)
         {
+            if (methodBase.IsRawSummaryText)
+            {
+                return writer.WriteRawXmlDocumentation(methodBase.Description);
+            }
```
ArcturusZhang, line 287: OK this is how it is used.
But I do not really like we put a flag here.
What would it happen if we just use the same method as we write ordinary descriptions?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -88,6 +115,21 @@ private static string GetDefaultNamespace()
...
+        {
+            get
+            {
+                // This can only be used for Mgmt now, because there are custom/hand-written code in HLC can't be loaded into CsharpType such as generic methods
+                // TODO: enable this for DPG, and check Configuration.Generate1ConvenientClient to disable it for HLC
+                if (!Configuration.AzureArm)
+                {
+                    return null;
+                }
+                return _signatureType ??= new SignatureType(_typeFactory, ShouldNotBeUsedForOutput().Select(x => (MethodSignature)x.Signature).ToList(), _sourceInputModel, DefaultNamespace, DefaultName);
```
ArcturusZhang, line 129: we should not really use `DefaultNamespace` and `DefaultName` here.
DefaultNamespace and DefaultName are kind of like "requested" namespace and name.
The generator might not generate those namespaces or names, for instance, when there is an existing type with `CodeGenType`, and it has a different namespace and name. When this happens, the generator will just write the namespace and name of the existing type ignoring the default namespace and default name.

Here I believe we should use `Declaration.Namespace` and `Declaration.Name`.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,318 @@
...
+        private readonly string _defaultName;
+        private readonly SignatureType? _customization;
+        private readonly SignatureType? _baselineContract;
+        private readonly MethodChangeset? _methodChangeset;
+
+        // Missing means the method with the same name is missing from the current contract
+        // Updated means the method with the same name is updated in the current contract, and the list contains the previous method and current methods including overload ones
+        private record MethodChangeset(IReadOnlyList<MethodSignature> Missing, IReadOnlyList<(List<MethodSignature> Current, MethodSignature Previous)> Updated) { }
+
+        public SignatureType(TypeFactory typeFactory, IReadOnlyList<MethodSignature> methods, SourceInputModel? sourceInputModel, string defaultNamespace, string defaultName)
```
ArcturusZhang, line 40: I believe we should remove the word default here (see the comment in src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs)

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,318 @@
...
+    ///     - current
+    ///     - custom
+    ///     - baseline contract
+    ///     current union custom compare with baseline contract outputs the changeset, we can apply different rules with it.
+    /// </summary>
+    internal class SignatureType
+    {
+        private readonly TypeFactory _typeFactory;
+        private readonly string _defaultNamespace;
+        private readonly string _defaultName;
```
ArcturusZhang, line 31: we should simply have:
```suggestion
        private readonly string _namespace;
        private readonly string _name;
```

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        private IReadOnlyList<Method>? _overloadMethods;
+        public IReadOnlyList<Method> OverloadMethods => _overloadMethods ??= EnsureOverloadMethods();
+
+        private IReadOnlyList<Method> EnsureOverloadMethods()
+        {
+            var overloadMethods = new List<Method>();
+            var updated = _methodChangeset?.Updated;
+            if (updated is null)
```
ArcturusZhang, line 67: nit: we could have:
```suggestion
            if (_methodChangeset is not {Updated: {} updated})
```
or
```suggestion
            if (_methodChangeset?.Updated is not {} updated)
```

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,318 @@
...
+
+        private IReadOnlyList<ValueExpression> BuildOverloadMethodParameters(OverloadMethodSignature overloadMethodSignature)
+        {
+            var parameters = new List<ValueExpression>();
+            var set = overloadMethodSignature.MissingParameters.ToHashSet(Parameter.TypeAndNameEqualityComparer);
+            foreach (var parameter in overloadMethodSignature.MethodSignature.Parameters)
+            {
+                if (set.Contains(parameter))
+                {
+                    parameters.Add(new PositionalParameterReference(parameter.Name, Default.CastTo(parameter.Type)));
```
ArcturusZhang, line 94: We usually prefer null if the type could accept null instead of using default everywhere.
I think we could have a `GetDefaultExpression` method on `CSharpType` maybe - something like this:
```csharp
public partial class CSharpType
{
	public ValueExpression GetDefaultExpression()
	{
		if (IsValueType && !IsNullable)
			return Snippets.Default.CastTo(this);

		return Snippets.Null.CastTo(this);
	}
}
```
or introduce a method in `Snippets` so that we could have a null or default with type
We could list this as a follow up refactor because in the sample generation part we have a lot of if else to get default or null with types.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,318 @@
...
+                // We can't use CsharpType.Equals here because they could have different implementations from different versions
+                if (previousMethod.ReturnType is null && item.ReturnType is not null)
+                {
+                    continue;
+                }
+                if (previousMethod.ReturnType is not null && !previousMethod.ReturnType.EqualsByName(item.ReturnType))
+                {
+                    continue;
+                }
+
```
ArcturusZhang, line 136: could we add some descriptions to these branches?

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,318 @@
...
+                        missing.Add(item);
+                    }
+                }
+            }
+            return new(missing, updated);
+        }
+
+        public IReadOnlyList<MethodSignature> Methods { get; }
+
+        private IReadOnlyList<MethodSignature> PopulateMethodsFromCompilation(Compilation? compilation)
```
ArcturusZhang, line 203: I believe this method should be static since it is using in the ctor only.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
```diff
@@ -0,0 +1,318 @@
...
+            {
+                modifiers |= MethodSignatureModifiers.Virtual;
+            }
+            if (methodSymbol.IsOverride)
+            {
+                modifiers |= MethodSignatureModifiers.Override;
+            }
+            return modifiers;
+        }
+    }
```
ArcturusZhang, line 317: we have something similar here: https://github.com/Azure/autorest.csharp/blob/6f0917d17343ee5f12d60636f14e74d26783f44e/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs#L311
maybe we could find somewhere like `BuilderHelpers` to have a generic method for this purpose and just use it here.

ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: The existing one only get accessibility, what about the others?

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: This can't be static, because `PopulateMethods` method within is using instance field `_typeFactory`, it can't be static.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
live1206: got it, thanks, updated.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: updated

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
live1206: Will create a new PR for this change.

live1206: The main change is tied to this PR, updated here but get more change.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: I think these code branches should explain themselves, otherwise we should have a better name for the methods.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
live1206: The xml documentation from base contract contains both summary description and parameter list.
Another option is to extract summary description only and use it with the original `WriteXmlDocumentationSummary`.

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: Would like to postpone this refactoring later when we have this `GetDefaultExpression` in place.

ArcturusZhang: yeah we could do that later

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
ArcturusZhang: I am OK with this right now, but I believe later we need a clean up on this because this flag does not make much sense.

src/AutoRest.CSharp/Common/Output/Models/Shared/CSharpAttribute.cs
```diff
@@ -1,9 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
 using AutoRest.CSharp.Generation.Types;
 
 namespace AutoRest.CSharp.Output.Models.Shared
 {
-    internal record CSharpAttribute(CSharpType Type, params string[] Arguments);
+    internal record CSharpAttribute(CSharpType Type, params ValueExpression[] Arguments);
```
ArcturusZhang, line 9: well if possible let us take this oppotunity to change this to `IReadOnlyList<ValueExpression>`?

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
ArcturusZhang: well I mean adding some description as "why" not just explaining what condition this branch is checking, which just works like a "document"

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/SignatureType.cs
live1206: Moved this to `BuilderHelpers`, the only usage is here for now.

live1206:
Pending on https://github.com/Azure/azure-sdk-for-net/pull/41322

PR #4027, opened 2023-12-09
Consolidate csproj writing

ArcturusZhang:
# Description

## Reason of making this change

[This change](https://github.com/Azure/autorest.csharp/pull/4026) uncovers an issue of the sequence of reading and writing `csproj`.
The autorest pipeline might parallelize the plugins when they do not have dependencies to each other, in our case, it is `csharpgen` and `csharpproj`, they are running in parallel - and in [this change](https://github.com/Azure/autorest.csharp/pull/4026), we read csproj in `csharpgen`, and write `csproj` in `csharpproj` which might lead to read file exception because the file is locked to be edited.

We once introduced a partial solution in [this issue](https://github.com/Azure/autorest.csharp/issues/3697) when we first find the issue with the same root cause, by introducing a `CSProjConfiguration` class to hold those configurations that are needed when writing csproj file.

Since these configuration should never have a different value from the one when writing source code, and we will never have a concurrent issue when we put them into the same plugin, this class (and its corresponding test cases) is not need therefore removed in this PR.

## Changes included in this PR

1. remove the `csharpproj` plugin (and all the other types and test cases related to it), and move the functionality of this plugin (writes a csproj file) to `csharpgen` plugin.
2. consolidate the way of writing csproj and other xml document to use `XmlWriter` instead to make it to have semantics.
3. (WIP) consolidate the writer of writing csproj files into the same class (`NewProjectScaffolding` class) because we previously have multiple classes doing the same thing.

Follow ups needed to be done after this PR: https://github.com/Azure/autorest.csharp/issues/4037

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Builders/CSProjBuilder.cs
```diff
@@ -0,0 +1,187 @@
...
+    public IList<CSProjCompileIncldue> CompileIncludes { get; }
+
+    public string Build()
+    {
+        var builder = new StringBuilder();
+        using var writer = XmlWriter.Create(builder, new XmlWriterSettings
+        {
+            OmitXmlDeclaration = true,
+            Indent = true
+        });
```
ArcturusZhang, line 53: I choose `XmlWriter` instead of constructing a tree of `XmlElement` or `XElement` (which is much more convenient)
because these `Element` stuff does not support we to add empty lines between those elements.
And we need to add blank lines in csproj files to make them more readable.
For the same reason, `Microsoft.Build` package is not useful here even if it could construct a csproj file in a way with semantics.

ArcturusZhang:
We have another issue that now the csproj of test projects are not generated.
Our test projects now work fine just because it previously has a csproj file. If we deleted it and regenerate, the csproj in test project will not be generated.

eng/Generate.ps1
```diff
@@ -9,7 +9,6 @@ $ErrorActionPreference = 'Stop'
 $repoRoot = Resolve-Path (Join-Path $PSScriptRoot '..')
 
 $swaggerDefinitions = @{};
-$swaggerTestDefinitions = @{};
```
ArcturusZhang, line 12: this and all its related contents are removed.
The initial purpose of this is to support the test generation for MPG, which is generated separately from the source code. We have to run autorest in the source code directory to generate the source code, and then run autorest in the test project directory to generate the samples/tests.
In our previous change, the test project in MPG is now generated along with the source code, we no longer need to invoke autorest twice to generate the test project, therefore these are not needed any more.

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -91,96 +41,153 @@ internal static string GetVersion()
...
+                writeFile(FormatPath(Path.Combine(testGenProjectFolder, $"{Configuration.Namespace}.Tests.csproj")), testCSProjContent);
             }
+        }
+
+        private static string FormatPath(string? path)
+        {
+            if (string.IsNullOrEmpty(path))
+                return path ?? "";
+            return Path.GetFullPath(path.TrimEnd('/', '\\')).Replace("\\", "/");
+        }
```
ArcturusZhang, line 90: autorest uses JRPC to get interprocess messaging which works in the form of json, and the default path separator on windows `\` needs to be escaped otherwise on autorest side they got a JSON error because of the escaping.
Here we introduce this method to manually replace the symbol to `/` which does not requires an escaping.

We have the same method in mgmt test gen target, I just copied it here. We need a follow up there to clean things up: https://github.com/Azure/autorest.csharp/issues/4037

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpProj.cs
```diff
@@ -3,75 +3,25 @@
...
+using AutoRest.CSharp.Common.Input;
+using AutoRest.CSharp.Common.Output.Builders;
 
 namespace AutoRest.CSharp.AutoRest.Plugins
 {
-    // ReSharper disable once StringLiteralTypo
-    [PluginName("csharpproj")]
-    // ReSharper disable once IdentifierTypo
-    internal class CSharpProj : IPlugin
+    internal class CSharpProj
```
ArcturusZhang, line 13: I believe there should be follow ups of this PR to eventually remove this class and combine its functionality into `NewProjectScaffolding` because they have overlapped responsibilities.

That is kind of a big change in our source code, and the main purpose of this PR is to remove the `csharpproj` plugin to fix the concurrent issue we found. Other parts work fine - therefore we could have other PRs to refactor that part.

archerzz (approved):

src/AutoRest.CSharp/Common/Output/Builders/CSProjBuilder.cs
live1206: Will this be used for production csproj file generation?
Or how important is it to keep the empty lines between elements? 
If we can leverage existing packages, it will reduce the code complexity significantly.

ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Output/Builders/CSProjBuilder.cs
ArcturusZhang: It is not used in our SDK.
It might be used by external customers. It is true that this blank line has relatively lower priority.
But I would like to reduce the amount of changes.
There will be follow up PRs on the csproj related things, we could change it there.

ArcturusZhang: On the other hand, since XmlWriter provides more low-level API, it will have less allocation and better performance then we are using XmlElement or XElement or Microsoft.Build package.
It is also kind of straightforward enough to write the xml file.

live1206: Not sure how this might be used for external customers since the class is `internal`.
And if this is used for external customers, I would prefer not to re-invent wheels to avoid maintenance effects.
I think we need to evaluate different ways by gathering more data, such as performance and allocation.
Given this is only used for testing internally, I am OK to merge it firstly and we revisit it with more data.

ArcturusZhang: No this is not the point.
The classes in our repo, are always only used by us.
The feature of generating csproj, is used by us only when generating test projects in this repo, and by external customers when they are generating their libraries.
This feature of generating csproj is controlled by config `skip-csproj`. It is always set to true when generating the libraries in `azure-sdk-for-net`.

We do not need to consider the possibility of "external customer needs more things so that they extend our generator to do that". That is not possible now

PR #4030, opened 2023-12-11
Update TypeFactory

live1206:
Changes for `TypeFactory.TryCreateType`
- Add support for IArrayTypeSymbol whose ElementType is framework type.
- Return false for not supported types instead of throwing exception, which causes more side effects, since CSharpType can only represent `INameTypeSymbol` for now.
- Return false if any `TypeArgument` can't be created


src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -454,6 +467,53 @@ public bool TryCreateType(ITypeSymbol symbol, Func<System.Type, bool> validator,
...
+            // For IArrayTypeSymbol, we can only handle it when the element type is a framework type.
+            var arrayType = TryGetFrameworkType(arrayTypeSymbol);
+            if (arrayType is not null)
+            {
+                type = new CSharpType(arrayType, false);
+                if (!type.IsValueType &&
+                    symbol.NullableAnnotation != NullableAnnotation.NotAnnotated)
+                {
+                    type = type.WithNullable(true);
+                }
```
ArcturusZhang, line 487: we could simplify this to calculate isNullable ahead of time.
Since `CSharpType.IsValueType` always has the same value as `Type.IsValueType`, therefore we could have it in this way:
```suggestion
                var isNullable = !arrayType.IsValueType && symbol.NullableAnnotation != NullableAnnotation.NotAnnotated;
                type = new CSharpType(arrayType, isNullable);
```

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -454,6 +467,53 @@ public bool TryCreateType(ITypeSymbol symbol, Func<System.Type, bool> validator,
...
+            {
+                if (!TryCreateType(typeArgtment, validator, out CSharpType? type))
+                {
+                    return false;
+                }
+                result.Add(type);
+            }
+            arguments = result.ToArray();
+            return true;
+        }
```
ArcturusZhang, line 515: We could save a `ToArray` here if we have this PR merged in first: https://github.com/Azure/autorest.csharp/pull/4004
and we could do this:
```suggestion
        private bool TryPopulateArguments(ImmutableArray<ITypeSymbol> typeArguments, Func<Type, bool> validator, [NotNullWhen(true)] out IReadOnlyList<CSharpType> arguments)
        {
            argument = null;
            var result = new List<CSharpType>();
            foreach (var typeArgtment in typeArguments)
            {
                if (!TryCreateType(typeArgtment, validator, out CSharpType? type))
                {
                    return false;
                }
                result.Add(type);
            }
            arguments = result;
            return true;
        }
```

ArcturusZhang: and when we have the attribute `NotNullWhen`, we do not really need to add the question symbol on the type.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -465,6 +525,13 @@ private string GetFullMetadataName(ISymbol namedTypeSymbol)
 
         private void GetFullMetadataName(StringBuilder builder, ISymbol symbol)
```
ArcturusZhang, line 526: I suggest we rename this to `BuildFullMetadataName` to avoid confusion

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -413,34 +414,46 @@ public bool TryCreateType(ITypeSymbol symbol, [NotNullWhen(true)] out CSharpType
...
+
+            // We can only handle IArrayTypeSymbol of framework type and INamedTypeSymbol for now since CSharpType can't represent other types such as IArrayTypeSymbol of user types
+            // Instead of throwing an exception, wihch causes more side effects, we just return false and let the caller handle it.
+            if (symbol is not INamedTypeSymbol namedTypeSymbol)
             {
-                throw new InvalidCastException($"Unexpected type {symbol}");
+                return false;
             }
 
+            return TryCreateTypeForINamedTypeSymbol(symbol, validator, out type, namedTypeSymbol);
```
ArcturusZhang, line 429: could we change this to a switch so that it would be more clear and much easier for us to add more case
such as:
```csharp
=> symbol switch
{
IArrayTypeSymbol arraySymbol => TryCreateTypeForIArrayTypeSymbol(arraySymbol, out type),
INamedTypeSymbol namedTypeSymbol => TryCreateTypeForINamedTypeSymbol(namedTypeSymbol, validator, out type),
_ => false,
}
```

ArcturusZhang: And we could change the signature of `TryCreateTypeForIArrayTypeSymbol` to only accept `IArraySymbol`

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -413,34 +414,46 @@ public bool TryCreateType(ITypeSymbol symbol, [NotNullWhen(true)] out CSharpType
...
+        }
+
+        private bool TryCreateTypeForINamedTypeSymbol(ITypeSymbol symbol, Func<Type, bool> validator, [NotNullWhen(true)] out CSharpType? type, INamedTypeSymbol namedTypeSymbol)
```
ArcturusZhang, line 432: From where this method is called, the parameter `symbol` and `namedTypeSymbol` is actually the same instance, maybe we could change it like this:
```suggestion
        private bool TryCreateTypeForINamedTypeSymbol(INamedTypeSymbol namedTypeSymbol, Func<Type, bool> validator, [NotNullWhen(true)] out CSharpType type)
```

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -454,6 +467,53 @@ public bool TryCreateType(ITypeSymbol symbol, Func<System.Type, bool> validator,
...
+        {
+            type = null;
+            if (symbol is not IArrayTypeSymbol arrayTypeSymbol)
+            {
+                return false;
+            }
+
+            // For IArrayTypeSymbol, we can only handle it when the element type is a framework type.
+            var arrayType = TryGetFrameworkType(arrayTypeSymbol);
+            if (arrayType is not null)
```
ArcturusZhang, line 480: I believe we also should have the validator here as well

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: sounds good

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: makes sense

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: refactoring this with above `switch` change, but `CSharpType?` needs to be nullable

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: yeah, sorry forgot it again...

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: We can have it this way for now, but once we support IArrayTypeSymbol with custom types, we need to do it same as INamedTypeSymbol since we need to construct CsharpType firstly and then check the nullability.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: If `arguments` is not nullable, how could you set it to null?

ArcturusZhang: > If `arguments` is not nullable, how could you set it to null?

The attribute does the magic.
For instance: https://github.com/Azure/autorest.csharp/blob/07e8ae69bdcd6d0b58d45d40f7ba38d8a6c37256/src/AutoRest.CSharp/Mgmt/Decorator/TypeExtensions.cs#L63
This is using a different attribute, but I believe it should follow the same rule

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
ArcturusZhang: OK - if compiler throws the error, maybe `NotNullWhen` attribute works differently from the attribute `MaybeNullWhen` because the latter does not require the question mark (with or without both work fine)

live1206: Interesting, I would prefer to have the explicit nullable symbol anyway for readability.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -414,61 +415,119 @@ public bool TryCreateType(ITypeSymbol symbol, [NotNullWhen(true)] out CSharpType
...
+            var result = new List<CSharpType>();
+            foreach (var typeArgtment in typeArguments)
+            {
+                if (!TryCreateType(typeArgtment, validator, out CSharpType? type))
+                {
+                    return false;
+                }
+                result.Add(type);
+            }
+            arguments = result.ToArray();
```
ArcturusZhang, line 505: since [this PR](https://github.com/Azure/autorest.csharp/pull/4004) has been merged, this `ToArray` could be removed.

ArcturusZhang (approved):

ArthurMa1978 (approved):

Issue #3819, opened 2023-10-11
Add PrivilegedOperation support to all mgmt sdk


ArthurMa1978:
This is a sub task of https://github.com/Azure/azure-sdk/issues/5589 which we will enable audit logging for privileged calls.

ArthurMa1978:
Here is the change for Storage and Tables https://github.com/Azure/azure-sdk-for-net/pull/39345

PR #4031, opened 2023-12-11
support privileged operation directive in mgmt codegen

RodgeFu:
fixes https://github.com/Azure/autorest.csharp/issues/3819

support privileged operation directive in mgmt codegen


src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -85,18 +85,44 @@ private MgmtClientOperation(IReadOnlyList<MgmtRestOperation> operations, Formatt
...
+                        .Select(op =>
+                        {
+                            var arg = Configuration.MgmtConfiguration.PrivilegedOperations[op.OperationId];
+                            MgmtReport.Instance.TransformSection.AddTransformLog(
+                                new TransformItem(TransformTypeName.PrivilegedOperations, op.OperationId, arg),
+                                op.Operation.GetFullSerializedName(),
+                                $"Operation {op.OperationId} is marked as Privileged Operation");
+                            return new CSharpAttribute(typeof(Azure.Core.CallerShouldAuditAttribute), arg);
+                        })
+                        .ToList();
```
ArcturusZhang, line 107: I believe all these content could be calculated when this `MgmtClientOperation` instance is created.
If we do in that way, we could significantly simply the code here.

ArcturusZhang: such as we add a field:
```
private readonly IReadOnlyList<CSharpAttribute> _attributes;

// intialize this in the ctor by calling a static method
```
and we could have this property almost unchanged, just add a `_attributes` argument.

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -73,6 +73,7 @@ private MgmtClientOperation(IReadOnlyList<MgmtRestOperation> operations, Formatt
             _extensionParameter = extensionParameter;
             IdVariableName = idVariableName;
             IsConvenientOperation = isConvenientOperation;
+            MethodSignature = EnsureMethodSignature();
```
ArcturusZhang, line 76: well this is not what I meant and this might lead to issues since the properties used to construct the MethodSignature (such as `ReturnType`) are lazy loaded, to ensure everything is always working, it should lazy load.

I mean the new part you added could be determined in the ctor (the assignment of attributes).

Or it is OK that we make MethodSignature to be lazy loaded, since no other place will need those attributes.

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -209,6 +198,36 @@ private FormattableString BuildDescription()
             return parameterMappings;
         }
 
+        private MethodSignature EnsureMethodSignature()
+        {
+            var attris = this._operations
```
ArcturusZhang, line 203: we really do not need to use a short form here, please just use `attributes` it is not really long.
```suggestion
            var attributes = this._operations
```

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -209,6 +198,36 @@ private FormattableString BuildDescription()
...
+                .Where(op => Configuration.MgmtConfiguration.PrivilegedOperations.ContainsKey(op.OperationId))
+                .Select(op =>
+                {
+                    var arg = Configuration.MgmtConfiguration.PrivilegedOperations[op.OperationId];
+                    MgmtReport.Instance.TransformSection.AddTransformLog(
+                        new TransformItem(TransformTypeName.PrivilegedOperations, op.OperationId, arg),
+                        op.Operation.GetFullSerializedName(),
+                        $"Operation {op.OperationId} is marked as Privileged Operation");
+                    return new CSharpAttribute(typeof(Azure.Core.CallerShouldAuditAttribute), arg);
+                })
```
ArcturusZhang, line 213: I also have a question here:
This class is a wrapper of rest operations, there are cases that we have multiple rest operations wrapped in this same client operation.
For instance, we have two rest operations wrapped inside this client operation: `Foo_A` and `Foo_B`.
And we have the configuration assigning them both:
```
privileged-operations:
- Foo_A: "operation A"
- Foo_B: "operation B"
```

Reading the code here, this scenario, we will generate one method like this:
```csharp
[CallerShouldAudit("operation A")]
[CallerShouldAudit("operation B")]
public async Task<Response> MethodName(...)
```

Is this expected? Will this compile? By default, I believe the `AllowMultiple` on attributes is set to false, therefore a method cannot have multiple same attributes

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
RodgeFu: then i think it's better to have the init in one place than put them separated. I will revert the change back.

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
RodgeFu: i thought about the scenario but decided not to handle it specially because 1. in most case, user can avoid it easily by only configuring one operation, 2. if there are cases we do want to config two for some reason, there are chance the right fix is to support AllowMultiple which is hard to determine automatically so i would prefer to let things broken and add a proper handling base on the real case then

ArcturusZhang (approved):

ArcturusZhang:
Please be sure to run the regen before you merge this PR. It should contain no change.

ArthurMa1978 (approved):

Issue #3932, opened 2023-11-15
Serialization/Deserialization with different media types 


pshao25:
Goal: go through several common media types by content-type/accept, and how the serialization and deserialization looks like with different bodies.

### Media type: application/json, body type: model

- [ ] Already perfect.

### Media type: application/json, body type: primitive type (string, int, double, etc)

- [ ] **Serialization:** 

Actual: `RequestContentHelper.FromObject(body);`
Expected: payload is expected, but there is boxing operation, we will optimize at https://github.com/Azure/autorest.csharp/issues/3915

- [ ] **Deserialization:**

Actual: `response.Content.ToObjectFromJson<string>()`

### Media type: application/json, body type: bytes
- [ ] **Serialization:** 

Actual: `RequestContentHelper.FromObject(body);`
Expected: same as above with boxing operation

- [ ] **Deserialization:**

Actual: `response.Content`

pshao25:
> How will this work once azure-sdk-for-net starts consuming the new version of BinaryData with [MediaType property](https://learn.microsoft.com/en-us/dotnet/api/system.binarydata.mediatype?view=dotnet-plat-ext-8.0)?

@JoshLove-msft Thank you for the information! Could you help me understand how this new property is expected to be used?

chunyu3:
how to model bytes: https://github.com/microsoft/typespec/issues/2703

PR #4039, opened 2023-12-12
Support text return type

pshao25:
Scope: handle thesection "Media type: Text, body type: primitive type (string, int, double, etc)" in https://github.com/Azure/autorest.csharp/issues/3932
1. When return type is `string` and `accept` is text type like `text/plain`, we need to support it. An example spec is
```
@route("/PetName")
@get
op getPetName(): {
  @statusCode statusCode: 200;
  @header contentType: "text/plain";
  @body body: string;
};
```
2. When request type is `string` and `content-type` is text type like `text/plain`, which is similar


src/AutoRest.CSharp/Common/Output/Expressions/Azure/AzureExtensibleSnippets.AzureRestOperationsSnippets.cs
```diff
@@ -37,9 +38,13 @@ public override TypedValueExpression GetTypedResponseFromEnum(EnumType enumType,
                 return ResponseExpression.FromValue(EnumExpression.ToEnum(enumType, rawResponse.Content.ToObjectFromJson(typeof(string))), rawResponse);
             }
 
-            public override TypedValueExpression GetTypedResponseFromBinaryDate(Type responseType, TypedValueExpression response)
+            public override TypedValueExpression GetTypedResponseFromBinaryDate(Type responseType, TypedValueExpression response, string? contentType = null)
```
ArcturusZhang, line 41: here is a typo in the method name
```suggestion
            public override TypedValueExpression GetTypedResponseFromBinaryData(Type responseType, TypedValueExpression response, string? contentType = null)
```

src/AutoRest.CSharp/Common/Output/Expressions/KnownValueExpressions/BinaryDataExpression.cs
```diff
@@ -13,6 +13,8 @@ internal sealed record BinaryDataExpression(ValueExpression Untyped) : TypedValu
         public FrameworkTypeExpression ToObjectFromJson(Type responseType)
             => new(responseType, new InvokeInstanceMethodExpression(Untyped, nameof(BinaryData.ToObjectFromJson), Array.Empty<ValueExpression>(), new[] { new CSharpType(responseType) }, false));
 
+        public StringExpression ToText() => new(Invoke(nameof(BinaryData.ToString)));
```
ArcturusZhang, line 16: if we want to invoke the `ToString` method, 
1. we should name this as `ToString` and disambiguate to `InvokeToString`.
2. since every object has a `ToString` method, we should not define this method on this type. We should define it on `ValueExpression`
3. `ValueExpression` already has a method `InvokeToString`.

Therefore we should remove this method, and where this method is called, we should use `InvokeToString` instead.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -349,6 +349,12 @@ private ReturnTypeChain BuildReturnTypes()
             return null;
         }
 
+        private string[]? GetReturnedResponseContentType()
```
ArcturusZhang, line 352: is it possible that we return `IReadOnlyList<string>`?

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -96,26 +96,20 @@ public static FormattableString GetConversionFormattable(this Parameter paramete
         {
             if (toType.EqualsIgnoreNullable(Configuration.ApiTypes.RequestContentType))
             {
-                switch (parameter.Type)
+                if (parameter.Type.IsFrameworkType)
```
archerzz, line 99: why revert the pattern matching? It seems the if-else can be done using pattern matching.

src/AutoRest.CSharp/Common/Input/InputTypes/OperationResponse.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace AutoRest.CSharp.Common.Input;
 
-internal record OperationResponse(IReadOnlyList<int> StatusCodes, InputType? BodyType, BodyMediaType BodyMediaType, IReadOnlyList<OperationResponseHeader> Headers, bool IsErrorResponse)
+internal record OperationResponse(IReadOnlyList<int> StatusCodes, InputType? BodyType, BodyMediaType BodyMediaType, IReadOnlyList<OperationResponseHeader> Headers, bool IsErrorResponse, string[]? ContentTypes)
```
archerzz, line 9: leaning towards an empty collection
```suggestion
internal record OperationResponse(IReadOnlyList<int> StatusCodes, InputType? BodyType, BodyMediaType BodyMediaType, IReadOnlyList<OperationResponseHeader> Headers, bool IsErrorResponse, IReadOnlyList<string> ContentTypes)
```

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -133,8 +127,33 @@ public static FormattableString GetConversionFormattable(this Parameter paramete
             return $"{parameter.Name:I}{conversionMethod}";
         }
 
+        private static FormattableString GetConversionFromFrameworkToRequestContent(this Parameter parameter, string? contentType)
+        {
+            if (TypeFactory.IsReadWriteDictionary(parameter.Type))
```
chunyu3, line 132: Does array (such as string[] )can be text? current we convert all array/list to Json.
Same consider about the enum.

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
pshao25: We need to check whether the type is `string`. switch not compiles on `typeof(string)`

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
pshao25: All of these are out of the scope of this PR.

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
ArcturusZhang: but when supports that.
You just need:
```csharp
case { IsFrameworkType: true } when parameter.Type.FrameworkType == typeof(string):
```

ArcturusZhang: or 
```csharp
case {} when parameter.Type.EqualsIgnoreNullable(typeof(string)):
```

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -145,12 +163,23 @@ private static bool IsContentTypeBinary(string contentType)
...
+                return BodyMediaType.Binary;
+            }
+
+            if (type == "text")
+            {
+                return BodyMediaType.Text;
             }
 
-            return false;
+            throw new NotSupportedException($"Content type {contentType} is not supported.");
```
lirenhe, line 182: What is the impact of throw exception here?  Will it cause the whole codegen to fail?  
If so, we should avoid that to happen.

pshao25: Yes, I would expect this to happen, because all the content types going to this line means they are not in the scopes that we have discussed, so we don't know what to generate. If exceptions come up then customers go to us and we will know that one is important.

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -134,8 +127,33 @@ public static FormattableString GetConversionFormattable(this Parameter paramete
...
+            {
+                return $"{typeof(RequestContentHelper)}.{nameof(RequestContentHelper.FromEnumerable)}({parameter.Name})";
+            }
+
+            BodyMediaType? mediaType = contentType == null ? null : ToMediaType(contentType);
+            if (parameter.RequestLocation == RequestLocation.Body && mediaType == BodyMediaType.Binary)
+            {
+                return $"{parameter.Name:I}";
+            }
+            if (parameter.RequestLocation == RequestLocation.Body && mediaType == BodyMediaType.Text && parameter.Type.FrameworkType == typeof(string))
```
lirenhe, line 147: what if the frameType is not string? e.g.  bool, int,  DatetimeOffset etc.  Are those valid case to support?

pshao25: No, we don't know how to support. We haven't discussed it.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/utils.ts
```diff
@@ -88,3 +94,38 @@ export function getSerializeName(
         type.name
     );
 }
+
+export function createContentTypeOrAcceptParameter(
```
lirenhe, line 98: Could you add some unit tests in emitter for this?

lirenhe: Dont' quite understand the name of this function.

pshao25: This function is for creating an `InputParameter`  of content type or accept for tspCodeMode. It's just a creating process with some arguments exposed. Nothing special. I don't think it is meaningful to add unit tests. What's more, this logic exists before, just because two places are calling this, I extract it into the utils file.

chunyu3 (approved):

ArcturusZhang (approved):

lirenhe (approved):
LGTM, please resolve CI failures before merging this PR. thanks

Issue #3454, opened 2023-05-31
[Feature] support AdditionalProperties


chunyu3:
DPG will support a model which may contain addition properties.

## AdditionalProperties definition in spec
1. swagger
```
"Foo": {
      "type": "object",
      "properties": {
        "known": {
          "type": "string"
        }
      },
      "required": [
        "known"
      ],
      "additionalProperties": true
    }
```
2. typespec (under discussion)
```
model Foo extends Record<unknown> {
  known: string;
}

model Foo is Record<unknown> {
  @doc("The name property")
  known: string;
}

```

DPG will generate a model which may accept additional properties.

chunyu3:
## Model generated
same as HLC

```
    public partial class TestModel
    {
        /// <summary> Initializes a new instance of TestModel. </summary>
        public TestModel()
        {
            AdditionalProperties = new ChangeTrackingDictionary<string, object>();
        }

        /// <summary> Initializes a new instance of TestModel. </summary>
        /// <param name="id"></param>
        /// <param name="name"></param>
        /// <param name="additionalProperties"> Additional Properties. </param>
        internal TestModel(string id, string name, IDictionary<string, object> additionalProperties)
        {
            Id = id;
            Name = name;
            AdditionalProperties = additionalProperties;
        }

        /// <summary> Gets or sets the id. </summary>
        public string Id { get; set; }
        /// <summary> Gets or sets the name. </summary>
        public string Name { get; set; }
        /// <summary> Additional Properties. </summary>
        public IDictionary<string, object> AdditionalProperties { get; }
    }


    public partial class TestModel : IUtf8JsonSerializable
    {
        void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
        {
            writer.WriteStartObject();
            if (Optional.IsDefined(Id))
            {
                writer.WritePropertyName("id"u8);
                writer.WriteStringValue(Id);
            }
            if (Optional.IsDefined(Name))
            {
                writer.WritePropertyName("name"u8);
                writer.WriteStringValue(Name);
            }
            foreach (var item in AdditionalProperties)
            {
                writer.WritePropertyName(item.Key);
                writer.WriteObjectValue(item.Value);
            }
            writer.WriteEndObject();
        }

        internal static TestModel DeserializeTestModel(JsonElement element)
        {
            if (element.ValueKind == JsonValueKind.Null)
            {
                return null;
            }
            Optional<string> id = default;
            Optional<string> name = default;
[...]

PR #4047, opened 2023-12-14
Support AdditionalProperties

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3454

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -1,14 +1,13 @@
...
 
-using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace AutoRest.CSharp.Common.Input
 {
-    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, bool IsNullable)
+    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, InputDictionaryType? InheritedDictionaryType, bool IsNullable)
```
ArcturusZhang, line 10: this change is taken from [Alexander's PR](https://github.com/Azure/autorest.csharp/pull/3726).

In most cases, it should work perfectly fine that we could get `InheritedDictionaryType` from the `BaseModel` property, especially in the scenario of using typespec, because in typespec, a model could only have one base model.
This is a backward compat thing because in swagger, a model could have multiple base models, and this property is not null when any of thme is a dictionary

ArcturusZhang: We could consider change this to a calculated property, for instance, for the case we have now, it could simply be:
```csharp
internal record InputModelType(/*parameters unchanged*/)
{
public InputDictionaryType InheritDictionaryType => BaseModel as InputDictionaryType;
}
```
and update the implementation here when we support the full conversion from swagger to input types.

But as for now, to reduce the amount of changes I will just leave it here, and may go over this again when the feature is done.

ArcturusZhang: And we cannot just make the InputDictionaryType the base model - because BaseModel only accepts `InputModelType`.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -347,12 +341,12 @@ export function getInputType(
 
     function getInputModelType(m: Model): InputType {
         /* Array and Map Type. */
-        if (!isNeverType(m)) {
```
ArcturusZhang, line 350: we only get into this function when `m.kind === "Model"`, and the `isNeverType` is checking `m.kind === "intrinsic"` which will never be true, therefore here we do not need to check this.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -558,14 +554,15 @@ export function getInputType(
...
+                Description: getDoc(program, m) ?? "",
                 IsNullable: false,
                 DiscriminatorPropertyName: discriminator?.propertyName,
                 DiscriminatorValue: getDiscriminatorValue(m, baseModel),
                 Usage: Usage.None,
+                InheritedDictionaryType: inheritedDictionaryType, // InheritedDictionaryType represent the type of additional properties property
                 BaseModel: baseModel, // BaseModel should be the last but one assigned to model
                 Properties: properties // Properties should be the last assigned to model
-            } as InputModelType;
+            };
```
ArcturusZhang, line 565: `as` kind of by-passes the static type validation.
For instance, here the variable `model` has a type of `InputModelType`, which requires a string for `Description`. But here we are creating an object, with `Description` of type `string | undefined`.
Without the `as`, this is a compilation error, if we have the `as` statement, the error is actually suppressed.
We should not do that, therefore I removed it, and assign empty string if description is empty.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Models/ChatCompletionFunctions.cs
```diff
@@ -24,7 +23,7 @@ public partial class ChatCompletionFunctions
         /// `{\"type\": \"object\", \"properties\": {}}`.
         /// </param>
         /// <exception cref="ArgumentNullException"> <paramref name="name"/> or <paramref name="parameters"/> is null. </exception>
-        public ChatCompletionFunctions(string name, IDictionary<string, BinaryData> parameters)
+        public ChatCompletionFunctions(string name, ChatCompletionFunctionParameters parameters)
```
ArcturusZhang, line 26: this changes because in its [tsp definition](https://github.com/Azure/autorest.csharp/blob/5ed56117b265c61877a8c2b8b9892266071f1f0c/test/UnbrandedProjects/Platform-OpenAI-TypeSpec/completions/models.tsp#L204) we have:
```
model ChatCompletionFunctionParameters is Record<unknown>;
```
when we do not have additional properties, we parse this as a dictionary, but when we support additional properties, this should be a model with additional properties property, therefore we have this change.

ArcturusZhang: and if they only want a dictionary in their tsp, they could 
1. just use plain `Record<unknown>` on the property
2. use alias like this:
```
alias ChatCompletionFunctionParameters = Record<unknown>;
```

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/modelType.test.ts
```diff
@@ -76,7 +76,6 @@ op test(@body input: Pet): Pet;
                         Kind: "String",
                         IsNullable: false
                     },
-                    IsNullable: false,
```
ArcturusZhang, line 79: this is removed because we never have a `IsNullable` property in the type `InputModelProperty`.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -667,40 +666,65 @@ export function getInputType(
...
                 `No specified type for discriminator property '${model.DiscriminatorPropertyName}'. Assume it is a string.`
             );
-            const discriminatorProperty = {
+            const discriminatorProperty: InputModelProperty = {
                 Name: model.DiscriminatorPropertyName,
                 SerializedName: model.DiscriminatorPropertyName,
                 Description: "Discriminator",
                 IsRequired: true,
                 IsReadOnly: false,
-                IsNullable: false,
```
ArcturusZhang, line 676: this is removed because we never have a `IsNullable` property in `InputModelProperty` type.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -667,40 +666,65 @@ export function getInputType(
...
-            return undefined;
+    function getInputModelBaseType(
+        m: Model
+    ): [InputModelType | undefined, InputDictionaryType | undefined] {
+        const baseModel = m.baseModel;
+        const sourceModel = m.sourceModel;
+
+        // we cannot have both `extends` and `is`, therefore only one of baseModel and sourceModel can be defined
+        if (
+            sourceModel !== undefined &&
```
chunyu3, line 695:
```suggestion
            !sourceModel &&
```

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -667,40 +666,65 @@ export function getInputType(
...
+                    sourceModel.indexer.key,
+                    sourceModel.indexer.value
+                )
+            ];
         }
 
-        // TypeSpec "primitive" types can't be base types for models
-        if (program.checker.isStdType(m)) {
-            return undefined;
+        if (baseModel !== undefined) {
```
chunyu3, line 707:
```suggestion
        if (!baseModel) {
```

chunyu3 (approved):
Approved with some change request.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
ArcturusZhang: well when I wrote this, I kind of did not figure out how to simplify this expression and ended up writing this
I will update

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -667,40 +666,62 @@ export function getInputType(
...
-        if (!m) {
-            return undefined;
-        }
 
-        // Arrays and dictionaries can't be a base type
-        if (m.indexer) {
-            return undefined;
+    function getInputModelBaseType(
+        m: Model
+    ): [InputModelType | undefined, InputDictionaryType | undefined] {
```
pshao25, line 689: Just curious, could array be a base type?

lirenhe, line 689: Can we return the union of  'InputModelType |  InputDictionaryType | undefined' for this method?

src/TypeSpec.Extension/Emitter.Csharp/src/type/inputType.ts
```diff
@@ -62,7 +63,7 @@ export interface InputEnumType extends InputType {
     Namespace?: string;
     Accessibility?: string;
     Deprecated?: string;
-    Description: string;
+    Description?: string;
```
lirenhe, line 66: Is this change related to the support of addtionalProperties?

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/modelType.test.ts
```diff
@@ -76,7 +76,6 @@ op test(@body input: Pet): Pet;
                         Kind: "String",
                         IsNullable: false
```
lirenhe, line 77: Can we have some tests in Emitter to cover  additionalProperties?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
ArcturusZhang: No. We tried in playground, it is OK to write this:
```
model Foo extends Array<string> {}
```
but Timothe confirms this is a bug.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Models/ChatCompletionFunctions.Serialization.cs
```diff
@@ -21,25 +21,7 @@ void IUtf8JsonWriteable.Write(Utf8JsonWriter writer)
                 writer.WriteStringValue(Description);
             }
             writer.WritePropertyName("parameters"u8);
-            writer.WriteStartObject();
-            foreach (var item in Parameters)
-            {
-                writer.WritePropertyName(item.Key);
-                if (item.Value == null)
```
lirenhe, line 28: Seems the null value handling was removed in the serialization of additionalProperties, is this expected?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
ArcturusZhang: Then we will have to write this in the place where we call this:
```
if (baseModel === undefined)
{
// undefined case
}
else if (isModelType(baseModel))
{
// do base model
}
else if (isInputDictionaryType(baseModel))
{
// do inherited dictionary type
}
```
I believe this looks way more ugly.

src/TypeSpec.Extension/Emitter.Csharp/src/type/inputType.ts
ArcturusZhang: No this is a clean up to align with the definitions.

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/modelType.test.ts
ArcturusZhang: Sure, I will add some

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Models/ChatCompletionFunctions.Serialization.cs
ArcturusZhang: This change is expected. But `the null value handling was removed in the serialization of additionalProperties` is not the correct way to understand this line of change.
This model has a property `Parameters`. Previously, we have:
```csharp
class ChatCompletionFunctions
{
public IDictionary<string, BinaryData> Parameters {get;}
}
```
and now we have:
```csharp
class ChatCompletionFunctions
{
public ChatCompletionFunctionParameters Parameters {get;}
}
```
and `ChatCompletionFunctionParameters` is a new model, therefore this bunch of code changes to just `WriteObjectValue` which calls the serialization of the model.

ArcturusZhang: I see. The serialization of `AdditionalProperties` of type `IDictionary<string, BinaryData>` whihc is the same as we changed before, but their serialization are different.
The serialization for additional properties is missing a `item == null` check.
I will not change it on this PR because other libraries have a lot of use cases of additional properties property, changing it causes a lot change to this PR.
I will follow up on this issue after this merges.

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/modelType.test.ts
ArcturusZhang: tests added

lirenhe (approved):

PR #4053, opened 2023-12-15
add doc for privileged operation

RodgeFu:
add doc for privileged operation


docs/mgmt/readme.md
```diff
@@ -599,6 +600,24 @@ public partial class VirtualMachineResource : ArmResource
 }
 ```
 
+### Mark Privileged Operations
+
+We provide a new configuration 'privileged-operations' to mark operations as privileged, which is a dictionary from the operationId to the reason why the operation should be marked as privileged
```
ArcturusZhang, line 605:
```suggestion
We provide a new configuration `privileged-operations` to mark operations as privileged, which is a dictionary from the operationId to the reason why the operation should be marked as privileged
```

ArcturusZhang (approved):

ArcturusZhang (approved):

ArthurMa1978 (approved):

PR #4054, opened 2023-12-15
fix the `GetHashCode` of `CSharpType`

ArcturusZhang:
# Description

In the [previous PR](https://github.com/Azure/autorest.csharp/pull/4004), we change the type of `Arguments` from `CSharpType[]` to `IReadOnly<CSharpType>`.
This change changes the fact that before this change, it is always an array, therefore its previous implementation of `GetHashCode` should always work, because arrays implicitly implement the interface `IStructuralEquatable`.
But that is not true after we change it to `IReadOnlyList<CSharpType>` because this interface does not implement the `IStructuralEquatable` interface and this might leads to a runtime cast exception.

# Checklist

# Ready to Land?


ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -135,7 +135,23 @@ private bool ArgumentsEquals(IReadOnlyList<Type> genericArguments)
             return true;
         }
 
-        public override int GetHashCode() => HashCode.Combine(_implementation, _type, ((System.Collections.IStructuralEquatable)Arguments).GetHashCode(EqualityComparer<CSharpType>.Default));
+        private int? _hashCode;
```
m-nash, line 138: Please add a unit test in Autorest.Csharp.Tests to ensure regressions like this are caught.

ArcturusZhang: tests are added

Issue #3650, opened 2023-08-09
Provide more information in the API doc for operations


ArthurMa1978:
- [x]  Add verbose information.
- [x]  Add ResourceType information, this will help on `SetApiVersion`, here is a [sample](https://github.com/Azure/azure-sdk-for-net/issues/35152).
- [x]  Update SetApiVersion comment about how to use these added information

RodgeFu:
Reopen because we also need to update Set-ApiVersion's comments to let user know how to figure out the ResourceType from the Operation's tooltip after the Resource information are available in most major SDK.

PR #4062, opened 2023-12-19
add resource type into method's summary

RodgeFu:
First step to fixes https://github.com/Azure/autorest.csharp/issues/3650. add resource type into method's summary

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -134,14 +134,23 @@ public MethodSignature MethodSignature
         private FormattableString BuildDescription()
         {
             var pathInformation = _operations.Select(operation =>
-                (FormattableString)$@"<item>
+            {
+                string resourceTypeItem = "";
```
ArcturusZhang, line 138: please use `string.Empty` when we need an empty string

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -134,14 +134,23 @@ public MethodSignature MethodSignature
...
+                if (operation.Resource != null)
+                    resourceTypeItem = $@"
+<item>
+<term>Resource Type</term>
+<description>{operation.Resource.ResourceType}</description>
+</item>";
```
ArcturusZhang, line 144: do we need to add something to indicate that this operation does not correspond to a resource type so that it does not support api-version overriding?

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
RodgeFu, line 138:
```suggestion
                string resourceTypeItem = String.Empty;
```

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
RodgeFu: I think it's better to add the comment to the method of overriding api-version about how to figure out the ResourceType or whether some methods are supported/unsupported. The ResourceType listed with the method is more like general information about the method like the OperationId and RequestPath. They are there to help user get them easily and correctly no matter who want to use them, so it's better for the methods that use them to determine/comment how they are expected to be used or supported.

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
RodgeFu: fixed. thanks.

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -134,14 +134,23 @@ public MethodSignature MethodSignature
         private FormattableString BuildDescription()
         {
             var pathInformation = _operations.Select(operation =>
-                (FormattableString)$@"<item>
+            {
+                string resourceTypeItem = String.Empty;
```
ArcturusZhang, line 138: well in this repo we prefer the keyword instead of their real type names for cases like this.
```suggestion
                string resourceTypeItem = string.Empty;
```
like `string` vs `System.String`, `int` vs `System.Int32`, `float` vs `System.Single`, etc.
Those real type names require you to include `using System`, but the keywords do not.

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetCollection.cs
```diff
@@ -65,6 +65,10 @@ internal static void ValidateResourceId(ResourceIdentifier id)
         /// <term>Operation Id</term>
         /// <description>AvailabilitySets_CreateOrUpdate</description>
         /// </item>
+        /// <item>
+        /// <term>Resource Type</term>
+        /// <description>Microsoft.Compute/availabilitySets</description>
```
ArcturusZhang, line 70: I think our customers might get confused why an operation has a "resource type".
More useful information for our customers might be "the default api-version" of this method which makes total sense to our customers.
And since this PR is trying to "resolve" issue https://github.com/Azure/autorest.csharp/issues/3650 which comes from an inquiry of "how to set the api-version of an operation". If the content of this PR is only showing up "resource type" of an operation, I cannot admit that this PR **resolves** the issue.
This PR does provide more information, but does not really resolve the issue.

RodgeFu: As I said, the "Resource Type" is a general information related to the operation, I am fine to improve it to something like "The Resource Type providing the Operation" or something like that. but i dont think we should provide the information explicitly for set-apiversion api which is just one kind of usage of that information and also each api should decide their own behavior by themselves. Other api shouldn't comment on how set-apiversion API would work or use what kind of information, otherwise we are adding unnecessary coupling and making it harder to maintain both APIs in the future.

As for the issue #3650, this is like the first step of the fix. i am fine to keep it open until we add more comment to the set-apiVersion API when the "Resource Type" is available in most important sdk.

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -134,14 +134,23 @@ public MethodSignature MethodSignature
...
         {
             var pathInformation = _operations.Select(operation =>
-                (FormattableString)$@"<item>
+            {
+                string resourceTypeItem = string.Empty;
+                if (operation.Resource != null)
+                    resourceTypeItem = $@"
+<item>
+<term>Resource Type</term>
+<description>{operation.Resource.ResourceType}</description>
```
ArcturusZhang, line 143: if you want to add a cref of a type, you could just write:
FormattableString s = $"{operation.Resource:C}";
when writer writes this, we will get:
<see cref="Namespace.Name"/>

ArcturusZhang:
default api-version

ArcturusZhang (approved):

ArthurMa1978 (approved):

PR #4066, opened 2023-12-22
Improve performance of TransformSection.cs and TransformableItem.cs

nisha-bhatia:
# Description

# Checklist

# Ready to Land?


nisha-bhatia:
Improve performance by using a hashset instead of a list and caching the results

nisha-bhatia:
Before function counts:
image
image

After improvements:
image
image
image

ArcturusZhang (approved):

src/AutoRest.CSharp/Mgmt/Report/TransformableItem.cs
```diff
@@ -18,20 +18,28 @@ public TransformableItem(string fullSerializedName, TransformSection transformSe
...
-                return r.Count == 0? null : r;
+                {
+                    if (_appliedTransformLogs is null)
+                    {
+                        var r = _transformSection.GetAppliedTransformLogs(
+                        this.FullSerializedName, this.TransformTypeWhiteList)
+                        .OrderBy(item => item.Log.Index)
+                        .Select(item => $"[{item.Log.Index}][{item.Transform}] {item.Log.LogMessage}").ToList();
+                        // return null when it's an empty list so that it will be ignored in Json
+                        _appliedTransformLogs = r.Count == 0 ? null : r;
```
pallavit, line 39: Going purely by the comment here, do we want to cache the result as empty and return null instead when the expression is indeed empty? That way you would cache empty value too.

src/AutoRest.CSharp/Mgmt/Report/TransformSection.cs
```diff
@@ -18,7 +18,7 @@ public TransformSection(string name)
         {
         }
 
-        public IEnumerable<(TransformItem Transfom, TransformLog Log)> GetAppliedTransformLogs(string targetSerializedName, List<string>? transformTypes = null)
+        public IEnumerable<(TransformItem Transform, TransformLog Log)> GetAppliedTransformLogs(string targetSerializedName, HashSet<string>? transformTypes = null)
```
pallavit, line 21: Also I don't see you doing any conversion here other than changing the signature type? What benefit are you seeing from this change if any? The implementation is still enumerating it using an Enumerable - no?

nisha-bhatia: Hashset look-up/adding item is more efficient than a list. Using a Hashset also removes duplicates from the list so the Hashset could be smaller.

pallavit: I did not see any callers of these methods being updated, does that mean they were all already hashsets?
Also, how big do you expect these collections to be?

JoshLove-msft: It looks like the one caller of this method was updated to pass a HashSet - https://github.com/Azure/autorest.csharp/pull/4066/files#diff-6a9dca84faeb14085ed31735635c276edead4e7fc0659c1e672a9bdbca9e0034R21. That said, I don't really understand how this property gets populated as I don't see any other usages - https://grep.app/search?q=transformtypewhitelist&filter[repo][0]=Azure/autorest.csharp.

@ArcturusZhang - do you know how the property gets set? Is it via config?

src/AutoRest.CSharp/Mgmt/Report/TransformableItem.cs
```diff
@@ -18,20 +18,28 @@ public TransformableItem(string fullSerializedName, TransformSection transformSe
         }
 
         public string FullSerializedName { get; set; }
-        protected virtual List<string>? TransformTypeWhiteList { get { return null; } }
+        protected virtual HashSet<string>? TransformTypeWhiteList { get { return null; } }
```
JoshLove-msft, line 21: Consider renaming to TransformTypeAllowList.

src/AutoRest.CSharp/Mgmt/Report/TransformableItem.cs
nisha-bhatia: Updated the code to set the value of _appliedTransformLogs value immediately after running GetAppliedTransformLogs once and doing the null check in the return statement.

pallavit: Not sure how this change helps? You are still assigning _appliedTransformLogs to null when r is empty so you will continue to reevaluate it every time r.count is empty - no?

JoshLove-msft: I think you are looking at the outdated code.

m-nash (approved):

PR #4068, opened 2023-12-24
introduce the design spec for implementation of model reader writer interfaces in the generator

ArcturusZhang:
# Description

# Checklist

# Ready to Land?


docs/design/model-reader-writer-implementation-design.md
```diff
@@ -0,0 +1,244 @@
...
+}
+```
+
+The default behavior of the most widely used JSON serializer and deserializer is usually: they serialize classes and structs in the same way, assuming that they are models containing properties. This is true for our classes, but for structs, it is not always true. The extensible enums are modeled as structs, but they are actually serialized to a string value in our internal implementation when building the payload or parsing the response. For example, the `Foo` model above are expected to be serialized to the following JSON:
+```json
+{
+    "bar": "value"
+}
+```
+but in the default behavior of `System.Text.JsonSerializer`, it will be serialized to:
```
ArthurMa1978, line 45: We need mention that the `ModelReaderWriter` does not support `System.Text.JsonSerializer` now, so even we implement it in .Net SDK but still can't serialize the model properly by using `System.Text.JsonSerializer`.

ArthurMa1978 (approved):

docs/design/model-reader-writer-implementation-design.md
```diff
@@ -0,0 +1,244 @@
...
+If the model supports `xml` format, it will contain the following methods:
+1. `IXmlSerializable.Write(XmlWriter writer, string nameHint)` method.
+2. `WriteInternal(XmlWriter writer, string nameHint, ModelReaderWriterOptions options)` method.
+3. the static `Deserialize{ModelName}(XElement element, ModelReaderWriterOptions options = null)` method.
+
+#### The changes on models
+
+All models will now have a field `private IDictionary<string, BinaryData> _serializedAdditionalRawData`, if the model has a base type, the field should change to `private protected IDictionary<string, BinaryData> _serializedAdditionalRawData`.
+
+This private field works exactly the same as additional properties property, it should vanish when the model or any of its base type has an additional properties property defined.
```
lirenhe, line 123: What if the base type has an additional properties defined as:
```
public IDictionary<string, string> AdditionalProperties { get; }  ?
```
Will we convert the type into  
```
private IDictionary<string, BinaryData>?
```

ArcturusZhang: this is indeed an issue. We need to resolve this later.
I will create an issue for this.

ArcturusZhang: Issue created: https://github.com/Azure/autorest.csharp/issues/4128

docs/design/model-reader-writer-implementation-design.md
ArcturusZhang: added a paragraph about this

Issue #4078, opened 2023-12-26
[Bug] Cannot create an instance of the abstract type or interface.


asaflevi-ms:
For polymorphism (models with discriminator), the base model is abstract, and also it will generate a Unknown*** model to handle models with unknown discriminator value.

unknown types are designed to solve the above issue of Cannot create an instance of the abstract type or interface 
somehow, in this case, the unknown type is not properly referenced here, which is the bug.

Please check this tsp-location
https://github.com/Azure/azure-sdk-for-net-pr/blob/asaflevi/healthinsights-sdk-for-net-v2/sdk/healthinsights/Azure.Health.Insights.RadiologyInsights/tsp-location.yaml

-----------------------------------------------
tsp-location
directory: specification/ai/HealthInsights/HealthInsights.RadiologyInsights
additionalDirectories:
  - specification/ai/HealthInsights/HealthInsights.Common
  - specification/ai/HealthInsights/HealthInsights.OpenAPI
commit: 1f3bf4e0b64d0eee6ffa88b92154dce1f4718f78
repo: asaflevi-ms/azure-rest-api-specs-pr
cleanup: false
--------------------------------------------------------

After generating the code, I am getting  
Error	CS0144	Cannot create an instance of the abstract type or interface 'RadiologyInsightsInference' 

This service should be public next month so please prioritize.

CC: @chunyu3 , @ArcturusZhang, @catalinaperalta

PR #4080, opened 2023-12-27
Fix the instantiate abstract type issue

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4078

# Description

There are a few inconsistencies in our previous implementation on "the discriminated set base class should be abstract" and there should be an unknown class to instantiate the abstract class.

To keep things simple on the generator side, even if the base class is not abstract, we will still use the Unknown class to instantiate the base during deserialization. This is not the change that this PR is trying to introduce, we do that before this change.
This PR fixes the condition when we are building the deserialization code. The core change is this line: https://github.com/Azure/autorest.csharp/pull/4080/files#r1437461666

This PR also uncovers a long-existed bug in mgmt plane when the "resource data" is the base type in a discriminated set because resource data schema usually has a base type.

# Checklist

# Ready to Land?


test/TestProjects/Inheritance/Generated/Models/BaseClassWithDiscriminator.Serialization.cs
```diff
@@ -97,132 +94,7 @@ internal static BaseClassWithDiscriminator DeserializeBaseClassWithDiscriminator
...
-                {
-                    if (property.Value.ValueKind == JsonValueKind.Null)
-                    {
-                        continue;
-                    }
-                    dfeUri = JsonSerializer.Deserialize<DataFactoryElement<Uri>>(property.Value.GetRawText());
-                    continue;
-                }
-            }
-            return new BaseClassWithDiscriminator(baseClassProperty.Value, dfeString.Value, dfeDouble.Value, dfeBool.Value, dfeInt.Value, dfeObject.Value, dfeListOfT.Value, dfeListOfString.Value, dfeKeyValuePairs.Value, dfeDateTime.Value, dfeDuration.Value, dfeUri.Value, discriminatorProperty);
```
ArcturusZhang, line 225: we got lucky that previously this could compile because this package is HLC and HLC does not enroll the feature of "making the base class in a discriminated set abstract"

chunyu3 (approved):

archerzz (approved):

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -333,7 +333,9 @@ private static IEnumerable<MethodBodyStatement> BuildDeserializeBody(Parameter e
                     new SwitchStatement(discriminatorElement.GetString(), GetDiscriminatorCases(jsonElement, discriminator).ToArray())
                 };
             }
-            if (discriminator is not null && !serialization.Type.HasParent && discriminator.DefaultObjectType != null && !serialization.Type.Equals(discriminator.DefaultObjectType.Type))
+            // we redirect the deserialization to the `DefaultObjectType` (the unknown version of the discriminated set) if possible.
+            // We could only do this when there is a discriminator, and the discriminator does not have a value (having a value indicating it is the child instead of base), and there is an unknown default object type to fall back, and I am not that fallback type.
+            if (discriminator is { Value: null, DefaultObjectType: { } defaultObjectType } && !serialization.Type.Equals(defaultObjectType.Type))
```
ArcturusZhang, line 338: Our previous condition has a check `!serializationType.HasParent`, meaning that we do not use the unknown to deserialize when this type has a parent.

This is true for the normal models, and the derived types in a discriminated set. But for the base type in the discriminated set, it can have parent model.
The issue occurs here when the base class in the set has a base (without a discriminator), and it is abstract - because the condition here does not match and we will deserialize it as if it is a normal model and call its ctor which leads to the exception.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -82,7 +82,7 @@ public SchemaObjectType(ObjectSchema objectSchema, BuildContext context)
         private bool _hasCalculatedDefaultDerivedType;
         public SerializableObjectType? DefaultDerivedType => _defaultDerivedType ??= BuildDefaultDerivedType();
 
-        protected override bool IsAbstract => !Configuration.SuppressAbstractBaseClasses.Contains(DefaultName) && ObjectSchema.Discriminator?.All != null && ObjectSchema.Parents?.All.Count == 0;
+        protected override bool IsAbstract => !Configuration.SuppressAbstractBaseClasses.Contains(DefaultName) && ObjectSchema.Discriminator?.All != null && ObjectSchema.DiscriminatorValue == null;
```
ArcturusZhang, line 85: we make a class abstract when it is the base class in the discriminated set.
Here we check if it is a base type by checking if it has no parent models, this is true for very limited scenarios.
This PR changes it to check the discriminatorValue is null because the base model in the set does not have the value, while its derived types have.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -199,8 +199,8 @@ private bool IsDescendantOf(SchemaObjectType schemaObjectType)
 
             // We need this redundant check as the internal backing schema will not be a part of the discriminator implementations of its base type.
             if (ObjectSchema.DiscriminatorValue == "Unknown" &&
-                ObjectSchema.Parents?.All.Count == 1 &&
-                ObjectSchema.Parents.All.First().Equals(schemaObjectType.ObjectSchema))
+                ObjectSchema.Parents?.Immediate.Count == 1 &&
+                ObjectSchema.Parents.Immediate.First().Equals(schemaObjectType.ObjectSchema))
```
ArcturusZhang, line 203: we should not use `All` here because All recursively collect all the parent schemas of myself, therefore even if this is a unknown, inheriting from the base, if the base has a parent schema, here `ObjectSchema.Parents.All` will have two elements

test/TestProjects/Inheritance/Generated/Models/BaseClassWithDiscriminator.Serialization.cs
ArcturusZhang: This type `BaseClassWithDiscriminator` has a base type, and in our previous implementation, even if this is a base type in a discriminated set, it is not made abstract.
I believe this is a bug, and now this is fixed in this PR. Now this type is abstract and use the `unknown` version to deserialize.

test/TestProjects/Inheritance/Generated/Models/BaseClassWithDiscriminator.cs
```diff
@@ -16,10 +16,10 @@ namespace Inheritance.Models
     /// Please note <see cref="BaseClassWithDiscriminator"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
     /// The available derived classes include <see cref="ClassThatInheritsFromBaseClassWithDiscriminator"/> and <see cref="ClassThatInheritsFromBaseClassWithDiscriminatorAndSomeProperties"/>.
     /// </summary>
-    public partial class BaseClassWithDiscriminator : BaseClass
+    public abstract partial class BaseClassWithDiscriminator : BaseClass
```
ArcturusZhang, line 19: See this: https://github.com/Azure/autorest.csharp/pull/4080#discussion_r1437472803

Previously this is not abstract because it has a base class. This PR fixes this.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
ArthurMa1978: Is empty string a possible value for an abstract base class return from service side?

ArthurMa1978 (approved):

ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
ArcturusZhang: no. The service does not accept a "type", it accepts a json with a valid discriminator value. If the service team does not listed empty string as a valid discriminator value, it does work. Therefore we should not do that.

PR #4089, opened 2024-01-03
Pass azure-arm to generator from emitter

live1206:
We will support TSP for MPG in the near future, the generator will need to be able to distinguish between DPG and MPG TSP input.

TCGC already has `arm` option from `SDKContext`, this PR passes `azure-arm` option from emitter to generator to prepare for the TSP support of MPG.


ArthurMa1978 (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -58,7 +58,7 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
 
     if (!program.compilerOptions.noEmit && !program.hasError()) {
         // Write out the dotnet model to the output path
-        const root = createModel(context);
+        const [root, azureArm] = createModel(context);
         if (
```
pshao25, line 62: From your description, we could pass this flag by configuration, why need to be part of the created model?

live1206: It is from `SdkContext.arm` in tcgc, which is invoked during `createModel`

pshao25 (dismissed):
I think it is not a good way to put the flag into the created model.

src/TypeSpec.Extension/Emitter.Csharp/src/options.ts
```diff
@@ -33,6 +33,7 @@ export type NetEmitterOptions = {
     branded?: boolean;
     generateTestProject?: boolean;
     "use-model-reader-writer"?: boolean;
+    arm?: boolean;
```
ArcturusZhang, line 36: could we keep the name aligned with the config in swagger? It is called `azure-arm`.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
pshao25: Mingzhe is working on singleton sdkContext. After that, you could read the sdkContext.arm without changing the return type.

src/TypeSpec.Extension/Emitter.Csharp/src/options.ts
live1206: `arm` is defined in sdkContext, which is not in our control.
I think this option is a cmobination of sdkContext and .net generator options.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
live1206: That would be great, @archerzz any ETA for the singleton sdkContext to be available?

src/TypeSpec.Extension/Emitter.Csharp/src/options.ts
live1206: yes, it should be, will update it.

live1206: Actually this is not needed, because `arm` is already defined in `SdkEmitterOptions`

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -80,7 +80,7 @@ export function createModel(
 export function createModelForService(
     context: EmitContext<NetEmitterOptions>,
     service: Service
-): CodeModel {
+): [CodeModel, boolean?] {
     const emitterOptions = resolveOptions(context);
     const program = context.program;
     const sdkContext = createSdkContext(
```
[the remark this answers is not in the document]
live1206: sounds good

src/TypeSpec.Extension/Emitter.Csharp/src/options.ts
pshao25: I'm confused how user pass in `arm`? They only way I know is from `tspconfig.yaml`. Do we have other ways?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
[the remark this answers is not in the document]
live1206: did not change the `createModel` signature, construct `sdkContext` in emitter instead, because `context` is needed anyway.
We do have some duplicates of constructing, such as options, will not resole them in this PR.

src/TypeSpec.Extension/Emitter.Csharp/src/options.ts
live1206: user doesn't need to pass in `arm`, it is from `sdkContext.arm` from tcgc

live1206: tcgc already knows if the input TSP is `arm`

pshao25: I see, it is assigned after we call listClients.  https://github.com/Azure/typespec-azure/blob/0a5020b2964d7fc565ddc42ad09147b4b22cf3f4/packages/typespec-client-generator-core/src/decorators.ts#L209. So I'm afraid your code only works when there is only one sdkContext instance. Put it another way, you still need Mingzhe's change.

pshao25: BTW, why not we call `isArm(service.type)` ourselves? Why we have to read it from sdkContext?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
live1206: This change is not needed anymore

src/TypeSpec.Extension/Emitter.Csharp/src/options.ts
live1206: Please check the latest code, we don't have `service` in `emitter.ts`

live1206: > I see, it is assigned after we call listClients. https://github.com/Azure/typespec-azure/blob/0a5020b2964d7fc565ddc42ad09147b4b22cf3f4/packages/typespec-client-generator-core/src/decorators.ts#L209. So I'm afraid your code only works when there is only one sdkContext instance. Put it another way, you still need Mingzhe's change.

This works as long as we are using the same `sdkContext` to get `sdkContext.arm`, which is the case now.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
archerzz: > That would be great, @archerzz any ETA for the singleton sdkContext to be available?

@live1206 It's in https://github.com/Azure/autorest.csharp/pull/4091

live1206: @pshao25 
Right now, we have duplicated `createSdkContext` in `clientModelBuilder.ts` and `operation.ts`.
In #4091, we are consolidating to have only 1 in `clientModelBuilder.ts` and pass along it to `operation.ts`.
In this PR, I moved `createSdkContext` to `emitter.ts` and pass along it to `clientModelBuilder.ts`.
In the end, there is still a single `createSdkContext` call, no dependency from each other.

pshao25: The latest code looks good. Just one more comment: sdkContext.emitContext = context, so createModel only need one parameter sdkContext.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Configuration.json
```diff
@@ -6,5 +6,6 @@
     "../../../../../artifacts/bin/AutoRest.CSharp/Debug/net7.0/Generator.Shared",
     "../../../../../artifacts/bin/AutoRest.CSharp/Debug/net7.0/Azure.Core.Shared"
   ],
-  "branded": false
+  "branded": false,
+  "azure-arm": false
```
lirenhe, line 10: This looks strange, we don't want the name of Azure to exist in non-branding libraries.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -161,7 +166,8 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
                     options["generateTestProject"] === true
                         ? undefined
                         : options["generateTestProject"],
-                "use-model-reader-writer": options["use-model-reader-writer"]
+                "use-model-reader-writer": options["use-model-reader-writer"],
+                "azure-arm": sdkContext.arm
```
lirenhe, line 170: can it have a default value so we don't to upgrade all configuration.json.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
live1206: updated

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
live1206: updated

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Configuration.json
live1206: removed

src/TypeSpec.Extension/Emitter.Csharp/src/options.ts
```diff
@@ -129,7 +129,7 @@ const defaultOptions = {
     generateTestProject: true
 };
 
-export function resolveOptions(context: EmitContext<NetEmitterOptions>) {
+export function resolveOptions(context: EmitContext<any>) {
```
pshao25, line 132: Personally I don't like it. I'll let others to give opinion.

live1206: This is because `sdkContext.emitContext` is `EmitContext<Record<string, any>>`, which does not comply with `EmitContext<NetEmitterOptions>`

live1206: And we can't change it to `EmitContext<NetEmitterOptions | Record<string, any>>`, because some properties only exist in NetEmitterOptions, it will cause compile error.
Another option is to have a different method accepting different types, which would cause duplicate code...

pshao25: Huh? Why [sdkContext.emitContext](https://github.com/Azure/typespec-azure/blob/0a5020b2964d7fc565ddc42ad09147b4b22cf3f4/packages/typespec-client-generator-core/src/interfaces.ts#L34) is `EmitContext<Record<string, any>>` ?  It should be `EmitContext<NetEmitterOptions>`

live1206: Because [`NetEmitterOptions`](https://github.com/Azure/autorest.csharp/blob/4e2f9487d2b08f1909a1d520c70b38052ca2056e/src/TypeSpec.Extension/Emitter.Csharp/src/options.ts#L7) is defined by us, not in tcgc.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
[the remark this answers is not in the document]
live1206: updated

src/TypeSpec.Extension/Emitter.Csharp/src/options.ts
live1206: set generic type for SdkContext as `SdkContext<NetEmitterOptions>` could work

live1206: forgot to revert this, fixed now

ArcturusZhang (approved):

pshao25 (approved):

Issue #4092, opened 2024-01-03
[Dpg][Bug] `An item with the same key has already been added` reported from HealthInsights


ArcturusZhang:
Minimal tsp to reproduce this issue:
```
model Base
{
	kind: string;
}
@discriminator("kind")
model DiscriminatedBase extends Base {}

model Foo extends DiscriminatedBase
{
	kind: "foo";
}

op Bar(@body body: Foo): Foo;
```

ArcturusZhang:
The root cause of this issue is that
The discriminator is not defined on the model with the `discriminator` decorator, instead, it is inherited from a base model of this model.
Swagger cannot handle this properly as well.

ArcturusZhang:
The same issue should also happen when we try to override a property in the derived type, such as:
```
model Foo
{
p: int64;
}

model Bar extends Foo {
p: int32;
}
```
because tsp supports we could narrow down the type of a property in derived types.

ArcturusZhang:
This issue also involves with the issue of usage.
For instance, with this typespec:
```
model Base {
	name: string;
}

model Derived extends Base {
	age?: int32;
}

op Foo(@body body: Derived): Derived;
```

Our current behavior:
- usage of `Derived` is roundtrip. This is expected.
- usage of `Base` is none. This is unexpected.

Reason:
Usage is used to control our public API, including:
1. if the usage does not have input, the model will not have public ctors (both its ctors will be internal)
2. if the usage does not have input, all its properties will be generated as if they are all readonly.

Therefore the above tsp will generate these models:
```csharp
public partial class Base
{
	internal Base(string name) {} // this is internal which is fine in this scenario
	internal Base(string name, IDictionary<string, BinaryData> serializedAdditionalRawData) {} // this is for deserialization, irrelevant in this case.
	public Name {get;} // this does not have setter because its usage is None.
}
```

catalinaperalta:
> Minimal tsp to reproduce this issue:
> 
> ```
> model Base
> {
> 	kind: string;
> }
> @discriminator("kind")
> model DiscriminatedBase extends Base {}
> 
> model Foo extends DiscriminatedBase
> {
> 	kind: "foo";
> }
> 
> op Bar(@body body: Foo): Foo;
> ```

Wouldnt the issue be resolved by spreading `Base` into `DiscriminatedBase`? Or is there a reason to define that relationship explicitly? @asaflevi-ms please let me know if this is intended

catalinaperalta:
[...]

Issue #4096, opened 2024-01-04
[Dpg][Bug] Support type narrow down


ArcturusZhang:
**Describe the issue or request**

For instance we could have this in tsp:
```
model Foo
{
p: int64;
}

model Bar extends Foo {
p: int32;
}
```
in which the model `Bar` extends `Foo` and narrows down the type of property `p` inherited from `Foo`.

We currently are not supporting this properly (the generator throws key already added exception).

ArcturusZhang:
Narrow down could also happen with union types, such as:
```
model Foo {
	p: string | int32;
}

model Bar extends Foo {
	p: string;
}
```
or with literal unions
```
model Foo {
	p: "a" | "b";
}

model Bar extends Foo {
	p: "b";
}
```

ArcturusZhang:
Even this could compile as a tsp:
```
model Foo {
  p: string;
}

model Bar extends Foo {
  p?: string;
}
```

PR #4097, opened 2024-01-04
Fix the discriminator property inheritance issue

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4092
Depending on https://github.com/Azure/autorest.csharp/pull/4124

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonAdditionalPropertiesSerialization.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace AutoRest.CSharp.Output.Models.Serialization.Json
 {
-    internal class JsonAdditionalPropertiesSerialization : JsonPropertySerialization
+    internal record JsonAdditionalPropertiesSerialization : JsonPropertySerialization
```
ArcturusZhang, line 11: changing this to record does not really change anything in the runtime's perspective.
But record has its own "ToString" implementation, which simplifies the debugging process.
Therefore I leave these changes in this PR.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -155,7 +156,7 @@ private ModelTypeProviderFields EnsureFields()
         private ConstructorSignature EnsurePublicConstructorSignature()
         {
             //get base public ctor params
-            GetConstructorParameters(Fields.PublicConstructorParameters, out var fullParameterList, out var parametersToPassToBase, out var baseInitializers, true);
+            GetConstructorParameters(Fields.PublicConstructorParameters, out var fullParameterList, out var baseInitializers, true);
```
ArcturusZhang, line 159: last time I change this method, this output parameter is not used anywhere but I forgot to remove it. Here we just remove it

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -283,11 +303,29 @@ private void GetConstructorParameters(IEnumerable<Parameter> parameters, out IRe
...
             }
             parameterList.AddRange(parameters);
-            fullParameterList = parameterList;
+            fullParameterList = parameterList.DistinctBy(p => p.Name).ToArray(); // we filter out the parameters with the same name since we might have the same property both in myself and my base.
+        }
+
+        private bool? _isDiscriminatorInheritedOnBase;
+        internal bool IsDiscriminatorInheritedOnBase => _isDiscriminatorInheritedOnBase ??= EnsureIsDiscriminatorInheritedOnBase();
+
+        private bool EnsureIsDiscriminatorInheritedOnBase()
```
chunyu3, line 315: Just want to confirm if we need to consider override, as following, and what's the expected behavior?
```
model base {
  kind: string;
}

@discriminator("kind")
model baseWithDiscriminator extends base {
  kind: string,
  name: string;
}

model derived extends baseWithDiscriminator {
  kind: "A",
  age: int32;
}
```

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -399,10 +438,25 @@ private IReadOnlyList<ObjectPropertyInitializer> GetPropertyInitializers(IReadOn
...
+        }
+
         protected override IEnumerable<ObjectTypeProperty> BuildProperties()
         {
+            var existingPropertyNames = GetParentPropertyNames();
             foreach (var field in Fields)
-                yield return new ObjectTypeProperty(field, Fields.GetInputByField(field));
+            {
+                var property = new ObjectTypeProperty(field, Fields.GetInputByField(field));
+                if (existingPropertyNames.Contains(property.Declaration.Name))
```
lirenhe, line 456: To confirm, if one property whose name also exists in both ParentPropertyNames but with different type, which one will you use?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -573,6 +625,18 @@ private bool TryGetPropertyForInputModelProperty(Func<ObjectTypeProperty, bool>
             );
         }
 
+        private ObjectTypeProperty GetPropertyForDiscriminator(string inputPropertyName)
+        {
+            foreach (var obj in EnumerateHierarchy())
+            {
+                var property = obj.Properties.FirstOrDefault(p => p.InputModelProperty?.Name == inputPropertyName);
```
lirenhe, line 632: will this ensure we pick the right one if there is a duplicate name of the properties for the discriminator?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
ArcturusZhang: we have exactly the same case in the `Models-TypeSpec` case.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
ArcturusZhang: we use the inherited one.
Overriding a property and changing its type (type narrow down) is not supported.
C# does not support type narrow down, therefore to support it, we need to design what does it look like first.
Issue tracking that: https://github.com/Azure/autorest.csharp/issues/4096
Now we just ignore all attempts to override the property in base.

For example:
```
model Base {
	a: string | int;
}

model Derived extends Base {
	a: string;
}
```
We will ignore the `a` defined in `Derived` (this piece of code did this).

When we need to support [type narrow down](https://github.com/Azure/autorest.csharp/issues/4096), we will go over the design first, and then change this part to implement it.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
ArcturusZhang: This is filtering through `Properties` on `ObjectType`, which we have already filtered the properties with duplicated names (see [this comment](https://github.com/Azure/autorest.csharp/pull/4097#discussion_r1449859334))

lirenhe (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
chunyu3: @ArcturusZhang  The one you added is a normal case, that the `kind` is in the parent of the base model, but the base model does not override `kind` property.

PR #4100, opened 2024-01-04
Onboard cadl-ranch for non-azure services

### LINKED ISSUE #4095: Enable cadl-ranch for non-azure

### LINKED ISSUE #4103: Incorrect client hierarchy for non-azure
Choose arbitrary spec, e.g., https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/authentication/http/custom/main.tsp, set it branded `false` and generate.

Actual: In this PR https://github.com/Azure/autorest.csharp/pull/4100/files, it generated as this structure
image

Expected: should be the same as azure one as
image

pshao25:
1. Fix #4095
2. Fix #4103

The dashboard is not ready so the report will not upload. Track https://github.com/Azure/cadl-ranch/pull/486


test/AutoRest.TestServer.Tests/Infrastructure/CadlRanchNonAzureServer.cs
```diff
@@ -0,0 +1,42 @@
...
+            return Path.Combine(GetSpecDirectory(), "http");
+        }
+        internal static string GetCoverageFilePath()
+        {
+            return Path.Combine(GetCoverageDirectory(), "cadl-ranch-coverage-csharp-nonAzure.json");
+        }
+
+        protected override void Stop(Process process)
+        {
+            Process.Start(new ProcessStartInfo("node", $"{Path.Combine(GetNodeModulesDirectory(), "@azure-tools", "cadl-ranch", "dist", "cli", "cli.js")} server stop --port {Port}"));
```
ArcturusZhang, line 38: should we wait for this to exit before we exit?

test/AutoRest.TestServer.Tests/Infrastructure/CadlRanchNonAzureServerSession.cs
```diff
@@ -0,0 +1,23 @@
...
+        public static CadlRanchNonAzureServerSession Start()
+        {
+            var server = new CadlRanchNonAzureServerSession();
+            return server;
+        }
+
+        public override ValueTask DisposeAsync()
+        {
+            Return();
+            return new ValueTask();
```
ArcturusZhang, line 20: I believe we should return `ValueTask.CompletedTask`

test/AutoRest.TestServer.Tests/Infrastructure/CadlRanchNonAzureServer.cs
pshao25: Wait for what? We don't have any information to get.

pshao25: WaitForExit is a blocked call. Do you want an unblocked one?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -197,58 +192,38 @@ export function createModelForService(
...
-                : client.name; //quick fix for reserved namespace need something more robust
+    function getClientName(client: SdkClient | SdkOperationGroup): string {
+        if (client.kind === ClientKind.SdkClient) {
+            return client.name;
         }
+
+        var pathParts = client.groupPath.split(".");
+        return pathParts?.length >= 3
+            ? pathParts.slice(pathParts.length - 2).join("")
+            : client.type.name === "Models"
```
pshao25, line 217: Full tests will be in https://github.com/Azure/autorest.csharp/pull/4082

chunyu3: I have concern to add this hard-code here which will make our customer confusing.

pshao25: Yes, we would discuss and handle that in #4082

test/AutoRest.TestServer.Tests/Infrastructure/CadlRanchNonAzureServer.cs
pshao25: We have two processes in the test run.
1. Process 1: the cadl ranch server process, which starts at the beginning.
2. Process 2: the cadl ranch server stop process, which starts when we exit the test run and it will send a signal to stop Process 1.

Line 38 is starting the Process 2 and line 39 is waiting for Process 1 to exit. So.
1. Process 1 is exactly following what you proposed above.
2. Process 2 we don't have a waitforexit. I don't think we need this because our goal is to just stop Process 1.

ArcturusZhang (approved):

eng/Generate.ps1
```diff
@@ -72,7 +72,12 @@ function Add-CadlRanch-TypeSpec([string]$testName, [string]$projectPrefix, [stri
     $tspMain = Join-Path $cadlRanchFilePath $testName "main.tsp"
     $clientTsp = Join-Path $cadlRanchFilePath $testName "client.tsp"
     $mainTypeSpecFile = If (Test-Path $clientTsp) { Resolve-Path $clientTsp } Else { Resolve-Path $tspMain }
-    Add-TypeSpec "$projectPrefix$testName" $projectDirectory $mainTypeSpecFile "$configString--option @azure-tools/typespec-csharp.new-project=true" "-n"
+    if ($projectPrefix -eq "typespec-nonAzure-") {
```
lirenhe, line 75: I am fine to use nonAzure for now, but that name is subject to change in the future.

lirenhe (approved):

PR #4132, opened 2024-01-14
Introduce interfaces to replace the input types

ArcturusZhang:
# Description

This PR introduces an extra layer of interfaces to our input type system, which is trying to solve our immutable issue for our input types.

The interfaces will have immutable properties, such as:
```
public interface IType
{
	string Name {get;}
}
```
and its implementation could have settable properties, such as:
```
public record InputType : IType
{
	public string Name {get;set;}
}
```
which introduces a flexibility, when we need the name to be set in some limited scenarios, we could cast the instance to `InputType` and set the value of the property. The places using these types should always get the interfaces so that when using these types, we never have a setter to ensure those are immutable.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/CodeModelInputs.cs
ArcturusZhang, line 1: this file is renamed to match its only type

ArcturusZhang:
since we are planning to put the input types in a separate csproj, we no longer need this change as a long term solution, closing

PR #4144, opened 2024-01-18
Upload cadl ranch non-azure report

pshao25:
Now we will have two reports
1. `cadl-ranch-coverage-csharp-azure.json` for azure version report
2. `cadl-ranch-coverage-csharp-standard.json` for non-azure version report

After each build, these two reports will be uploaded to the artifacts.

After merge each PR, two commands will be called to upload these two reports in artifacts to storage account
1. `npx cadl-ranch upload-coverage --coverageFile ./artifacts/coverage/cadl-ranch-coverage-csharp-azure.json --generatorName csharp --storageAccountName azuresdkcadlranch --generatorVersion $(node -p -e "require('./src/TypeSpec.Extension/Emitter.Csharp/package.json').version") --generatorMode azure`
2. `npx cadl-ranch upload-coverage --coverageFile ./artifacts/coverage/cadl-ranch-coverage-csharp-standard.json --generatorName csharp --storageAccountName azuresdkcadlranch --generatorVersion $(node -p -e "require('./src/TypeSpec.Extension/Emitter.Csharp/package.json').version") --generatorMode standard`

The diff from previous command is that there will be a new argument called `generatorMode`.

Guidance for other languages
---
1. Start a cadl-ranch server for all the azure tests, you will get a coverage report for azure. 
2. Start a cadl-ranch server **again** for all the non-azure tests, you will get another coverage report for non-azure. 
3. (Optional) Upload these two reports to build artifacts as what you are currently doing.
4. Call `npx cadl-ranch upload-coverage --coverageFile {your azure report} --generatorMode azure {other arguments}`
5. Call `npx cadl-ranch upload-coverage --coverageFile {your non-azure report} --generatorMode standard {other arguments}`

### DIFF
--- a/eng/pipelines/build.yml
+++ b/eng/pipelines/build.yml
@@ -124,17 +124,28 @@ stages:
               failOnStderr: false
               displayName: 'Publish to GitHub, npm and coverage report'
               workingDirectory: $(Build.SourcesDirectory)/autorest.csharp
-          - publish: $(Build.SourcesDirectory)/autorest.csharp/artifacts/coverage/cadl-ranch-coverage-csharp.json
+          - publish: $(Build.SourcesDirectory)/autorest.csharp/artifacts/coverage/cadl-ranch-coverage-csharp-azure.json
+            artifact: CoverageReport
+          - publish: $(Build.SourcesDirectory)/autorest.csharp/artifacts/coverage/cadl-ranch-coverage-csharp-standard.json
             artifact: CoverageReport
           - ${{if eq(variables['System.TeamProject'], 'internal')}}:
             - task: AzureCLI@2
-              displayName: 'Upload to Cadl Ranch Coverage Report'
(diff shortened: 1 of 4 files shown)


chunyu3 (approved):

lirenhe (approved):

weidongxu-microsoft:
Does "--generatorMode standard" need to be called, when there is no "standard" test in codegen (only "azure")?

pshao25:
> Does "--generatorMode standard" need to be called, when there is no "standard" test in codegen (only "azure")?

no standard then only azure

weidongxu-microsoft:
OK, I get it refreshed now https://azure.github.io/cadl-ranch/

Seems the logic of "overwrite" get changed a bit. Previously if we re-run a client version 0.12.2, it would overwrite the previous 0.12.2 report. But now seems we need a new typespec-java version e.g. 0.12.3

Issue #3463, opened 2023-06-02
The CodeGenMemberSerializationHooks attribute cannot work on properties inherited from base types


ArcturusZhang:
**Describe the issue or request**

Currently to assign CodeGenMemberSerializationHooks attribute, we have to move the definition of the property into customize code, therefore it is impossible to change properties inherited because it is not defined on this model.

We need another attribute to allow that.

ArcturusZhang:
Proposal:

New attribute (or modify on existing attribute) on class
Note the `CodeGenMemberSerialization` and `CodeGenMemberSerializationHooks` can only be added on properties. The new attribute is on classes, or we could change the existing two to allow them to be added on classes.

Usage:
```csharp
[CodeGenMemberSerializationHooks(nameof(PropertyToChange), SerializationValueHook = nameof(SerializationValueHook), DeserializationValueHook = nameof(DeserializationValueHook))]
public partial Model
{
	private void SerializationValueHook(Utf8JsonWriter writer) {}
	private void DeserializationValueHook(JsonProperty property, ref Optional<string> propertyToChange) {}
}
```

PR #4159, opened 2024-01-26
Update attribute `CodeGenMemberSerializationHooks` to `CodeGenSerialization`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/3463

# Description

The PR removes the attribute `CodeGenMemberSerialization` and `CodeGenMemberSerializationHooks`.

This PR introduces a new attribute `CodeGenSerialization` which contains the functionalities of both removed attributes.

The new attribute could only be adde to a class or struct with a required parameter of `PropertyName` which represent the property or field you would like to change.
This is extremely useful when we want to change the serialization of an inherited property (in MPG, usually the `Id` inherited from `ResourceData`).

Generated code and documents are changed accordingly

A scenario of using this feature is demonstrated in the test projects and could be found here: [Azure/azure-sdk-for-net@`be68e6c` (#41635)](https://github.com/Azure/azure-sdk-for-net/pull/41635/commits/be68e6c8f51bf3a1272c6411148188d40a1bc3f5)

# Checklist

# Ready to Land?


src/assets/Generator.Shared/CodeGenMemberSerializationHooksAttribute.cs
```diff
@@ -26,5 +32,10 @@ internal class CodeGenMemberSerializationHooksAttribute : Attribute
         public CodeGenMemberSerializationHooksAttribute()
         {
         }
+
+        public CodeGenMemberSerializationHooksAttribute(string propertyName)
```
live1206, line 36: is it necessary to have this constructor?

ArthurMa1978: This is for property which is not in this class but inherit from base class

live1206: yes, I mean we can just have parameterless constructor since `PropertyName` has a setter already.
Why we have this constructor is because the user can just write `[CodeGenMemberSerializationHooks(nameof(Id),...)` instead of `[CodeGenMemberSerializationHooks(PropertyName: nameof(Id),...)`.

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
```diff
@@ -66,19 +66,28 @@ public bool TryGetCodeGenMemberSerializationAttributeValue(AttributeData attribu
...
         {
+            propertyName = null;
             serializationHook = null;
             deserializationHook = null;
             if (!CheckAttribute(attributeData, CodeGenMemberSerializationHooksAttribute))
             {
                 return false;
             }
 
+            if (attributeData.ConstructorArguments.Length > 0)
```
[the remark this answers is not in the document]
ArcturusZhang: it is ok, but as I said in this comment: https://github.com/Azure/autorest.csharp/pull/4159#discussion_r1469735875
I would like to use this ctor to emphasize this is required

readme.md
```diff
@@ -876,11 +876,18 @@ namespace Azure.Service.Models
 
 #### Change the implementation of serialization/deserialization method of one particular property
 
-If you want to change the implementation of serialization/deserialization method of one particular property, you could define your own hook methods and assign them to a property using the `CodeGenMemberSerializationHooks` attribute.
+If you want to change the implementation of serialization/deserialization method of one particular property, you could define your own hook methods and assign them to the property by using the `CodeGenMemberSerializationHooks` attribute.
```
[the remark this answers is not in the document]
ArcturusZhang: well it is the backward compatibility consideration, therefore I introduced this flexibility.
We have quite a few usage in the sdk already adding this attribute to a property.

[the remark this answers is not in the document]
live1206: yeah, I got the idea behind it.
But I feel like the same attribute on different places with different required fields is kind of confusing.
Is it possible to change the existing attribute we have in the custom code? Is it part of the public API already?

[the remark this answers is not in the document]
ArcturusZhang: And we do not only have this attribute, we have other similar attributes that only applies to property.

Another reason we should allow this attribute to be applied to a property is, we have the scenario that we are adding new properties and want to add it to the serialization. In this scenario, it would be intuitive if this attribute could apply to the property they just added. Same thing happens to the case when we are changing the type of an existing property.

ArthurMa1978 (approved):

ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Input/Source/ModelTypeMapping.cs
```diff
@@ -61,28 +47,37 @@ public ModelTypeMapping(CodeGenAttributes codeGenAttributes, INamedTypeSymbol ex
                     Usage = usage;
                     Formats = formats;
                 }
+
+                // handle CodeGenMemberSerializationHooks attribute
```
live1206, line 51: comment needs to be updated

test/TestProjects/MgmtCustomizations/Customization/PetStoreData.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Text.Json;
+using Azure.Core;
+
+namespace MgmtCustomizations
+{
+    [CodeGenSerialization(nameof(Id), SerializationValueHook = nameof(WriteId), DeserializationValueHook = nameof(ReadId))]
+    public partial class PetStoreData
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal void WriteId(Utf8JsonWriter writer)
```
live1206, line 16: If this is not in use, can just remove it?

test/TestProjects/Customizations-TypeSpec/src/Generated/Models/RenamedModel.cs
```diff
@@ -43,22 +43,22 @@ public partial class RenamedModel
         /// </list>
         /// </para>
         /// </summary>
-        private IDictionary<string, BinaryData> _serializedAdditionalRawData;
+        private protected IDictionary<string, BinaryData> _serializedAdditionalRawData;
```
live1206, line 46: if this is protected, do we still need private here?

ArcturusZhang: protected members are considered as part of the public API, because users of the library could inherit this class and use them, changing this will break their code. This is the reason why we want this field to be **private** protected

readme.md
```diff
@@ -874,13 +852,15 @@ namespace Azure.Service.Models
...
 </details>
 
-#### Change the implementation of serialization/deserialization method of one particular property
+#### Change the implementation of serialization/deserialization method of a property
 
-If you want to change the implementation of serialization/deserialization method of one particular property, you could define your own hook methods and assign them to a property using the `CodeGenMemberSerializationHooks` attribute.
+If you want to change the implementation of serialization/deserialization method of a property, you could define your own hook methods and assign them to the `CodeGenSerialization` attribute.
 
-<details>
+This `CodeGenMemberSerializationHooks` comes with two different usages, you could either add the attribute to the containing type of the property you would like to change, or add the attribute to a specified property.
```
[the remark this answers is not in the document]
ArcturusZhang: I forgot to remove this line. This will be removed.

src/AutoRest.CSharp/Common/Input/Source/ModelTypeMapping.cs
ArcturusZhang: updated

test/TestProjects/Customizations-TypeSpec/src/Generated/Models/RenamedModel.cs
ArcturusZhang: And the reason here we added `protected` is:
I change the spec to let another model extends this one. In public serialization implementation, when a model inherits from another, it does not have its own serialized additional raw data field, it will inherit one from the base type, therefore the field is `private` if no other model inherits it, it changes to `private protected` if this model is inherited by another model.

test/TestProjects/MgmtCustomizations/Customization/PetStoreData.cs
ArcturusZhang: it is used, in the attribute.
This project does not enable the mrw feature, therefore we did not see the write method to call this, but if we enable that, this will have usages.

live1206 (approved):

Issue #4044, opened 2023-12-13
[Feature Request] allow customization code to change a property in the model to a user-defined type


ArcturusZhang:
The following is only describing a scenario when we have this requirement, but the issue is actually more generic.

For instance we have this model:
```
model Foo
{
	data: int | float | string | unknown;
}
```
Our current generated code will generate this like:
```
public partial class Foo
{
public BinaryData Data {get; set;}
}
```
Despite we already have handy documentations to tell you what kind of object this property could accept, we still could customize this property by creating my own type to contain it.
If we do this:
```csharp
public class Data
{
// handles the union here, for instance by creating a few ctors to accept the types we accept.
}

public partial class Foo
{
[CodeGenModel("Data")]
[CodeGenMemberSerializationHooks(SerializationHook = nameof(WriteData), DeserializationHook = nameof(ReadData)]
public Data Data {get;set;}
}
```
which looks fine but it fails with the following exception:
```
Unhandled exception. System.InvalidOperationException: Unable to find a model or framework type that corresponds to A
  zure.Communication.JobRouter.RouterValue
     at AutoRest.CSharp.Generation.Types.TypeFactory.CreateType(ITypeSymbol symbol) in D:\a\_work\1\s\autorest.csharp\s
  rc\AutoRest.CSharp\Common\Generation\Types\TypeFactory.cs:line 401
     at AutoRest.CSharp.Output.Builders.BuilderHelpers.GetTypeFromExisting(ISymbol existingMember, CSharpType defaultTy
  pe, TypeFactory typeFactory) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Output\Builders\BuilderHelp
  ers.cs:line 215
     at AutoRest.CSharp.Output.Models.Types.ModelTypeProviderFields.CreateFieldFromExisting(ISymbol existingMember, Sou
  rcePropertySerializationMapping serialization, CSharpType originalType, InputModelType inputModel, InputModelProperty
   inputModelProperty, TypeFactory typeFactory, Boolean optionalViaNullability) in D:\a\_work\1\s\autorest.csharp\src\A
  utoRest.CSharp\Common\Output\Models\Types\ModelTypeProviderFields.cs:line 198
```

ArcturusZhang:
We will not do it here for autorest.csharp. Instead we will introduce this feature in the new generator.

PR #4160, opened 2024-01-26
Support fully hand-written types as property type

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4044

# Description

## Problem to solve

Our generator supports customization code, and you could introduce new properties or replace existing properties to the generated models using customization code, but with a constraint: the type of your new property or replaced property must be either framework type in the **generator's** assembly, or a generated type by the generator.
If you included a new dependency, and use a type from that library, you will get exceptions complaining that the generator cannot figure out the type.
Same thing happens when you write a new type by yourself in the customization code, and use that new type as a property of a model to customize.

This PR fixes this.

The most common use case when this is needed should be the union case.
The generator now generates union as `BinaryData` type. Our customer might have the need for writing a common type for the union, and use that as the type of those union property.
For example:
```
model Foo {
	a: string | int32;
}
```
by default, this will generate:
```csharp
public partial class Foo
{
	public BinaryData A {get;set;}
}
```
We could define a union type by ourselves in customization code, or introduce a dependency that provides a generic union type for C#, and then write the following customization code:
```csharp
public partial class Foo
{
	[CodeGenMemberSerializationHooks(SerializationValueHook = nameof(WriteA), DeserializationValueHook = nameof(ReadA))]
	public Union A {get;set;}

	[MethodImpl(MethodImplOptions.AggressiveInlining)]
	internal void WriteA(Utf8JsonWriter writer) {}

	[MethodImpl(MethodImplOptions.AggressiveInlining)]
	internal static void ReadA(JsonProperty, ref Union a) {}
}
```
The `CodeGenMemberSerializationHooks` is required in this use case, because there is no way the generator could know how this fully customized type should serialize and deserialize.

## Design details

The `CSharpType` carries all our metadata of a type of type, property, parameter, etc. It comes as a "union" of `System.Type` and `TypeProvider`:
https://github.com/Azure/autorest.csharp/blob/67c87784b786b2f2d23a3e38a5b48e86b75846aa/src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs#L19-L20
and
https://github.com/Azure/autorest.csharp/blob/67c87784b786b2f2d23a3e38a5b48e86b75846aa/src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs#L91-L93

This PR introduces another option for `CSharpType`, and now it could accept `INamedTypeSymbol` which represents a type read from customization code.
Previously, the generator also reads the customization code by `INamedTypeSymbol` [here](https://github.com/Azure/autorest.csharp/blob/67c87784b786b2f2d23a3e38a5b48e86b75846aa/src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs#L429), it will try to parse this symbol and see if it could consolidate to a known `System.Type` which we call a "framework type" or a known `TypeProvider` which represents a type being generated right now.
This method eventually will return `false` if the symbol cannot be consolidated to a `System.Type` or a `TypeProvider` and we get the exception.
This PR makes this method calls the new ctor of `CSharpType` to construct a type from the symbol, therefore this scenario is supported.

After the change, now `CSharpType` is a union of `System.Type`, `TypeProvider` or `INamedTypeSymbol`. Previously since it is only a union of two different types, we only need one flag to distinguish from what is the real type underlying - `IsFrameworkType`, but now since we have 3 options when have a `CSharpType` instance, we must add some new flags.
This PR adds another two flags: `IsTypeProvider` and `IsSymbolType`.
Previously there is a constraint that if we want to use the `FrameworkType` property, the `IsFrameworkType` flag must be `true`, and to use the `Implementation` property, the `IsFrameworkType` flag must be `false`.
Now since we have 3 options, this becomes more intuitive:
if we want to use property `FrameworkType`, the `IsFrameworkType` flag must be `true`.
if we want to use property `Implementation`, the `IsTypeProvider` flag must be `true`.
if we want to use property `SourceType`, the `IsSourceType` flag must be `true`.

Majority of the source code change in this PR is fixing this pattern, because previously we want to use the `Implementation` property on `CSharpType`, we are using this way:
```csharp
if (type is {IsFrameworkType: false, Implementation: {} implementation})
```
Now, when `IsFrameworkType` is `false`, there is a chance that we could have `SourceType` behind the curtain, therefore we have to change all of those to:
```csharp
if (type is {IsTypeProvider: true, Implementation: {} implementation})
```

Here is an demonstration of a scenario why this feature is useful: [Azure/azure-sdk-for-net@`0ceaf00` (#41632)](https://github.com/Azure/azure-sdk-for-net/pull/41632/commits/0ceaf00d8fd957eb5bbdc21d20e4cd2d2cdba25f)

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -90,13 +104,14 @@ public CSharpType(TypeProvider implementation, string ns, string name, bool isVa
...
         public Type FrameworkType => _type ?? throw new InvalidOperationException("Not a framework type");
+        public bool IsTypeProvider => _implementation != null;
         public TypeProvider Implementation => _implementation ?? throw new InvalidOperationException($"Not implemented type: '{Namespace}.{Name}'");
+        public bool IsSymbolType => _symbolType != null;
+        public INamedTypeSymbol SymbolType => _symbolType ?? throw new InvalidOperationException($"Not symbol type: '{Namespace}.{Name}'");
         public bool IsNullable { get; }
 
         public Type? SerializeAs { get; init; }
 
-        public bool HasParent => IsFrameworkType ? false : Implementation is ObjectType objectType ? objectType.Inherits is not null : false;
```
ArcturusZhang, line 98: remove this because the API is not used anywhere

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -90,13 +104,14 @@ public CSharpType(TypeProvider implementation, string ns, string name, bool isVa
         public IReadOnlyList<CSharpType> Arguments { get; } = Array.Empty<CSharpType>();
         public bool IsFrameworkType => _type != null;
         public Type FrameworkType => _type ?? throw new InvalidOperationException("Not a framework type");
+        public bool IsTypeProvider => _implementation != null;
         public TypeProvider Implementation => _implementation ?? throw new InvalidOperationException($"Not implemented type: '{Namespace}.{Name}'");
```
ArcturusZhang, line 108: I am not quite sure if we should keep the current names because `Implementation` is kind of not so intuitive
Should we align these two? If so, which should we take?
`IsTypeProvider` and `TypeProvider` or `IsImplemenation` and `Implementation`?

test/TestProjects/TypeSchemaMapping/SomeFolder/Generated/Models/CustomizedModel.cs
```diff
@@ -5,60 +5,24 @@
...
 
-using System;
-using System.Collections.Generic;
 using NamespaceForEnums;
 
 namespace CustomNamespace
 {
     /// <summary> The Model. </summary>
-    internal partial class CustomizedModel
+    internal partial class CustomizedModel : BaseClassForCustomizedModel
```
ArcturusZhang, line 13: we now have this change, because we have this base type in the customization code (existed before, not added in this update), but previously we cannot read that from the customization code, the code is generated as if this model does not have a base type.
Now we could recognize types like this `BaseClassForCustomizedModel` which does not correspond to a schema, or a framework type

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
```diff
@@ -115,27 +115,6 @@ private static void WriteBodyToRequestContent(CodeWriter writer, CodeWriterDecla
         {
             writer.Line($"using {Configuration.ApiTypes.RequestContentType} {contentVariable:D} = {requestContentValue};");
         }
-
-        public bool IsDeprecatedForExamples()
```
ArcturusZhang, line 119: removed because not used anywhere

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -11,13 +11,27 @@
 using AutoRest.CSharp.Input.Source;
 using AutoRest.CSharp.Output.Models.Shared;
 using AutoRest.CSharp.Output.Models.Types;
+using Microsoft.CodeAnalysis;
 
 namespace AutoRest.CSharp.Generation.Types
 {
     internal class CSharpType
     {
         private readonly TypeProvider? _implementation;
```
live1206, line 20: shall we change the naming to be more self-explaining?
```csharp
private readonly TypeProvider? _typeProvider;
private readonly Type? _frameworkType;
```
So that, we know the first one is for type created by the generator, the second one is the framework type.

live1206 (approved):

ArthurMa1978 (approved):

archerzz:
Looks like we may need a refactoring **in the future** to make `CSharpType` as abstract class, and have 3 sub-classes: `AssemblyCSharpType`, `InputCSharpType`, and `CustomCSharpType`

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
archerzz: We should make the names more straightforward.

ArcturusZhang:
> Looks like we may need a refactoring **in the future** to make `CSharpType` as abstract class, and have 3 sub-classes: `AssemblyCSharpType`, `InputCSharpType`, and `CustomCSharpType`

This is really a good idea. I will track and apply this idea

archerzz (approved):

PR #4174, opened 2024-01-30
Generate bicep serialization for ARM models

JoshLove-msft:


src/AutoRest.CSharp/Common/Generation/Writers/CodeWriter.cs
```diff
@@ -482,6 +482,7 @@ public CodeWriter Literal(object? o)
                 decimal d => SyntaxFactory.Literal(d).ToString(),
                 double d => SyntaxFactory.Literal(d).ToString(),
                 float f => SyntaxFactory.Literal(f).ToString(),
+                char c => SyntaxFactory.Literal(c).ToString(),
```
JoshLove-msft, line 485: This was needed to support generating the following:
```c#
string indent = new string(' ', spaces);
```

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -36,8 +38,13 @@
 using Azure.Core;
 using Azure.ResourceManager.Resources.Models;
 using Microsoft.CodeAnalysis;
+using YamlDotNet.Serialization.NodeTypeResolvers;
 using static AutoRest.CSharp.Common.Output.Models.Snippets;
+using ConstantExpression = AutoRest.CSharp.Common.Output.Expressions.ValueExpressions.ConstantExpression;
+using MemberExpression = AutoRest.CSharp.Common.Output.Expressions.ValueExpressions.MemberExpression;
 using SerializationFormat = AutoRest.CSharp.Output.Models.Serialization.SerializationFormat;
+using SwitchCase = AutoRest.CSharp.Common.Output.Expressions.Statements.SwitchCase;
```
JoshLove-msft, line 46: Looks like there are now some ambiguous references with the new Linq APIs so the type alias was necessary.

src/AutoRest.CSharp/Common/Output/Models/Types/SerializableObjectType.cs
```diff
@@ -58,13 +70,33 @@ protected SerializableObjectType(string defaultNamespace, SourceInputModel? sour
...
+
         protected abstract JsonObjectSerialization? BuildJsonSerialization();
         protected abstract XmlObjectSerialization? BuildXmlSerialization();
 
+        protected BicepObjectSerialization? BuildBicepSerialization()
+        {
+            // if this.Usages does not contain Output bit, then return null
+            // alternate - is one of ancestors resource data or contained on a resource data
+            var usage = GetUsage();
+
```
JoshLove-msft, line 91: This should be updated to only build bicep for models that are either ResourceData (or derive from ResourceData) or are properties on a ResourceData (or a model that derives from ResourceData).

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -193,6 +196,44 @@ public static JsonSerialization BuildSerialization(Schema schema, CSharpType typ
             }
         }
 
+        public static BicepSerialization BuildBicepSerialization(Schema schema, CSharpType type, bool isCollectionElement)
+        {
+            switch (schema)
```
ArcturusZhang, line 201: this could aslo be a switch expression I think?

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
```diff
@@ -123,7 +123,18 @@ public static void WriteMethodBodyStatement(this CodeWriter writer, MethodBodySt
...
                     }
-
+                    break;
+                case ForStatement forStatement:
+                    using (writer.AmbientScope())
+                    {
+                        writer.AppendRaw("for (");
+                        writer.Append($"int {forStatement.Indexer:D} = 0; {forStatement.Indexer} < {forStatement.Enumerable}.Length; {forStatement.Indexer}++)");
+                        writer.LineRaw("{");
+                        writer.Append($"{forStatement.ItemType} {forStatement.Item:D} = {forStatement.Enumerable}[{forStatement.Indexer}];");
```
ArcturusZhang, line 133: I believe this is not a good practice for a new statement.
This statement corresponds to 
```
for (<declOfIndexer>; <conditionOfIndexer>; <steperOfIndexer>)
{
// statements
}
```
I believe we should define this statement as this.
<declOfIndexer> is an "AssignmentExpression" which we do not have right now, we could create one.
<conditionOfIndexer> is a `BoolExpression`
<steperOfIndexer> is just simply a `ValueExpression`, and when construct this, we could use `UnaryExpression` to have `i++`.

And this line:
```
$"{forStatement.ItemType} {forStatement.Item:D} = {forStatement.Enumerable}[{forStatement.Indexer}];"
```
should not show up here, instead if we need this line in your scenario, we should always add this statement to the for statement as its first member

src/AutoRest.CSharp/Common/Output/Models/Serialization/Bicep/BicepObjectSerialization.cs
```diff
@@ -0,0 +1,79 @@
...
+                        property.SchemaProperty.Schema,
+                        property.Declaration.Type,
+                        false);
+                }
+                else if (property.InputModelProperty != null)
+                {
+                    bicepSerialization = SerializationBuilder.BuildBicepSerialization(property.InputModelProperty!.Type,
+                        property.Declaration.Type, false, property.SerializationFormat);
+                }
+                else if (property.Declaration.Name == "AdditionalProperties")
```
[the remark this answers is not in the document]
JoshLove-msft: Thanks, I reworked the implementation to reuse the JsonObjectSerialization types to avoid having to handle all of these complexities that have already been solved.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
JoshLove-msft: This method will go away

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -235,8 +237,7 @@ private IEnumerable<JsonPropertySerialization> CreatePropertySerializations()
                         valueSerialization,
                         property.IsRequired,
                         ShouldExcludeInWireSerialization(property, inputModelProperty),
-                        customSerializationMethodName: serializationMapping?.SerializationValueHook,
-                        customDeserializationMethodName: serializationMapping?.DeserializationValueHook,
+                        serializationMapping: serializationMapping,
```
JoshLove-msft, line 240: @ArcturusZhang I reworked this to pass the whole serializationMapping into PropertySerialization so that it can be used for bicep as well.

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.Methods.cs
JoshLove-msft: Thanks, I've updated based on your guidance.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
ArcturusZhang: well we kind of do not want to use this type here because `serializationMapping` here is considered as an "input model", and the serialization object we built here should be considered as "output model"

JoshLove-msft: I see your point, but the mapping contains metadata is used by the PropertySerialization. What if I encapsulated the part that contains the serialization hooks into a new record and instead pass that through?

JoshLove-msft: Potentially consider making the attribute into a dictionary/tuple input where users pass the key is the MRW format, and the value is the method name.

test/TestProjects/MgmtDiscriminator/Generated/Configuration.json
```diff
@@ -9,6 +9,7 @@
   "azure-arm": true,
   "public-clients": true,
   "skip-csproj-packagereference": true,
+  "use-model-reader-writer": true,
```
m-nash, line 12: Not saving the new flag

src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
```diff
@@ -50,12 +50,13 @@ public bool TryGetCodeGenMemberAttributeValue(AttributeData attributeData, [Mayb
             return name != null;
         }
 
-        public bool TryGetCodeGenSerializationAttributeValue(AttributeData attributeData, [MaybeNullWhen(false)] out string propertyName, out IReadOnlyList<string>? serializationNames, out string? serializationHook, out string? deserializationHook)
+        public bool TryGetCodeGenSerializationAttributeValue(AttributeData attributeData, [MaybeNullWhen(false)] out string propertyName, out IReadOnlyList<string>? serializationNames, out string? serializationHook, out string? deserializationHook, out string? bicepSerializationHook)
```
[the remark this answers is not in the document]
JoshLove-msft: I think this would make sense if we can also use a dictionary for the attribute properties. Right now it is 1:1 with the attribute properties. I can file an issue to explore updating the attribute to use a dictionary.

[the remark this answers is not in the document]
JoshLove-msft: https://github.com/Azure/autorest.csharp/issues/4207

m-nash (approved):

PR #4180, opened 2024-02-01
Add null check for optional request body

### LINKED ISSUE #3878: NRE when input body of convenience method is null
```C#
internal virtual async Task<Response> ReclassifyJobAsync(string id, IDictionary<string, string> reclassifyJobRequest = null, CancellationToken cancellationToken = default)
{
using RequestContent content = RequestContentHelper.FromDictionary(reclassifyJobRequest);
}
```

When passing `null` to reclassifyJobRequest, `RequestContentHelper.FromDictionary` throws NRE.

Solution 1:
In `RequestContentHelper.FromDictionary` we add:
```C#
if (dictionary == null)
    return null;
```

Solution 2:
In convenience method, we add:
```C#
if (reclassifyJobRequest == null)
{
    return await ReclassifyJobAsync(id, (RequestContent)null, cancellationToken).ConfigureAwait(false);
}
```

pshao25:
Fix #3878


archerzz (commented):
can we have a test server mock api?

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -134,12 +134,22 @@ private static FormattableString GetConversionFromFrameworkToRequestContent(this
         {
             if (TypeFactory.IsReadWriteDictionary(parameter.Type))
             {
-                return $"{typeof(RequestContentHelper)}.{nameof(RequestContentHelper.FromDictionary)}({parameter.Name})";
+                var conversionString = $"{typeof(RequestContentHelper)}.{nameof(RequestContentHelper.FromDictionary)}({parameter.Name})";
+                if (parameter.IsOptionalInSignature)
```
ArcturusZhang, line 138: even if the parameter is not optional in signature, it still could be null.
Here we should check the validation of this parameter is not "AssertNotNull or AssertNotNullOrEmpty" or just check if the validation is "None"

pshao25: If the parameter is optional in signature, we will check null, so it cannot be null.

ArcturusZhang: well never mind - we could do that later maybe

pshao25:
> can we have a test server mock api?

First, we are not testing payload. Second, per Michael, mockapi should be removed because it confuses customers. So I think we don't need that test.

src/AutoRest.CSharp/Common/Generation/Writers/FormattableStringHelpers.cs
```diff
@@ -134,12 +134,22 @@ private static FormattableString GetConversionFromFrameworkToRequestContent(this
...
+                {
+                    conversionString = $"{parameter.Name} != null ? {conversionString} : null";
+                }
+                return $"{conversionString}";
```
chunyu3, line 142: can we declare `conversionString` as `FormattableString` type, then it may not need to wrap again?

chunyu3 (approved):

lirenhe (approved):

PR #4186, opened 2024-02-02
Refactor the static methods in `TypeFactory` to properties of `CSharpType`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4194

# Description

We have quite a few static method in `TypeFactory` class to query the attribute of a `CSharpType`, for example:
```csharp
public static bool IsList(CSharpType type)
```
which tells you if this type is a list.

`CSharpType` works as a immutable data structure containing the type information, things like `IsList` is describing attributes of this type itself, therefore it should be a property.

This PR removes those static methods, and move them to `CSharpType` as properties.

Also this PR is trying to align the implementation between `CSharpType` in our repo, and [`CSharpType` in MGC](https://github.com/jorgerangel-msft/Microsoft.Generator.CSharp/blob/8265ded5efcf48af82bddbba5628ba7152b7ada3/Microsoft.Generator.CSharp/src/CSharpType.cs#L17)

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -56,30 +61,60 @@ public CSharpType(Type type, bool isNullable, IReadOnlyList<CSharpType> argument
             IsValueType = type.IsValueType;
             IsEnum = type.IsEnum;
             IsPublic = type.IsPublic && arguments.All(t => t.IsPublic);
+
+            #region Assign the attributes of this type
```
live1206, line 65: how about use a method instead of region?

pshao25:
> Actually, from OOP's perspective, these things should be properties on CSharpType class.

Do you have any reference to this statement?

Now you allocate several new spaces and need to do several new calculations for every type. Previous we only need to do this when needed.

src/AutoRest.CSharp/Mgmt/Decorator/ParameterMappingBuilder.cs
```diff
@@ -200,16 +200,21 @@ private static IEnumerable<List<Segment>> SplitDiffIntoPairs(RequestPath diff)
 
         public static FormattableString GetValueExpression(CSharpType type, FormattableString rawExpression)
         {
-            if (TypeFactory.IsStringLike(type))
+            // if the type is string
+            if (type.EqualsIgnoreNullable(typeof(string)))
+                return rawExpression;
+
+            // or if the type is extensible enum and its underlying type is also string
+            if (type is { IsFrameworkType: false, Implementation: EnumType { IsExtensible: true, ValueType: { } enumValueType } } && enumValueType.EqualsIgnoreNullable(typeof(string)))
```
live1206, line 208: how about having a private method to check this and the above condition?

live1206 (approved):

test/AutoRest.CSharp.UnitTests/Common/Generation/Types/CSharpTypeTests.cs
```diff
@@ -175,5 +176,42 @@ public void GetGenericTypeDefinitionForConstructedType()
...
+            => Assert.AreEqual(new CSharpType(type).IsAsyncPageable, expected);
+
+        [TestCase(typeof(int), false)]
+        [TestCase(typeof(AsyncPageable<>), false)]
+        [TestCase(typeof(AsyncPageable<int>), false)]
+        [TestCase(typeof(TestAsyncPageable<>), false)]
+        [TestCase(typeof(TestAsyncPageable<int>), false)]
+        [TestCase(typeof(Pageable<>), true)]
+        [TestCase(typeof(Pageable<int>), true)]
+        [TestCase(typeof(TestPageable<>), false)]
```
lirenhe, line 198: Why TestPageable<> is not 'Pageable' here?

ArcturusZhang: This represents a derived type for `Pageable<T>`, and we need this method to return false.

This is not new test case, I just move the test from another class here

src/AutoRest.CSharp/Mgmt/Decorator/ParameterMappingBuilder.cs
ArcturusZhang: well since we only use this once, I do not think it is worthy to write a method for it

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -56,30 +61,60 @@ public CSharpType(Type type, bool isNullable, IReadOnlyList<CSharpType> argument
...
+            // TODO -- these are azure-specific. We might need to generalize these using ApiTypes
+            IsResponse = type == typeof(Response);
+            IsResponseOfT = type == typeof(Response<>);
+            IsPageable = type == typeof(Pageable<>);
+            IsAsyncPageable = type == typeof(AsyncPageable<>);
+            IsOperation = type == typeof(Operation);
+            IsOperationOfT = type == typeof(Operation<>);
+            IsOperationOfAsyncPageable = IsOperationOfT && arguments.Count == 1 && arguments[0].IsAsyncPageable;
+            IsOperationOfPageable = IsOperationOfT && arguments.Count == 1 && arguments[0].IsPageable;
+            #endregion
```
ArcturusZhang, line 90: I do not have preference but Alexander usually calls out we should not abuse the lazy-loaded property, and assign everything to them when constructed.

This is usually valuable for immutable types such as this, I read something related on this topic as a comparison between C++ and rust.
It is the best practice to avoid funny issues that we always initialize everything for a type when constructing it especially when it is imuttable

chunyu3:
I think we can define a property `typeKind` to store this kind of information
enum CsharpTypeKind
{
            IEnumerableType = "IEnumerable";
           IEnumerableOfTType ="IEnumerable<>";
           IAsyncEnumerableOfTType = "IAsyncEnumerable<>";
           IReadOnlyDictionaryType = "IReadOnlyDictionary<,>";
            ......
}

internal class CSharpType
{
  ......
  public CsharpTypeKind typeKind;
}


//usage, check the CSharpType  property
if (cshartype.CSharpType == CsharpTypeKind.IEnumerableType ) {
}

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
live1206, line 65: would prefer a method than region here

ArcturusZhang: fixed.

chunyu3 (approved):

PR #4187, opened 2024-02-04
Update generator for LRO rehydration

live1206:
- Add serialization implementation for `Resource` types
- Consume [RehdyrationToken](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/RehydrationToken.cs) during generation


live1206:
Pending on https://github.com/Azure/azure-sdk-for-net/pull/41088

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -43,6 +43,61 @@ namespace AutoRest.CSharp.Common.Output.Builders
...
+
+
+            // void IJsonModel<T>.Write(Utf8JsonWriter writer, ModelReaderWriterOptions options)
+            var writer = new Utf8JsonWriterExpression(KnownParameters.Serializations.Utf8JsonWriter);
+            yield return new Method(
+                new MethodSignature(nameof(IJsonModel<object>.Write), null, null, MethodSignatureModifiers.None, null, null, new[] { KnownParameters.Serializations.Utf8JsonWriter, KnownParameters.Serializations.Options }, ExplicitInterface: jsonModelInterface),
+                new MethodBodyStatement[]
+                {
+                    // writer.WriteStringValue(ModelReaderWriter.Write(Data, options));
+                    new InvokeInstanceMethodStatement(writer, "WriteStringValue", new ValueExpression[] { new InvokeStaticMethodExpression(modelReaderWriter, "Write", new ValueExpression[] { new MemberExpression(This, "Data"), options }, new List<CSharpType> { resourceDataType }) }, false)
```
ArcturusZhang, line 63: I believe there is a method `WriteStringValue` on `Utf8JsonWriterExpression` so that you could just write:
```
writer.WriteStringValue(...the parameter expressions...)
```
as if we are really calling the writer.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -43,6 +43,61 @@ namespace AutoRest.CSharp.Common.Output.Builders
...
+            var resourceDataType = resource.ResourceData.Type;
+            var jsonModelInterface = new CSharpType(typeof(IJsonModel<>), resourceDataType);
+            var options = new ModelReaderWriterOptionsExpression(KnownParameters.Serializations.Options);
+            var modelReaderWriter = new CSharpType(typeof(ModelReaderWriter));
+            var iModelTInterface = new CSharpType(typeof(IPersistableModel<>), resourceDataType);
+            var data = new BinaryDataExpression(KnownParameters.Serializations.Data);
+
+
+            // void IJsonModel<T>.Write(Utf8JsonWriter writer, ModelReaderWriterOptions options)
+            var writer = new Utf8JsonWriterExpression(KnownParameters.Serializations.Utf8JsonWriter);
```
ArcturusZhang, line 57: there is a reference of this parameter in KnownParameters static class

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -43,6 +43,61 @@ namespace AutoRest.CSharp.Common.Output.Builders
...
+            yield return new Method(
+                new MethodSignature(nameof(IJsonModel<object>.Write), null, null, MethodSignatureModifiers.None, null, null, new[] { KnownParameters.Serializations.Utf8JsonWriter, KnownParameters.Serializations.Options }, ExplicitInterface: jsonModelInterface),
+                new MethodBodyStatement[]
+                {
+                    // writer.WriteStringValue(ModelReaderWriter.Write(Data, options));
+                    new InvokeInstanceMethodStatement(writer, "WriteStringValue", new ValueExpression[] { new InvokeStaticMethodExpression(modelReaderWriter, "Write", new ValueExpression[] { new MemberExpression(This, "Data"), options }, new List<CSharpType> { resourceDataType }) }, false)
+                });
+
+            // T IJsonModel<T>.Create(ref Utf8JsonReader reader, ModelReaderWriterOptions options)
+            var reader = new Parameter("reader", null, typeof(Utf8JsonReader), null, CSharp.Output.Models.Shared.ValidationType.None, null);
```
ArcturusZhang, line 67: there is a reference of this parameter in KnownParameters static class

ArcturusZhang: And we should have a type Utf8JsonReaderExpression (if we do not, maybe we should create one)

live1206: The reason I did not use KnowParameters is because the reader has `ref`, changed it to a method instead.

live1206: And don't want to add more complexity to this PR, let's add `Utf8JsonReaderExpression` in later PR and refactor the existing code.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
live1206: updated

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
```diff
@@ -363,7 +363,7 @@ private string GetBrandedTestCSProj()
             var writer = new CSProjWriter()
             {
                 TargetFrameworks = "$(RequiredTargetFrameworks)",
-                NoWarn = new("$(NoWarn);CS1591", "We don't care about XML doc comments on test types and members")
+                NoWarn = new("$(NoWarn);CS1591;Nu1605", "We don't care about XML doc comments on test types and members. We also don't care about System.ClientModel downgrade in the test project.")
```
m-nash, line 366: This should definitely not be here when we merge.  I am working on upgrading the System.ClientModel to 1.1.0-* https://github.com/Azure/autorest.csharp/issues/4242

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -220,6 +226,50 @@ public static void WriteRequestCreation(CodeWriter writer, RestClientMethod clie
...
         }
 
+        private static void WriteUriCreation(CodeWriter writer, RestClientMethod clientMethod, string methodAccessibility, ClientFields? fields, IReadOnlyList<Parameter>? clientParameters = null)
+        {
+            using var methodScope = writer.AmbientScope();
+            var methodName = CreateRequestUriMethodName($"{clientMethod.Name}");
+            writer.Append($"{methodAccessibility} {typeof(RequestUriBuilder)} {methodName}(");
+            foreach (Parameter clientParameter in clientMethod.Parameters)
+            {
+                writer.Append($"{clientParameter.Type} {clientParameter.Name:D},");
```
m-nash, line 236: We should use the expressions here instead of raw strings

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/SampleArmOperation.cs
```diff
@@ -43,6 +46,9 @@ internal SampleArmOperation(ClientDiagnostics clientDiagnostics, HttpPipeline pi
         public override string Id => throw new NotImplementedException();
 #pragma warning restore CA1822
 
+        /// <inheritdoc />
+        public override RehydrationToken? GetRehydrationToken() => _rehydrationToken;
```
m-nash, line 50: I don't think you can store this and will need to calculate it every time since each time Update is called the value can change

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetCollection.cs
```diff
@@ -101,7 +101,9 @@ public virtual async Task<ArmOperation<AvailabilitySetResource>> CreateOrUpdateA
             try
             {
                 var response = await _availabilitySetRestClient.CreateOrUpdateAsync(Id.SubscriptionId, Id.ResourceGroupName, availabilitySetName, data, cancellationToken).ConfigureAwait(false);
-                var operation = new SampleArmOperation<AvailabilitySetResource>(Response.FromValue(new AvailabilitySetResource(Client, response), response.GetRawResponse()));
```
m-nash, line 104: I don't think we should change this line here.  It should still work exactly as it did before since this is not a rehydration scenario.

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetResource.Serialization.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            return ModelReaderWriter.Write(Data, options);
+        }
+
+        AvailabilitySetData IPersistableModel<AvailabilitySetData>.Create(BinaryData data, ModelReaderWriterOptions options)
+        {
+            return ModelReaderWriter.Read<AvailabilitySetData>(data, options);
+        }
+
+        string IPersistableModel<AvailabilitySetData>.GetFormatFromOptions(ModelReaderWriterOptions options) => "J";
```
m-nash, line 36: this should return Data.GetFormatFromOptions

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetResource.Serialization.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.ClientModel.Primitives;
+using System.Text.Json;
+
+namespace Azure.ResourceManager.Sample
+{
+    public partial class AvailabilitySetResource : IJsonModel<AvailabilitySetData>
+    {
+        void IJsonModel<AvailabilitySetData>.Write(Utf8JsonWriter writer, ModelReaderWriterOptions options)
+        {
+            writer.WriteStringValue(ModelReaderWriter.Write(Data, options));
```
m-nash, line 18: I think we should forward via the interface here instead of calling back into MRW.

`((IJsonModel<AvailabilitySetData>)Data).Write(writer, options);`

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetResource.Serialization.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        AvailabilitySetData IJsonModel<AvailabilitySetData>.Create(ref Utf8JsonReader reader, ModelReaderWriterOptions options)
+        {
+            return ModelReaderWriter.Read<AvailabilitySetData>(new BinaryData(reader.ValueSequence), options);
```
live1206, line 23: This introduces memory allocation.

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/SampleArmOperation.cs
live1206: Actually, we don't have a way to re-calculate it within `ArmOperation`. We don't get the instance of `NextLinkOperationImplementation` in place, we only get internal `IOperation` back, which we don't want to add `GetRehdyrationToken()`.
It's OK to have the old token, the drawback of using it is, it will cause extra http traffic forwarding to latest polling URL.

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetCollection.cs
live1206: The reason we pass along the `rehydrationToken` here is, we need to get the `HttpMethod` for fake LRO scenario [here](https://github.com/Azure/azure-sdk-for-net/blob/0d9b36ffce7de9fcdd9d255635bb8ea467c845ea/sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs#L275).
If we can get the HttpMethod with other ways, we can revert this change.
I will think about it if I can find another way.

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/SampleArmOperation.cs
live1206: never mind, we can cast it back and re-calculate rehydration token.

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
live1206: Postpone this. If we can find a way to get Http method for fake LRO, we don't need this change.

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetResource.Serialization.cs
live1206: updated

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetResource.Serialization.cs
live1206: updated

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetResource.Serialization.cs
live1206: updated

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetCollection.cs
live1206: Added `RequestMethod` instead of `RehydrationToken` to `OperationInternalBase`

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
live1206: change reverted

live1206: Need to add this back, and this is just part of the [existing code](https://github.com/Azure/autorest.csharp/blob/49a5dfc9d7634bb901d53f20b25ceb4239201456/src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs#L72). Rewrite them into expressions takes more effort than expected. I would prefer to defer this until we convert the entire code into expression.

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/SampleArmOperation.cs
live1206: Update to get latest calculation for incomplete operation, get previous cache for complete operation.

samples/Azure.ResourceManager.Sample/Generated/AvailabilitySetCollection.cs
live1206: We still need to update this.
`rehydrationToken` is needed for fake LRO, because there is no way to recalculate the rehydration token for fake LRO and it won't change anyway since it is complete.

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
live1206: Pending on `System.ClientModel 1.1.0-*` release

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
live1206: create an issue for this

samples/Azure.ResourceManager.Sample/Generated/LongRunningOperation/SampleArmOperation.cs
```diff
@@ -20,20 +20,31 @@ internal class SampleArmOperation : ArmOperation
...
         internal SampleArmOperation(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, Request request, Response response, OperationFinalStateVia finalStateVia, bool skipApiVersionOverride = false, string apiVersionOverrideValue = null)
         {
             var nextLinkOperation = NextLinkOperationImplementation.Create(pipeline, request.Method, request.Uri.ToUri(), response, finalStateVia, skipApiVersionOverride, apiVersionOverrideValue);
+            if (nextLinkOperation is NextLinkOperationImplementation nextLinkOperationValue)
+            {
+                _nextLinkOperation = nextLinkOperationValue;
+            }
+            else
+            {
+                _rehydrationToken = NextLinkOperationImplementation.GetRehydrationToken(request.Method, request.Uri.ToUri(), response, finalStateVia, skipApiVersionOverride, apiVersionOverrideValue);
```
live1206, line 46: rename this

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
live1206: https://github.com/Azure/autorest.csharp/issues/4335

samples/AzureSample.ResourceManager.Sample/src/Generated/LongRunningOperation/SampleArmOperation.cs
```diff
@@ -20,28 +22,52 @@ internal class SampleArmOperation : ArmOperation
...
             _operation = new OperationInternal(nextLinkOperation, clientDiagnostics, response, "SampleArmOperation", fallbackStrategy: new SequentialDelayStrategy());
         }
 
+        private string GetOperationId(RehydrationToken? rehydrationToken)
+        {
+            if (rehydrationToken is null)
+            {
+                return null;
+            }
+            var lroDetails = ModelReaderWriter.Write(rehydrationToken, ModelReaderWriterOptions.Json).ToObjectFromJson<Dictionary<string, string>>();
```
live1206, line 63: Once we finish https://github.com/Azure/azure-sdk-for-net/issues/43260, we can just use `rehydrationToken.Id`.

ArthurMa1978 (approved):

PR #4188, opened 2024-02-04
Refactors on `SignatureType`

ArcturusZhang:
# Description

This PR introduces refactors for `SignatureType`.
Now the previous `SignatureType` has been split into two classes: new `SignatureType` and `BreakingChangeResolver`.

The new `SignatureType` class reads the type from the customization code and from the baseline contract to produce a list of `MethodSignature` for us to compare.
The `BreakingChangeResolver` class consumes the list of generated methods and produces a list of methods that the write should write, in which those "should skip" method are not included and the "overload methods" to mitigate the breaking changes are included

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/Source/SourceTypeMapping.cs
```diff
@@ -6,29 +6,36 @@
 
 namespace AutoRest.CSharp.Input.Source
 {
-    public class ModelTypeMapping
+    public class SourceTypeMapping
```
ArcturusZhang, line 9: I rename this type to `SourceTypeMapping` because it does not only apply to model

src/AutoRest.CSharp/Common/Output/Models/Types/TypeProvider.cs
```diff
@@ -15,6 +15,9 @@ namespace AutoRest.CSharp.Output.Models.Types
     internal abstract class TypeProvider
     {
         private readonly Lazy<INamedTypeSymbol?> _existingType;
+        private readonly Lazy<INamedTypeSymbol?> _contractType;
```
live1206, line 18: I don't feel like it's better to add such many additional fields/properties to the base `TypeProvider`, this affects every derived type since it is the base type of generator output types.

I tried to use `SignatureType` to hold the breakingchange overload before, it isolates the functionality in a central place.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -31,30 +31,19 @@ internal sealed class ModelFactoryTypeProvider : TypeProvider
         protected override string DefaultName { get; }
         protected override string DefaultAccessibility { get; }
 
-        // TODO: remove this intermediate state once we generate it before output types
         private IReadOnlyList<Method>? _methods;
+        public IReadOnlyList<Method> Methods => _methods ??= BuildMethods().ToArray();
 
-        // This method should only be called from OutputMethods as intermediate state.
-        private IReadOnlyList<Method> ShouldNotBeUsedForOutput([CallerMemberName] string caller = "")
```
live1206, line 38: This private method can be removed because of the change of `SignatureType`, which introduced additional properties to the base `TypeProvider`, I explained in the other comment.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -31,30 +31,19 @@ internal sealed class ModelFactoryTypeProvider : TypeProvider
...
+        private IEnumerable<Method> BuildMethods()
         {
-            Debug.Assert(caller == nameof(Methods) || caller == nameof(SignatureType), $"This method should not be used for output. Caller: {caller}");
-            return _methods ??= _models.Select(CreateMethod).ToList();
-        }
-
-        // TODO: remove this intermediate state once we generate it before output types
```
live1206, line 44: So, it seems like we would never be able to move this before constructing the output types because we need to have the generated method signatures to get the changeset combined with previous contract and custom potion.

But, is it possible to construct signatures beforehand and move this breaking change resolving functionality ahead? And then, we just need to input the prepared methods & properties into the output types instead of doing them here. That's the final vision from @m-nash. 

Again, that's why I would like to put this logic in a central place to be easily moved around later.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelFactoryTypeProvider.cs
```diff
@@ -115,20 +104,8 @@ private static string GetDefaultNamespace()
 
         public HashSet<MethodInfo> ExistingModelFactoryMethods { get; }
 
-        private SignatureType? _signatureType;
-        public override SignatureType? SignatureType
-        {
-            get
-            {
-                // This can only be used for Mgmt now, because there are custom/hand-written code in HLC can't be loaded into CsharpType such as generic methods
```
live1206, line 123: Are these concerns resolved already?

live1206 (changes requested):
We need to agree on the design decision firstly.

PR #4191, opened 2024-02-04
Refactor serialization generation

ArcturusZhang:
# Description

Previously the serialization code was built inside the `SerializationWriter` class.

Since the model with the serialization actually owns those methods, this PR changes the code so that the `SerializableObjectType` now produces a list of methods to contain those interface implementation methods and the `SerializationWriter` could just dumbly write whatever methods it gets, which is our expectation for a writer to do.

This PR introduces a new type `ObjectTypeSerialization` to collect all possible serializations, in order to make it "open to expand", by making all possible serializations as properties of the new type.

For instance, before change, if we would like to add a new format `bicep`, a ton of method signature needs to be changed, and now we just need to add a new property in `ObjectTypeSerialization`, build a proper value for it, and then write a new builder to build the specific part of that (new switch-cases should be added).

This PR also introduces a clean-up for those interface implementation methods. Previously the `JsonSerializationMethodsBuilder` builds the methods for `IJsonModel<T>`, which makes sense, but it also builds the methods for `IPersistableModel<T>`, which does not make much sense because this method is not only meant for Json, for example, xml serialization also must be considered as part of its implementation.
The clean up is that this PR creates a new class `SerialzationMethodsBuilder` which will collect the methods created by other serialization method builds (Json and Xml), and it handles the creation for `IPersistableModel<T>`.

# Checklist

# Ready to Land?


lirenhe (approved):

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -28,65 +25,71 @@
...
 using Azure.ResourceManager.Resources.Models;
 using Microsoft.CodeAnalysis;
+using static AutoRest.CSharp.Common.Input.Configuration;
+using static System.Runtime.InteropServices.JavaScript.JSType;
 using static AutoRest.CSharp.Common.Output.Models.Snippets;
 using SerializationFormat = AutoRest.CSharp.Output.Models.Serialization.SerializationFormat;
 
 namespace AutoRest.CSharp.Common.Output.Builders
 {
     internal static class JsonSerializationMethodsBuilder
```
live1206, line 42: Thought `SerializationMethodsBuilder` will be the entry for writer to consume. `JsonSerializationMethodsBuilder` and `XmlSerializationMethodsBuilder` are like private builders to it, but seems not, they are consumed by writers as well? how should they be used?

src/AutoRest.CSharp/Common/Output/Models/Types/JsonConverterProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+                });
+            var writer = new Utf8JsonWriterExpression(KnownParameters.Serializations.Utf8JsonWriter);
+            MethodBodyStatement writeMethodBody;
+            if (_includeSerializer)
+            {
+                writeMethodBody = writer.WriteObjectValue(modelParameter);
+            }
+            else
+            {
+                writeMethodBody = Throw(New.Instance(typeof(NotImplementedException)));
```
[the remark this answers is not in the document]
ArcturusZhang: this is the previous behavior.
When we do not have MRW, a model might not have serializer or deserializer. But with MRW, every model will have both, we no longer have that not implemented branch in generated code.

src/AutoRest.CSharp/Common/Output/Models/Serialization/SerializationInterfaces.cs
```diff
@@ -0,0 +1,102 @@
...
+using AutoRest.CSharp.Common.Input;
+using AutoRest.CSharp.Generation.Types;
+
+namespace AutoRest.CSharp.Output.Models.Serialization
+{
+    internal record SerializationInterfaces : IEnumerable<CSharpType>
+    {
+        public SerializationInterfaces(bool includeSerializer, bool isStruct, CSharpType modelType, bool hasJson, bool hasXml)
+        {
+            // TODO -- includeSerializer could be removed after the use-model-reader-writer configuration is removed because it is always true
```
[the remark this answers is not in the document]
ArcturusZhang: We enable MRW by `use-model-reader-writer`, and this is how we get this flag:
image

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
```diff
@@ -4,16 +4,9 @@
 using System;
 using System.ClientModel.Primitives;
 using System.Linq;
```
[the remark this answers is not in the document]
ArcturusZhang: we should do that.
But will not included in this PR.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: builders are not consumed by writers. TypeProviders should use those builders to build their content.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -28,65 +25,71 @@
...
 using AutoRest.CSharp.Output.Models.Serialization.Json;
-using AutoRest.CSharp.Output.Models.Serialization.Xml;
 using AutoRest.CSharp.Output.Models.Shared;
 using AutoRest.CSharp.Output.Models.Types;
 using AutoRest.CSharp.Utilities;
 using Azure;
 using Azure.Core;
 using Azure.ResourceManager.Resources.Models;
 using Microsoft.CodeAnalysis;
+using static AutoRest.CSharp.Common.Input.Configuration;
```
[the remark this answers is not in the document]
ArcturusZhang: cleaned up

src/AutoRest.CSharp/Common/Output/Models/Serialization/SerializationInterfaces.cs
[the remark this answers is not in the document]
live1206: Yes, I mean when can we remove the flag? That last time I asked your answer is we need to wait for most RP enabled MRW. My question is why not remove this flag and make public serialization enabled globally?

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
live1206: Sorry, I meant how typeProvider should use these builders.
My actual question is, how do we decide the responsibility of `SerializationMethodsBuilder`, `JsonSerializationMethodsBuilder` and `XmlSerializationMethodsBuilder`? 

> JsonSerializationMethodsBuilder builds the methods for IJsonModel<T>, which makes sense, but it also builds the methods for IPersistableModel<T>, which does not make much sense because this method is not only meant for Json, for example, xml serialization also must be considered as part of its implementation.
The clean up is that this PR creates a new class SerialzationMethodsBuilder which will collect the methods created by other serialization method builds (Json and Xml), and it handles the creation for IPersistableModel<T>.

Does this mean `JsonSerializationMethodsBuilder` build methods for `IJsonModel<T>`?
What does `XmlSerializationMethodsBuilder` build methods for?
`SerialzationMethodsBuilder` collect methods created by other serialization method builds, what does `other` contain here? Only for `IPersistableModel<T>`?
The responsibility is not clear from the naming.

src/AutoRest.CSharp/Common/Output/Models/Types/JsonConverterProvider.cs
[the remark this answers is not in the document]
live1206: If `use-model-reader-writer` is not set to true for a RP, that means MRW is not enabled for that RP.
And if MRW is enabled, `_includeSerializer` is not needed anymore, we will not go to the else branch.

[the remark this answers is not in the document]
ArcturusZhang: Yes, what is your point

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: These builders are not building methods for interfaces.
Please note the `IPersistableModel<T>` interface does not correspond to a specific format.
`IJsonModel<T>` corresponds to json format yes, but `IPersistableModel<T>` does not.
Therefore the way I am trying to introduce is that:
these builders build methods for that format, instead of interfaces.
Json is special because it has an interface. (skip this)
How is IPersistableModel<T> implementation built?
This is how it looks like:
```
BinaryData IPersistableModel<T>.Write(ModelReaderWriterOptions options)
{
    // get the format

    switch (format)
    {
        case "J":
            // write Json part
        case "X":
            // write Xml part
        case "other":
            // if we have other formats to support, they show up here
        default:
            // this always throws
    }
}
```

Therefore the builder for a specific format, should build a `SwitchCase` for `Write` and `Create`.

`SerializationMethodsBuilder` is the builder a `TypeProvider` should use.

Please do not worry if we have to write guards on these methods so that `JsonSerializationMethodsBuilder` will not be used anywhere else other than in `SerializationMethodsBuilder`. It is impossible to prevent every possibility of misuse of code especially at a cost of reducing readability.

Also I made the `BuildSerialization` method private only for `SerializableObjectType`.

ArcturusZhang: About the question of who should use which.

I am trying to establish a standard that, the `TypeProvider` should only use `SerializationMethodsBuilder`.
And all those builders for specific format, such as `JsonSerializationMethodsBuilder` (short for json builder) and xml builder, should only be called in `SerializationMethodsBuilder`.

There are not such accessibility modifiers to mark those classes are only available for one class, unless we made them private classes and putting them inside the `SerializationMethodsBuilder` class. I could do that, but I do not really like the idea.

I really think we should not worry too much if people will use things wrong. It is impossible to avoid all possibilities to use things wrong, especially when it has a cost of readability or other cost.

live1206: Actually, this is not about wrong usage, since the same functionality only exists in one place.
Since this refactoring is trying to split the logic into different classes, I am trying to understand what is the standard. Next time, we add a new serialization functionality, where it should belong to. Let's sync about this offline.

chunyu3: I think we can define each `XXXSerialilzationMethodsBuilder ` for each format, such as `JsonSerializationMethodsBuilder` for json format, `XmlSerializationMethodsBuilder` for xml format, and they inheritance from the base class `SerializationMethodsBuilder` which will contain common logics. In the future, if we need to support a new format, just add a new `XXXSerializationMethodsBuilder`

chunyu3:
LGTM, just one suggestion about the MethodsBuilder classes hierarchy.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: > I think we can define each `XXXSerialilzationMethodsBuilder ` for each format, such as `JsonSerializationMethodsBuilder` for json format, `XmlSerializationMethodsBuilder` for xml format, and they inheritance from the base class `SerializationMethodsBuilder` which will contain common logics. In the future, if we need to support a new format, just add a new `XXXSerializationMethodsBuilder`

this is the thing this PR is trying to do.

ArcturusZhang: Unfortunately since some methods in `JsonSerializationMethodsBuilder` are used somewhere else, therefore we cannot refactor our code to do this:
`XXXSerializationMethodsBuilder` classes are only used in `SerializationMethodsBuilder`.

We should do it but since we cannot do that right now, I will keep it as is and let future refactors do that.

Issue #4193, opened 2024-02-06
[Bug] Serialization of nested models are not entirely correct


ArcturusZhang:
**Describe the issue or request**

When a model contains a property of another model, this is how we are doing to write it:
https://github.com/Azure/autorest.csharp/blob/d72fed646c36e24cc899f499954ac1eef6f215f9/test/TestProjects/FirstTest-TypeSpec/src/Generated/Models/RoundTripModel.Serialization.cs#L244
This is using `writer.WriteObjectValue` which is an extension method:
https://github.com/Azure/autorest.csharp/blob/d72fed646c36e24cc899f499954ac1eef6f215f9/src/assets/Azure.Core.Shared/Utf8JsonWriterExtensions.cs#L71
But this method does not take an `options` parameter, therefore the option is missing when serializing this (and possible nested models this would also have)
We should fix this.

There are a few approaches of doing this:
1. Update the method in shared code to make it accept an optional options
2. Update our generator code, to make our generator know of when a case of model happens, it uses `ModelReaderWriter.Write` and then writes the `BinaryData` into the writer. For other cases, we still use `writer.WriteObjectValue` because this method is meant to be writing values with unknown types (`object`)

In fact, approach 1 might not be acceptable because shared code locates in `Azure.Core`, adding a new parameter creates a dependency to `System.ClientModel` in our shared code - which would be a nightmare because there is no dependency of System.ClientModel for the libraries that are not enabled with MRW.

live1206:
We are adding `System.ClientModel` as package reference to `Azure.Core` in https://github.com/Azure/azure-sdk-for-net/pull/41655/files#diff-e9213ad79eedc5d615c9c6f35698e02f01cb9644614ee8cdf2eed9f62f4d12e4 .
So, maybe option 1 is not acceptable?

ArcturusZhang:
We need to resolve https://github.com/Azure/autorest.csharp/issues/4246 first before we resolve this issue

ArcturusZhang:
Fixed in https://github.com/Azure/autorest.csharp/pull/4447

PR #4196, opened 2024-02-06
Fix nested model serialization issue

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4193

# Description

In Json serializations, when using `ModelReaderWriter.Write` on a model, it will call the `IJsonModel<T>.Write` method with the passed in `options`.
In our previous implementation (details in the above issue), on a property with a type of generated model, we will call `writer.WriteObjectValue`. This method switches the type of the model, and calls the corresponding method. In this case, it calls the `IUtf8JsonSerializable.Write` method, but this method, per our current implementation, will use the `W` options, instead of passing the `options` along the way.

This will introduce issues when some one serializing a model using `J` format because the format is not passed along.

# Checklist

# Ready to Land?


samples/AnomalyDetector/src/Generated/Models/AnomalyDetectionModel.Serialization.cs
```diff
@@ -39,7 +39,7 @@ void IJsonModel<AnomalyDetectionModel>.Write(Utf8JsonWriter writer, ModelReaderW
             if (Optional.IsDefined(ModelInfo))
             {
                 writer.WritePropertyName("modelInfo"u8);
-                writer.WriteObjectValue(ModelInfo);
+                ((IJsonModel<ModelInfo>)ModelInfo).Write(writer, options);
```
ArcturusZhang, line 42: there are three kinds of changes in this PR.
This PR only changes the code when the type of the property or the element of collection is a generated model (therefore it should implement the `IJsonModel<T>` interface)
This is the first kind: optional property.
We do not need the null check because `Optional.IsDefined` already did that

samples/AnomalyDetector/src/Generated/Models/AnomalyState.Serialization.cs
```diff
@@ -32,15 +32,22 @@ void IJsonModel<AnomalyState>.Write(Utf8JsonWriter writer, ModelReaderWriterOpti
...
-                    writer.WriteObjectValue(item);
+                    if (item != null)
+                    {
+                        ((IJsonModel<ErrorResponse>)item).Write(writer, options);
+                    }
+                    else
+                    {
+                        writer.WriteNullValue();
+                    }
                 }
```
ArcturusZhang, line 51: this is the second kind of the change: for elements of collections
We need to check null here because there is a chance that the value is null (csharp language allows that)

samples/AnomalyDetector/src/Generated/Models/MultivariateDetectionResult.Serialization.cs
```diff
@@ -33,12 +33,26 @@ void IJsonModel<MultivariateDetectionResult>.Write(Utf8JsonWriter writer, ModelR
...
             writer.WritePropertyName("summary"u8);
-            writer.WriteObjectValue(Summary);
+            if (Summary != null)
+            {
+                ((IJsonModel<MultivariateBatchDetectionResultSummary>)Summary).Write(writer, options);
+            }
+            else
+            {
+                writer.WriteNullValue();
+            }
```
ArcturusZhang, line 43: this is the third kind of the changes: required property.
Despite when constructing the model using public ctor, this property might not be allowed to be null, we have other ways to construct a model, in which almost any value is allowed for a property as long as csharp language allows, for example, the model factory. Using the model factory, you could construct an instance of a model with all of its properties to be null, no matter it is required or optional.
The null check here is required otherwise if we have an instance constructed by model factory, this serialization will throw NRE.

samples/AnomalyDetector/src/Generated/Models/MultivariateDetectionResult.Serialization.cs
```diff
@@ -101,7 +115,14 @@ internal static MultivariateDetectionResult DeserializeMultivariateDetectionResu
...
                     {
-                        array.Add(AnomalyState.DeserializeAnomalyState(item));
+                        if (item.ValueKind == JsonValueKind.Null)
+                        {
+                            array.Add(null);
+                        }
+                        else
+                        {
+                            array.Add(AnomalyState.DeserializeAnomalyState(item));
+                        }
```
ArcturusZhang, line 125: this is a change coming from symmetry between serialization and deserialization.
Before this change, we also add null - the `AnomalyState.DeserializeAnomalyState` method will return null value if `item.ValueKind` is null, this change only takes that part out here.

PR #4217, opened 2024-02-12
Support parameter overrides in bicep serialization

JoshLove-msft:


test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj
```diff
@@ -5,7 +5,7 @@
     <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net461</TargetFrameworks>
     <IsPackable>false</IsPackable>
     <Nullable>annotations</Nullable>
-    <NoWarn>SA1649;SA1633</NoWarn>
+    <NoWarn>SA1649;SA1633;NU1605</NoWarn>
```
m-nash, line 8: can we add this in a common place so that when we remove it we don't miss places?

JoshLove-msft: Not really a great common place as there is one src project and one test project so there would still need to be two conditions added into Directory.Build.Props. Also, we will be able to see the diff when merging back to feature/v3.

PR #4219, opened 2024-02-14
Fix Unknown variant deserialization issue in MRW

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4220

# Description

The unknown variant is working as the "deserialization proxy" of the base model (which is usually abstract).
Therefore in previous implementation, every model `Foo` should implement the interfaces with itself as generic arguments, such as:
```csharp
public partial class Foo : IJsonModel<Foo>
```
as a proxy, the unknown implements interfaces with its base model as generic arguments:
```csharp
public partial class BaseModel : IJsonModel<BaseModel> {}
internal partial class UnknownBaseModel : IJsonModel<BaseModel> {} // instead of IJsonModel<UnknownBaseModel>
```
This is the behavior we already implemented in previous MRW features.

There is a mistake, despite the interface is using `BaseModel` as the generic parameter, the deserialization implementation is calling the Deserialize**UnknownBaseModel** method instead of `DeserializeBaseModel` method.

This PR changes this part to let it use `DeserializeBaseModel`  method instead.

This is how MRW works in brief:
When we call `ModelReaderWriter.Read<T>(BinaryData, ModelReaderWriterOptions)` method, it returns you a new instance of `T` with the deserialized data from the `BinaryData`.
Here `T` inherits `IPersistableModel<T>`. To deserialize, we are creating an instance, but the deserialization method defined in an interface which requires an instance to call. dotnet 7 introduces interfaces to contain abstract static method, but MRW library must compile to netstandard 2.0 therefore we cannot use this feature and the deserialize method on the interface must be an instance method.
The `ModelReaderWriter.Read` method will first use the `Activator` to create an instance of type `T`. and then call the `Create` method on the created instance, which returns you the final result. The instance created by the `Activator` is only used to call the method, it will never be returned, and the returned instance is not the same instance with its properties changed.
When the type `T` is abstract, to create an instance of `T` will throw exceptions, therefore we introduced the `[PersistableModelProxy(Type)]` attribute. The generator will always add this attribute to the base model with a discriminator because this is the only abstract type we generate.
If we have this attribute on type `T`, the `ModelReaderWriter` will use the type specified in the attribute to create an instance with the `Activator` and we will not have the exception.

When we call `ModelReaderWriter.Read` on the base model with a json value, we are actually expecting to get a derived type depending on the discriminator value. For example, if we have this definition in tsp:
```
@discriminator("kind")
model Base
{
kind: string;
}

model ModelX : Base
{
kind: "X";
}

model ModelY : Base
{
kind: "Y";
}
```
And when we deserialize JSON `{ "kind": "X" }`, the method should return an instance of `ModelX`, and when we deserialize JSON `{ "kind": "Y" }`, the method should return an instance of `ModelY`, and when we deserialize a JSON with an unknown discriminator value, such as `{ "kind": "Z" }`, the method should return an instance of `UnknownBase`.

Some tests fail because we fix this because in the deserialization we are not passing the options along. If we fix the entire issue in this PR, it introduces 5000+ changed files in the regen, therefore I decided that we separate the fix in two steps, the first step is this PR, whose regen has ~400 files changed which is relatively a less amount of files. The second step is [here](https://github.com/Azure/autorest.csharp/pull/4232) which makes sure that we pass the options along the nested models during deserialization. The second PR will inevitably have 5000+ changed files but they will have the same pattern.

# Checklist

# Ready to Land?


test/TestProjects/ModelReaderWriterValidation-TypeSpec/src/Generated/Models/UnknownBaseModel.Serialization.cs
```diff
@@ -61,7 +61,7 @@ BaseModel IJsonModel<BaseModel>.Create(ref Utf8JsonReader reader, ModelReaderWri
             }
 
             using JsonDocument document = JsonDocument.ParseValue(ref reader);
-            return DeserializeUnknownBaseModel(document.RootElement, options);
+            return DeserializeBaseModel(document.RootElement, options);
```
ArcturusZhang, line 64: Here we fix the issue #4220

ArcturusZhang: calling this method instead of changing the implementation of `DeserializeUnknownBaseModel` because of the following reasons:
1. We will have a pattern that if we have `T IJsonModel<T>.Create(ref Utf8JsonReader reader, ModelReaderWriterOptions options)`, we call `DeserializeT`.
2. `DeserializeModel` should have single responsibility: deserialize the json element and return an instance of myself, therefore `DeserializeUnknownBaseModel` should only return an instance of `UnknownBaseModel`

ArcturusZhang:
@m-nash please have a review

test/AutoRest.TestServer.Tests/ModelReaderWriterValidation/BaseModelTests.cs
```diff
@@ -8,54 +8,55 @@
...
 namespace AutoRest.TestServer.Tests
 {
-    internal class BaseModelTests : ModelJsonTests<BaseModel>
-    {
-        protected override BaseModel GetModelInstance()
-        {
-            var typeToActivate = typeof(BaseModel).Assembly.GetTypes().FirstOrDefault(t => t.Name == $"Unknown{typeof(BaseModel).Name}") ?? throw new InvalidOperationException("Unable to find BaseModel type");
-            return Activator.CreateInstance(typeToActivate, true) as BaseModel ?? throw new InvalidOperationException($"Unable to create model instance of BaseModel");
-        }
+    // TODO -- uncomment when https://github.com/Azure/autorest.csharp/issues/4221 is resolved
```
ArcturusZhang, line 11: we have to remove this entire test case because it needs the fix in [this PR](https://github.com/Azure/autorest.csharp/pull/4232) to work properly.
We will enable these test cases and add more

lirenhe (approved):

test/TestProjects/ModelReaderWriterValidation-TypeSpec/src/Generated/Models/UnknownBaseModel.Serialization.cs
m-nash: This still doesn't quite match what we did in System.ClientModel.

There we had IJsonModel.Create from the unknown type call DeserializeUnknown.  This method always forwards to DeserializeBase.

There is no implementation for DeserializeUnknown because you always need the logic on the base to run to pick which type to create.

Is there a reason we are leaving this implementation here?  If its because the default case on the DeserializeBase is supposed to call this that seems very strange that we are bouncing back and forth between the two classes.

m-nash: Lets open a tracking issue to discuss further.

m-nash (approved):

test/TestProjects/ModelReaderWriterValidation-TypeSpec/src/Generated/Models/UnknownBaseModel.Serialization.cs
ArcturusZhang: > Lets open a tracking issue to discuss further.

created: https://github.com/Azure/autorest.csharp/issues/4254

PR #4252, opened 2024-02-21
Don't serialize empty objects

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/42018


src/AutoRest.CSharp/Common/Output/Builders/BicepSerializationMethodsBuilder.cs
```diff
@@ -336,25 +343,27 @@ private static IEnumerable<MethodBodyStatement> WriteAppendChildObject()
...
             StringBuilderExpression stringBuilder,
+            string propertyName,
             BicepArraySerialization arraySerialization,
             EnumerableExpression array,
             int spaces)
         {
             string indent = new string(' ', spaces);
             return new[]
             {
-                stringBuilder.AppendLine(" ["),
```
JoshLove-msft, line 346: all these single spaces are now added when writing property name

m-nash (approved):

PR #4267, opened 2024-02-22
Add flags to split HLC codeflow into individual features

AlexanderSher:
Flags are for the output model only, so no changes in the config files for now.
"generation1-convenience-client" flag will be fully replaced during the final stage of consolidation


src/AutoRest.CSharp/Common/Input/Configuration.cs
```diff
@@ -264,6 +270,11 @@ internal static (string AbsoluteProjectFolder, string RelativeProjectFolder) Par
         public static bool SkipSerializationFormatXml { get; private set; }
         public static bool DisablePaginationTopRenaming { get; private set; }
 
+        public static bool AssertModelConstructorParametersForReadOnlyProperties { get; private set; }
+        public static bool PublicRestClients { get; private set; }
+        public static bool GenerateLongRunningOperationTypes { get; private set; }
+        public static bool GenerateResponseHeadersModels { get; private set; }
```
m-nash, line 276: nit: `GenerateResponseHeaderModels`

src/AutoRest.CSharp/Common/Input/Configuration.cs
```diff
@@ -264,6 +270,11 @@ internal static (string AbsoluteProjectFolder, string RelativeProjectFolder) Par
         public static bool SkipSerializationFormatXml { get; private set; }
         public static bool DisablePaginationTopRenaming { get; private set; }
 
+        public static bool AssertModelConstructorParametersForReadOnlyProperties { get; private set; }
+        public static bool PublicRestClients { get; private set; }
```
m-nash, line 274: I am wondering if we don't do this and simple add custom code to make it internal in the HLC libraries that will need this.

Proportionally this feels like a tiny % of the total libraires we have will want this.

src/AutoRest.CSharp/Common/Input/Configuration.cs
```diff
@@ -264,6 +270,11 @@ internal static (string AbsoluteProjectFolder, string RelativeProjectFolder) Par
         public static bool SkipSerializationFormatXml { get; private set; }
         public static bool DisablePaginationTopRenaming { get; private set; }
 
+        public static bool AssertModelConstructorParametersForReadOnlyProperties { get; private set; }
```
m-nash, line 273: What happens if we stop asserting here for the libraries that currently do?  This doesn't feel like a breaking change.  I wonder if we should choose a single best practice here and use it everywhere.

AlexanderSher: Consolidated. Flag is not needed anymore.

m-nash (approved):

PR #4306, opened 2024-02-29
Remove generator dependency on `RequestContentHelper`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4301
# Description

# Checklist

# Ready to Land?


test/TestProjects/Lro-TypeSpec/Lro-TypeSpec.tsp
```diff
@@ -27,7 +27,7 @@ using Azure.Core.Traits;
     OAuth2Auth<[AuthFlow]> | ApiKeyAuth<ApiKeyLocation.header, "x-ms-api-key">
 )
 @useDependency(Azure.Core.Versions.v1_0_Preview_1)
-namespace lro;
+namespace LroInTypeSpec;
```
ArcturusZhang, line 30: we changed the namespace here, because in test server test projects, there is another test project with the generated namespace of `lro`.
And when the pipeline is building the test projects to run test cases, they are compiled together, therefore since they both have a partial class `RequestContentHelper`, they somehow compile into one class, and this class has same methods defined in the same class, which leads to a compilation error.

The simplest fix here is to change the namespace of one of the test projects.

This issue only happens in our test projects, the real generated libraries never have this issue.

test/AutoRest.Shared.Tests/RequestContentHelperTests.cs
```diff
@@ -1,149 +0,0 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Text.Json;
-using System.Xml;
-using System.Xml.Linq;
-using NUnit.Framework;
-
-namespace Azure.Core.Tests
```
m-nash, line 9: Lets not remove these tests, lets just pull in one instance of the internal class to use as the test source.

This should be happening for you with this https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.Shared.Tests/AutoRest.Shared.Tests.csproj#L23.

We definitely want to unit tests the generated classes.

m-nash (approved):
Feel free to merge after fixing the test comment

test/AutoRest.Shared.Tests/RequestContentHelperTests.cs
ArcturusZhang: introduced back this test case

PR #4308, opened 2024-03-01
Remove shared source dependency on ChangeTrackingList

m-nash:
# Description
Fixes https://github.com/Azure/autorest.csharp/issues/4321

# Checklist

# Ready to Land?


JoshLove-msft (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
```diff
@@ -12,29 +13,52 @@ namespace AutoRest.CSharp.Output.Models.Types
...
+        protected ExpressionTypeProvider(string defaultNamespace, SourceInputModel? sourceInputModel)
+            : base(defaultNamespace, sourceInputModel)
         {
+            DeclarationModifiers = ClassSignatureModifiers.Partial | ClassSignatureModifiers.Public;
         }
 
         private IReadOnlyList<string>? _usings;
         public IReadOnlyList<string> Usings => _usings ??= BuildUsings().ToArray();
 
+        public ClassSignatureModifiers DeclarationModifiers { get; init; }
```
ArcturusZhang, line 25: I think we should make it `protected init;` because we never want to accidentally assign the value of this when new an derived type of this

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
```diff
@@ -12,29 +13,52 @@ namespace AutoRest.CSharp.Output.Models.Types
     // TODO -- eventually we should combine everything in this class into TypeProvider
     internal abstract class ExpressionTypeProvider : TypeProvider
     {
-        protected ExpressionTypeProvider(string defaultNamespace, SourceInputModel? sourceInputModel) : base(defaultNamespace, sourceInputModel)
+        protected ExpressionTypeProvider(string defaultNamespace, SourceInputModel? sourceInputModel)
+            : base(defaultNamespace, sourceInputModel)
         {
+            DeclarationModifiers = ClassSignatureModifiers.Partial | ClassSignatureModifiers.Public;
```
ArcturusZhang, line 19: I like this idea, but I think we should call this `TypeDeclarationModifier` or `TypeSignatureModifier` since it might apply to struct and enum as well

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
```diff
@@ -12,29 +13,52 @@ namespace AutoRest.CSharp.Output.Models.Types
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override string DefaultAccessibility { get; } = "public";
+
         public virtual CSharpType? Inherits { get; protected init; }
 
         private IReadOnlyList<CSharpType>? _implements;
         public virtual IReadOnlyList<CSharpType> Implements => _implements ??= BuildImplements().ToArray();
 
+        private IReadOnlyList<PropertyDeclaration>? _properties;
+        public virtual IReadOnlyList<PropertyDeclaration> Properties => _properties ??= BuildProperties().ToArray();
```
ArcturusZhang, line 41: I think in the near future we might need to let `PropertyDeclaration` and `FieldDeclaration` to inherit the same base type such as `MemberDeclaration` because in a lot of cases properties and fields are alike and could be replaced with each other - for instance the `AdditionalProperties` property and the `_serializedAdditionalRawData` field - one exists and the other disappears
I will create an issue for this, and we might need a discussion on how we define a TypeProvider when migrating these to the new generator

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
ArcturusZhang: Issue here: https://github.com/Azure/autorest.csharp/issues/4311

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
ArcturusZhang: Issue here: https://github.com/Azure/autorest.csharp/issues/4312

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+
+namespace AutoRest.CSharp.Common.Output.Models.Types
+{
+    internal class ChangeTrackingListProvider : ExpressionTypeProvider
+    {
+        private static readonly Lazy<ChangeTrackingListProvider> _instance = new(() => new ChangeTrackingListProvider(Configuration.Namespace, null));
+        private readonly MethodSignature _ensureListSignature = new MethodSignature("EnsureList", null, null, MethodSignatureModifiers.Public, typeof(IList<>), null, Array.Empty<Parameter>());
+        private readonly MethodSignature _getEnumeratorSignature = new MethodSignature("GetEnumerator", null, null, MethodSignatureModifiers.Public, typeof(IEnumerator<>), null, Array.Empty<Parameter>());
+        private readonly TypeProvider _t;
+        private readonly FieldDeclaration _innerlListField;
```
ArcturusZhang, line 26: here is a typo
```suggestion
        private readonly FieldDeclaration _innerListField;
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        protected override string DefaultName => "ChangeTrackingList";
+
+        protected override IEnumerable<Method> BuildConstructors()
+        {
+            yield return new Method(new ConstructorSignature(Type, null, null, MethodSignatureModifiers.Public, Array.Empty<Parameter>()), Snippets.EmptyStatement);
+            var iListParam = new Parameter("innerList", null, typeof(IList<>), null, ValidationType.None, null);
+            var iListSignature = new ConstructorSignature(Type, null, null, MethodSignatureModifiers.Public, new Parameter[] { iListParam });
+            var iListVariable = new ParameterReference(iListParam);
+            var iListbody = new MethodBodyStatement[]
```
ArcturusZhang, line 53: typo here, should be `iListBody`

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+            {
+                new IfStatement(Snippets.NotEqual(iListVariable, Snippets.Null))
+                {
+                    new AssignValueStatement(_innerList, iListVariable)
```
ArcturusZhang, line 57: There is a method that we could do this in a more semantic way
```suggestion
                    Assign(_innerList, iListVariable)
```
you need to using static Snippets; <- VS should have intellisence to automatically complete this because I do not remember the full namespace here

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        protected override IEnumerable<CSharpType> BuildTypeArguments()
+        {
+            yield return _t.Type;
+        }
+
+        protected override IEnumerable<CSharpType> BuildImplements()
+        {
+            yield return new CSharpType(typeof(IList<>));
```
ArcturusZhang, line 83: since this interface should use the same argument of the generic parameter of this type itself, I suggest we change it this way:

```suggestion
            yield return new CSharpType(typeof(IList<>), _t.Type);
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        public static ChangeTrackingListProvider Instance => _instance.Value;
+
+        private ChangeTrackingListProvider(string defaultNamespace, SourceInputModel? sourceInputModel)
+            : base(defaultNamespace, sourceInputModel)
+        {
+            _innerlListField = new FieldDeclaration(FieldModifiers.Private, typeof(IList<>), "_innerList");
+            _innerList = new VariableReference(new CSharpType(typeof(IList<>)), _innerlListField.Declaration);
+            _t = new GenericParameterTypeProvider("T", DefaultNamespace, null);
+            _tArray = new GenericParameterTypeProvider("T[]", DefaultNamespace, null);
```
ArcturusZhang, line 40: I think that we never use the `TypeProvider` of `T` and `T[]` - could we change the type of `_t` and `_tArray` to `CSharpType` and here we could have:
```suggestion
            _t = new GenericParameterTypeProvider("T", DefaultNamespace, null).Type;
            _tArray = new GenericParameterTypeProvider("T[]", DefaultNamespace, null).Type;
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+        {
+            yield return _innerlListField;
+        }
+
+        protected override IEnumerable<PropertyDeclaration> BuildProperties()
+        {
+            yield return new PropertyDeclaration(null, MethodSignatureModifiers.Public, typeof(bool), "IsUndefined", new ExpressionPropertyBody(Snippets.Equal(_innerList, Snippets.Null)));
+            yield return BuildCount();
+            yield return BuildIsReadOnly();
+            yield return BuildIndexer();
```
ArcturusZhang, line 97: I think we should make these properties as fields or lazy loaded properties of this class because we are using them in other places, for example, the property `IsUndefined` is used in a lot of methods and other properties.
So that we could just reference the `PropertyDeclaration` instance instead of hard-coding a lot of strings (like `IsUndefined`) everywhere, we only need to put the hard-code string in one place.

In the meantime, we have an implicit operator from `FieldDeclartion` to `ValueExpression`, therefore I think we could have one for `PropertyDeclaration` as well.

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+                    Snippets.Literal(0),
+                    new MemberExpression(new InvokeInstanceMethodExpression(Snippets.This, _ensureListSignature), "Count"))));
+        }
+
+        private PropertyDeclaration BuildIndexer()
+        {
+            return new PropertyDeclaration(null, MethodSignatureModifiers.Public, _t.Type, "this", new MethodPropertyBody(
+                new MethodBodyStatement[]
+                {
+                    new IfStatement(new BoolExpression(new MemberExpression(Snippets.This, "IsUndefined")))
```
ArcturusZhang, line 123: there is a possible refactor here to improve the readability here:
```suggestion
                    new IfStatement(new BoolExpression(IsUndefinedProperty))
```
if we define this `IsUndefined` property as a property of `PropertyDeclaration` in this class, and we should add an implicit cast operator from `PropertyDeclaration` to `ValueExpression` as we already have for `FieldDeclaration`.

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+                    {
+                        Snippets.Throw(Snippets.New.Instance(typeof(ArgumentOutOfRangeException), Snippets.Nameof(new ParameterReference(_indexParam))))
```
ArcturusZhang, line 125: Usually we add the `using static Snippets` using statement to improve the readability of method body statements

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+                    },
+                    Snippets.Return(new ArrayElementExpression(new InvokeInstanceMethodExpression(Snippets.This, _ensureListSignature), new ParameterReference(_indexParam))),
```
ArcturusZhang, line 127: Also to improve readability:

```suggestion
                    Return(new ArrayElementExpression(This.Invoke(_ensureListSignature), _indexParam)),
```

src/AutoRest.CSharp/Common/Output/Models/Types/GenericParameterTypeProvider.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright(c) Microsoft Corporation.All rights reserved.
+// Licensed under the MIT License.
+
+using AutoRest.CSharp.Input.Source;
+
+namespace AutoRest.CSharp.Output.Models.Types
+{
+    internal class GenericParameterTypeProvider : TypeProvider
+    {
+        public GenericParameterTypeProvider(string name, string defaultNamespace, SourceInputModel? sourceInputModel)
```
ArcturusZhang, line 10: I am curious, if we define the type in this way, it should have a namespace.
But as an open generic parameter such as `T`, it should not have namespaces.
I am not sure if this could be an issue.

src/AutoRest.CSharp/Common/Output/Models/ClassSignatureModifiers.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace AutoRest.CSharp.Common.Output.Models
+{
+    [Flags]
+    internal enum ClassSignatureModifiers
```
ArcturusZhang, line 9: As I mention in an above comment, I think we should call this as `TypeDeclarationModifiers` or `TypeSignatureModifiers`

src/AutoRest.CSharp/Common/Output/Expressions/Statements/DeclareLocalFunctionStatement.cs
```diff
@@ -9,5 +9,12 @@
 
 namespace AutoRest.CSharp.Common.Output.Expressions.Statements
 {
-    internal record DeclareLocalFunctionStatement(CodeWriterDeclaration Name, IReadOnlyList<Parameter> Parameters, CSharpType ReturnType, ValueExpression Body) : DeclarationStatement;
+    internal record DeclareLocalFunctionStatement(CodeWriterDeclaration Name, IReadOnlyList<Parameter> Parameters, CSharpType ReturnType, ValueExpression? BodyExpression, MethodBodyStatement? BodyStatement) : DeclarationStatement
```
ArcturusZhang, line 12: this syntax also creates a ctor that could take both `BodyExpression` and `BodyStatement`, we should not have this ctor.

src/AutoRest.CSharp/Common/Output/Expressions/Statements/DeclareFieldStatement.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
+
+namespace AutoRest.CSharp.Common.Output.Expressions.Statements
+{
+    internal record DeclareFieldStatement(VariableReference Variable) : MethodBodyStatement;
```
ArcturusZhang, line 8: I did not find a place where we are using this, maybe we could remove it?
Also I think we already have a similar statement, we do not have to have a declaration statement specifically for fields.

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -470,7 +471,7 @@ private static void WriteQueryParameter(CodeWriter writer, CodeWriterDeclaration
 
                 writer.Append($" != null && !(");
                 writer.WriteValueExpression(valueExpressoin);
```
ArcturusZhang, line 473: maybe we could also fix the typo here BTW

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -298,5 +298,17 @@ public bool EqualsByName(CSharpType? other)
 
             return true;
         }
+
+        internal CSharpType WithArguments(IReadOnlyList<CSharpType> arguments)
```
ArcturusZhang, line 302: I think this should be `public`, and also could we name it as `MakeGenericType` to mimic the same behavior in `System.Type`? I think this method serves the same purpose to replace the open generic parameter to a concrete type, such as create `IList<int>` out of `IList<T>`

src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -702,13 +702,24 @@ public static void WriteProperty(this CodeWriter writer, PropertyDeclaration pro
                 .AppendRawIf("static ", modifiers.HasFlag(MethodSignatureModifiers.Static))
                 .AppendRawIf("virtual ", modifiers.HasFlag(MethodSignatureModifiers.Virtual)); // property does not support other modifiers, here we just ignore them if any
 
-            writer.Append($"{property.PropertyType} {property.Declaration:I}"); // the declaration order here is quite anonying - we might need to assign the values to those properties in other places before these are written
+
+
```
ArcturusZhang, line 706: we should reduce the empty lines here

src/AutoRest.CSharp/Common/Output/Expressions/Snippets.Linq.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Linq;
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
+
+namespace AutoRest.CSharp.Common.Output.Models
+{
+    internal static partial class Snippets
+    {
+        public static class Linq
+        {
+            public static ValueExpression ToList(ValueExpression expression)
```
ArcturusZhang, line 13: since linq methods are all extension methods, I think we could define our expressions as extension methods here as well, and this should only applies to `EnumerableExpression`, therefore
```suggestion
            public static EnumerableExpression ToList(this EnumerableExpression expression)
```

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
m-nash: I prefer to keep these separate.  Right now field declaration has a lot of stuff like write as property which we will remove in the new generator.

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
m-nash: I will try this, right now the writer automatically adds `T` but to your point if I wanted to call it TMyName then we would want that to flow through.

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
m-nash: The generic function does not like this for some reason.  New AssignValueStatement seems to work fine though.

src/AutoRest.CSharp/Common/Output/Expressions/Statements/DeclareLocalFunctionStatement.cs
m-nash: The idea is one or the other will not be null

src/AutoRest.CSharp/Common/Output/Expressions/Snippets.Linq.cs
m-nash: the `expression` being passed in here is a parameter reference which doesn't need to be enumerable.  Not sure we can make this change

src/AutoRest.CSharp/Common/Output/Expressions/Statements/DeclareLocalFunctionStatement.cs
ArcturusZhang: Well this ctor never prevent the user to give values to both statement and body. The ctors of this type should look like this:
```csharp
public DeclarreLocalFunctionStatement(Name, Parameters, ReturnType, BodyExpression) : this(Name, Parameters, ReturnType, BodyExpression, null)
public DeclarreLocalFunctionStatement(Name, Parameters, ReturnType, BodyStatement) : this(Name, Parameters, ReturnType, null, BodyStatement)
private DeclarreLocalFunctionStatement(Name, Parameters, ReturnType, BodyExpression?, BodyStatement?)
```
We did this in `Method` class, therefore I believe we should do the same here.

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
ArcturusZhang: I think we forgot to change this?

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ChangeTrackingListProvider.cs
ArcturusZhang: Well just because their type is different and we should do this:
```suggestion
                    Assign<ValueExpression>(_innerList, iListVariable)
```
or we could change the type of field `_innerList` to `ValueExpression` because usually we do not need an explicit `VariableReference`.

src/AutoRest.CSharp/Common/Output/Expressions/Statements/DeclareLocalFunctionStatement.cs
m-nash: I don't think this works on a record.  Once we switch to classes we can do this but I don't want to take that change in this PR.

Issue #4209, opened 2024-02-08
Need to handle TypeSpec union type with extensible enums and enum as closed enumeration..


lirenhe:
Please follow https://github.com/Azure/typespec-azure-pr/issues/3827.

The latest TCGC 0.39.0 should already contain this change but considering dotnet codegen not fully apply that version, we need to figure out how to support this.

PR to update https://github.com/Azure/azure-rest-api-specs/pull/27687/files already created, so this is likely to cause breaking change if this work is not finished soon.

PR #4315, opened 2024-03-02
Support open union to extensible enum

pshao25:
First step of #4209.

1. Generate extensible enum for open union.
2. Change all the existing enum in spec to open union.
3. Give enum a description if it does not exist in the spec.

Some float type is not correct now. See details in https://github.com/Azure/typespec-azure/issues/353.

Still naked num after this change: versions in the `@versioned`


src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
```diff
@@ -59,10 +60,9 @@ public static InputEnumType CreateEnumType(ref Utf8JsonReader reader, string? id
             name = name ?? throw new JsonException("Enum must have name");
             // TODO: roll back to throw JSON error when there is linter on the upstream to check enum without @doc
             //description = description ?? throw new JsonException("Enum must have a description");
-            if (description == null)
+            if (description.IsNullOrEmpty())
             {
-                description = "";
-                Console.Error.WriteLine($"[Warn]: Enum '{name}' must have a description");
```
chunyu3, line 65: We need to give out a message for null or empty description.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -805,6 +801,11 @@ export function getInputType(
     }
 
     function getInputTypeForUnion(union: Union): InputUnionType | InputType {
+        var clientType = getClientType(context, union);
+        if (clientType.kind === "enum" && clientType.isFixed === false) {
```
chunyu3, line 805: Why does we check `isFixed` here? Can it be a Fixed enum if it defined as Union? If so, shall we emit fixed enum and do not ignore?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -805,6 +801,11 @@ export function getInputType(
     }
 
     function getInputTypeForUnion(union: Union): InputUnionType | InputType {
+        var clientType = getClientType(context, union);
```
chunyu3, line 804: Does `getClientType` handle all kinds of Union, and convert it to correct type such as enum, union or nullable type? If so, maybe we just need to call `getClientType` and remove duplicate handling below.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
pshao25: What message? I copy the behavior from swagger.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
pshao25: > Can it be a Fixed enum if it defined as Union

Yes.

> shall we emit fixed enum and do not ignore

Fixed enum is next step. We only handle extensible enum in this PR.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
pshao25: It is expected to. But currently it is only prepared for extensible enum, so extensible enum is the scope of this PR. When it is prepared for other types, we would integrate it.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
chunyu3: Why remove the Error log?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
chunyu3: I don't quite understand. So we can use Union to define a fixed enum, but now we will not support this feature, and return Union as type as before?

pshao25: Yes, if you define `"a"|"b"`, it is still union.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
pshao25: Added back.

chunyu3 (approved):

lirenhe (approved):

lirenhe:
This should solve the dotnet SDK generation error in https://github.com/Azure/azure-rest-api-specs/pull/27687
cc @lmazuel

Issue #4318, opened 2024-03-04
Emit static helpers


m-nash:
In order to maintain code readability and consistency with custom code we will emit the IsDefined and IsCollection helpers from Optional, but won't carry over the `To` methods or the struct.  Instead in serialization code will simply declare optional structs like `int?`

PR #4320, opened 2024-03-04
Emit helpers for Optional.IsDefined and Optional.IsCollectionDefnied

m-nash:
# Description
Fixes https://github.com/Azure/autorest.csharp/issues/4318

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Types/GenericParameterTypeProvider.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright(c) Microsoft Corporation.All rights reserved.
+// Licensed under the MIT License.
+
+using AutoRest.CSharp.Common.Input;
+using AutoRest.CSharp.Input.Source;
+
+namespace AutoRest.CSharp.Output.Models.Types
+{
+    internal class GenericParameterTypeProvider : TypeProvider
```
ArcturusZhang, line 9: I think we should make this sealed since this has its own special logic in `CodeWriter`

eng/scripts/preview/Download-SharedSource.ps1
```diff
@@ -65,7 +65,6 @@ $files = @(
     'OperationInternal.cs',
     'OperationInternalOfT.cs',
     'OperationPoller.cs',
-    'Optional.cs',
```
ArcturusZhang, line 68: we should also remove this file from the now-in-using-Download scripts, this script is a preview one.
And the file is not removed on your PR yet.

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneTarget.cs
```diff
@@ -25,6 +21,10 @@ public static void Execute(GeneratedCodeWorkspace project, CodeModel codeModel,
             var headerModelModelWriter = new DataPlaneResponseHeaderGroupWriter();
             var longRunningOperationWriter = new LongRunningOperationWriter();
 
+            var helperWriter = new CodeWriter();
+            new ExpressionTypeProviderWriter(helperWriter, OptionalTypeProvider.Instance).Write();
+            project.AddGeneratedFile($"Internal/{OptionalTypeProvider.Instance.Type.Name}.cs", helperWriter.ToString());
```
ArcturusZhang, line 26: I suggest we could add a method such as `AddHelperFile` on `GeneratedCodeWorkspace` so that we could do something different from usual generated files, for instance in there we could let the post-processor to ignore all helper files.

ArcturusZhang (approved):

ArcturusZhang (approved):
A few minor comments

src/AutoRest.CSharp/Common/Output/Models/Types/GenericParameterTypeProvider.cs
ArcturusZhang: Or we could always just get a generic parameter definition from the compiler so that we do not have to have this class.

eng/scripts/preview/Download-SharedSource.ps1
m-nash: It was removed in a previous pr.  Optional is not in either download script

src/AutoRest.CSharp/DataPlane/AutoRest/DataPlaneTarget.cs
m-nash: Having this in here lets the post processor omit any helpers that aren't actually used.

Issue #4316, opened 2024-03-04
Emit Argument helper into internal code


m-nash:
In order to ensure consistent error messages we should have helpers for asserting arguments instead of inlining.

PR #4324, opened 2024-03-05
Emit Argument as code

m-nash:
# Description
Fixes https://github.com/Azure/autorest.csharp/issues/4316

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -25,6 +25,8 @@ internal record Parameter(string Name, FormattableString? Description, CSharpTyp
         public CSharpAttribute[] Attributes { get; init; } = Array.Empty<CSharpAttribute>();
         public bool IsOptionalInSignature => DefaultValue != null;
 
+        public Parameter WithRef(bool isRef = true) => this with { IsRef = isRef };
```
ArcturusZhang, line 28: I think maybe we could check if isRef is the same as the input, if so we could return `this` as we did in `WithAsync` in method signature

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ArgumentProvider.cs
```diff
@@ -0,0 +1,285 @@
...
+                    ThrowArgumentException(NullCoalescing(new ParameterReference(messageParam), Literal("Value must be null.")))
+                }
+            });
+        }
+
+        private Method BuildCheckNotNullOrEmptyString()
+        {
+            var valueParam = new Parameter("value", null, typeof(string), null, ValidationType.None, null);
+            var signature = GetSignature("CheckNotNull", new[] { valueParam, _nameParam }, returnType: typeof(string));
+            var valueParamRef = new ParameterReference(valueParam);
```
ArcturusZhang, line 98: I would like to name this expression with the same name it is written, for instance here we could name this as `value`
```suggestion
            var value = new ParameterReference(valueParam);
```
and then when we are building the statements, the statements would look like the code we write.
```csharp
var value = new ParameterReference(valueParam);
var name = new ParameterReference(_nameParam);
return new Method(signature, new MethodBodyStatement[]
{
    AssertNotNullOrEmpty(value, name),
    Return(value)
});
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ArgumentProvider.cs
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        private Method BuildAssertNotNullOrWhiteSpace()
+        {
+            var valueParam = new Parameter("value", null, typeof(string), null, ValidationType.None, null);
+            var signature = GetSignature("AssertNotNullOrWhiteSpace", new[] { valueParam, _nameParam });
+            return new Method(signature, new MethodBodyStatement[]
+            {
+                AssertNotNullSnippet(valueParam),
+                new IfStatement(new BoolExpression(new InvokeStaticMethodExpression(typeof(string), "IsNullOrWhiteSpace", new[] { new ParameterReference(valueParam) })))
```
ArcturusZhang, line 187: I think we should add a `IsNullOrWhiteSpace` static method to `StringExpression` to introduce more readability here, and this method could return `BoolExpression` because `string.IsNullOrWhiteSpace` is returning `bool`.
```suggestion
                new IfStatement(StringExpression.IsNullOrWhiteSpace(value))
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ArgumentProvider.cs
```diff
@@ -0,0 +1,285 @@
...
+            var valueParam = new Parameter("value", null, typeof(string), null, ValidationType.None, null);
+            var signature = GetSignature(AssertNotNullOrEmptyMethodName, new[] { valueParam, _nameParam });
+            return new Method(signature, new MethodBodyStatement[]
+            {
+                AssertNotNullSnippet(valueParam),
+                new IfStatement(Equal(new MemberExpression(new ParameterReference(valueParam), "Length"), Literal(0)))
+                {
+                    ThrowArgumentException("Value cannot be an empty string.")
+                }
+            });
```
ArcturusZhang, line 205: I think we should define a `value` expression from the parameter first:
```suggestion
            var signature = GetSignature(AssertNotNullOrEmptyMethodName, new[] { valueParam, _nameParam });
            var value = new StringExpression(valueParam);
            return new Method(signature, new MethodBodyStatement[]
            {
                AssertNotNullSnippet(value),
                new IfStatement(Equal(value.Length, Literal(0)))
                {
                    ThrowArgumentException("Value cannot be an empty string.")
                }
            });
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ArgumentProvider.cs
```diff
@@ -0,0 +1,285 @@
...
+
+        private Method BuildAssertNotNullStruct()
+        {
+            var valueParam = new Parameter("value", null, _nullableT, null, ValidationType.None, null);
+            var signature = GetSignature(AssertNotNullMethodName, new[] { valueParam, _nameParam }, new[] { _t }, new Dictionary<CSharpType, FormattableString>() { { _t, $"struct" } });
+            return new Method(signature, new MethodBodyStatement[]
+            {
+                new IfStatement(Not(new BoolExpression(new MemberExpression(new ParameterReference(valueParam), "HasValue"))))
+                {
+                Throw(New.ArgumentNullException(_nameParamRef, false))
```
ArcturusZhang, line 252: here is an indent issue

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -93,6 +94,8 @@ public CSharpType(TypeProvider implementation, string ns, string name, bool isVa
         public TypeProvider Implementation => _implementation ?? throw new InvalidOperationException($"Not implemented type: '{Namespace}.{Name}'");
         public bool IsNullable { get; }
 
+        public WhereExpression IsType(CSharpType type) => new WhereExpression(this, type);
```
ArcturusZhang, line 97: I do not think this method belong to this class.
We should add it to `Snippets.Where` since it is building a where expression, for example:
```csharp
public WhereExpression Inherits(CSharpType type, CSharpType inherits)
```
or we could make a `params` and call it `Implements`

ArcturusZhang (approved):

ArcturusZhang:
I think it would be better if `Argument` is a partial class so that when our customer writing customized code, they could add things into `Argument` and use them

PR #4363, opened 2024-03-07
feat(configuration): replace `branded` with `flavor`

### LINKED ISSUE #4309: Change default generation flavor and flags for generic codegen
- Emitter should generate "unbranded" code by default
- Add a new flag `flavor` that generates azure code for the value `azure`. This is not case-sensitive. Any other value does nothing, and therefore generates unbranded code.
- Don't remove yet currently supported flag `branded` to help the transition to the new system. If both flags are set, read `flavor` first and ignore `branded` boolean flag.
- If `package-dir` starts with `azure`, auto-set `flavor` to `azure` if `flavor` was not set already. If `package-dir` is anything else, no specific behavior.

archerzz:
# Description

- replace `branded` flag with `flavor`
- update typespec test project configuration
- update `eng\Generate.ps1`
- regen test projects

resolve #4309

---

Some deviation from original requirement:
- remove `branded` flag instead of still keeping it, because the change will mandate change in SDK project configuration. If we have to update the configuration files, there is no need to still keep `branded`.
- deduce azure project by namespace instead of `package-dir` which is NOT used by csharp generator.

# Checklist

# Ready to Land?


archerzz:
Regen preview PR: https://github.com/Azure/azure-sdk-for-net/pull/42534

src/AutoRest.CSharp/Common/Input/Configuration.cs
```diff
@@ -174,7 +179,7 @@ public enum UnreferencedTypesHandlingOption
             _modelsToTreatEmptyStringAsNull = new HashSet<string>(modelsToTreatEmptyStringAsNull);
             _intrinsicTypesToTreatEmptyStringAsNull.UnionWith(additionalIntrinsicTypesToTreatEmptyStringAsNull);
             _methodsToKeepClientDefaultValue = methodsToKeepClientDefaultValue ?? Array.Empty<string>();
-            _apiTypes = branded ? new AzureApiTypes() : new SystemApiTypes();
+            _apiTypes =  "azure".Equals(flavor, StringComparison.InvariantCultureIgnoreCase) ? new AzureApiTypes() : new SystemApiTypes();
```
pshao25, line 182: I think it is case sensitive.

archerzz: > I think it is case sensitive.

It's not. Laurent has confirmed that, though I think it's better to be case sensitive.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -157,11 +155,12 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
                 "head-as-boolean": options["head-as-boolean"],
                 "deserialize-null-collection-as-null-value":
                     options["deserialize-null-collection-as-null-value"],
+                flavor:
+                    options["flavor"] ??
+                    (namespace.toLowerCase().startsWith("azure.")
+                        ? "azure"
+                        : undefined),
```
pshao25, line 162: If user inputs `flavor = non-azure` and the namespace starts with `azure`, it would be `azure`?

eng/Generate.ps1
```diff
@@ -74,10 +74,10 @@ function Add-CadlRanch-TypeSpec([string]$testName, [string]$projectPrefix, [stri
     $clientTsp = Join-Path $cadlRanchFilePath $testName "client.tsp"
     $mainTypeSpecFile = If (Test-Path $clientTsp) { Resolve-Path $clientTsp } Else { Resolve-Path $tspMain }
     if ($projectPrefix -eq "typespec-nonAzure-") {
-        Add-TypeSpec "$projectPrefix$testName" $projectDirectory $mainTypeSpecFile "$configString--option @azure-tools/typespec-csharp.new-project=true --option @azure-tools/typespec-csharp.branded=false" "-n"
+        Add-TypeSpec "$projectPrefix$testName" $projectDirectory $mainTypeSpecFile "$configString--option @azure-tools/typespec-csharp.new-project=true" "-n"
     }
     else {
-        Add-TypeSpec "$projectPrefix$testName" $projectDirectory $mainTypeSpecFile "$configString--option @azure-tools/typespec-csharp.new-project=true" "-n"
+        Add-TypeSpec "$projectPrefix$testName" $projectDirectory $mainTypeSpecFile "$configString--option @azure-tools/typespec-csharp.new-project=true --option @azure-tools/typespec-csharp.flavor=""azure""" "-n"
```
ArcturusZhang, line 80: is the quote really required here?
I believe if we have
```
--option xxxx=azure
```
it should also work perfectly fine.

src/AutoRest.CSharp/Common/Input/Configuration.cs
ArcturusZhang: this line needs to be formatted

src/AutoRest.CSharp/Common/Input/Configuration.cs
```diff
@@ -600,7 +603,10 @@ private static void WriteConfiguration(Utf8JsonWriter writer)
             {
                 MgmtTestConfiguration.SaveConfiguration(writer);
             }
-            WriteIfNotDefault(writer, Options.Branded, ApiTypes is AzureApiTypes);
+            if (ApiTypes is AzureApiTypes)
+            {
+                writer.WriteString(Options.Flavor, "azure");
+            }
```
ArcturusZhang, line 609: I think we could remove this check.
When it is `AzureApiTypes`, the value is azure, and when it is not, we should output whatever it is (and omit it if it is null).

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
ArcturusZhang: if flavor has value, the flavor will be the value. There is a null coalescing here

src/TypeSpec.Extension/Emitter.Csharp/src/type/configuration.ts
```diff
@@ -5,7 +5,7 @@ export interface Configuration {
     "output-folder": string;
     namespace: string;
     "library-name": string | null;
-    branded?: boolean;
+    flavor?: string;
```
chunyu3, line 8: And there is another requirement , the default flavor is not azure. Do you handle this?
https://github.com/Azure/autorest.csharp/issues/4309
And it said should keep Branded flag for a while

src/AutoRest.CSharp/Common/Input/Configuration.cs
```diff
@@ -366,7 +371,7 @@ public static void Initialize(IPluginCommunication autoRest, string defaultNames
                 methodsToKeepClientDefaultValue: autoRest.GetValue<string[]?>(Options.MethodsToKeepClientDefaultValue).GetAwaiter().GetResult() ?? Array.Empty<string>(),
                 mgmtConfiguration: MgmtConfiguration.GetConfiguration(autoRest),
                 mgmtTestConfiguration: MgmtTestConfiguration.GetConfiguration(autoRest),
-                branded: GetOptionBoolValue(autoRest, Options.Branded),
```
lirenhe, line 369: Can we keep branded for now?
'Don't remove yet currently supported flag branded to help the transition to the new system. If both flags are set, read flavor first and ignore branded boolean flag.'

src/TypeSpec.Extension/Emitter.Csharp/src/type/configuration.ts
archerzz: > And there is another requirement , the default flavor is not azure. Do you handle this? #4309 And it said should keep Branded flag for a while

Yes. And Laurent has confirmed that we should remove the flag.

src/AutoRest.CSharp/Common/Input/Configuration.cs
archerzz: > Can we keep branded for now? 'Don't remove yet currently supported flag branded to help the transition to the new system. If both flags are set, read flavor first and ignore branded boolean flag.'

He has changed the requirement, but the issue description is not updated.

src/TypeSpec.Extension/Emitter.Csharp/src/type/configuration.ts
chunyu3: Requirement: Emitter should generate "unbranded" code by default.
I couldn't see any source code support this.

src/AutoRest.CSharp/Common/Input/Configuration.cs
```diff
@@ -164,6 +164,11 @@ public enum UnreferencedTypesHandlingOption
                 }
             }
 
+            if (flavor is not null && !"azure".Equals(flavor, StringComparison.InvariantCultureIgnoreCase))
+            {
+                throw new Exception($"Invalid value of setting 'flavor': {flavor}");
```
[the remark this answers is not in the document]
archerzz: > Can we follow this?
> 
> ' Any other value does nothing, and therefore generates unbranded code.'

I tend to adopt strict validation here. We should not allow free input for configuration properties.

src/TypeSpec.Extension/Emitter.Csharp/src/type/configuration.ts
archerzz: `flavor` is an optional string. If you don't specify it, then it's undefined, and in the generator we will not adopt `AzureApiType`.

eng/Generate.ps1
archerzz: removed double quote

src/AutoRest.CSharp/Common/Input/Configuration.cs
archerzz: change to save the `flavor` value if possible

src/AutoRest.CSharp/Common/Input/Configuration.cs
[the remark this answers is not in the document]
archerzz: can you give me an example of non `azure` value? I thought one flavor defines a set of code generation template. So if we only have two templates right now (e.g. `null` and `azure`), then we should not accept other values.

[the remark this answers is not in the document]
chunyu3: Arch is discussing the different values of `flavor`. The exact values are not finalized, but it could have other values besides `azure`.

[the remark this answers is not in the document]
archerzz: IMHO, the forward compatibility is useless here and may cause confusion. If in the future we add `openai` flavor, the old emitter will generate unbranded codes without error. Better to just throw error so that users know they need to upgrade the emitter.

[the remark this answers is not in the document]
archerzz: I removed the validation since we're approaching the deadline.

chunyu3 (approved):

lirenhe (approved):

PR #4371, opened 2024-03-09
Fix datetime formatting in samples

AlexanderSher:
Default format for datetime in headers is ~~RFC1123~~ RFC7231
Default format for datetime in JSON is RFC3339


src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputParameterConverter.cs
```diff
@@ -119,5 +117,16 @@ public static InputParameter CreateInputParameter(ref Utf8JsonReader reader, str
...
             return parameter;
         }
+
+        private static InputType FixInputParameterType(InputType parameterType, RequestLocation requestLocation)
+            => parameterType switch
+            {
+                InputLiteralType literalType => literalType.LiteralValueType,
+                InputListType listType => listType with { ElementType = FixInputParameterType(listType.ElementType, requestLocation) },
+                InputDictionaryType dictionaryType => dictionaryType with { ValueType = FixInputParameterType(dictionaryType.ValueType, requestLocation) },
+                InputPrimitiveType { Kind: InputTypeKind.DateTime } when requestLocation == RequestLocation.Header => InputPrimitiveType.DateTimeRFC1123,
```
chunyu3, line 127: In previous design:Default format for datetime in headers is RFC7231
Default format for datetime in JSON is RFC3339.
Why do we change this?
And we apply this default logic when set the serializationFormat for parameter in `InputParameterGetSerializationFormat`. If we correct the DateTime type in converter, we may need to remove the duplicate and inconsistent behavior there.

AlexanderSher: Fixed

test/CadlRanchProjects/encode/datetime/src/Generated/Docs/Header.xml
```diff
@@ -7,15 +7,15 @@ This sample shows how to call DefaultAsync.
 <code><![CDATA[
 Header client = new DatetimeClient().GetHeaderClient();
 
-Response response = await client.DefaultAsync(DateTimeOffset.Parse("2022-05-10T14:57:31.2311892-04:00"));
```
m-nash, line 10: isn't this a breaking change if it happens in serialization scenarios?

AlexanderSher: Because of the `InputParameterGetSerializationFormat`, there is no change in serialization code - only in samples. However, if we switch HLC to input model, there will be breaking changes in serialization, hence the PR to avoid them.

AlexanderSher: Sure: https://github.com/Azure/azure-sdk-for-net/pull/42602/files

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputParameterConverter.cs
AlexanderSher: RFC7231 is correct default for headers

m-nash (approved):

PR #4382, opened 2024-03-11
Support extensible enum value type as a type

pshao25:
We don't support extensible value type now by throwing error. Below is [an example in cadl-ranch](https://github.com/Azure/cadl-ranch/blob/8fa11387a5e6b2d0d0f0e09d619f4325ab08d205/packages/cadl-ranch-specs/http/type/property/value-types/main.tsp#L278):
```
union ExtendedEnum {
  string,
  EnumValue2: "value2",
}

model UnionEnumValueProperty {
  property: ExtendedEnum.EnumValue2;
}
```
which is a real case in OpenAI: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/OpenAI.Inference/models/completions/extensions/oyd/azure_search_options.tsp#L21

This PR is to deal with this case.

What we are going to generate for the model `UnionEnumValueProperty` is like
```C#
public partial class UnionEnumValueProperty
{
    public UnionEnumValuePropertyProperty Property { get; } = UnionEnumValuePropertyProperty.Value2;
}
public readonly partial struct UnionEnumValuePropertyProperty 
{
    private const string Value2Value = "value2";
}
```
Pay attention to the value of `UnionEnumValuePropertyProperty`, we only have "value2" even users define a lot in the `union ExtendedEnum` because we only use "value2".


lirenhe (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/model.ts
```diff
@@ -1091,6 +1091,9 @@ export function getFormattedType(program: Program, type: Type): FormattedType {
     const format = getFormat(program, type);
     if (type.kind === "ModelProperty") {
         targetType = type.type;
+        if (targetType.kind === "UnionVariant") {
```
chunyu3, line 1094: we may need to add this new switch case in `getInputType`, when kind == `UnionVariant`, we use union variant type.
In that way we will support more case which use a UnionVariant as type, not just model property.

chunyu3:
What expected behavior:
```
model ModelT
{
  name: string
}
union UnionModel {
  item1: ModelT,
  item2: string,
}

model UnionEnumValueProperty {
  property: UnionModel .item1;
}
```

I think this is rare case, but what's the expected behavior?

pshao25:
> What expected behavior:
> 
> ```
> model ModelT
> {
>   name: string
> }
> union UnionModel {
>   item1: ModelT,
>   item2: string,
> }
> 
> model UnionEnumValueProperty {
>   property: UnionModel .item1;
> }
> ```
> 
> I think this is rare case, but what's the expected behavior?

My expectation for the type of `property` is `ModelT`, and luckily the actual behavior is `ModelT` too.

chunyu3 (approved):

PR #4384, opened 2024-03-12
Fix model factory for abstract types with discriminator

AlexanderSher:


test/TestProjects/Models-TypeSpec/src/Generated/ModelsTypeSpecModelFactory.cs
```diff
@@ -373,16 +391,6 @@ public static Int32ValuesFacet Int32ValuesFacet(string field = null, IEnumerable
             return new Int32ValuesFacet(field, serializedAdditionalRawData: null, values?.ToList(), value, kind);
         }
 
-        /// <summary> Initializes a new instance of <see cref="Models.DerivedWithDiscriminatorDefinedOnBase"/>. </summary>
-        /// <param name="optionalString"> Optional string. </param>
-        /// <param name="requiredString"> Required string. </param>
-        /// <param name="optionalInt"> Optional int. </param>
-        /// <returns> A new <see cref="Models.DerivedWithDiscriminatorDefinedOnBase"/> instance for mocking. </returns>
-        public static DerivedWithDiscriminatorDefinedOnBase DerivedWithDiscriminatorDefinedOnBase(string optionalString = null, string requiredString = null, int? optionalInt = null)
```
AlexanderSher, line 381: This change is expected.
`DerivedWithDiscriminatorDefinedOnBase` has one required property that can be set via public constructor, and two optional properties that have public setters, so it doesn't need factory method to get instantiated.

AlexanderSher:
Ignore [this](https://github.com/Azure/azure-sdk-for-net/pull/42618) preview, it must be rerun after azure-sdk-for-net is updated

test/TestProjects/MgmtMockAndSample/src/Generated/ArmMgmtMockAndSampleModelFactory.cs
```diff
@@ -572,7 +572,7 @@ public static FirewallPolicyData FirewallPolicyData(ResourceIdentifier id = null
         /// <param name="priority"> Priority of the Firewall Policy Rule Collection Group resource. </param>
         /// <param name="ruleCollections">
         /// Group of Firewall Policy rule collections.
-        /// Please note <see cref="FirewallPolicyRuleCollection"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
+        /// Please note <see cref="Models.FirewallPolicyRuleCollection"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
```
m-nash, line 575: Is this needed?  This class is already in Models namespace.

PR #4386, opened 2024-03-12
Throw on empty ResourceType

live1206:
Improvement for https://github.com/Azure/azure-sdk-for-net/issues/40050

We should throw on empty ResourceType while generating the SDK, the duplicated operations should be removed from the spec instead to fix the scope operation.


src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
```diff
@@ -367,6 +367,12 @@ protected FormattableString GetResourceTypeExpression(ResourceTypeSegment resour
             {
                 return $"{candidates.First().Type}.ResourceType";
             }
+
+            if (string.IsNullOrEmpty(resourceType.SerializedType))
+            {
+                throw new InvalidOperationException($"ResourceType is empty");
```
ArthurMa1978, line 373: Regen the sdk repo

ArthurMa1978 (approved):

src/AutoRest.CSharp/Mgmt/Generation/MgmtClientBaseWriter.cs
live1206: There is no SDK change in the regen [PR](https://github.com/Azure/azure-sdk-for-net/pull/42735)

PR #4390, opened 2024-03-13
Add endpoint if user not defined in the @server

### LINKED ISSUE #4389: Assume endpoint is a required parameter if none is specified in the @server decorator
As an example this spec https://github.com/bterlson/typespec-todo/blob/main/main.tsp should result in all client ctor overloads with a required `Uri endpoint`.

Currently we get this
```c#
// <auto-generated/>

#nullable disable

using System;
using System.ClientModel;
using System.ClientModel.Primitives;
using System.ClientModel.Primitives.Pipeline;
using System.Threading;

namespace Todo
{
    // Data plane generated client.
    /// <summary> The Todo service client. </summary>
    public partial class TodoClient
    {
        private const string AuthorizationHeader = "session-id";
        private readonly KeyCredential _keyCredential;
        private readonly MessagePipeline _pipeline;

        /// <summary> The ClientDiagnostics is used to provide tracing support for the client library. </summary>
        internal TelemetrySource ClientDiagnostics { get; }

        public virtual MessagePipeline Pipeline => _pipeline;

        /// <summary> Initializes a new instance of TodoClient for mocking. </summary>
        protected TodoClient()
        {
        }

        /// <summary> Initializes a new instance of TodoClient. </summary>
        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
        /// <exception cref="ArgumentNullException"> <paramref name="credential"/> is null. </exception>
        public TodoClient(KeyCredential credential) : this(credential, new TodoClientOptions())
        {
        }

        /// <summary> Initializes a new instance of TodoClient. </summary>
        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
        /// <param name="options"> The options for configuring the client. </param>
        /// <exception cref="ArgumentNullException"> <paramref name="credential"/> is null. </exception>
        public TodoClient(KeyCredential credential, TodoClientOptions options)
        {
            Argument.AssertNotNull(credential, nameof(credential));
            options ??= new TodoClientOptions();

            ClientDiagnostics = new TelemetrySource(options, true);
            _keyCredential = credential;
            _pipeline = MessagePipeline.Create(options, new IPipelinePolicy<PipelineMessage>[] { new KeyCredentialPolicy(_keyCredential, AuthorizationHeader) }, Array.Empty<IPipelinePolicy<PipelineMessage>>());
        }

        private Users _cachedUsers;
        private TodoItems _cachedTodoItems;

        /// <summary> Initializes a new instance of Users. </summary>
        public virtual Users GetUsersClient()
        {
            return Volatile.Read(ref _cachedUsers) ?? Interlocked.CompareExchange(ref _cachedUsers, new Users(ClientDiagnostics, _pipeline, _keyCredential), null) ?? _cachedUsers;
        }

        /// <summary> Initializes a new instance of TodoItems. </summary>
        public virtual TodoItems GetItemsClient()
        {
            return Volatile.Read(ref _cachedTodoItems) ?? Interlocked.CompareExchange(ref _cachedTodoItems, new TodoItems(ClientDiagnostics, _pipeline, _keyCredential), null) ?? _cachedTodoItems;
        }
    }
}

```

pshao25:
Fix #4389 

This applies to both azure and non-azure.

### DIFF
--- a/src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
+++ b/src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
@@ -209,7 +209,7 @@ namespace AutoRest.CSharp.Output.Models
             }
             operationsMap.Add(operation, updatedOperation);
 
-            if (!updatedOperation.Parameters.Any(p => p.IsEndpoint))
+            if (!updatedOperation.Parameters.Any(p => p.IsEndpoint || (p.Name.ToLower() == "endpoint" && p.Kind == InputOperationParameterKind.Client)))
             {
                 updatedOperation = updatedOperation with
                 {
--- a/src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
+++ b/src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
@@ -209,7 +209,7 @@ namespace AutoRest.CSharp.Output.Models
             }
             operationsMap.Add(operation, updatedOperation);
 
-            if (!updatedOperation.Parameters.Any(p => p.IsEndpoint || (p.Name.ToLower() == "endpoint" && p.Kind == InputOperationParameterKind.Client)))
+            if (!updatedOperation.Parameters.Any(p => p.IsEndpoint || (p.Location == RequestLocation.Uri && p.Kind == InputOperationParameterKind.Client)))
             {
                 updatedOperation = updatedOperation with
                 {
--- a/src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
+++ b/src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
@@ -209,8 +209,24 @@ namespace AutoRest.CSharp.Output.Models
             }
             operationsMap.Add(operation, updatedOperation);
 
+            if (!updatedOperation.Parameters.Any(p => p.Name == KnownParameters.Endpoint.Name))
+            {
+                updatedOperation = updatedOperation with
+                {
+                    Parameters = AddEndpointParameter(updatedOperation.Parameters)
+                };
+            }
+
             return updatedOperation;
 
+        }
+
+        private static IReadOnlyList<InputParameter> AddEndpointParameter(IReadOnlyList<InputParameter> operationParameters)
+        {
+            return new List<InputParameter>(operationParameters)
+            {
+                new InputParameter(KnownParameters.Endpoint.Name, "endpoint", $"{KnownParameters.Endpoint.Description}", new InputPrimitiveType(InputTypeKind.Uri, false), RequestLocation.Uri, null, null, null, InputOperationParameterKind.Client, true, false, false, false, true, false, false, null, null)
+            };
         }
 
         private static string UpdateOperationName(InputOperation operation, string clientName)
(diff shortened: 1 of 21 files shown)


lirenhe (approved):

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -209,10 +209,26 @@ private static InputOperation UpdateOperation(InputOperation operation, string c
...
+
             return updatedOperation;
 
         }
 
+        private static IReadOnlyList<InputParameter> AddEndpointParameter(IReadOnlyList<InputParameter> operationParameters)
+        {
+            return new List<InputParameter>(operationParameters)
+            {
+                new InputParameter(KnownParameters.Endpoint.Name, "endpoint", $"{KnownParameters.Endpoint.Description}", new InputPrimitiveType(InputTypeKind.Uri, false), RequestLocation.Uri, null, null, null, InputOperationParameterKind.Client, true, false, false, false, true, false, false, null, null)
```
chunyu3, line 228:
```suggestion
                new InputParameter(KnownParameters.Endpoint.Name,KnownParameters.Endpoint.Name, $"{KnownParameters.Endpoint.Description}", new InputPrimitiveType(InputTypeKind.Uri, false), RequestLocation.Uri, null, null, null, InputOperationParameterKind.Client, true, false, false, false, true, false, false, null, null)
```

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -209,10 +209,26 @@ private static InputOperation UpdateOperation(InputOperation operation, string c
             }
             operationsMap.Add(operation, updatedOperation);
 
+            if (!updatedOperation.Parameters.Any(p => p.IsEndpoint || (p.Location == RequestLocation.Uri && p.Kind == InputOperationParameterKind.Client)))
```
chunyu3, line 212: Suggest to group this into `UpdateOperationParameters`, in their include all the update logic for the parameter list.

pshao25: That is for paging operations. We don't call it.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
pshao25: It is not model property, we don't even use this `NameInRequest`.

chunyu3: Even we does not use it, we may better to keep consistent.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
chunyu3: It is better to combine all those in the `UpdateOperationParameters` for all operations (paging or non-paging), and in this method, switch the operation type to apply needed changes, and keep common changes for all type operation.
Otherwise, there are duplicate code and spread changes in several places, hard to maintain

pshao25: Synced offline. We could improve this part later.

Issue #4208, opened 2024-02-07
[CDK] Generate AppendChildObject only once per project


JoshLove-msft:
This method is currently generated in each model that can be serialized to bicep, but the implementation is identical.

PR #4399, opened 2024-03-14
Fix ResourceData algorithm and add separate helper class for AppendChildObject

JoshLove-msft:
Fixes https://github.com/Azure/autorest.csharp/issues/4208


src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/BicepSerializationTypeProvider.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        private BicepSerializationTypeProvider(string defaultNamespace, SourceInputModel? sourceInputModel)
+            : base(defaultNamespace, sourceInputModel)
+        {
+            DeclarationModifiers = TypeSignatureModifiers.Internal | TypeSignatureModifiers.Static;
+        }
+
+        protected override string DefaultName => "BicepSerializationHelpers";
+
+        protected override IEnumerable<Method> BuildMethods() => BicepSerializationMethodsBuilder.BuildPerAssemblyBicepSerializationMethods();
```
m-nash, line 30: we should move that method here since this type is meant to encapsulate all of that construction logic

src/AutoRest.CSharp/Common/Output/Expressions/Snippets.Bicep.cs
```diff
@@ -0,0 +1,26 @@
...
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
+using AutoRest.CSharp.Output.Models.Types;
+
+namespace AutoRest.CSharp.Common.Output.Models
+{
+    internal static partial class Snippets
+    {
+        public static class InvokeBicep
+        {
+            public static InvokeStaticMethodStatement AppendChildObject(
```
m-nash, line 15: Instead of an extra snippet here people can just call the instance method on the type provider directly

PR #4424, opened 2024-03-19
Support property overrides for flattened properties

JoshLove-msft:


src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
```diff
@@ -124,9 +124,7 @@ private void WriteProperty(CodeWriter writer, ObjectTypeProperty property, Objec
             writer.WriteXmlDocumentationSummary(CreatePropertyDescription(property));
             if (Configuration.EnableBicepSerialization && objectType.Declaration.Accessibility == "public" && property.Declaration.Accessibility == "public")
             {
-                string? wrapper = property.SchemaProperty?.FlattenedNames?.FirstOrDefault();
-                wrapper = wrapper is null ? string.Empty : $"{wrapper}.";
-                writer.Line($"[WirePath(\"{wrapper}{property.SerializedName}\")]");
+                writer.Line($"[WirePath(\"{property.GetWirePath()}\")]");
```
JoshLove-msft, line 127: Left this change in but the wirepath isn't needed for the changes in this PR. @m-nash let me know if I should remove it.

samples/AzureSample.ResourceManager.Sample/src/Generated/Models/AdditionalCapabilities.Serialization.cs
```diff
@@ -125,6 +130,10 @@ private BinaryData SerializeBicep(ModelReaderWriterOptions options)
             return BinaryData.FromString(builder.ToString());
         }
 
+        private void TransformFlattenedOverrides(BicepModelReaderWriterOptions bicepOptions, IDictionary<string, string> propertyOverrides)
```
m-nash, line 133: can we omit if there aren't any instead of adding this dead code?

JoshLove-msft: When emitting the methods we won't know if it is needed or not.

PR #4425, opened 2024-03-19
Update Azure.Core

live1206:
The latest `Azure.Core.1.38.0` depends on `System.ClientModel.1.0.0`, but generator depends on types from `System.ClientModel.1.0.0-beta.4`, it will throw warning on version downgrade for `System.ClientModel`.
```csharp
<PackageReference Update="System.ClientModel" Version="1.0.0-beta.4" NoWarn="NU1605" />
```
Ignore `NU1605` for `System.ClientModel` only.


src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
```diff
@@ -355,6 +355,7 @@ private string GetUnbrandedSrcCSProj()
 
         private static readonly IReadOnlyList<CSProjWriter.CSProjDependencyPackage> _brandedTestDependencyPackages = new CSProjWriter.CSProjDependencyPackage[]
         {
+            new("System.ClientModel", "1.0.0"),
```
live1206, line 358: Need to specify the version for `System.ClientModel` to avoid warning thrown for version downgrade.

test/AutoRest.TestServer.Tests/AutoRest.TestServer.Tests.csproj
```diff
@@ -24,6 +24,7 @@
     <PackageReference Include="NUnit3TestAdapter" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" />
     <PackageReference Include="Moq" Version="4.20.69" />
+    <PackageReference Include="System.ClientModel" Version="1.0.0-beta.4" NoWarn="NU1605" />
```
live1206, line 27: `System.ClientModel.1.0.0-beta.4` is needed for generation tests.

JoshLove-msft (approved):

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
m-nash: I thought this wasn't needed any longer?  Shouldn't this pull in transitively from Azure.Core

m-nash (changes requested):

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
JoshLove-msft, line 358:
```suggestion
```

JoshLove-msft: removed

m-nash (approved):

PR #4432, opened 2024-03-20
Mgmt schema consolidation

live1206:
Consolidate `CodeModel.Schema` into `InputType`


src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -48,10 +48,10 @@ public TypeFactory(OutputLibrary library, Type unknownType)
             InputListType { IsEmbeddingsVector: true } listType => new CSharpType(typeof(ReadOnlyMemory<>), listType.IsNullable, CreateType(listType.ElementType)),
             InputListType listType => new CSharpType(typeof(IList<>), listType.IsNullable, CreateType(listType.ElementType)),
             InputDictionaryType dictionaryType => new CSharpType(typeof(IDictionary<,>), inputType.IsNullable, typeof(string), CreateType(dictionaryType.ValueType)),
-            InputEnumType enumType => _library.ResolveEnum(enumType).WithNullable(inputType.IsNullable),
+            InputEnumType enumType => _library.ResolveEnum(enumType).IsNullable == inputType.IsNullable ? _library.ResolveEnum(enumType) : _library.ResolveEnum(enumType).WithNullable(inputType.IsNullable),
```
ArcturusZhang, line 51: we could just simplify this as 
```suggestion
            InputEnumType enumType => _library.ResolveEnum(enumType).WithNullable(inputType.IsNullable),
```
because `WithNullable` has the check if `isNullable` is the same as the nullable value we have in this type, return `this`

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -48,10 +48,10 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
             // TODO -- this is a temporary solution until we refactored the type replacement to use input types instead of code model schemas
             InputModelType { Namespace: "Azure.Core.Foundations", Name: "Error" } => SystemObjectType.Create(AzureResponseErrorType, AzureResponseErrorType.Namespace!, null).Type,
-            InputModelType model => _library.ResolveModel(model).WithNullable(inputType.IsNullable),
+            InputModelType model => _library.ResolveModel(model).IsNullable == inputType.IsNullable ? _library.ResolveModel(model) : _library.ResolveModel(model).WithNullable(inputType.IsNullable),
```
ArcturusZhang, line 54: same here

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
```diff
@@ -45,7 +45,8 @@ private void WriteObjectSchema(CodeWriter writer, ObjectType schema)
                     implementsTypes.Add(schema.Inherits);
                 }
 
-                writer.WriteXmlDocumentationSummary($"{schema.Description}");
+                var description = string.IsNullOrEmpty(schema.Description.ToString()) ? $"The {schema.Declaration.Name}." : schema.Description;
```
ArcturusZhang, line 48: we have a `FormattableStringHelper.IsEmpty` to see if a formattablestring is empty, so that we do not have to `ToString` it.
Also I am curious that previously we should have some fallback mechanism in model building to automatically give them this description when the schema does not have one, we do not have that on `ModelTypeProvider` now?

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
```diff
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Security.Cryptography.X509Certificates;
+
 namespace AutoRest.CSharp.Common.Input;
 
-internal abstract record InputType(string Name, bool IsNullable)
+internal abstract record InputType(string Name, bool IsNullable, string? OriginalName = default)
```
ArcturusZhang, line 8: so we added a `OriginalName` for all the types to store the name before we do any transformations.
It does not look elegant - but I believe we need it because we have to write the swagger names such as when `mgmt-debug.show-serialized-names` is on.
If there are other approaches, we might want to try that, such as let MPG output library to hold a mapping of name transformations?
It is ok to keep as is

src/AutoRest.CSharp/Common/Input/InputTypes/InputListType.cs
```diff
@@ -3,4 +3,4 @@
 
 namespace AutoRest.CSharp.Common.Input;
 
-internal record InputListType(string Name, InputType ElementType, bool IsEmbeddingsVector, bool IsNullable) : InputType(Name, IsNullable) { }
+internal record InputListType(string Name, InputType ElementType, bool IsEmbeddingsVector, bool IsNullable, string? OriginalName = null) : InputType(Name, IsNullable, OriginalName) { }
```
ArcturusZhang, line 6: same as dictionary, we do not need to add OriginalName here

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -42,6 +43,8 @@ internal void SetBaseModel(InputModelType? baseModel, [CallerFilePath] string fi
 
         public IEnumerable<InputModelType> GetAllBaseModels() => EnumerateBase(BaseModel);
 
+        public IReadOnlyList<InputModelType> GetImmediateBaseModels() => EnumerateBase(BaseModel, "AzureResourceBase").ToArray();
```
ArcturusZhang, line 46: this looks very like a workaround, what scenario does this method serve?

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -292,7 +301,7 @@ public static FormattableString CreateDerivedTypesDescription(CSharpType type)
         public static string CreateDescription(this Schema schema)
         {
             return string.IsNullOrWhiteSpace(schema.Language.Default.Description) ?
-                $"The {schema.Name}." :
+                string.Empty :
                 EscapeXmlDocDescription(schema.Language.Default.Description);
         }
```
ArcturusZhang, line 306: I understand because we want the generated description to match its real generated class name (which might be different from the name of its schema), we remove this fallback and let it to be empty, but could we just remove this method and let wherever calling this to handle its description?
When we are converting schemas to input types, we should not convert anything in the description, including escaping them.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -883,7 +883,7 @@ private static ValueExpression GetDeserializeImplementation(TypeProvider impleme
                 case Resource { ResourceData: SerializableObjectType resourceDataType } resource:
                     return New.Instance(resource.Type, new MemberExpression(null, "Client"), SerializableObjectTypeExpression.Deserialize(resourceDataType, element));
 
-                case MgmtObjectType mgmtObjectType when TypeReferenceTypeChooser.HasMatch(mgmtObjectType.ObjectSchema):
+                case MgmtObjectType mgmtObjectType when TypeReferenceTypeChooser.HasMatch(mgmtObjectType.InputModel):
```
ArcturusZhang, line 886: I should have seen exactly same code in `JsonCodeWriteExtensions`, could we add some todo to mark it as "to be refactored" in case we forget it?

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -32,65 +31,51 @@ namespace AutoRest.CSharp.Output.Models.Types
...
-        private readonly OutputLibrary? _library;
-        private readonly SchemaTypeUsage _usage;
-
-        private readonly IReadOnlyList<KnownMediaType> _supportedSerializationFormats;
 
         private ObjectTypeProperty? _additionalPropertiesProperty;
         private CSharpType? _implementsDictionaryType;
 
-        protected SchemaObjectType(ObjectSchema objectSchema, string defaultNamespace, TypeFactory typeFactory, SchemaUsageProvider schemaUsageProvider, OutputLibrary? library, SourceInputModel? sourceInputModel)
+        protected SchemaObjectType(InputModelType inputModel, string defaultNamespace, TypeFactory typeFactory, SourceInputModel? sourceInputModel, SerializableObjectType? defaultDerivedType = null)
```
ArcturusZhang, line 40: I think this is our first step of consolidation? By only modifying `SchemaObjectType` to consume input types instead of schemas.
I assume we should have a next step that we unify these two classes of `ModelTypeProvider` and `SchemaObjectType`, is this true?

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -156,86 +155,24 @@ public static ValidationType GetValidation(CSharpType type, RequestLocation requ
...
-                ChoiceSchema choiceSchema when type.IsFrameworkType => AddAllowedValues(description, choiceSchema.Choices),
-                SealedChoiceSchema sealedChoiceSchema when type.IsFrameworkType => AddAllowedValues(description, sealedChoiceSchema.Choices),
+                InputEnumType choiceSchema when type.IsFrameworkType => AddAllowedValues(description, choiceSchema.AllowedValues),
                 _ => description
             };
 
-            static FormattableString AddAllowedValues(FormattableString description, ICollection<ChoiceValue> choices)
+            static FormattableString AddAllowedValues(FormattableString description, IReadOnlyList<InputEnumTypeValue> choices)
             {
                 var allowedValues = string.Join(" | ", choices.Select(c => c.Value).Select(v => $"\"{v}\""));
```
ArcturusZhang, line 177: I think we could do better here by joining them using the `Join` method in our `FormattableStringHelper` static class because here we are assuming the allowed values here are all strings (we are adding quotes), but actually, we could do that using
```csharp
.Select(v => (FormattableString)$"{v:L}")
```
and then joining together to build another `FormattableString` instance using `FormattableStringHelper.Join` method, in this way, the write will automatically handle all values in literal, which adds quotes when it is a string, and handles correctly for other type of values.

src/AutoRest.CSharp/Common/Output/Models/Types/BuildContext.cs
```diff
@@ -10,23 +10,20 @@ namespace AutoRest.CSharp.Output.Models.Types
 {
     internal class BuildContext
     {
-        public BuildContext(CodeModel codeModel, SourceInputModel? sourceInputModel) : this(codeModel, sourceInputModel, Configuration.LibraryName, Configuration.Namespace)
+        public BuildContext(InputNamespace inputNamespace, SourceInputModel? sourceInputModel) : this(inputNamespace, sourceInputModel,Configuration.LibraryName, Configuration.Namespace)
```
ArcturusZhang, line 13: would this PR be an oppotunity to remove this class? Will removing this class cause a lot of effort?

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -107,7 +92,7 @@ protected SchemaObjectType(ObjectSchema objectSchema, string defaultNamespace, T
                 var existingMember = ModelTypeMapping?.GetMemberByOriginalName("$AdditionalProperties");
 
                 _additionalPropertiesProperty = new ObjectTypeProperty(
-                    BuilderHelpers.CreateMemberDeclaration("AdditionalProperties", ImplementsDictionaryType, "public", existingMember, _typeFactory),
+                    BuilderHelpers.CreateMemberDeclaration("AdditionalProperties", ImplementsDictionaryType, "public", existingMember, MgmtContext.TypeFactory),
```
ArcturusZhang, line 95: so we are changing this type to be "MPG only type provider"? But I still see we have a `_typeFactory` field in this class, we might need to align

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -737,24 +695,24 @@ protected bool TryGetPropertyForSchemaProperty(Func<ObjectTypeProperty, bool> pr
 
         protected override FormattableString CreateDescription()
         {
-            return $"{ObjectSchema.CreateDescription()}";
+            return $"{InputModel.DerivedModels}";
```
ArcturusZhang, line 698: would this work properly?

src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -33,16 +34,16 @@ internal static class CodeWriterExtensions
         /// <param name="type"></param>
         /// <param name="includeCollectionInitialization"></param>
         /// <returns></returns>
-        public static CodeWriter AppendExampleValue(this CodeWriter writer, ExampleValue exampleValue, CSharpType? type = null, bool includeCollectionInitialization = true)
+        public static CodeWriter AppendExampleValue(this CodeWriter writer, InputExampleValue exampleValue, CSharpType? type = null, bool includeCollectionInitialization = true)
```
ArcturusZhang, line 37: since we have the type unified, maybe we could try to remove this class and replace it by this class: https://github.com/Azure/autorest.csharp/blob/f823fd228ed51588df6703d8b35d00d0764818a6/src/AutoRest.CSharp/LowLevel/Extensions/Snippets.ExampleValues.cs#L26
which is designed to be a replacement of this
But this could happen later since we already have a really big PR

ArcturusZhang: please add a TODO here if we plan this replacement to happen later

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
live1206: That is true

src/AutoRest.CSharp/Common/Input/InputTypes/InputListType.cs
live1206: I have a plan to remove OriginalName from InputType related types, basically it's only used for generation report.

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
live1206: oh, this is for typespec input, we should not have them in this PR

src/AutoRest.CSharp/Common/Output/Models/Types/BuildContext.cs
live1206: Given the existing changed files, I will not remove it in this PR. Let's see if it gets easier after this PR

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
live1206: Good catch, this should be a bug

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
live1206: was in an intermediate stage to have `MgmtContext`, updated to use `_typeFactory` instead

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
live1206: Updated, but the default value is needed

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: removed

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
live1206: `schema.Description.IsNullOrEmpty()` is not working as expected, added TODO for later investigation

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
live1206: If we can have a rename map somewhere, we don't even need to make settable name properties in InputType

src/AutoRest.CSharp/Common/Input/InputTypes/InputListType.cs
live1206: removed

src/AutoRest.CSharp/Common/Input/InputTypes/InputDictionaryType.cs
```diff
@@ -3,4 +3,4 @@
 
 namespace AutoRest.CSharp.Common.Input;
 
-internal record InputDictionaryType(string Name, InputType KeyType, InputType ValueType, bool IsNullable) : InputType(Name, IsNullable) { }
+internal record InputDictionaryType(string Name, InputType KeyType, InputType ValueType, bool IsNullable, string? OriginalName = null) : InputType(Name, IsNullable, OriginalName) { }
```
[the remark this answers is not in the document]
live1206: removed

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
live1206: simplified

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
live1206: Created https://github.com/Azure/autorest.csharp/pull/4501 for this simplification.

src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs
live1206: We can't use `FormattableStringHelpers.IsNullOrEmpty` here, below `FormattableString` is not null or empty based on the logic but it is an empty string
image

live1206: >  Also I am curious that previously we should have some fallback mechanism in model building to automatically give them this description when the schema does not have one, we do not have that on ModelTypeProvider now?

The fallback was [there ](https://github.com/Azure/autorest.csharp/blob/0e54c4edd7bf20fc99267772a62e1e49fd31f9b9/src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs#L303) and removed in this commented code: https://github.com/Azure/autorest.csharp/pull/4432#discussion_r1542231569

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
live1206: not sure which piece of code you are talking about, could you share a link to it?

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
```diff
@@ -27,4 +29,8 @@ internal InputType GetCollectionEquivalent(InputType inputType)
                 return inputType;
         }
     }
+
+    public string Name { get; internal set; } = Name;
+
+    public string? OriginalName { get; internal set; } = OriginalName;
```
pshao25, line 35: 1. I don't think we should have any `set` for any `InputType`, which deviates from why we design it as `record`.
2. I think we should do something like:
```C#
internal abstract record InputType(string Name, bool IsNullable)
{
    public string CalculatedName => /*Some calculation*/;
    public string OriginalName => Name;
}
```
Because emitter doesn't do such calculation.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
pshao25, line 40: You won't want to do HLC at the same time, so we should not delete and API.

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelProperty.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace AutoRest.CSharp.Common.Input;
 
-internal record InputModelProperty(string Name, string SerializedName, string Description, InputType Type, InputConstant? ConstantValue, bool IsRequired, bool IsReadOnly, bool IsDiscriminator, IReadOnlyList<string>? FlattenedNames = null)
+internal record InputModelProperty(string Name, string SerializedName, string Description, InputType Type, InputConstant? ConstantValue, bool IsRequired, bool IsReadOnly, bool IsDiscriminator, IReadOnlyList<string>? FlattenedNames = null, string? Format = null, string? ElementTypeFormat = null)
```
live1206, line 9: store formats info for dfe

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -4,17 +4,19 @@
...
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Runtime.CompilerServices;
 
 namespace AutoRest.CSharp.Common.Input
 {
-    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, InputDictionaryType? InheritedDictionaryType, bool IsNullable)
-        : InputType(Name, IsNullable)
+    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, InputDictionaryType? InheritedDictionaryType, bool IsNullable, string[] SerializationFormats, bool IsEmpty = false, string? OriginalName = null)
```
live1206, line 12: SerializationFormats needs discussion with typespec team
IsEmpty can be replaced
`Original` should be set to `Name` in the constructor and have a getter for it, remove this from constructor, rename this to `SpecName`

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelTypeUsage.cs
```diff
@@ -11,5 +11,6 @@ internal enum InputModelTypeUsage
     None = 0,
     Input = 1,
     Output = 2,
-    RoundTrip = Input | Output
+    RoundTrip = Input | Output,
+    Converter
```
live1206, line 15: will have a decorator for `converter` in typespec

live1206: we will have decorator(key, value pair list) in InputType

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
```diff
@@ -30,7 +30,9 @@ internal record InputOperation(
     OperationPaging? Paging,
     bool GenerateProtocolMethod,
     bool GenerateConvenienceMethod,
-    bool KeepClientDefaultValue)
+    bool KeepClientDefaultValue,
+    string? OperationId,
```
live1206, line 34: we don't need Operationid, we can change the description

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
```diff
@@ -30,7 +30,9 @@ internal record InputOperation(
...
+    string? OriginalName)
```
live1206, line 35: follow the change in model

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
```diff
@@ -23,7 +23,8 @@ internal record InputParameter(
     bool SkipUrlEncoding,
     bool Explode,
     string? ArraySerializationDelimiter,
-    string? HeaderCollectionPrefix)
+    string? HeaderCollectionPrefix,
+    string? Format = null)
```
live1206, line 27: Format is needed for dfe, needs confirmation from typespec team

src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs
```diff
@@ -43,6 +44,11 @@ internal record InputParameter(
         SkipUrlEncoding: false,
         Explode: false,
         ArraySerializationDelimiter: null,
-        HeaderCollectionPrefix: null)
+        HeaderCollectionPrefix: null,
+        Format: null)
     { }
+
+    public string Name { get; internal set; } = Name;
+    public string NameInRequest { get; internal set; } = NameInRequest;
```
live1206, line 52: follow the patten in model

src/AutoRest.CSharp/Common/Input/InputTypes/InputSystemType.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace AutoRest.CSharp.Common.Input
+{
+    internal record InputSystemType(Type Type, InputType ElementType, bool IsNullable) : InputType(Type.Name, IsNullable);
```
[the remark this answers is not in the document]
live1206: remove it

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
```diff
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Security.Cryptography.X509Certificates;
+
 namespace AutoRest.CSharp.Common.Input;
 
-internal abstract record InputType(string Name, bool IsNullable)
+internal abstract record InputType(string Name, bool IsNullable, string? OriginalName = null)
```
live1206, line 8: follow the patten in model

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelTypeUsage.cs
live1206: we can work on this right away

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
pshao25: IReadonlyList<string>

src/AutoRest.CSharp/Mgmt/Decorator/BodyParameterNormalizer.cs
```diff
@@ -45,56 +46,55 @@ internal static void UpdateUsingReplacement(RequestParameter bodyParameter, IDic
...
         }
 
-        internal static void UpdateParameterNameOnly(RequestParameter bodyParam, IDictionary<string, HashSet<OperationSet>> resourceDataDictionary, Operation operation)
+        internal static void UpdateParameterNameOnly(InputParameter bodyParam, IDictionary<string, HashSet<OperationSet>> resourceDataDictionary, InputOperation operation)
         {
-            string oriName = bodyParam.Language.Default.Name;
-            bodyParam.Language.Default.SerializedName ??= bodyParam.Language.Default.Name;
-            bodyParam.Language.Default.Name = NormalizeParamNames.GetNewName(bodyParam.Language.Default.Name, bodyParam.Schema.Name, resourceDataDictionary);
+            string oriName = bodyParam.Name;
+            bodyParam.Name = NormalizeParamNames.GetNewName(bodyParam.Name, bodyParam.Type.Name, resourceDataDictionary);
```
live1206, line 57: `bodyParam.Type.Name` is causing parameter name changing from `anyObject` to `unknown`.
Because we are changing from `AnyObjectSchame` to `InputIntrinsicType.Unknown`.

src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
live1206: Created https://github.com/Azure/autorest.csharp/issues/4717 for tracking

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
live1206: would like to keep `OperationId` for swagger input, because the user is using this to locate the corresponding operation in swagger.

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
```diff
@@ -10,50 +10,79 @@
...
+    string path,
+    string? externalDocsUrl,
+    IReadOnlyList<string>? requestMediaTypes,
+    bool bufferResponse,
+    OperationLongRunning? longRunning,
+    OperationPaging? paging,
+    bool generateProtocolMethod,
+    bool generateConvenienceMethod,
+    bool keepClientDefaultValue,
+    string? operationId = null)
```
live1206, line 36: would like to keep `OperationId` for swagger input, as the user is using it to locate the corresponding operation in swagger.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -38,10 +37,10 @@ public TypeFactory(OutputLibrary library, Type unknownType)
         /// </summary>
         /// <param name="inputType">The input type to convert.</param>
         /// <returns>The <see cref="CSharpType"/> of the input type.</returns>
-        public CSharpType CreateType(InputType inputType) => inputType switch
+        public CSharpType CreateType(InputType inputType, string? format = null) => inputType switch
```
ArcturusZhang, line 40: I think this is not a good practice.
Could we consolidate those xms format to corresponding input types?
If this is for DataFactory, we could create some new types such as `InputDataFactoryElementType` in the converter

live1206: DataFacotry mdoels have been converted into InputGenericType in CodeModelConverter, `format` is needed for a few other cases

live1206: Guess we can try to handle them in CodeModelConverter as well, will try it out when I resolve all the generation errors

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelProperty.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace AutoRest.CSharp.Common.Input;
 
-internal record InputModelProperty(string Name, string SerializedName, string Description, InputType Type, InputConstant? ConstantValue, bool IsRequired, bool IsReadOnly, bool IsDiscriminator, IReadOnlyList<string>? FlattenedNames = null)
+internal record InputModelProperty(string Name, string SerializedName, string Description, InputType Type, InputConstant? ConstantValue, bool IsRequired, bool IsReadOnly, bool IsDiscriminator, bool? IsNullable = null, IReadOnlyList<string>? FlattenedNames = null, IReadOnlyList<string>? GroupParameterNames = null)
```
live1206, line 9: nullable should exist in type not property

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: resolved

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -41,14 +40,16 @@ public TypeFactory(OutputLibrary library, Type unknownType)
         public CSharpType CreateType(InputType inputType) => inputType switch
         {
             InputLiteralType literalType => CSharpType.FromLiteral(CreateType(literalType.LiteralValueType), literalType.Value),
-            InputUnionType unionType => CSharpType.FromUnion(unionType.UnionItemTypes.Select(CreateType).ToArray(), unionType.IsNullable),
+            InputUnionType unionType => CSharpType.FromUnion(unionType.UnionItemTypes.Select(x => CreateType(x)).ToArray(), unionType.IsNullable),
```
ArcturusZhang, line 43: looks like we do not need to change this line now

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -41,14 +40,16 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
             InputListType { IsEmbeddingsVector: true } listType => new CSharpType(typeof(ReadOnlyMemory<>), listType.IsNullable, CreateType(listType.ElementType)),
             InputListType listType => new CSharpType(typeof(IList<>), listType.IsNullable, CreateType(listType.ElementType)),
             InputDictionaryType dictionaryType => new CSharpType(typeof(IDictionary<,>), inputType.IsNullable, typeof(string), CreateType(dictionaryType.ValueType)),
+            // TODO: consolidate the behavior when we handle type nullability in DPG
```
ArcturusZhang, line 47: do we have an issue for this?

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -4,11 +4,12 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Runtime.CompilerServices;
 
 namespace AutoRest.CSharp.Common.Input
 {
-    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, InputDictionaryType? InheritedDictionaryType, bool IsNullable)
+    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, InputDictionaryType? InheritedDictionaryType, bool IsNullable, IReadOnlyList<InputModelType>? Parents = null, bool IsBasePolyType = false)
```
ArcturusZhang, line 12: `Parents` reads quite confusing, could we use the `CompositionModels` as Alexander introduced previously to represent multiple `allOf` in swaggers?

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
```diff
@@ -27,4 +36,8 @@ internal InputType GetCollectionEquivalent(InputType inputType)
                 return inputType;
         }
     }
+
+    public bool IsNullable { get; init; }
+    public string Name { get; internal set; }
+    public string? SpecName { get; init; }
```
ArcturusZhang, line 42: considering this issue: https://github.com/Azure/autorest.csharp/issues/4728
if a type does not have to have a name, it does not have to have a spec name.
We might need some discussion on this property

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: Here: https://github.com/Azure/autorest.csharp/blob/ced9b4b84e81a2daac4f7033ffd3693cd179ac67/src/AutoRest.CSharp/Common/Generation/Writers/JsonCodeWriterExtensions.cs#L160

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -476,9 +474,9 @@ private static IEnumerable<PathSegment> GetPathSegments(string httpRequestUri, I
             };
         }
 
-        public virtual Parameter BuildConstructorParameter(InputParameter operationParameter)
+        public Parameter BuildConstructorParameter(InputParameter operationParameter, TypeFactory typeFactory)
```
ArcturusZhang, line 477: considering this is an instance method, if we need typeFactory, could we just use the field `_typeFactory` directly? If so we do not need to change this signature.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -496,7 +494,7 @@ public virtual Parameter BuildConstructorParameter(InputParameter operationParam
         public static bool IsIgnoredHeaderParameter(InputParameter operationParameter)
             => operationParameter.Location == RequestLocation.Header && IgnoredRequestHeader.Contains(operationParameter.NameInRequest);
 
-        private Parameter BuildParameter(in InputParameter operationParameter, Type? typeOverride = null)
+        private Parameter BuildParameter(in InputParameter operationParameter, TypeFactory typeFactory, Type? typeOverride = null)
```
ArcturusZhang, line 497: same here

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -449,7 +449,7 @@ private IReadOnlyList<ObjectPropertyInitializer> GetPropertyInitializers(IReadOn
                 {
                     // For structs all properties become required
                     Constant? defaultParameterValue = null;
-                    if (property.SchemaProperty?.ClientDefaultValue is { } defaultValueObject)
+                    if (property.InputModelProperty?.DefaultValue is object defaultValueObject)
```
ArcturusZhang, line 452: would we write in this way and the type of this local variable `defaultValueObject` changes to `object` type?
Or it looks like to be.
I think we should still use `{}` which is just asserting this is not null

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -702,7 +702,7 @@ protected override IEnumerable<ObjectTypeConstructor> BuildConstructors()
             return _inputModelSerialization.Xml is { } xml ? SerializationBuilder.BuildXmlObjectSerialization(xml.Name ?? _inputModel.Name, this, _typeFactory) : null;
         }
 
-        protected override bool EnsureIncludeSerializer()
+    protected override bool EnsureIncludeSerializer()
```
ArcturusZhang, line 705: this needs to format

src/AutoRest.CSharp/Mgmt/Decorator/ResourceDetection.cs
```diff
@@ -6,28 +6,27 @@
...
 using AutoRest.CSharp.Utilities;
+using Azure.Core;
 
 namespace AutoRest.CSharp.Mgmt.Decorator
 {
     internal static class ResourceDetection
     {
         private const string ProvidersSegment = "/providers/";
-        private static ConcurrentDictionary<string, (string Name, ObjectSchema? Schema)?> _resourceDataSchemaCache = new ConcurrentDictionary<string, (string Name, ObjectSchema? Schema)?>();
+        private static ConcurrentDictionary<string, (string Name, InputModelType? Schema)?> _resourceDataSchemaCache = new ConcurrentDictionary<string, (string Name, InputModelType? Schema)?>();
```
ArcturusZhang, line 19: maybe we could rename this `Schema` to `Model` here

src/AutoRest.CSharp/Mgmt/Decorator/ResourceDetection.cs
```diff
@@ -110,9 +109,9 @@ private static bool CheckEvenSegments(string requestPath)
             return segments.Length % 2 == 0;
         }
 
-        private static bool TryOperationWithMethod(this OperationSet set, HttpMethod method, [MaybeNullWhen(false)] out ObjectSchema resourceSchema)
+        private static bool TryOperationWithMethod(this OperationSet set, RequestMethod method, [MaybeNullWhen(false)] out InputModelType resourceType)
```
ArcturusZhang, line 112: resourceType might be misunderstood in the context of MPG, could we change to `resourceModel` here?

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: yes, after removing format

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
live1206: I don't think we need to consider all TypeSpec alignments with this PR, this PR works based on the existing input types.
And we already had a discussion with Michael before regarding this `SpecName`.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
live1206: typeFactory is not in use anyway, removed.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
live1206: typeFactory is not in use anyway, removed.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
live1206: reverted

src/AutoRest.CSharp/Mgmt/Decorator/ResourceDetection.cs
live1206: renamed to `InputModel`

src/AutoRest.CSharp/Mgmt/Decorator/ResourceDetection.cs
live1206: renamed to `inputModel`

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: I thought we already have an issue regarding nullability for property and type

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
live1206: `CompositionModels` only represents `Immediate` parents, not `All` parents. If you think the naming is better, I can use that instead and change the value accordingly.

live1206: Renamed `Parents` to `AllBaseModels`, and get `CompositionModels` from `AllBaseModels`

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
m-nash: Lets link the issue here for reference.

m-nash (commented):
I didn't see an on / off switch for this.  Does that mean this is already on for everything and there are zero diffs?

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -84,60 +85,11 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
             },
             InputGenericType genericType => new CSharpType(genericType.Type, CreateType(genericType.ArgumentType)).WithNullable(inputType.IsNullable),
-            InputIntrinsicType { Kind: InputIntrinsicTypeKind.Unknown } => UnknownType,
-            CodeModelType cmt => CreateType(cmt.Schema, cmt.IsNullable),
+            // TODO: consolidate the behavior when we handle type nullability in DPG
+            InputIntrinsicType { Kind: InputIntrinsicTypeKind.Unknown } => Configuration.AzureArm ? new CSharpType(UnknownType, inputType.IsNullable) : UnknownType,
             _ => throw new Exception("Unknown type")
         };
 
-        public CSharpType CreateType(Schema schema, bool isNullable, string? formatOverride = default, Property? property = default) => CreateType(schema, formatOverride ?? schema.Extensions?.Format, isNullable, property);
```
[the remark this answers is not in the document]
live1206: They are not in use anymore, all schema consumption has been converted to InputType in `CodeModelConverter.cs`

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -289,11 +296,7 @@ public static FormattableString CreateDerivedTypesDescription(CSharpType type)
...
 
         public static string CreateDescription(this Schema schema)
-        {
```
live1206, line 292: This causes [Description update](https://github.com/Azure/azure-sdk-for-net/pull/44148/files#diff-327d085438f0bbac252b734c863541084d36954d7ed0966205310e77e5c5b1deL14)

src/AutoRest.CSharp/Mgmt/Models/MgmtClientOperation.cs
```diff
@@ -146,12 +146,12 @@ private FormattableString BuildDescription()
...
                 }
                 FormattableString defaultApiVersion = $"";
-                if (operation.Operation.ApiVersions.Any())
+                if (operation.Operation.Parameters.Where(p => p.IsApiVersion).Any())
                 {
                     defaultApiVersion = $@"
 <item>
 <term>Default Api Version</term>
-<description>{string.Join(", ", operation.Operation.ApiVersions.Select(v => v.Version))}</description>
+<description>{string.Join(", ", operation.Operation.Parameters.Where(p => p.IsApiVersion).Select(x => x.DefaultValue?.Value))}</description>
```
live1206, line 154: This is causing [description change](https://github.com/Azure/azure-sdk-for-net/pull/44148/files#diff-3aa60dfdb9a3c22c0d6e23cb9a6e5b33738d46bb939af869191bb273e7c0c9d8L67-L70), because we don't store api versions in operation anymore, only in operation parameters, some operation doesn't have api version parameter, but have operation-wise api versions.

src/AutoRest.CSharp/Mgmt/Models/MgmtRestOperation.cs
```diff
@@ -182,17 +184,17 @@ public MgmtRestOperation(MgmtRestOperation other, string nameOverride, CSharpTyp
 
         private CSharpType? GetFinalResponse()
         {
-            var finalSchema = Operation.LongRunningFinalResponse.ResponseSchema;
+            var finalSchema = Operation.LongRunning?.FinalResponse.BodyType;
```
live1206, line 187: This is causing [additional internal rest operation](https://github.com/Azure/azure-sdk-for-net/pull/44148/files#diff-6dc9899281ca92ce54ccab5d43c123a817258d507983dd49e397475f839073a7).
With the[ swagger definition](https://github.com/Azure/azure-rest-api-specs/blob/9061674ac55af0ec23e12e4f9f505010ab428935/specification/automation/resource-manager/Microsoft.Automation/stable/2023-11-01/runbook.json#L146-L148), `ResponseSchema` from M4 is null, not same as `BodyType`.

live1206:
> I didn't see an on / off switch for this. Does that mean this is already on for everything and there are zero diffs?

Yes. With swagger input, we have `CodeModel.Schema` from M4 and then we convert all the schema related models into existing input types, which we apply with TypeSpec input. And then, this PR updates all the consumption of CodeModel.Schema to InputModel in all modules for the MPG generator.

There are some minor diffs, which I think are acceptable:
- [Description update](https://github.com/Azure/autorest.csharp/pull/4432/files#r1607530341)
- [Another description update](https://github.com/Azure/autorest.csharp/pull/4432/files#r1607549779)
- [Parameter anyObject renaming to unknown](https://github.com/Azure/autorest.csharp/pull/4432#discussion_r1601074584)
- [Additional internal rest operation](https://github.com/Azure/autorest.csharp/pull/4432/files#r1607561201)

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -23,7 +24,12 @@ internal record InputModelType(string Name, string? Namespace, string? Accessibi
         /// <summary>
         /// Types provided as immediate parents in spec that aren't base model
         /// </summary>
-        public IReadOnlyList<InputModelType> CompositionModels { get; init; } = Array.Empty<InputModelType>();
+        public IEnumerable<InputModelType> CompositionModels => AllBaseModels.Where(x => x != BaseModel && x.Name != "AzureResourceBase");
+
+        public IReadOnlyList<InputModelType> AllBaseModels { get; init; } = Array.Empty<InputModelType>();
+
+        // TODO: remove the workaround for immediate base models
+        public IEnumerable<InputModelType> ImmediateBaseModels => AllBaseModels.Where(x => x.Name != "AzureResourceBase");
```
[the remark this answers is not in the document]
live1206: - `AllBaseModels` corresponds to [Schema.Parents.All](https://github.com/Azure/autorest.csharp/blob/8184be8ff9895b9600555aac2359aee868e0dca2/src/AutoRest.CSharp/Common/Input/Generated/CodeModel.cs#L1376)
- `ImmediateBaseModels` corresponds to [Schema.Parents.Immediate](https://github.com/Azure/autorest.csharp/blob/8184be8ff9895b9600555aac2359aee868e0dca2/src/AutoRest.CSharp/Common/Input/Generated/CodeModel.cs#L1372), and I found it is just `AllBaseModels` excluding `AzureResourceBase`
- `CompositionModel` corresponds to `ImmediateBaseModels` excluding `BaseModel`, it;s only used in `ModelTypeProvider` for DPG.

I guess we can consolidate them further while we consolidate `SchemaObjectType` and `ModelTypeProvider`.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: https://github.com/Azure/autorest.csharp/issues/4701

live1206: put the work item in the comment

live1206: drop the TODO and always call WithNullable

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
live1206: If this is only used for comment, we should use property.serializedName instead

src/AutoRest.CSharp/Common/Input/InputTypes/Examples/InputOperationExample.cs
```diff
@@ -6,5 +6,5 @@
 namespace AutoRest.CSharp.Common.Input.Examples
 {
     // TODO -- currently we do not need the responses. In the case that in the future we need to handle the responses in examples, we need to add a new class InputExampleResponse and a new property here
-    internal record InputOperationExample(InputOperation Operation, IReadOnlyList<InputParameterExample> Parameters);
+    internal record InputOperationExample(InputOperation Operation, IReadOnlyList<InputParameterExample> Parameters, string? Name = null, string? OriginalFile = null);
```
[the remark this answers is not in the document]
live1206: rename to ExampleFilePath

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelProperty.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace AutoRest.CSharp.Common.Input;
 
-internal record InputModelProperty(string Name, string SerializedName, string Description, InputType Type, InputConstant? ConstantValue, bool IsRequired, bool IsReadOnly, bool IsDiscriminator, IReadOnlyList<string>? FlattenedNames = null)
+internal record InputModelProperty(string Name, string SerializedName, string Description, InputType Type, InputConstant? ConstantValue, bool IsRequired, bool IsReadOnly, bool IsDiscriminator, IReadOnlyList<string>? FlattenedNames = null, IReadOnlyList<string>? GroupParameterNames = null)
```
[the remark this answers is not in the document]
live1206: try to find out the swagger

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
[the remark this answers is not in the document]
live1206: get the composition properties instead of composition models

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -184,8 +185,14 @@ private static bool IsEscapedMatch(ref ReadOnlySpan<char> span, int i, string es
         public static string CSharpName(this Property property) =>
             (property.Language.Default.Name == null || property.Language.Default.Name == "null") ? "NullProperty" : property.Language.Default.Name.ToCleanName();
 
-        public static string CSharpName(this Operation operation) =>
-            operation.Language.Default.Name.ToCleanName();
+        public static string CSharpName(this InputModelProperty property) =>
```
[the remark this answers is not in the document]
live1206: add a follow up item to reorganize this

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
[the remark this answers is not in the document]
live1206: All removed

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
live1206: Add an internal `OperationId` property
```csharp
internal string OperationId => ResourceName is null ? Name : $"{ResourceName}_{Name}";
```

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
live1206: Removed all of these, build properties in advance in `CodeModelConverter`

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelProperty.cs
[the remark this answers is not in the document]
live1206: We already have [InputParameter.GroupedBy](https://github.com/Azure/autorest.csharp/blob/b1eff796298fdec2f03e14f580167009d79ec689/src/AutoRest.CSharp/Common/Input/InputTypes/InputParameter.cs#L15)

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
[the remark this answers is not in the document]
live1206: Created https://github.com/Azure/autorest.csharp/issues/4767 and referenced it in code

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
live1206: Can't remove `SpecName`/`OriginalName` for now.
Since we are dealing with nullability by creating a new instance of the type, it is used in https://github.com/Azure/autorest.csharp/pull/4432/files#diff-46494a1ddd104f848f61dc5ede6cb7aa5f98c01813e9e9d67108e02a55299e58R841.

Heard that we will use Nullable to hold the original instance for nullability, then we can still refer to the same instance of the same type. Will rework on this when nullability is handled properly.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
live1206: Created https://github.com/Azure/autorest.csharp/issues/4777 for it.

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelProperty.cs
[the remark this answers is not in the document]
live1206: replace the grouping parameters with created type in converter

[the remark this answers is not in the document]
live1206: Added TODO in code and created https://github.com/Azure/autorest.csharp/issues/4779 for tracking.

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
live1206: Created https://github.com/Azure/autorest.csharp/issues/4780 for tracking, we will keep this workaround for now.

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
live1206: This is not only used in output models, such as https://github.com/Azure/autorest.csharp/blob/03d3cc83dd33cce6347227622e7a0eb4ec4ac544/src/AutoRest.CSharp/Mgmt/Decorator/OperationExtensions.cs#L58 and https://github.com/Azure/autorest.csharp/blob/03d3cc83dd33cce6347227622e7a0eb4ec4ac544/src/AutoRest.CSharp/Mgmt/Output/Resource.cs#L333.
Would prefer to expose this as a public property instead of constructing it in every single usage.

m-nash (approved):

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -523,18 +464,15 @@ protected override IEnumerable<ObjectTypeProperty> BuildProperties()
             var propertiesFromSpec = GetParentPropertyDeclarationNames();
             var existingProperties = GetParentPropertySerializedNames();
 
-            foreach (var objectSchema in GetCombinedSchemas())
+            foreach (var property in UpdateInputModelProperties())
```
ArcturusZhang, line 467: well I have some questions on this.

PR #4434, opened 2024-03-20
add the new `ClientUriBuilder` helper to replace `RequestUri` in non-branded libraries

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4431

# Description

# Checklist

# Ready to Land?


samples/Azure.Network.Management.Interface/Generated/Internal/ModelSerializationExtensions.cs
```diff
@@ -245,7 +245,7 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value)
             }
         }
 
-        private static class TypeFormatters
+        internal static class TypeFormatters
```
annelo-msft, line 248: It looks like this may apply the change to Azure libraries - the new ClientUriBuilder should be used for unbranded only for now until we have time to do the validation that changing the solution in Azure clients won't cause perf regressions.

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+#nullable disable
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NoTestTypeSpec
+{
+    internal class ClientUriBuilder
```
annelo-msft, line 12: Just wondering because of my lack of familiarity with autorest.csharp code -- have we added tests to validate all the different scenarios that this type addresses?  If so, that's great - would you be able to share a link for my learning?  This is a non-blocking comment, but anything you provide will help me understand whether we have good validation of the helper type implementation.

annelo-msft (commented):
One more non-blocking question for my learning -- is it easy to point to code that changes a call to the old RequestUri.AppendRaw to one of the new variants that calls either AppendPath, AppendQuery, or Reset(Uri)?  I am interested to understand whether this proposal was able to address all the cases where we were calling AppendRaw, and if so, what those changes look like and how much work it was to implement.  Many thanks!

samples/Azure.Network.Management.Interface/Generated/Internal/ModelSerializationExtensions.cs
m-nash: typeformatters is shared between them so we need to mark it as internal for all

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
m-nash: No, we have very little testing in the unbranded path but the tests should be added here like we did for TypeFormatter https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.Shared.Tests/TypeFormatterTests.cs

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            _uriBuilder = new UriBuilder(uri);
+            _pathBuilder = new StringBuilder(UriBuilder.Path);
+            _queryBuilder = new StringBuilder(UriBuilder.Query);
+        }
+
+        public void AppendPath(string value, bool escape)
+        {
+            Argument.AssertNotNullOrWhiteSpace(value, nameof(value));
+
```
annelo-msft, line 39: We should add the following lines here:

```

        if (PathBuilder[PathBuilder.Length - 1] == '/' && value[0] == '/')
        {
            PathBuilder.Remove(PathBuilder.Length - 1, 1);
        }
```

JoshLove-msft (approved):

Issue #4234, opened 2024-02-19
Enable CodeGenSerialization full functions for MPG


ArthurMa1978:
Enable [Add a new property to the model with serialization/deserialization](https://github.com/Azure/autorest.csharp/?tab=readme-ov-file#add-a-new-property-to-the-model-with-serializationdeserialization)

ArthurMa1978:
Fixed

PR #4435, opened 2024-03-20
support add property into serialization for mgmt plane

RodgeFu:
Fix issue https://github.com/Azure/autorest.csharp/issues/4234

support adding property in serialization for Mgmt Plane by using CodeGenSerialization attribute


src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -417,14 +421,30 @@ private IEnumerable<JsonPropertySerialization> GetPropertySerializationsFromBag(
...
+                {
+                    addToPropertyBag(objectTypeLevel, objectTypeProperty);
+                }
+            }
+            foreach (var objectTypeLevel in objectType.EnumerateHierarchy())
+            {
+                foreach (var objectTypeProperty in objectTypeLevel.SerializationConstructor.Initializers.Select(i => i.Property))
+                {
+                    addToPropertyBag(objectTypeLevel, objectTypeProperty);
                 }
```
ArcturusZhang, line 448: why here we iterate over some properties twice?

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -417,14 +421,30 @@ private IEnumerable<JsonPropertySerialization> GetPropertySerializationsFromBag(
...
-                    if (objectTypeProperty.SchemaProperty != null)
+                    var ms = objectType.GetForMemberSerialization(curProp.Declaration.Name);
+                    if (ms == null && curObj is SchemaObjectType sot)
                     {
-                        propertyBag.Properties.Add(objectTypeProperty, objectType.GetForMemberSerialization(objectTypeProperty.Declaration.Name));
+                        ms = sot.GetForMemberSerialization(curProp.Declaration.Name);
                     }
+                    propertyBag.Properties.Add(curProp, ms);
+                }
+            };
```
ArcturusZhang, line 435: Instead of a lambda, we should define a local method here according to the scenario here. Such as
```suggestion
            static void AddToPropertyBag(PropertyBag propertyBag, ObjectType curObj, ObjectTypeProperty curProp)
            {
                if (curProp.SchemaProperty != null && !propertyBag.Properties.ContainsKey(curProp))
                {
                    var ms = objectType.GetForMemberSerialization(curProp.Declaration.Name);
                    if (ms == null && curObj is SchemaObjectType sot)
                    {
                        ms = sot.GetForMemberSerialization(curProp.Declaration.Name);
                    }
                    propertyBag.Properties.Add(curProp, ms);
                }
            };
```

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -382,7 +384,9 @@ private IEnumerable<JsonPropertySerialization> GetPropertySerializationsFromBag(
                 var serializedName = serializationMapping?.SerializationPath?[^1] ?? schemaProperty.SerializedName;
                 var isRequired = schemaProperty.IsRequired;
                 var shouldExcludeInWireSerialization = (schemaProperty.IsDiscriminator == null || !schemaProperty.IsDiscriminator.Value) && property.InitializationValue is null && schemaProperty.IsReadOnly;
-                var serialization = BuildSerialization(schemaProperty.Schema, property.Declaration.Type, false);
+                var serialization = schemaProperty.Schema.GetType() == typeof(Schema)
+                    ? BuildJsonSerialization(BuilderHelpers.GetInputTypeFromCSharpType(property.Declaration.Type), property.Declaration.Type, false, property.SerializationFormat)
+                    : BuildSerialization(schemaProperty.Schema, property.Declaration.Type, false);
```
ArcturusZhang, line 389: we should not have this if.
Instead we should construct the property with proper schema when the property is added by customization code.

In the near future, schema related things should be removed, we should not introduce more issues for that work item to resolve.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -223,6 +224,51 @@ protected override ObjectTypeConstructor BuildSerializationConstructor()
...
+                        _ => throw new NotSupportedException($"'{propertyWithSerialization.ContainingType.Name}.{propertyWithSerialization.Name}' must be either field or property.")
+                    };
+                    var xml = propertyWithSerialization.GetDocumentationCommentXml();
+                    string comment = "";
+                    if (!string.IsNullOrEmpty(xml))
+                    {
+                        var xd = new global::System.Xml.XmlDocument();
+                        xd.LoadXml(xml);
+                        comment = xd.SelectSingleNode("/member/summary")?.InnerText.Trim() ?? "";
+                    }
```
ArcturusZhang, line 247: this is redundant - we are adding this property because this property has been defined in the customized code, therefore no matter how well we did for this property which will be written into the generated code, it will be removed again in post processor, because this property has been defined in the customization code and it cannot be defined in two places.
Please refer how it is done in `ModelTypeProviderFields`: https://github.com/RodgeFu/autorest.csharp/blob/seralize-add-property/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs#L305
Here we just need to put a placeholder for summary, description, etc.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -223,6 +224,51 @@ protected override ObjectTypeConstructor BuildSerializationConstructor()
...
+                {
+                    if (serializationInitializers.Exists(i => i.Property.Declaration.Name == propertyWithSerialization.Name))
+                        continue;
+                    var csharpType = BuilderHelpers.GetTypeFromExisting(propertyWithSerialization, typeof(object), MgmtContext.TypeFactory);
+                    var isReadOnly = propertyWithSerialization switch
+                    {
+                        IPropertySymbol propertySymbol => propertySymbol.SetMethod == null,
+                        IFieldSymbol fieldSymbol => fieldSymbol.IsReadOnly,
+                        _ => throw new NotSupportedException($"'{propertyWithSerialization.ContainingType.Name}.{propertyWithSerialization.Name}' must be either field or property.")
+                    };
```
ArcturusZhang, line 239: I believe this is identical to whatever we have in `ModelTypeProviderField`, we could extract this as a method in `BuilderHelpers`.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -223,6 +224,51 @@ protected override ObjectTypeConstructor BuildSerializationConstructor()
...
+                        new MemberDeclarationOptions(propertyWithSerialization.DeclaredAccessibility.ToString(), propertyWithSerialization.Name, csharpType),
+                        comment,
+                        isReadOnly,
+                        new Property()
+                        {
+                            Required = true,
+                            ReadOnly = isReadOnly,
+                            IsDiscriminator = false,
+                            SerializedName = propertyWithSerialization.Name,
+                        });
```
ArcturusZhang, line 258: This is the place I believe we should assign a property `Schema` for this made-up `Property` because the serialization builder needs this to identify if this should be serialized as a list or dictionary or normal property.
Just make a new method following the same pattern as we have in `ModelTypeProviderFields.GetInputTypeFromExistingMemberType` method to construct a new schema for it.

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -220,6 +221,21 @@ public static CSharpType GetTypeFromExisting(ISymbol existingMember, CSharpType
...
+                return new InputListType("Array", GetInputTypeFromCSharpType(type.Arguments[0]), TypeFactory.IsReadOnlyMemory(type), false);
+            }
+
+            if (TypeFactory.IsDictionary(type))
+            {
+                return new InputDictionaryType("Dictionary", InputPrimitiveType.String, GetInputTypeFromCSharpType(type.Arguments[1]), false);
+            }
+
+            return InputPrimitiveType.Object;
+        }
```
ArcturusZhang, line 237: because of the reason I said [here](https://github.com/Azure/autorest.csharp/pull/4435/files#r1535335683), please move this method back to where it was.

test/TestProjects/MgmtMockAndSample/src/Customize/ApplicationRule.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System.Collections.Generic;
+using Azure.Core;
+using MgmtMockAndSample;
+
+namespace MgmtMockAndSample.Models
+{
+    [CodeGenSerialization(nameof(NewIntSerializeProperty), "newIntSerializeProperty")]
+    [CodeGenSerialization(nameof(NewGeneratedTypeSerializeProperty), "newGeneratedTypeSerializeProperty")]
+    public partial class ApplicationRule : FirewallPolicyRule
```
ArcturusZhang, line 14: we also need to test this functionality in a project that enables model-reader-writer (short forMRW)
or
since in the near future, model-reader-writer will be enabled by default and disabling it will not be an option, I suggest we enable model-reader-writer for this project.

MRW can be enabled by adding a configuration
```
use-model-reader-writer: true
```
in the according `autorest.md`

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
RodgeFu: we iterate them twice to keep the original property order in serialization code. they can be merged if we dont need to keep the order. just let me know.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
RodgeFu: any existing code to generate schema from CSharpType?

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
RodgeFu: use lamda can further limit the impact scope of the method. but anyway, make it a private method to make it the same style of the rest code.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
RodgeFu: Do you mean have a switch for list/dict/object/string/int/binary/... to generate corresponding schema? any existing code to do so? or any sample code i can refer to cover all the possibility of the switch?

test/TestProjects/MgmtMockAndSample/src/Customize/ApplicationRule.cs
RodgeFu: verified the feature with use-model-reader-writer: true and the result is positive. But I would prefer not to include the change of the config in this PR which will introduce change in hundreds of test files which is not related to this feature. thanks.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
RodgeFu: seems the comment here will impact the parameter's comment in the generated code like below. please help to confirm.
image

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
ArcturusZhang: Please add a TODO item to explain that
in the meantime I still do not understand why there could be a change.
Your change should only append some new properties to the list of properties **only when** there are attributes assigning serialization metadata to the customized properties.
Nothing should happen to existing properties. I think we need to figure out why changes are happening to existing properties because this feature should change nothing on existing properties (including their order).

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
ArcturusZhang: We need to write one, taking the `GetInputTypeFromCSharpType` as reference.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
ArcturusZhang: looks like we did not consider it.
Then we should have this, and we should change it as well in `ModelTypeProviderFields`.
Please write a method in a common place to get the description from a symbol and use it here and in `ModelTypeProviderFields`

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
ArcturusZhang: I need to correct my previous statement: we actually do not need a `Property` here. The `SchemaProperty` property in `ObjectTypeProperty` is nullable, we just need to put it as null here.
This is the same thing we did for `ModelTypeProvider`.
You need to go over how we did it for `ModelTypeProvider` and use the same approach for `SchemaObjectType`.

How the property is built: https://github.com/Azure/autorest.csharp/blob/23dcd81c26074c182961ae4a10319915ec4026ff/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs#L132-L151
How the serialization is constructed: https://github.com/Azure/autorest.csharp/blob/23dcd81c26074c182961ae4a10319915ec4026ff/src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs#L302-L317
In `ModelTypeProvider` we just put the `InputModelProperty` property as null value, to indicate that this property is not included in the spec (which is the case you are trying to solve: add a new property to serialization, otherwise it is "changing the serialization of an existing property")

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
ArcturusZhang: well I just noticed that because of a recent update, this method is not used anywhere, therefore we should remove it.

test/TestProjects/MgmtMockAndSample/src/Customize/ApplicationRule.cs
ArcturusZhang: If so, please add the test to another project that enables MRW

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
RodgeFu: the order is changed because the order in SerializatoinCustructor is different in the order or property when there is inheritance. actually, only the 2nd foreach enough, the 1st foreach is only for keeping the original order of property

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
RodgeFu: look into the GetInputTypeFromCSharpType, it seems ObjectSchema is used for all type except list and dictionary, is that expected? shouldn't stringSchema used for string and so on?

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -447,7 +468,9 @@ private static void PopulatePropertyBag(SerializationPropertyBag propertyBag, in
         {
             foreach (var (property, serializationMapping) in propertyBag.Properties.ToArray())
             {
-                var schemaProperty = property.SchemaProperty!; // we ensure this is not null when we build the array
+                if (property.SchemaProperty == null)
+                    continue; // skip properties not from specification
+                Property schemaProperty = property.SchemaProperty;
```
ArcturusZhang, line 473: we should not skip the property when it is not from the specification.
The attribute `CodeGenSerialization` is able to assign an array of serialization names which mean the full serialization path

if you check the code below, the schemaProperty variable is barely used, we only use it as fallback value when the serializationMapping is null. When the property comes from the customization only, this mapping cannot be null, otherwise we will not have it here.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
RodgeFu: discussed offline, we should change the property directly, so no need this part change anymore

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
RodgeFu: set the description to string.empty to keep things simple consider all these methods are internal

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
RodgeFu: update to null. and found additionalProperties would also have schemaProperty=null. Some check is also added for that.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
RodgeFu: revert the change. we will use schemaProperty == null to do the check

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
RodgeFu: removed

test/TestProjects/MgmtMockAndSample/src/Customize/ApplicationRule.cs
RodgeFu: test added in MgmtCustomization

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
RodgeFu: fixed.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -478,8 +479,18 @@ private ObjectTypeDiscriminatorImplementation[] CreateDiscriminatorImplementatio
...
+            return EnumerateHierarchy()
+                .Skip(1)
+                .SelectMany(type => type.Properties)
+                .Select(p => p.Declaration.Name)
+                .ToHashSet();
+        }
+
         protected override IEnumerable<ObjectTypeProperty> BuildProperties()
         {
+            var propertiesFromSpec = GetParentPropertyDeclarationNames();
```
ArcturusZhang, line 493: well in the `ModelTypeProvider` variant, we are using the `ISymbol` to indicate whether the member has been handled. Here we are using their names, which I believe should work fine - still I have a concern when there is name collisions, whether our system could handle that edge case (symbol definitely will not have this issue).
It is fine that we keep it as is here right now, because in a near future, we will not have this class any more, but if there is any issue, we might keep in mind that this implementation potenially has this issue.

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -220,6 +221,13 @@ public static CSharpType GetTypeFromExisting(ISymbol existingMember, CSharpType
             return PromoteNullabilityInformation(newType, defaultType);
         }
 
+        public static bool IsTypeFromExistingReadOnly(ISymbol existingMember) => existingMember switch
```
ArcturusZhang, line 224: should this be `IsReadOnlyFromExisting`?
We are getting if the "property/field" is readonly instead of the type.

ArcturusZhang (approved):

PR #4459, opened 2024-03-25
Replace LookupDictionary with built-in Dictionary

live1206:
Similar to https://github.com/Azure/autorest.csharp/pull/4336


ArcturusZhang (approved):
I am not sure why we have to remove this but approved.
It ensures that the two dictionaries will always have the same/compatible content to have the class.

live1206:
> I am not sure why we have to remove this but approved. It ensures that the two dictionaries will always have the same/compatible content to have the class.

I found it's hard to extend it during https://github.com/Azure/autorest.csharp/pull/4432, for instance I would like to have a non-default comparer for the internal dictionaries.
If we take a look at the implementation of [LookupDictionary](https://github.com/Azure/autorest.csharp/pull/4459/files#diff-ebda8d24f48cd1a81392acc7ae1a7020f428af796c2701aaa889f9f3724145fdL30-L31), it holds two dictionaries, which is the same as we have in `MgmtOutputLibrary` now.

PR #4468, opened 2024-03-26
[experiment2][dpg] do not convert enum client parameters into string

archerzz:
# Description
Instead of checking the enum value type against protocol parameter type, we just not check types when building operation chain.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -498,7 +498,8 @@ public static bool IsIgnoredHeaderParameter(InputParameter operationParameter)
 
         private Parameter BuildParameter(in InputParameter operationParameter, Type? typeOverride = null)
         {
-            CSharpType type = typeOverride != null ? new CSharpType(typeOverride, operationParameter.Type.IsNullable) : _typeFactory.CreateType(operationParameter.Type);
+            CSharpType type = typeOverride != null ? new CSharpType(typeOverride, operationParameter.Type.IsNullable) :
+                operationParameter is { IsApiVersion: true, Type: InputEnumType enumType } ? _typeFactory.CreateType(enumType.EnumValueType) : _typeFactory.CreateType(operationParameter.Type);
```
archerzz, line 502: Special handling of `apiVersion`

src/AutoRest.CSharp/Common/Output/Models/ClientFields.cs
```diff
@@ -132,6 +133,9 @@ private ClientFields(IEnumerable<Parameter> parameters, InputAuth? authorization
         public FieldDeclaration? GetFieldByParameter(Parameter parameter)
             => GetFieldByParameter(parameter.Name, parameter.Type);
 
+        public FieldDeclaration? GetFieldByName(string name)
```
archerzz, line 136: Provide this method for cases when we don't check type matching. I cannot remove `GetFieldByPrarameter`, since in some cases we still need to check the type. For example, `NamesConflict` test project provide a case when client parameter and query parameter have the same name. We actually rely on the type check to decide which should be used. Not sure if that's the best way to do.

PR #4469, opened 2024-03-26
Support enum in spec to fixed enum in SDK

pshao25:
Part of #4209 

Because we still have a lot of known issues/gaps in TCGC, we fix at our side in this PR.


test/CadlRanchProjects/server/path/multiple/src/Generated/tspCodeModel.json
```diff
@@ -20,7 +20,7 @@
    ],
    "Namespace": "Server.Path.Multiple",
    "Description": "Service versions",
-   "IsExtensible": true,
+   "IsExtensible": false,
```
lirenhe, line 23: could you confirm whether this change is caused by some tsp file is not updated from enum to Union?
If that is the case, shall we update the cadl-ranch case directly?

test/TestProjects/FirstTest-TypeSpec/FirstTest-TypeSpec.tsp
```diff
@@ -517,7 +513,6 @@ op enumParameter(@path p1: ProjectedEnum ): void;
 
 
 @doc("enum with projected name used in a model")
-@fixed
 @projectedName("csharp", "CsProjectedEnumInModel")
```
lirenhe, line 516: we shall also clean up our usage for @projectedName in another PR.

test/CadlRanchProjects/server/path/multiple/src/Generated/tspCodeModel.json
pshao25: This is API version. Compiler only allows `enum` to define an API version. I don't think we should change cadl-ranch. And it's OK for us to make it `IsExtensible` false, because we handle it in a separate way.

test/TestProjects/FirstTest-TypeSpec/FirstTest-TypeSpec.tsp
pshao25: Sure, but we could still hold it until it is deprecated because we currently still support projected Name.

ArcturusZhang (approved):

lirenhe (approved):

PR #4472, opened 2024-03-26
encode-bytes test

mcgallan:
# Description

# Checklist

# Ready to Land?


test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -161,5 +161,42 @@ public class EncodeBytesTests : CadlRanchTestBase
             Response response = await new BytesClient(host, null).GetRequestBodyClient().Base64urlAsync(data);
             Assert.AreEqual(204, response.Status);
         });
+
+        [Test]
+        public Task Encode_Bytes_ResponseBody_default() => Test(async (host) =>
+        {
+            var response = await new BytesClient(host, null).GetResponseBodyClient().DefaultAsync();
+            BinaryData.Equals(BinaryData.FromObjectAsJson("dGVzdA=="), response.Value);
```
ArcturusZhang, line 169: as far as I know, `BinaryData` class does not have a static method `BinaryData.Equals(BinaryData, BinaryData)` defined, therefore here we are actually calling `object.Equals`, which is fine, but since we are in a test case, we need to assert that these two BinaryData instances are the same, we need to assert the result. This `Equals` method just returns true or false, it does not assert.

We should call either:
```csharp
Assert.AreEqual(BinaryData.FromObjectAsJson("dGVzdA=="), response.Value);
```
I believe this method will call the `Equals` method on the instance to compare the instances instead of just checking reference equal, or you could do this:
```csharp
Assert.IsTrue(BinaryData.Equals(BinaryData.FromObjectAsJson("dGVzdA=="), response.Value));
```
I prefer the first if it works because it has more semantics.

test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -161,5 +161,42 @@ public class EncodeBytesTests : CadlRanchTestBase
...
+        });
+
+        [Test]
+        public Task Encode_Bytes_ResponseBody_octetStream() => Test(async (host) =>
+        {
+            BinaryData data = new BinaryData(File.ReadAllBytes(SamplePngPath));
+            var response = await new BytesClient(host, null).GetResponseBodyClient().OctetStreamAsync();
+            BinaryData.Equals(data, response.Value);
```
ArcturusZhang, line 177: same here

test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -161,5 +161,42 @@ public class EncodeBytesTests : CadlRanchTestBase
...
+        });
+
+        [Test]
+        public Task Encode_Bytes_ResponseBody_customContentType() => Test(async (host) =>
+        {
+            BinaryData data = new BinaryData(File.ReadAllBytes(SamplePngPath));
+            var response = await new BytesClient(host, null).GetResponseBodyClient().CustomContentTypeAsync();
+            BinaryData.Equals(data, response.Value);
```
ArcturusZhang, line 185: same here

test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -161,5 +161,42 @@ public class EncodeBytesTests : CadlRanchTestBase
...
+        });
+
+        [Test]
+        public Task Encode_Bytes_ResponseBody_base64() => Test(async (host) =>
+        {
+            var response = await new BytesClient(host, null).GetResponseBodyClient().Base64Async();
+            BinaryData.Equals(BinaryData.FromObjectAsJson("dGVzdA=="), response.Value);
```
ArcturusZhang, line 192: same here

test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -161,5 +161,42 @@ public class EncodeBytesTests : CadlRanchTestBase
...
+        });
+
+        [Test]
+        public Task Encode_Bytes_ResponseBody_base64url() => Test(async (host) =>
+        {
+            var response = await new BytesClient(host, null).GetResponseBodyClient().Base64urlAsync();
+            BinaryData.Equals(BinaryData.FromObjectAsJson("dGVzdA"), response.Value);
```
ArcturusZhang, line 199: same here

test/CadlRanchProjects.Tests/encode-bytes.cs
mcgallan: fixed

test/CadlRanchProjects.Tests/encode-bytes.cs
mcgallan: fixed

test/CadlRanchProjects.Tests/encode-bytes.cs
mcgallan: fixed

test/CadlRanchProjects.Tests/encode-bytes.cs
mcgallan: fixed

test/CadlRanchProjects.Tests/encode-bytes.cs
mcgallan: fixed

test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -161,5 +162,44 @@ public class EncodeBytesTests : CadlRanchTestBase
             Response response = await new BytesClient(host, null).GetRequestBodyClient().Base64urlAsync(data);
             Assert.AreEqual(204, response.Status);
         });
+
+        [Test]
+        public Task Encode_Bytes_ResponseBody_default() => Test(async (host) =>
+        {
+            var response = await new BytesClient(host, null).GetResponseBodyClient().DefaultAsync();
+            var expected = BinaryData.FromObjectAsJson("dGVzdA==");
+            Assert.AreEqual(expected.ToString(), response.Value.ToString());
```
ArcturusZhang, line 171: well let us use the same way to assert binary data, or we could write a private static helper inside this class such as `AssertBinaryDataAreEqual` to assert them

mcgallan: fixed

test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -1,11 +1,14 @@
...
+using System.Buffers.Text;
 using System.IO;
 using System.Threading.Tasks;
 using AutoRest.TestServer.Tests.Infrastructure;
 using Azure;
 using Encode.Bytes;
 using Encode.Bytes.Models;
 using NUnit.Framework;
+using NUnit.Framework.Internal;
+using static System.Runtime.InteropServices.JavaScript.JSType;
```
ArcturusZhang, line 11: please clean up the unused usings

ArcturusZhang (approved):

test/CadlRanchProjects.Tests/encode-bytes.cs
mcgallan: fixed

test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -161,5 +160,43 @@ public class EncodeBytesTests : CadlRanchTestBase
             Response response = await new BytesClient(host, null).GetRequestBodyClient().Base64urlAsync(data);
             Assert.AreEqual(204, response.Status);
         });
+
+        [Test]
+        public Task Encode_Bytes_ResponseBody_default() => Test(async (host) =>
+        {
+            var response = await new BytesClient(host, null).GetResponseBodyClient().DefaultAsync();
+            CollectionAssert.AreEqual(BinaryData.FromObjectAsJson("dGVzdA==").ToString(), response.Value.ToString());
```
ArcturusZhang, line 168:
```suggestion
            CollectionAssert.AreEqual(BinaryData.FromObjectAsJson("dGVzdA==").ToArray(), response.Value.ToArray());
```

test/CadlRanchProjects.Tests/encode-bytes.cs
```diff
@@ -161,5 +160,43 @@ public class EncodeBytesTests : CadlRanchTestBase
             Response response = await new BytesClient(host, null).GetRequestBodyClient().Base64urlAsync(data);
             Assert.AreEqual(204, response.Status);
         });
+
+        [Test]
+        public Task Encode_Bytes_ResponseBody_default() => Test(async (host) =>
+        {
+            var response = await new BytesClient(host, null).GetResponseBodyClient().DefaultAsync();
```
chunyu3, line 167: The operation return Bytes, not Bytes Array.
You need to check the value, not convert it to array and check.

And all following tests have the same problem.

ArcturusZhang: this method is returning `Response<BinaryData>` therefore the result here is a `BinaryData`.

We use `ToArray` to verify the content of two `BinaryData` instances. The other approach is using `ToString`, these are identical.

What is your point?

chunyu3: Sorry. I made a mistake about the method `toArray`. Yes, we need to convert `BinaryData` to byte array and then compare.

PR #4491, opened 2024-03-27
Warn when client name is Models and model-namespace on

### LINKED ISSUE #4123: Remove hardcode for client name
Currently we have hard code "if client name == Models, then change it to ModelsOps" https://github.com/Azure/autorest.csharp/blob/50b816e46533a872f922976550b5c55909827695/src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts#L217-L219

We should remove it.

We do this hard code because if user defines a model named "Models", the compilation will throw error if `mode-namespace` is true. Like [this](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYAMBYAUBgRjz1QIE4AKAIkB4NwEH2qBKAbmNygEMBbAUwGcADhwDGPAAQBBKQDoAsgHs4PADZ88AbzwBfNnk69BI8VIliNeMZbEYxCpaotXNuK2J24tQA=).

Maybe could throw error if user defines a model named "Models" and `mode-namespace` is true. A step further, we could throw error if user defines a model whose name is the same as some part of namespace.

pshao25:
Fix #4123

After this change, our behavior is:
1. model-namespace is false: generate as defined in spec
2. model-namespace is true:
2.1 client name is "Models": give warning and rename to "ModelsOps"
2.2 client name is not "Models": generate as defined in spec


chunyu3 (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/clientModelBuilder.ts
```diff
@@ -219,7 +219,18 @@ export function createModelForService(
         }
 
         var clientName = getLibraryName(sdkContext, client.type);
-        return clientName === "Models" ? "ModelsOps" : clientName;
+        if (
+            clientName === "Models" &&
+            resolveOptions(sdkContext.emitContext)["model-namespace"] !== false
```
lirenhe, line 224: Do we have any case to verify the below?

model-namespace is true:
2.1 client name is "Models": give warning and rename to "ModelsOps"

pshao25: Yes. Platform-OpenAI-TypeSpec is a case.

PR #4494, opened 2024-03-27
Add Write methods to ValueExpression and MethodBodyStatement

AlexanderSher:
No changes in generated code


src/AutoRest.CSharp/Common/Generation/Writers/CodeWriterExtensions.cs
```diff
@@ -779,5 +776,52 @@ static void WritePropertyAccessorModifiers(CodeWriter writer, MethodSignatureMod
...
         }
+
+        public static void WriteTypeArguments(this CodeWriter writer, IEnumerable<CSharpType>? typeArguments)
+        {
+            if (typeArguments is null)
+            {
+                return;
+            }
+
+            writer.AppendRaw("<");
```
m-nash, line 787: do we need a TODO to normalize this with the other write arguments that handles the docs case where instead of `<` we use `{`?

AlexanderSher: Done

m-nash (approved):

PR #4497, opened 2024-03-28
Azure_Core_Basic and Client_Naming Test

mcgallan:
# Description

# Checklist

# Ready to Land?


test/CadlRanchProjects.Tests/azure-core-basic.cs
```diff
@@ -99,5 +99,35 @@ await foreach (Page<User> page in allPages.AsPages())
             User response = await new BasicClient(host, null).ExportAsync(1, "json");
             Assert.AreEqual("Madge", response.Name);
         });
+
+        [Test]
+        public Task Azure_Core_Basic_listWithParameters() => Test(async (host) =>
+        {
+            AsyncPageable<User> allPages = new BasicClient(host, null).GetWithParametersAsync(new ListItemInputBody("Madge"), ListItemInputExtensibleEnum.Second);
+            await foreach (Page<User> page in allPages.AsPages())
```
ArcturusZhang, line 107: we should just foreach over all the result, such as
```suggestion
            await foreach (var item in allPages)
```

test/CadlRanchProjects.Tests/azure-core-basic.cs
```diff
@@ -99,5 +99,35 @@ await foreach (Page<User> page in allPages.AsPages())
...
+            {
+                User firstUser = page.Values.First();
+                Assert.AreEqual(1, firstUser.Id);
+                Assert.AreEqual("Madge", firstUser.Name);
+                Assert.AreEqual("11bdc430-65e8-45ad-81d9-8ffa60d55b59", firstUser.Etag);
+            }
```
ArcturusZhang, line 113: This implementation implies that there should be only one item in the response.
We should also include this in the assertion.
Therefore, we should do something like this:
```suggestion
            AsyncPageable<User> allPages = new BasicClient(host, null).GetWithParametersAsync(new ListItemInputBody("Madge"), ListItemInputExtensibleEnum.Second);
            var results = new List<User>();
            await foreach (var item in allPages)
            {
            	results.Add(item);
            }
            Assert.AreEqual(1, results.Count);
            User user = results.First();
            Assert.AreEqual(1, user.Id);
            Assert.AreEqual("Madge", user.Name);
            Assert.AreEqual("11bdc430-65e8-45ad-81d9-8ffa60d55b59", user.Etag);
```
Same thing applies to the other test.

ArcturusZhang: I made some changes to the existing test cases: https://github.com/Azure/autorest.csharp/pull/4515

ArcturusZhang (approved):

test/CadlRanchProjects.Tests/client-naming.cs
```diff
@@ -60,10 +60,18 @@ public class ClientNamingTests: CadlRanchTestBase
...
-        [Ignore("https://github.com/Azure/autorest.csharp/issues/2493")]
         public Task Client_Naming_Header_response() => Test(async (host) =>
         {
             var response = await new NamingClient(host, null).ResponseAsync();
+            var results = response.Headers.Contains("default-name");
+            Assert.AreEqual(true, results);
+            var list = response.Headers.GetEnumerator();
+            list.MoveNext();
+            list.MoveNext();
+            var key = list.Current.Name;
```
pshao25, line 71: We should loop the headers and find a key named "default-name", then check whether its value is "true", instead of expecting it is the second item.

test/CadlRanchProjects.Tests/type-property-optionality.cs
```diff
@@ -385,7 +385,7 @@ public class TypePropertyOptionalTests : CadlRanchTestBase
         public Task Type_Property_Optional_UnionFloatLiteral_getAll() => Test(async (host) =>
         {
             Response response = await new OptionalClient(host, null).GetUnionFloatLiteralClient().GetAllAsync(new RequestContext());
-            BinaryData.Equals(BinaryData.FromObjectAsJson(2.375), UnionFloatLiteralProperty.FromResponse(response).Property);
+            Assert.AreEqual(BinaryData.FromObjectAsJson(2.375).ToString(), UnionFloatLiteralProperty.FromResponse(response).Property.ToString());
```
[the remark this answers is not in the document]
ArcturusZhang: Not the case. The `BinaryData` class does not have a `Equals` static method defined. Therefore this `BinaryData.Equals` is actually calling the `object.Equals` method, which compares the two argument by **reference**.
Please clarify, are we expecting the two BinaryData instance to be the same one? Or we are asserting "they have the same content".

Also I think we should use
```suggestion
            CollectionAssert.AreEqual(BinaryData.FromObjectAsJson(2.375).ToArray(), UnionFloatLiteralProperty.FromResponse(response).Property.ToArray());
```
because we are using this to assert the content of binarydata in the other PR

chunyu3 (approved):

test/CadlRanchProjects.Tests/type-property-optionality.cs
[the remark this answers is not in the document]
ArcturusZhang: @mcgallan please use this new method to assert BinaryData instances: https://github.com/Azure/autorest.csharp/pull/4570/files

test/CadlRanchProjects.Tests/client-naming.cs
mcgallan: fixed

test/CadlRanchProjects.Tests/type-property-optionality.cs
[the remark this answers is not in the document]
mcgallan: fixed

test/CadlRanchProjects.Tests/type-property-value-types.cs
```diff
@@ -401,7 +401,7 @@ public class TypePropertyValueTypesTests : CadlRanchTestBase
             var response = await new ValueTypesClient(host, null).GetUnionFloatLiteralClient().GetUnionFloatLiteralAsync();
             var expected = BinaryData.FromObjectAsJson(46.875);
             var actual = response.Value.Property;
-            BinaryData.Equals(expected, actual);
+            Assert.AreEqual(expected.ToString(), actual.ToString());
```
ArcturusZhang, line 404: we are missing one

mcgallan: fixed

ArcturusZhang (approved):

Issue #4484, opened 2024-03-26
[ClientModel-based clients] Move message.Apply(options) call to end of request-creation method


annelo-msft:
https://github.com/annelo-msft/gists/blob/main/clientmodel-client/readme.md#type-usage

Issue #4482, opened 2024-03-26
[ClientModel-based clients] Rename `context` parameter name in protocol methods to `options`


annelo-msft:
This is part of the public API surface, so should match the ClientModel type name.

Issue #4243, opened 2024-02-20
Enable the new cancellation token pattern in the debranded generator


pallavit:
More information of the pattern can be found at - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/ProtocolMethods.md

Also, given this change will be a breaking change it is something we will have to absorb before we GA a debranded SDK.

Scope unclear at the moment.

annelo-msft:
The pattern for CancellationTokens in unbranded clients is variously documented here:
- https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/core/System.ClientModel#customizing-http-requests
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/System.ClientModel/samples/ClientImplementation.md#implementing-protocol-methods
- https://github.com/annelo-msft/gists/blob/main/clientmodel-client/readme.md#client-apis

A short summary is:
- Convenience methods don't take CancellationTokens
- In unbranded clients, passing a CancellationToken is considered an "advanced scenario" -- users who want to do this must set the `CancellationToken` property on `RequestOptions` and call the protocol method.

Issue #3768, opened 2023-09-22
Consolidate Rest Client service method generation


AlexanderSher:
Consolidate generation of Rest Client (for DPG - public client is treated as Rest Client) service methods with ability to switch between HLC-style method structure, e.g.:

```
CreateXxxRequest(BinaryData bd)
{
    var rc = new RequestContent();
    rc.Writer.Write(bd.ToArray());
    request.RequestContent = rc;
}

Response<Y> Xxx(BinaryData bd)
{
    ///
    var message = CreateXxxRequest(bd);
    ///
}
```

to DPG-style method structure
```
Response<Y> XxxValue(BinaryData bd)
{
    var rc = new RequestContent();
    rc.Writer.Write(bd.ToArray());
    var response = Xxx(rc);
    return (Y)response;
}

Response Xxx(RequestContent rc)
{
    var message = CreateXxxRequest(rc);
    ///
}

CreateXxxRequest(RequestContent rc)
{
    request.RequestContent = rc;
}
```

ArthurMa1978:
Push to MGC plugin

PR #4502, opened 2024-03-29
move the `message.Apply` to the last call / change parameter name of `RequestOptions` to `options` in non-branded case

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4484
Fixes https://github.com/Azure/autorest.csharp/issues/4482

# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -28,13 +29,13 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal static class RequestWriterHelpers
     {
-        public static void WriteRequestCreation(CodeWriter writer, RestClientMethod clientMethod, string methodAccessibility, ClientFields? fields, string? responseClassifierType, bool writeSDKUserAgent, IReadOnlyList<Parameter>? clientParameters = null)
+        public static void WriteRequestCreation(CodeWriter writer, RestClientMethod clientMethod, ClientFields? fields, string? responseClassifierType, bool writeSDKUserAgent, IReadOnlyList<Parameter>? clientParameters = null)
```
ArcturusZhang, line 32: parameter `methodAccessibility` is removed here because in all the places we are invoking this method, we always pass in the value "internal".
Therefore for simplicity, I removed it and hard-coded it in the method.

We should schedule a refactor to this method to use the expressions to build the method instead of using strings, and when we do that, we could add the parameter back via `MethodSignatureModifiers`

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -28,13 +28,13 @@ namespace AutoRest.CSharp.Generation.Writers
 {
     internal static class RequestWriterHelpers
     {
-        public static void WriteRequestCreation(CodeWriter writer, RestClientMethod clientMethod, string methodAccessibility, ClientFields? fields, string? responseClassifierType, bool writeSDKUserAgent, IReadOnlyList<Parameter>? clientParameters = null)
+        public static void WriteRequestCreation(CodeWriter writer, RestClientMethod clientMethod, ClientFields fields, string? responseClassifierType, bool writeSDKUserAgent, IReadOnlyList<Parameter>? clientParameters = null)
```
ArcturusZhang, line 31: I change the parameter `fields` to non-nullable because in all the use case of this method, it is never null to reduce complexity.

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -252,7 +254,7 @@ private static ReferenceOrConstant GetReferenceForQueryParameter(ClientFields? f
             return fields.GetFieldByParameter(value.Reference.Name, value.Reference.Type) ?? value;
         }
 
-        public static void WriteHeaders(CodeWriter writer, RestClientMethod clientMethod, CodeWriterDeclaration request, bool content, ClientFields? fields)
+        public static void WriteHeaders(CodeWriter writer, RestClientMethod clientMethod, CodeWriterDeclaration request, bool content, ClientFields fields)
```
ArcturusZhang, line 257: make `fields` non-nullable because it is never null

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -296,7 +298,7 @@ private static MethodBodyStatement GetRequestContentForSerialization(CodeWriterD
             };
         }
 
-        internal static void WriteHeader(CodeWriter writer, CodeWriterDeclaration request, RequestHeader header, ClientFields? fields)
+        internal static void WriteHeader(CodeWriter writer, CodeWriterDeclaration request, RequestHeader header, ClientFields fields)
```
ArcturusZhang, line 301: make `fields` non-nullable because it is never null

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -333,10 +335,8 @@ internal static void WriteHeader(CodeWriter writer, CodeWriterDeclaration reques
             }
         }
 
-        internal static void WriteHeaderSystem(CodeWriter writer, CodeWriterDeclaration request, RequestHeader header, ClientFields? fields)
+        internal static void WriteHeaderSystem(CodeWriter writer, CodeWriterDeclaration request, RequestHeader header, ClientFields fields)
```
ArcturusZhang, line 338: make `fields` non-nullable because it is never null

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
```diff
@@ -333,10 +335,8 @@ internal static void WriteHeader(CodeWriter writer, CodeWriterDeclaration reques
...
         {
-            string? delimiter = header.Delimiter;
```
ArcturusZhang, line 338: removed because it is never used.

ArcturusZhang:
@m-nash @annelo-msft  could you please have a review?

chunyu3 (approved):

m-nash (approved):
LGTM @annelo-msft do the resulting client changes look correct?

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
m-nash: Can you link a tracking issue for this?

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/NoTestTypeSpecClient.cs
```diff
@@ -167,6 +163,10 @@ internal PipelineMessage CreateSayHiRequest(string headParameter, string queryPa
             request.Uri = uri.ToUri();
             request.Headers.Set("head-parameter", headParameter);
             request.Headers.Set("Accept", "application/json");
+            if (options != null)
```
annelo-msft, line 166: Why do we need the null check?  I think it is better to throw here to indicate an error.  If we instead don't apply the options to the message, we could get strange behavior from the pipeline without having identified the root of the problem -- it will make it harder to find bugs in either client or user code.

test/CadlRanchProjectsNonAzure/authentication/http/custom/src/Generated/CustomClient.cs
```diff
@@ -61,13 +61,13 @@ public CustomClient(Uri endpoint, ApiKeyCredential credential, CustomClientOptio
...
-        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <param name="options"> The request options, which can override default behaviors of the client pipeline on a per-call basis. </param>
         /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
         /// <returns> The response returned from the service. </returns>
-        public virtual async Task<ClientResult> ValidAsync(RequestOptions context = null)
+        public virtual async Task<ClientResult> ValidAsync(RequestOptions options = null)
         {
-            using PipelineMessage message = CreateValidRequest(context);
-            return ClientResult.FromResponse(await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false));
+            using PipelineMessage message = CreateValidRequest(options);
```
annelo-msft, line 69: In protocol methods where options can be null or default, we'll need to add this line:

```
options ??= new RequestOptions();
```

You can see https://github.com/annelo-msft/openai-in-typespec/pull/3/files#diff-e44e45d43a3dd8ac6d3c5c5fc6cacb2ce2b5f138500a60cc98237a4500828f89R86-R93 for details.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/AudioTranscriptions.cs
```diff
@@ -48,9 +48,9 @@ public virtual async Task<ClientResult<CreateTranscriptionResponse>> CreateAsync
         {
             Argument.AssertNotNull(audio, nameof(audio));
 
-            RequestOptions context = FromCancellationToken(cancellationToken);
+            RequestOptions options = FromCancellationToken(cancellationToken);
```
annelo-msft, line 51: Not sure if you plan to handle this in this or a later PR, but:
1. Convenience methods in unbranded clients should not take CancellationTokens
2. Once 1. is addressed, you can remove the `FromCancellationToken` helper method from unbranded clients

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/NoTestTypeSpecClient.cs
m-nash: It's because parameter from the public API is optional so this can be null here.

test/CadlRanchProjectsNonAzure/authentication/http/custom/src/Generated/CustomClient.cs
m-nash: I don't think we should be instantiating an object if we don't need that object.  It's just extra memory for now reason.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/AudioTranscriptions.cs
m-nash: Yes remove cancellation token is tracked here https://github.com/Azure/autorest.csharp/issues/4243

test/CadlRanchProjectsNonAzure/authentication/http/custom/src/Generated/CustomClient.cs
annelo-msft: @m-nash and I discussed this offline.  I see the concern and I think it's probably OK to do it the way it's proposed in this PR.  I want to double-check with @KrzysztofCwalina that this approach doesn't cause problems for the RequestOptions freezing story... I can't think offhand of how it would, but don't know if there is a case I haven't considered here.

We'll talk to @KrzysztofCwalina about it in the Azure.Core meeting on Thursday, and can loop back then.  Since I think it's reasonably low-risk, let's not block this PR on the question and we can update the generated code if we learn we need to.

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/NoTestTypeSpecClient.cs
annelo-msft: Same response as https://github.com/Azure/autorest.csharp/pull/4502#discussion_r1548462971

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/AudioTranscriptions.cs
annelo-msft: Sounds good.

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/NoTestTypeSpecClient.cs
ArcturusZhang: @annelo-msft please create an issue for this if we plan to change this.
I will resolve it for now - but if we decide to change this, we could do that as follow ups

test/CadlRanchProjectsNonAzure/authentication/http/custom/src/Generated/CustomClient.cs
ArcturusZhang: please create issues if follow ups are needed. I would like to resolve these to finish this PR to unblock others, thank you

src/AutoRest.CSharp/Common/Generation/Writers/RequestWriterHelpers.cs
ArcturusZhang: I believe it is tracking here by Alexander: https://github.com/Azure/autorest.csharp/issues/3768

PR #4507, opened 2024-03-31
Bicep - write flattened properties fix

JoshLove-msft:
Instead of https://github.com/Azure/autorest.csharp/pull/4503


samples/AzureSample.ResourceManager.Sample/src/Generated/ImageData.Serialization.cs
```diff
@@ -362,13 +357,18 @@ private BinaryData SerializeBicep(ModelReaderWriterOptions options)
...
             builder.AppendLine(" {");
-            hasPropertyOverride = hasObjectOverride && propertyOverrides.TryGetValue(nameof(SourceVirtualMachine), out propertyOverride);
+            hasPropertyOverride = hasObjectOverride && propertyOverrides.TryGetValue("SourceVirtualMachineId", out propertyOverride);
             if (Optional.IsDefined(SourceVirtualMachine) || hasPropertyOverride)
             {
                 builder.Append("    sourceVirtualMachine: ");
                 if (hasPropertyOverride)
                 {
-                    builder.AppendLine($"{propertyOverride}");
+                    builder.AppendLine("{");
```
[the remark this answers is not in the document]
JoshLove-msft: When the property gets a second value, it will become unflattened. The user's override would no longer work, but it would be difficult to handle this scenario because the user could override the entire envelope when the property is unflattened. We would need to come up with an algorithm for determining precedence. I would suggest that we punt this for now given that we can address it a non-breaking way if needed in the future.

test/TestProjects/MgmtDiscriminator/src/Generated/Models/Unflattened.cs
```diff
@@ -0,0 +1,71 @@
...
+        internal Unflattened(string unflattenedName, string value, IDictionary<string, BinaryData> serializedAdditionalRawData)
+        {
+            UnflattenedName = unflattenedName;
+            Value = value;
+            _serializedAdditionalRawData = serializedAdditionalRawData;
+        }
+
+        /// <summary> The name. </summary>
+        [WirePath("unflattenedName")]
+        public string UnflattenedName { get; set; }
```
m-nash, line 66: Can we call this `Name`?

test/TestProjects/MgmtDiscriminator/src/Generated/DeliveryRuleData.Serialization.cs
```diff
@@ -461,6 +477,33 @@ private BinaryData SerializeBicep(ModelReaderWriterOptions options)
                 }
             }
 
+            hasPropertyOverride = hasObjectOverride && propertyOverrides.TryGetValue("UnflattenedName", out propertyOverride);
+            if (Optional.IsDefined(Unflattened) || hasPropertyOverride)
```
m-nash, line 481: I feel like this if is redundant since its duplicated inside and can be removed.

test/TestProjects/MgmtDiscriminator/src/Generated/DeliveryRuleData.Serialization.cs
```diff
@@ -461,6 +477,33 @@ private BinaryData SerializeBicep(ModelReaderWriterOptions options)
...
+            {
+                builder.Append("  unflattened: ");
```
m-nash, line 483: If you do the above comment on line 481 this would need to go inside the if on 484 and the else on 501.

test/TestProjects/MgmtDiscriminator/src/Generated/DeliveryRuleData.Serialization.cs
```diff
@@ -461,6 +477,33 @@ private BinaryData SerializeBicep(ModelReaderWriterOptions options)
...
+                    }
+                    else
+                    {
+                        Dictionary<string, string> propertyDictionary = new Dictionary<string, string>();
+                        propertyDictionary.Add("UnflattenedName", propertyOverride);
+                        bicepOptions.PropertyOverrides.Add(Unflattened, propertyDictionary);
+                        BicepSerializationHelpers.AppendChildObject(builder, Unflattened, options, 2, false, "  unflattened: ");
+                    }
+                }
+                else
```
m-nash, line 501: if you remove the outer if this would need an `else if (Unflattened.IsDefined)`

test/TestProjects/MgmtDiscriminator/src/Generated/DeliveryRuleData.Serialization.cs
```diff
@@ -334,172 +345,217 @@ private BinaryData SerializeBicep(ModelReaderWriterOptions options)
...
-                if (hasPropertyOverride)
-                {
-                    builder.AppendLine($"{propertyOverride}");
-                }
-                else
+                builder.AppendLine(propertyOverride);
+            }
+            else
+            {
+                if (Optional.IsDefined(LocationWithCustomSerialization))
```
m-nash, line 386: nit: can we do an else if on the one line?

m-nash (approved):

PR #4510, opened 2024-04-01
Consolidate `AdditionalProperties` with raw data field (Phase 1)

### LINKED ISSUE #4128: [Bug][Serialization][AdditionalProperties Phase 1] Exception will be thrown when the model accepts AdditionalProperties with a value type of non-BinaryData
when a model defined with `string` typed additional properties property, we have the following deserialization code:
```csharp
internal static ExtendsStringAdditionalProperties DeserializeExtendsStringAdditionalProperties(JsonElement element, ModelReaderWriterOptions options = null)
{
    options ??= new ModelReaderWriterOptions("W");

    if (element.ValueKind == JsonValueKind.Null)
    {
        return null;
    }
    string name = default;
    IDictionary<string, string> additionalProperties = default;
    Dictionary<string, string> additionalPropertiesDictionary = new Dictionary<string, string>();
    foreach (var property in element.EnumerateObject())
    {
        if (property.NameEquals("name"u8))
        {
            name = property.Value.GetString();
            continue;
        }
        additionalPropertiesDictionary.Add(property.Name, property.Value.GetString());
    }
    additionalProperties = additionalPropertiesDictionary;
    return new ExtendsStringAdditionalProperties(name, additionalProperties);
}
```
if we use this json:
```
{
	"name": "the name",
	"addtional1": "1",
	"addtional2": 2
}
```
as input to deserialize the model, we will get a `JsonException` on this line:
```
                additionalPropertiesDictionary.Add(property.Name, property.Value.GetString());
```
when we try to consume "addtional2" property in the json (because it is a number).

Furthermore, the public serialization feature requires the models to be roundtrip when serializing and deserializing using the `J` format, in this case, it will never be roundtrip because no property or field could take the number value additional property `additional2`.

ArcturusZhang:
Fixes #4128 

For phases, please refer https://github.com/Azure/autorest.csharp/issues/4465

# Description

1. Support the `...Record<T>` syntax for additional properties.
2. Update the serialization code so that now when we have additional properties of primitive types, when invalid value comes in, we no longer throw exceptions.

Main changes in the generator:
1. some changes in the emitter to support `...Record<T>` syntax for additional properties. I think we do not have to have a perfect solution for that because the perfect solution for this is adopting `getAllModels` from TCGC, therefore that change is just making it work.
2. I changed the JSON serialization to contain two things: an `AdditionalPropertiesSerialization` and a `RawDataFieldSerialization` since a model might have both when its additionalProperties property cannot catch everything up.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonAdditionalPropertiesSerialization.cs
```diff
@@ -10,12 +10,12 @@ namespace AutoRest.CSharp.Output.Models.Serialization.Json
 {
     internal class JsonAdditionalPropertiesSerialization : JsonPropertySerialization
     {
-        public CSharpType Type { get; }
+        public CSharpType ImplementationType { get; }
```
ArcturusZhang, line 13: I renamed this property because this is confusing.
This `Type` property is not the actual type of the additional properties property (which should be `IDictionary<string, BinaryData>` for example), but it is the implementation type of the dictionary interface (`Dictionary<string, BinaryData>`) therefore here we add the prefix back.

ArcturusZhang:
Now BinaryData is hard-coded in ModelTypeProvider, but HLC also uses ModelTypeProvider now, but its corresponding type is `object`. I need to figure out how to do that.

ArcturusZhang:
Waiting for [this PR](https://github.com/Azure/autorest.csharp/pull/4603) to merge first...

lirenhe (approved):

src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonAdditionalPropertiesSerialization.cs
ArthurMa1978: The name is still a little bit weird, but fine if there is no better choice

ArthurMa1978 (approved):

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -358,5 +357,66 @@ public static MethodSignatureModifiers MapModifiers(ISymbol symbol)
...
+            var arguments = originalType.Arguments;
+            var keyType = arguments[0];
+            var valueType = arguments[1];
+
+            return originalType.MakeGenericType(new[] { ReplaceUnverifiableType(keyType, unknownType), ReplaceUnverifiableType(valueType, unknownType) });
+        }
+
+        private static CSharpType ReplaceUnverifiableType(CSharpType type, CSharpType unknownType)
+        {
+            // when the type is System.Object
```
live1206, line 373:
```suggestion
            // when the type is System.Object Or BinaryData
```

src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs
```diff
@@ -358,5 +357,66 @@ public static MethodSignatureModifiers MapModifiers(ISymbol symbol)
...
+            if (type.EqualsIgnoreNullable(unknownType))
+            {
+                return type;
+            }
+
+            // when the type is a verifiable type
+            if (BuilderHelpers.IsVerifiableType(type))
```
live1206, line 380:
```suggestion
            if (IsVerifiableType(type))
```

live1206 (approved):

m-nash (approved):

Issue #4385, opened 2024-03-12
Add support in codegen for unbranded codegen attributes (e.g., `CodeGenModel`, `CodeGenMember`, etc.)


joseharriaga:
Today, codegen checks for the fully qualified name of the attribute, which includes the Azure.Core namespace. See here, for example: https://github.com/Azure/autorest.csharp/blob/bdf44309240bc6f8347a737224de9614e88d59ac/src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs#L84

We need to update the search to not include namespace.

For testing purposes, it is possible to add custom code to one of the unbranded test projects to try it out locally.

PR #4511, opened 2024-04-01
Add codegen attribute tags

nisha-bhatia:
Fixes: https://github.com/Azure/autorest.csharp/issues/4385
# Description

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/Source/CodeGenAttributes.cs
```diff
@@ -15,25 +15,24 @@ public class CodeGenAttributes
...
-            CodeGenMemberAttribute = GetSymbol(compilation, typeof(CodeGenMemberAttribute));
-            CodeGenTypeAttribute = GetSymbol(compilation, typeof(CodeGenTypeAttribute));
-            CodeGenModelAttribute = GetSymbol(compilation, typeof(CodeGenModelAttribute));
-            CodeGenClientAttribute = GetSymbol(compilation, typeof(CodeGenClientAttribute));
-            CodeGenSerializationAttribute = GetSymbol(compilation, typeof(CodeGenSerializationAttribute));
+            CodeGenMemberAttributeName = "CodeGenMemberAttribute";
+            CodeGenTypeAttributeName = "CodeGenTypeAttribute";
+            CodeGenModelAttributeName = "CodeGenModelAttribute";
+            CodeGenClientAttributeName = "CodeGenClientAttribute";
+            CodeGenSerializationAttributeName = "CodeGenSerializationAttribute";
```
ArcturusZhang, line 22: we could make these as const strings as well

ArcturusZhang (commented):
As a general comment, I kind of suggest we unify the way of introducing customization code via attributes in branded and non-branded libraries.
The generator could generate those attributes, in this way we do not have to write the attributes (some of them contain quite a few properties).
If we choose this way, we might want to generate those attributes in a fixed namespace (such as something like `Generator.Internal`) instead of in the namespace of the library (because of the [apicomat settings](https://github.com/Azure/azure-sdk-for-net/blob/00af2ba6303ae616ba247cba0415151fe351ddb4/eng/ApiListing.exclude-attributes.txt#L6) in azure-sdk-for-net, as far as we know, this does not support wild cards)

m-nash:
@ArcturusZhang I do think unifying is what we want to achieve, however I think in mgc these will be public attributes instead of making customers create their own copies of them.

ArcturusZhang:
> @ArcturusZhang I do think unifying is what we want to achieve, however I think in mgc these will be public attributes instead of making customers create their own copies of them.

yeah making them public is a way, but I think if we let the generator to generate those attribute, in this way, we do not have to make those attribute public, which might be valuable since the public API is reduced, and those attributes are useless outside this generated library, therefore I think if possible to keep them internal, we should.

Customers are impossible to use these attributes without a document, therefore I am imagining this way of using them:
The generator will always generate them in the same namespace, but since no classes are using them, they will be removed by post processor (I think we still will have this in MGC, right?), and the customer reads the document to know that I need to add this attribute on the customized code and regenerate, they just write the code, despite the attribute does not exist, after they regenerate, the attribute now has a usage, and will no longer be removed, and everything should work as expected.

What do you think? I think this is a way to do both "not make them public" and "make everything to work properly" considering the fact that every time we add customization code, we will have to regenerate the library again

m-nash:
@ArcturusZhang opened a tracking issue to finalize on this https://github.com/Azure/autorest.csharp/issues/4551

m-nash (approved):

Issue #4485, opened 2024-03-26
[ClientModel-based clients] Update ClientPipelineExtensions to use ClientModel patterns


annelo-msft:
As described in:
- https://github.com/annelo-msft/openai-in-typespec/pull/1
- https://github.com/annelo-msft/gists/blob/main/clientmodel-client/readme.md#type-removal

PR #4516, opened 2024-04-02
resolves SCM changes in `ClientPipelineExtensions`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4485

# Description

# Checklist

# Ready to Land?


ArcturusZhang:
@m-nash @annelo-msft this PR is ready to review

test/CadlRanchProjectsNonAzure/authentication/http/custom/src/Generated/Internal/ClientPipelineExtensions.cs
```diff
@@ -2,53 +2,41 @@
...
 {
     internal static class ClientPipelineExtensions
     {
-        public static async ValueTask<PipelineResponse> ProcessMessageAsync(this ClientPipeline pipeline, PipelineMessage message, RequestOptions requestContext, CancellationToken cancellationToken = default)
+        public static async ValueTask<PipelineResponse> ProcessMessageAsync(this ClientPipeline pipeline, PipelineMessage message, RequestOptions options)
         {
             await pipeline.SendAsync(message).ConfigureAwait(false);
 
-            if (message.Response == null)
+            if (message.Response.IsError && options?.ErrorOptions == ClientErrorBehaviors.Default)
```
annelo-msft, line 17: Do we have unit tests validating the following cases?
1. If an end-user does not pass RequestOptions, the client will throw on an error response
2. If an end-user passes RequestOptions with ErrorOptions == Default, the client will throw on an error response
3. If an end-user passes RequestOptions with ErrorOptions == NoThrow, the client will _not_ throw on an error response

It would be good to validate these to ensure the new logic is correct.

m-nash (approved):

test/CadlRanchProjectsNonAzure/authentication/http/custom/src/Generated/Internal/ClientPipelineExtensions.cs
ArcturusZhang: Hi @annelo-msft I added some test cases here: [`7bf996f` (#4516)](https://github.com/Azure/autorest.csharp/pull/4516/commits/7bf996f55cc33df94f0bd1625ae0373ce0a5ca14)

but one of them `ValidateClientPipelineExtensions_Error_NoOptions` is not behaving like you described.
This case, when there is an error in response, and a request options is not specified, what should we do? Should we throw?
If so, the implementation in the extension should change, the current condition is:
```csharp
message.Response.IsError && options?.ErrorOptions == ClientErrorBehaviors.Default
```
when it is error, the first condition is `true`, and the second is never true when there is not a options specified.
We could change it in this way:
```csharp
message.Response.IsError && options?.ErrorOptions == ClientErrorBehaviors.NoThrow
```
to assert that we only omit the throwing when `NoThrow` is explicitly specified in the options. Or we add a check for options being null
```csharp
options == null || (message.Response.IsError && options.ErrorOptions == ClientErrorBehaviors.Default)
```
but this will let the pipeline always throw when an options is not specified even if the response is not an error.

How do you would like to make the change?

annelo-msft: Yeah, I like

> `message.Response.IsError && options?.ErrorOptions != ClientErrorBehaviors.NoThrow`

best.  I think it makes it clear that the client user has to explicitly pass a RequestOptions instance and that instance must explicitly set ErrorOptions to `NoThrow` to opt-out of the default exception-throwing behavior.

And actually, looking at it more closely, because [ClientErrorBehavior](https://apiview.dev/Assemblies/Review/1b123e7a51d44ebe945f0212ee039c65#System.ClientModel.Primitives.ClientErrorBehaviors) is a flags enum, I think we should maybe modify it a bit to 

```csharp
if (message.Response.IsError &&
    (options?.ErrorOptions & ClientErrorBehaviors.NoThrow) != ClientErrorBehaviors.NoThrow)
{
   // throw
}
```

The idea is if we later were to add a `NoLog` or `NoTrace` value to the flags enum, we would still only opt-out of throwing if `NoThrow` had been set explicitly by the end user.  I think `options?.ErrorOptions != ClientErrorBehaviors.NoThrow` alone would not accomplish this if both `NoThrow` and `NoLog` were set, for example.

Thanks!

ArcturusZhang: thank you for your suggestion. I did not notice the enum has `Flags`.
Code updated accordingly.

annelo-msft (approved):
I'm only reviewing the outputs under `test/`, but these LGTM.  Thanks for driving this @ArcturusZhang!

Issue #4480, opened 2024-03-26
[ClientModel-based clients] ClientUriBuilder does not escape special characters directly


annelo-msft:
This is because it uses the BCL UriBuilder type under the hood, which always escapes path values.  Escaping is not idempotent so double-escaping can cause functional problems.  We need to investigate the right thing for unbranded clients.

ArcturusZhang:
We need to investigate whether we need the `escape` parameter, and/or we need to manually escape values when append path/query
<To be edited later>

PR #4519, opened 2024-04-02
resolves SCM changes `ClientUriBuilder`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4567

# Description

We have this `if` statement in `ClientUriBuilder` to avoid duplicated separators:
```csharp
if (value[0] == PathSeparator)
{
    value = value.Substring(1);
}
```
but in some cases it removes the separator incorrectly when there is no duplicate (as described in the above issue).
This PR refines the check to ensure neither of these happens

# Checklist

# Ready to Land?


test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -42,11 +41,6 @@ public void AppendPath(string value, bool escape)
                 value = Uri.EscapeDataString(value);
             }
 
-            if (value[0] == PathSeparator)
-            {
-                value = value.Substring(1);
-            }
```
ArcturusZhang, line 48: this part might still be useful, because sometimes, we might have uris like this `//authentication/http/custom/valid` which leads to a request failure.

lirenhe (approved):

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -42,7 +42,7 @@ public void AppendPath(string value, bool escape)
                 value = Uri.EscapeDataString(value);
             }
 
-            if (value[0] == PathSeparator)
+            if (value[0] == PathSeparator && PathBuilder.Length > 0 && PathBuilder[PathBuilder.Length - 1] == PathSeparator)
```
annelo-msft, line 45: Would something like:

```csharp
        if (PathBuilder[PathBuilder.Length - 1] == '/' && value[0] == '/')
        {
            PathBuilder.Remove(PathBuilder.Length - 1, 1);
        }
```

I had started drafting some changes to this in https://github.com/Azure/azure-sdk-for-net/pull/43436, to implement test coverage I was thinking we could eventually move into the generator code base.

The reason to remove from the builder instead of calling value.Substring is that substring creates a new string instance which makes an allocation.

test/UnbrandedProjects.Tests/ClientUriBuilderTests.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+using NUnit.Framework;
+using UnbrandedTypeSpec;
+
+namespace UnbrandedProjects.Tests
+{
+    public class ClientUriBuilderTests
+    {
+        [TestCaseSource(nameof(AppendPathSource))]
+        public void ValidateAppendPath_ToUri(Uri expected, Uri endpoint, params string[] paths)
```
annelo-msft, line 13: Thanks for adding these!  I ❤️ good test coverage.

It would be good to ensure that the ClientUriBuilder type we're adding has full test coverage of all supported scenarios.  I had started adding some of these in this PR (https://github.com/Azure/azure-sdk-for-net/pull/43436/files#diff-2bc03b2074b965c69d675ff1d7caae6aa3d37d0210268944974f02213377c264R7) -- if we intend for this adapter type to support call cases from Azure.Core, Azure.Core's test suite has many tests for this.  I don't think we _do_ want full scenario parity across the Azure.Core type and the emitted type, but that's work that remains to be explored, esp around escaping -- i.e. we don't want to inadvertently ship a client that escapes a path segment multiple times and makes an invalid request to the service.

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
ArcturusZhang: Yeah I know that we should avoid of using substring - my original plan is to replace it with `ReadOnlySpan.Slice` but I think your solution is simpler, I will change the code to this.

test/UnbrandedProjects.Tests/ClientUriBuilderTests.cs
ArcturusZhang: Thank you! I am adding these test cases only to validate that my change could fix the issue I am trying to resolve because this work has an ETA of the end of this week.
Please let me know when your test cases are done, I could migrate them here as well.

test/UnbrandedProjects/NoTest-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
ArcturusZhang: BTW do we need to check `PathBuilder.Length > 0` first? Or this is somehow guaranteed always to be true?

test/UnbrandedProjects.Tests/ClientUriBuilderTests.cs
annelo-msft: If it's possible to add test cases for existing functionality that we want to prevent regressing, that would be desired to have.

test/UnbrandedProjects.Tests/ClientUriBuilderTests.cs
```diff
@@ -2,65 +2,460 @@
...
+            builder.AppendPath(value, escape);
+
+            Assert.AreEqual(expected, builder.ToUri().ToString());
+        }
+
+        [TestCase("http://localhost", "6/30/1905 1:14:00 PM +00:00", "D", true, "http://localhost/1905-06-30")]
+        [TestCase("http://localhost", "6/30/1905 1:14:00 PM +00:00", "U", true, "http://localhost/-2035622760")]
+        [TestCase("http://localhost", "6/30/1905 1:14:00 PM +00:00", "O", true, "http://localhost/1905-06-30T13%3A14%3A00.0000000Z")]
+        [TestCase("http://localhost", "6/30/1905 1:14:00 PM +00:00", "o", true, "http://localhost/1905-06-30T13%3A14%3A00.0000000Z")]
+        [TestCase("http://localhost", "6/30/1905 1:14:00 PM +00:00", "R", true, "http://localhost/Fri%2C 30 Jun 1905 13%3A14%3A00 GMT")] // TODO -- why spaces are not escaped?
```
ArcturusZhang, line 263: should we also escape the spaces? valid uri cannot contain spaces.

test/UnbrandedProjects.Tests/ClientUriBuilderTests.cs
```diff
@@ -0,0 +1,461 @@
...
+        [TestCase("http://localhost:12345/some/", "one|1", "http://localhost:12345/some/?one=1")]
+        [TestCase("http://localhost", "one|1|two|2", "http://localhost/?one=1&two=2")]
+        [TestCase("http://localhost:12345", "one|1|two|2", "http://localhost:12345/?one=1&two=2")]
+        [TestCase("http://localhost/some", "one|1|two|2", "http://localhost/some?one=1&two=2")]
+        [TestCase("http://localhost:12345/some", "one|1|two|2", "http://localhost:12345/some?one=1&two=2")]
+        [TestCase("http://localhost/", "one|1|two|2", "http://localhost/?one=1&two=2")]
+        [TestCase("http://localhost:12345/", "one|1|two|2", "http://localhost:12345/?one=1&two=2")]
+        [TestCase("http://localhost/some/", "one|1|two|2", "http://localhost/some/?one=1&two=2")]
+        [TestCase("http://localhost:12345/some/", "one|1|two|2", "http://localhost:12345/some/?one=1&two=2")]
+        public void ResetThenAppendQuery(string endpoint, string queryPart, string expected)
```
annelo-msft, line 74: Would we ever want a case where we append a query param where the parameter name has a leading '?' and the test validates that our type removes the leading query (i.e. only one '?' appears in the resulting URI) ?

test/UnbrandedProjects.Tests/ClientUriBuilderTests.cs
```diff
@@ -0,0 +1,461 @@
...
+        [TestCase("http://localhost:12345/with/", "/path", "http://localhost:12345/with/path")]
+        [TestCase("http://localhost", "/path/", "http://localhost/path/")]
+        [TestCase("http://localhost/", "/path/", "http://localhost/path/")]
+        [TestCase("http://localhost:12345", "/path/", "http://localhost:12345/path/")]
+        [TestCase("http://localhost:12345/", "/path/", "http://localhost:12345/path/")]
+        [TestCase("http://localhost/with", "/path/", "http://localhost/with/path/")]
+        [TestCase("http://localhost/with/", "/path/", "http://localhost/with/path/")]
+        [TestCase("http://localhost:12345/with", "/path/", "http://localhost:12345/with/path/")]
+        [TestCase("http://localhost:12345/with/", "/path/", "http://localhost:12345/with/path/")]
+        public void ResetThenAppendPath(string endpoint, string pathPart, string expected)
```
annelo-msft, line 45: At some point in the future, we should add tests to validate that special characters like '\u1234` and "%#?&" are not escaped when we pass escape=false.  But this work is represented by https://github.com/Azure/autorest.csharp/issues/4480 and does not need to block this PR.

test/UnbrandedProjects.Tests/ClientUriBuilderTests.cs
ArcturusZhang: yeah we may need those new test cases, and I am deferring your PR to consolidate those cases: https://github.com/Azure/azure-sdk-for-net/pull/43436

PR #4524, opened 2024-04-03
Enable the new cancellationToken pattern on non-azure libraries

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4243

# Description

Long story short: in non-azure libraries, convenience methods no longer have the parameter `cancellationToken`. Customers needing to provide a cancelling context should use the protocol methods instead for advanced scenarios.

# Checklist

# Ready to Land?


ArcturusZhang:
@m-nash @annelo-msft this PR is ready to review

test/CadlRanchProjectsNonAzure/authentication/http/custom/src/Generated/CustomClient.cs
```diff
@@ -96,7 +96,7 @@ public virtual ClientResult Valid(RequestOptions options = null)
         /// <list type="bullet">
         /// <item>
         /// <description>
-        /// This <see href="https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/ProtocolMethods.md">protocol method</see> allows explicit creation of the request and processing of the response for advanced scenarios.
+        /// This <see href="https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/System.ClientModel/samples/ServiceMethods.md#protocol-methods">protocol method</see> allows explicit creation of the request and processing of the response for advanced scenarios.
```
annelo-msft, line 99: Non-blocking question: do we want the string "Azure" to appear here?  We could use an aka.ms link instead of the full path if we wanted to remove Azure from unbranded client code and docs.

annelo-msft (approved):
Approving with regard to the output code, which looks correct to me.  I defer to others on the generator implementation.

test/CadlRanchProjectsNonAzure/authentication/http/custom/src/Generated/CustomClient.cs
ArcturusZhang: I would change this to [aka.ms/azsdk/net/protocol-methods](https://aka.ms/azsdk/net/protocol-methods) as you suggested in https://github.com/annelo-msft/openai-in-typespec/pull/1#discussion_r1550115744

archerzz (approved):
LGTM.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
```diff
@@ -1146,14 +1146,14 @@ internal virtual ClientResult InternalProtocol(BinaryContent content, RequestOpt
...
-        public virtual async Task<ClientResult> StillConvenientAsync()
+        public virtual async Task<ClientResult> StillConvenientValueAsync()
         {
             ClientResult result = await StillConvenientAsync().ConfigureAwait(false);
             return result;
         }
 
         /// <summary> When set protocol false and convenient true, the convenient method should be generated even it has the same signature as protocol one. </summary>
-        public virtual ClientResult StillConvenient()
+        public virtual ClientResult StillConvenientValue()
```
annelo-msft, line 1156: Curious: why change the name of this method instead of making `RequestOptions` a required parameter to the protocol method?

ArcturusZhang: I did not notice this change. I will investigate

ArcturusZhang: There is another issue in this PR when similar things happen:
image
This method is calling itself, but it should call its corresponding protocol method

test/UnbrandedProjects/Customized-TypeSpec/src/Generated/SuperClient.cs
```diff
@@ -414,20 +401,16 @@ public virtual ClientResult CreateLiteral(BinaryContent content, RequestOptions
         }
 
         /// <summary> Send literal parameters. </summary>
-        /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual async Task<ClientResult<Thing>> HelloLiteralAsync(CancellationToken cancellationToken = default)
+        public virtual async Task<ClientResult<Thing>> HelloLiteralAsync()
         {
-            RequestOptions options = FromCancellationToken(cancellationToken);
-            ClientResult result = await HelloLiteralAsync(options).ConfigureAwait(false);
+            ClientResult result = await HelloLiteralAsync().ConfigureAwait(false);
```
ArcturusZhang, line 406: this is terribly wrong, because this causes infinite invocation and leads to stack over flow.

ArcturusZhang:
The doc changes are distractions, I will extract those changes into a standalone PR and merge it first.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
ArcturusZhang: as for this, this is a bug according to the document @pshao25 has wrote: https://github.com/Azure/azure-sdk-for-net/pull/37795/files
In this case, we should not have this convenience method at all, but something is wrong and we have this.
We will track this issue in another issue, since the problem is not introduced in this PR, this PR will not fix it.

ArcturusZhang:
@annelo-msft FYI I updated this PR and now it only contains the change of removing cancellationToken (the other doc link change has been merged in another PR), it should be a lot more clear to review.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
pshao25: Though our principle is that we don't generate convenience method if there is no value (i.e. no models involved), this case makes the protocol method internal, so we have to still generated convenience method otherwise there would be no operations. Considering the potential ambiguity when one day user makes this protocol method public, we need to choose a way to handle it. Appending `Value` is the current behavior, we could adjust to make `RequestContext` required.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
```diff
@@ -265,20 +260,16 @@ public virtual ClientResult NoContentType(string p2, string p1, BinaryContent co
...
         }
 
         /// <summary> Return hi in demo2. </summary>
-        /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual ClientResult<Thing> HelloDemo2(CancellationToken cancellationToken = default)
+        public virtual ClientResult<Thing> HelloDemo2()
         {
-            RequestOptions options = FromCancellationToken(cancellationToken);
-            ClientResult result = HelloDemo2(options);
+            ClientResult result = HelloDemo2();
```
annelo-msft, line 272: Is this a recursive call now?  It may be that we still need to create an instance of RequestOptions in some cases, where the required RequestOptions is the only thing that differentiates a convenience method from a protocol method.

In this case, the correct thing to do here would either be:

```csharp
RequestOptions options = new();
ClientResult result = HelloDemo2(options);
```

or

```csharp
ClientResult result = HelloDemo2((RequestOptions)null);
```

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
annelo-msft: I agree that in the case where the convenience method doesn't take an input model and doesn't return an output model it is unnecessary to generate the convenience method.  

For the image in https://github.com/Azure/autorest.csharp/pull/4524/files#r1556782904, the convenience method is necessary because, even though there is no input model, the convenience method needs to return an output model.  We should make RequestOptions required on the protocol method and the convenience method should pass either a new instance of RequestOptions to call the convenience method instead of itself, or null cast to RequestOptions.

I have illustrated the proposed fix in https://github.com/Azure/autorest.csharp/pull/4524/files#r1557834090.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
ArcturusZhang: Yeah I am making a change to do the second.

ArcturusZhang:
The merge broke this PR right now :( need to fix.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
ArcturusZhang: I kind of get worried for issues like this, therefore I decided to change the generator so that when we call a protocol method here, we always pass all the parameters defined in the corresponding protocol method, no matter the parameter is optional or required (which should change nothing in run time, but this should make it more precise and acurate)

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
ArcturusZhang: since the "append value suffix" case is not introduced by this PR, I will not change it here.
@pshao25 please feel free to open an issue to discuss that case when protocol is internal, and it does not have models as input or output, should we make the request context required, or append value to convenience method.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
annelo-msft: I like this approach -- I can't think of any issues with it offhand, and I agree it's nice to make it more explicit.  The only suggestion I'd make here, is that it can be confusing for readers of the code when we pass nulls in parameter lists -- if it's not too hard to add in the explicit argument name, that might make the code more readable and thereby more maintainable.  For example, instead of 
 
```
ClientResult result = AnonymousBody(content, null);
```
 
do:
 
```
ClientResult result = AnonymousBody(content, options: null);
```
 
But either way should work if that's too much work for the generator.

PR #4557, opened 2024-04-09
Refactor spread converter generation

ArcturusZhang:
# Description

This PR refactors how we write the convenience methods in DPG.
Only normal methods, and LRO methods are changed. Pageable methods are handled in another class, therefore this PR does not touch that part.

When writing convenience, we are actually trying to build up a conversion from the current parameters (convenience parameters) to the arguments for the protocol parameters.
The refactor in general is iterating over all the parameters on the protocol method (because this is what we are trying to build), and find a conversion from the parameter in convenience method with the same name.

Previously we are assigning values for spread parameters using the initialization ctor of the backing model, which is complicated to do, because some values are assigned by parameters, some values are assigned by property initializers. This PR changes that to use the serialization ctor because it just takes all the properties via parameters.

Also by using the serialization ctor, this PR fixes an issue despite it was never reported.
For a case when there was an alias with optional collections defined, and the alias was spreaded into the operation, such as:
```tsp
alias Foo = {
	name: string;
	array?: string[];
}

op FooOp(...Foo): void;
```
this will generate code like this (before we change)
```csharp
public virtual Response FooOp(string name, IEnumerable<string> array = null, CancellationToken cancellationToken = default)
{
	// Argument.Assert part, skip
	FooOpRequest fooOpRequest = new FooOpRequest(name);
    if (array != null)
    {
        foreach (var value in array)
        {
            fooOpRequest.Array.Add(value);
        }
    }
    FooOpRequest fooOpRequest0 = fooOpRequest;
    Response response = await FooOp(fooOpRequest0.ToRequestContent(), context).ConfigureAwait(false);
    return response;
}
```
and the new code would be:
```csharp
public virtual Response FooOp(string name, IEnumerable<string> array = null, CancellationToken cancellationToken = default)
{
	// Argument.Assert part, skip
	var fooOpRequest = new FooOpRequest(name, array?.ToList() as IList<string> ?? new ChangeTrackingList<string>(), null);
    Response response = await FooOp(fooOpRequest.ToRequestContent(), context).ConfigureAwait(false);
    return response;
}
```
The difference is in how the values are populated into the `fooOpRequest` instance. There are three cases: when parameter `array` is `null`, empty, or with values.

1. `null`: the result is the same.
2. the input `array` parameter has values: in the old code, the instance will have an `ChangeTrackingList<string>` instance with values in it, therefore it serializes into `["1", "2"]`. In the new code, the instance will have an `List<string>` instance with values in it, therefore it serializes into the same thing.
3. when the `array` parameter is empty, things are different. The old code has a foreach loop inside the if, when array is empty, no `Add` calls will be executed, therefore the `ChangeTrackingList<string>` instance will remain "undefined", therefore when we serialize the request, we get an "undefined" - there is no `array` property in the payload. In the new code, an empty `List<string>` instance is assigned to `Array` property, therefore it serializes into an empty array `[]` in JSON, which should be correct.

# Checklist

# Ready to Land?


archerzz:
@ArcturusZhang What is the last `null` for? A typo?
> var fooOpRequest = new FooOpRequest(name, array?.ToList() as IList<string> ?? new ChangeTrackingList<string>(), null);

src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
```diff
@@ -8,6 +8,7 @@
 using System.Threading;
 using Autorest.CSharp.Core;
 using AutoRest.CSharp.Common.Input;
+using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
```
archerzz, line 11: This line can be removed?

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -365,7 +365,7 @@ private ReturnTypeChain BuildReturnTypes()
                     {
                         if (convenienceParameter.Type is { IsFrameworkType: false, Implementation: ModelTypeProvider model })
                         {
-                            var parameters = BuildSpreadParameters(model).OrderBy(p => p.DefaultValue == null ? 0 : 1);
+                            var parameters = BuildSpreadParameters(model).OrderBy(p => p.DefaultValue == null ? 0 : 1).ToArray();
```
archerzz, line 368: Why do we add `ToArray()`?

test/CadlRanchProjects/type/union/src/Generated/EnumsOnly.cs
```diff
@@ -140,7 +141,7 @@ public virtual async Task<Response> SendAsync(EnumsOnlyCases prop, CancellationT
         {
             Argument.AssertNotNull(prop, nameof(prop));
 
-            SendRequest6 sendRequest6 = new SendRequest6(prop);
+            SendRequest6 sendRequest6 = new SendRequest6(prop, null);
```
archerzz, line 144: Do we really need to add an explicit `null ` here?

ArcturusZhang:
> @ArcturusZhang What is the last `null` for? A typo?
> 
> > var fooOpRequest = new FooOpRequest(name, array?.ToList() as IList ?? new ChangeTrackingList(), null);

last null is for the raw data field. every model enabled MRW will have it.

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
ArcturusZhang: I changed the signature of another method which consumes parameters from `IEnumerable<Parameter>` to `IReadOnlyList<Parameter>`.
This does not impact the performance really because we will call `ToArray` any way later

test/CadlRanchProjects/type/union/src/Generated/EnumsOnly.cs
ArcturusZhang: this is for the raw data field.
Raw data field parameter will appear in every serialization ctor if MRW is enabled.
Previously we do not have to provide this because we are using initialization ctor.

ArcturusZhang: Since this part is totally generated by the generator, and the serialization ctor is always internal, we do not need to worry about it.

src/AutoRest.CSharp/Common/Generation/Writers/PageableMethodsWriterExtensions.cs
ArcturusZhang: will clean this up later

archerzz (approved):
LGTM

chunyu3 (approved):

Issue #4486, opened 2024-03-26
[ClientModel-based clients] Update comments and method names to convert models to BinaryContent


annelo-msft:
This is lower-priority and illustrated in https://github.com/annelo-msft/openai-in-typespec/pull/1/files#r1540124769

PR #4564, opened 2024-04-10
Update xml doc of `ToRequestContent` method in azure and non-azure libraries

### LINKED ISSUE #4553: [ClientModel-based clients] Setting `unreferenced-types-handling: keepAll` in tspconfig.yaml results in build error
If I set `unreferenced-types-handling: keepAll` in the tspconfig.yaml and re-generate, two internal files are added:
- src/Generated/Internal/RequestContentHelper.cs
- src/Generated/Internal/Utf8JsonRequestBody.cs

These files are not actually used, but the `Utf8JsonRequestBody` class causes a build error. More specifically, the problem is the following line in the constructor:
`_content = Create(_stream);` 

This is because the `Create` method that takes a `MemoryStream` does not exist.

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4553
Fixes https://github.com/Azure/autorest.csharp/issues/4486

# Description

Contains the following:
1. No longer try to generate the two helpers: `RequestContentHelper` and `Utf8JsonRequestContent` (fixes #4553 )
2. Rename `ToBinaryBody` to `ToBinaryContent` (fixes https://github.com/Azure/autorest.csharp/issues/4486)
3. Refine the xml doc of `ToRequestContent` and `ToBinaryContent`

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
```diff
@@ -32,6 +30,11 @@ internal static IEnumerable<ExpressionTypeProvider> GetHelperProviders()
                 yield return ClientPipelineExtensionsProvider.Instance;
                 yield return ClientUriBuilderProvider.Instance;
             }
+            if (Configuration.IsBranded)
```
lirenhe, line 33: I think we are using 'flavor' instead of 'branded' in our code, could you double check?

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/RequestContentHelperProvider.cs
```diff
@@ -25,15 +26,16 @@ internal class RequestContentHelperProvider : ExpressionTypeProvider
         private readonly MethodSignatureModifiers _methodModifiers = MethodSignatureModifiers.Public | MethodSignatureModifiers.Static;
         private RequestContentHelperProvider() : base(Configuration.HelperNamespace, null)
         {
+            // non-azure libraries do not need this type
+            Debug.Assert(Configuration.IsBranded);
```
lirenhe, line 30: Same above.  Suggest to use 'flavor'

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
ArcturusZhang: When we introduce `flavor`, in order to reduce the impact in our code, we still use this as internal implementation.
I think there is no point of changing our internal implementation.
The value of `IsBranded` is initialized as true if `flavor` is azure, otherwise false.

ArcturusZhang: If we want to use `flavor`, it will look like we are using
```
Configuration.Flavor == "azure"
```
everywhere. We should avoid hard-coded strings.
And to simplify, we need to introduce a new property, such as `IsBranded` on `Configuration`, so that we could use just
```
Configuration.IsBranded
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/RequestContentHelperProvider.cs
ArcturusZhang: I do not think so

src/AutoRest.CSharp/Common/Output/Models/Types/ExpressionTypeProvider.cs
ArcturusZhang: Also considering the existence of MGC, we definitely do not need to change it. It will automatically go away, because the non-azure plugin will take the false part, and our azure plugin takes the true part. There will no longer be a `if`

lirenhe (approved):

Issue #4571, opened 2024-04-12
[ClientModel-based clients] Introduce back `RequestContentHelper` and `Utf8JsonRequestContent`


ArcturusZhang:
when fixing [this issue](https://github.com/Azure/autorest.csharp/issues/4553), these two types are removed because their content is not correctly updated when we made the SCM update.
But these two classes are needed (at least one of them is definitely needed), we are lucky because existing test cases and libraries do not have a use case of them.

We need to find a way to update them accordingly and introduce them back, or find a way to replace them if we decide not to introduce them back.

PR #4573, opened 2024-04-12
Bring back `Utf8JsonRequestContent`/`Utf8JsonBinaryContent` and `RequestContentHelper`/`BinaryContentHelper` for non-azure libraries

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4571

# Description

The current implementation of `Utf8JsonRequestContent` is  using a `StreamRequestContent` inside it for its content, we only have a `BinaryContent.Create` method in SCM that consumes a `Stream` in version `1.1.0-beta.3`, this PR updates the version and use it.

# Checklist

# Ready to Land?


test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
```diff
@@ -1257,6 +1258,86 @@ public virtual ClientResult<bool> HeadAsBoolean(string id, RequestOptions option
             return ClientResult.FromValue(result.Value, result.GetRawResponse());
         }
 
+        /// <summary> put array. </summary>
+        /// <param name="body"> The <see cref="IEnumerable{T}"/> where <c>T</c> is of type <see cref="string"/> to use. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="body"/> is null. </exception>
+        public virtual async Task<ClientResult> PutArrayAsync(IEnumerable<string> body)
```
annelo-msft, line 1264: If we own this test case, it might be helpful to return an output model to make it clear this is a convenience method.  Sometimes PUT returns the updated resource in the response body, so this wouldn't be entirely unnatural.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/Internal/BinaryContentHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+
+using System;
+using System.ClientModel;
+using System.ClientModel.Primitives;
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace UnbrandedTypeSpec
+{
+    internal static class BinaryContentHelper
```
annelo-msft, line 13: Is this basically a copy of Azure.Core's `RequestContentHelper`?

I wonder if, when we work on these types for https://github.com/Azure/azure-sdk-for-net/issues/28369, it would make sense to create collection types that implement `IPersistableModel<T>` and/or `IJsonModel<T>` to handle these out of the box instead of with extension methods?

I don't think that blocks this implementation today, though, but wanted to flag it as future work.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
ArcturusZhang: This is only added to show that we still need these two files in non-azure libraries.
I will add some more context in the other comment

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/Internal/BinaryContentHelper.cs
ArcturusZhang: Yes.
We have removed this file in shared code, therefore it is always generated by the generator, we could be able to remove this file from shared file, and do not need to migrate it in the future.

This file is to serve the case that we are using types other than models (including arrays, dictionaries, other primitive types), such as operations defined in this way in tsp:
```tsp
op Foo(@body body: Foo[]): void;
```
Because they are built-in types in C#, we cannot let them to implement `IPersistableModel<T>` or `IJsonModel<T>`, also from the name of the interface, they are not "models" therefore it might not make sense if we could.

Our principal is that if someone could write an operation using tsp and the tsp could compile, we should generate proper code for it without violating our patterns of generating them and it should work just fine. The helper is enabling cases such as this.
We could optimize the implementation in the future, I agree with that.

test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecClient.cs
ArcturusZhang: Yes we have a lot similar cases in cadl-ranch, I also introduced them in this PR for non-azure cases

chunyu3 (approved):

lirenhe (approved):

PR #4576, opened 2024-04-15
[Mgmt]Build enums from code model converter

pshao25:
This PR is to gradually support code model conversion to input types with several PRs, instead of a huge PR at one time.

**Current behavior:**
We have
1. Different kinds of schemas in code model. (and operations)
2. Constructors for these schemas as `public TypeProvider(Schema schem){}`

**Final goal:**
We want:
1. Schame mapping to input type pseudo as `InputType inputType = CodeModelConverter.From(schema)`.
2. Constructors for these input types as `public TypeProvider(InputType inputType){}`

**Solution:**
We could support one kind of schema/input type pair at one time. For example, in this PR, it supports `ChoiceSchema/InputEnumType`. Therefore this PR would introduce (which actually already exists):
1. `InputEnumType inputType = CodeModelConverter.From(ChoiceSchema/SealedChoiceSchema)`.
2. `public EnumType(InputEnumType inputType){}`

To support one type at one time, we might need to expose `CodeModelConverter` and the mappings it builds temporarily as what this PR did. In the transition period, 
1. `CodeModelConverter` is exposed all the time.
2. We get type from input types if we support.
3. We get type from code model if we not support.

After we convert all the types to input types, we remove all the exposed things from `CodeModelConverter`.

### DIFF
--- a/src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
+++ b/src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
@@ -47,8 +47,10 @@ namespace AutoRest.CSharp.AutoRest.Plugins
                 }
                 else
                 {
-                    MgmtContext.Initialize(new BuildContext<MgmtOutputLibrary>(codeModel, sourceInputModel, schemaUsageProvider));
                     CodeModelTransformer.TransformForMgmt(codeModel);
+                    var codeModelConverter = new CodeModelConverter(codeModel, schemaUsageProvider);
+                    codeModelConverter.CreateNamespace();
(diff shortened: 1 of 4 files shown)


src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -47,8 +47,8 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(CodeModel codeModel)
                 }
                 else
                 {
-                    MgmtContext.Initialize(new BuildContext<MgmtOutputLibrary>(codeModel, sourceInputModel, schemaUsageProvider));
                     CodeModelTransformer.TransformForMgmt(codeModel);
+                    MgmtContext.Initialize(new BuildContext<MgmtOutputLibrary>(codeModel, sourceInputModel, schemaUsageProvider));
```
ArcturusZhang, line 51: I think before we go any further, we should remove this `MgmtContext` first to make everything explicit.
Also we should remove `BuildContext`.

PR #4598, opened 2024-04-18
fix(generator): do not keep azure.core models

### LINKED ISSUE #4597: [generator]: do not keep a class under `Azure.Core` namespace even if its accessibility is `internal`
## Description

`communication/Communication.JobRouter` has marked some operations `internal`: https://github.com/Azure/azure-rest-api-specs/blob/c4e661cdf92c8f579574008d0cd11874cc303da0/specification/communication/Communication.JobRouter/client.tsp#L25

Thus, the accessibility of some foundation models are marked as `internal`.
image

In current implementation, we'll keep a model as long as its accessibility is set. I think we should not keep any foundation model, no matter what the accessibility is. If it's really used, it should be referenced from `Azure.Core`, not from our generated codes.

archerzz:
# Description

Even the accessibility is set, we should not keep them since they should be referenced from Azure.Core

resolve #4597

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
+++ b/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
@@ -48,6 +48,9 @@ namespace AutoRest.CSharp.Output.Models.Types
         protected override bool IsAbstract => !Configuration.SuppressAbstractBaseClasses.Contains(DefaultName) && _inputModel.DiscriminatorPropertyName is not null && _inputModel.DiscriminatorValue is null;
 
         public ModelTypeProviderFields Fields => _fields ??= EnsureFields();
+
+        public string? Namespace => _inputModel.Namespace;
+
         private ConstructorSignature InitializationConstructorSignature => _publicConstructor ??= EnsurePublicConstructorSignature();
         private ConstructorSignature SerializationConstructorSignature => _serializationConstructor ??= EnsureSerializationConstructorSignature();
 
--- a/src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
+++ b/src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
@@ -58,7 +58,7 @@ namespace AutoRest.CSharp.Output.Models.Types
 
         private IEnumerable<string>? _accessOverriddenModels;
         public IEnumerable<string> AccessOverriddenModels => _accessOverriddenModels ??= Enums.Where(e => e.IsAccessibilityOverridden).Select(e => e.Declaration.Name)
-            .Concat(Models.Where(m => m.IsAccessibilityOverridden).Select(m => m.Declaration.Name));
+            .Concat(Models.Where(m => m.IsAccessibilityOverridden && (m.Namespace == null || !m.Namespace.StartsWith("Azure.Core"))).Select(m => m.Declaration.Name));
 
         private AspDotNetExtensionTypeProvider? _aspDotNetExtension;
         public AspDotNetExtensionTypeProvider AspDotNetExtension => _aspDotNetExtension ??= new AspDotNetExtensionTypeProvider(RestClients, Configuration.Namespace, _sourceInputModel);


archerzz:
I cherrypick this change to my TCGC PR.
-The regen PR is https://github.com/Azure/azure-sdk-for-net/pull/43495
- If you compare it with previous regen PR, you'll find the newly added `PageXXXX` models are gone. https://github.com/Azure/azure-sdk-for-net/pull/43463

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
```diff
@@ -58,7 +58,7 @@ public DpgOutputLibrary(string libraryName, IReadOnlyDictionary<InputEnumType, E
 
         private IEnumerable<string>? _accessOverriddenModels;
         public IEnumerable<string> AccessOverriddenModels => _accessOverriddenModels ??= Enums.Where(e => e.IsAccessibilityOverridden).Select(e => e.Declaration.Name)
-            .Concat(Models.Where(m => m.IsAccessibilityOverridden).Select(m => m.Declaration.Name));
+            .Concat(Models.Where(m => m.IsAccessibilityOverridden && (m.Namespace == null || !m.Namespace.StartsWith("Azure.Core"))).Select(m => m.Declaration.Name));
```
ArcturusZhang, line 61: Well if this is the reason, I think we do not need to fix this because this really looks like a workaround, and we should not do this.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -48,6 +48,9 @@ internal sealed class ModelTypeProvider : SerializableObjectType
         protected override bool IsAbstract => !Configuration.SuppressAbstractBaseClasses.Contains(DefaultName) && _inputModel.DiscriminatorPropertyName is not null && _inputModel.DiscriminatorValue is null;
 
         public ModelTypeProviderFields Fields => _fields ??= EnsureFields();
+
+        public string? Namespace => _inputModel.Namespace;
```
ArcturusZhang, line 52: this property is kind of misleading because the namespace of a model in its generated code should be found in `Declaration.Namespace` instead of this.

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
pshao25: If we have to fix, maybe we should let emitter not generate models from azure.core. There is a flag in TCGC that to filter out azure.core models.

archerzz: we cannot let TCGC filter out core models, because that will break other cases for emitter. and I remember in generator we already filter out core models?

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
archerzz: OK, I can change it into `OriginalNamespace` or something like that

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
ArcturusZhang: our emitter could do this.

src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
ArcturusZhang: that does not make sense either, this type is an "output type", it really should not provide anything that is not used in output

src/AutoRest.CSharp/LowLevel/AutoRest/DpgOutputLibrary.cs
archerzz: OK. Previously I thought our generator will remove core types. If that assumption is false, then it makes sense to filter out in emitter.

Issue #4339, opened 2024-03-06
Core library implementation for multipart/form-data


chunyu3:
```[tasklist]
### Tasks
- [ ] Emitter MultipartFormDataRequestContent for azure SDK
- [ ] Emitter MultipartFormDataBinaryContent for generic SDK
```

PR #4613, opened 2024-04-22
Multipart form-data

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/4339
Fix https://github.com/Azure/autorest.csharp/issues/4340
Fix https://github.com/Azure/autorest.csharp/issues/4338

- implement multipart/form-data
- Codegen will generate core library MultipartFormDataRequestConent for azure, and MultipartFormDataBinaryContent for non-azure
- each model will generate ToMultipartRequestContent to serialize model to multipart/form-data format
- Add `contentType` parameter in protocol method
- Convenience method will generate MultipartFormDataRequestContent/MultipartFormDataBinaryContent from model and call protocol method.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/ApiTypes.cs
```diff
@@ -98,5 +101,8 @@ public FormattableString GetHttpPipelineKeyCredentialString(string pipelineField
         public abstract ExtensibleSnippets ExtensibleSnippets { get; }
 
         public abstract string LicenseString { get; }
+        public abstract MethodBodyStatement GetMultipartFormDataRequestContentAddStatment(VariableReference multipartContent, ValueExpression content, ValueExpression name, ValueExpression? fileName, ValueExpression? contentType);
+        public abstract MethodBodyStatement GetMultipartFormDataRequestContentWriteToStatment(VariableReference multipartContent, ValueExpression stream, ValueExpression? cancellationToken);
+        public abstract MethodBodyStatement GetMultipartFormDataRequestContentWriteToAsyncStatment(VariableReference multipartContent, ValueExpression stream, ValueExpression? cancellationToken);
```
ArcturusZhang, line 106: multipartContent does not have to be a `VariableReference`.
```suggestion
        public abstract MethodBodyStatement GetMultipartFormDataRequestContentAddStatment(ValueExpression multipartContent, ValueExpression content, ValueExpression name, ValueExpression? fileName, ValueExpression? contentType);
        public abstract MethodBodyStatement GetMultipartFormDataRequestContentWriteToStatment(ValueExpression multipartContent, ValueExpression stream, ValueExpression? cancellationToken);
        public abstract MethodBodyStatement GetMultipartFormDataRequestContentWriteToAsyncStatment(ValueExpression multipartContent, ValueExpression stream, ValueExpression? cancellationToken);
```

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace AutoRest.CSharp.Common.Input
 {
-    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, InputDictionaryType? InheritedDictionaryType, bool IsNullable)
+    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, InputDictionaryType? InheritedDictionaryType, bool IsNullable, IReadOnlyList<string> MediaTypes)
```
ArcturusZhang, line 11: why do we need this `MediaTypes` on models?
Is the purpose to know whether this model is directly or indirectly used by a multipart operation?
If so, I remember last time Michael said we could put the information in `usage`.
In my opinion, this is just like "serialization", as we have for now, json and xml.
I am OK with this, but we might need to discuss how we should implement this, especially this is also related with xml

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -91,10 +99,25 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
...
+            }
+            reader.Read();
+
+            while (reader.TokenType != JsonTokenType.EndArray)
+            {
+                var type = reader.ReadWithConverter<string>(options);
+                mediaTypes.Add(type ?? throw new JsonException($"null item isn't allowed"));
+            }
+            reader.Read();
+        }
```
ArcturusZhang, line 119: well this is a list of string, I think we do not have to write this manually, just call
```
reader.TryReadWithConverter()
```
and give it a `IReadOnlyList<string>`, it will work.

src/AutoRest.CSharp/Common/Input/ApiTypes.cs
chunyu3: Hello @ArcturusZhang  Actually we enter the VariableReference, not typedValueExpression. Because we need to support both azure and non-azure, so in `WriteMultiParts` where whose method used, we cannot pass TypedValueExpression, instead `VariableReference`

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -132,6 +155,34 @@ private static void CreateProperties(ref Utf8JsonReader reader, ICollection<Inpu
...
+                    property = new InputModelProperty(property.Name,
+                                                    property.SerializedName,
+                                                    property.Description,
+                                                    propertyType,
+                                                    property.ConstantValue,
+                                                    property.IsRequired,
+                                                    property.IsReadOnly,
+                                                    property.IsDiscriminator,
+                                                    property.FlattenedNames);
+                }
```
ArcturusZhang, line 188: this looks so weird that we are changing the serialization result, which we should not.
We should do this when building the property of the generated model in `ModelTypeProviderFields`

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
chunyu3: Yes. after we adopt tcgc, we will get the `usage`. And current tcgc does not handle multiple-contentType senario ( maybe it will change in future), I think, we may not combine this in our inputModel, just recorded down the MediaType a model support.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+                    Configuration.ApiTypes.ToMultipartRequestContentName,
+                    null,
+                    null,
+                    MethodSignatureModifiers.Internal | MethodSignatureModifiers.Virtual,
+                    Configuration.ApiTypes.MultipartRequestContentType,
+                    null,
+                    Array.Empty<Parameter>()),
+                BuildToMultipartRequestContentMethodBody(multipart).ToArray());
+        }
+        public static SwitchCase BuildMultipartWriteSwitchCase(MultipartObjectSerialization multipart, ModelReaderWriterOptionsExpression options)
```
ArcturusZhang, line 50: this should be private

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
chunyu3: I will try.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
chunyu3: I will try

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+                            typeof(BinaryData),
+                            null,
+                            new[]
+                            {
+                                KnownParameters.Serializations.Options
+                            }),
+                        new ValueExpression[]
+                        {
+                            options
+                        })));
```
ArcturusZhang, line 71: if we want to invoke a method with a given parameter list, please just do:
```
new InvokeInstanceMethodExpression(null, "SerializeMultipart", new[] {options})
```
it has a ctor that just takes the method name.
The overload taking `MethodSignature` we should only use it when we have a `MethodSignature`.

ArcturusZhang: same comment applies to all the similar cases in this file.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+                        : NotEqual(property.Value, Null);
+
+                    yield return Serializations.WrapInCheckNotWire(
+                        property,
+                        Literal("MFD"),
+                        InvokeOptional.WrapInIsDefined(
+                            property,
+                    new IfElseStatement(checkPropertyIsInitialized,
+                    WritePropertySerialization(multipartContent, property), null)
+                    ));
```
ArcturusZhang, line 129: why are we calling the `check not wire` method here?
Do we need to check if this is not wire?
Also please put something for this `MFD`, we should not hard code it everywhere in case it needs to change.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+            {
+                contentExpression = BuildValueSerizationExpression(valueSerialization.Type, valueExpression);
+                if (valueSerialization.FileName != null)
+                {
+                    fileNameExpression = Literal(valueSerialization.FileName);
+                } else
+                {
+                    fileNameExpression = name.Add(Literal(".wav"));
+                }
+                if (valueSerialization.ContentType != "application/json")
```
ArcturusZhang, line 221: we have quite a few magic strings.
For `application/json`, I think we should define it somewhere.
For `.wav` is this correct? This looks so weird.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+            }
+
+            switch (valueSerialization.Type.Implementation)
+            {
+                case SerializableObjectType serializableObjectType:
+                    contentExpression = BuildValueSerizationExpression(serializableObjectType.Type, valueExpression);
+                    break;
+                case ObjectType:
+                    contentExpression = BuildValueSerizationExpression(valueSerialization.Type, valueExpression);
+                    break;
```
ArcturusZhang, line 242: would ther e be the second `ObjectType` case? because  `SerializableObjectType` inherits from `ObjectType`.
And we are switching on `valueSerialization.Type`, if this is a `SerializableObjectType`, we use the type. if it is a `ObjectType`, we use the type `valueSerialization.Type`. But they are the same, we should only use one of them.

I think we should only keep the `SerializableObjectType` and let `ObjectType` (if any) error out. Because this is for serialization, a type provider not inheriting `SerializableObjectType` as suggested by its name, should not support any type of serialization.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+                case EnumType { IsIntValueType: true, IsExtensible: false } enumType:
+                    contentExpression = BuildValueSerizationExpression(typeof(int), new CastExpression(valueExpression, enumType.ValueType));
+                    break;
+                case EnumType { IsNumericValueType: true } enumType:
+                    contentExpression = BuildValueSerizationExpression(typeof(float), new EnumExpression(enumType, valueExpression).ToSerial());
+                    break;
+                case EnumType enumType:
+                    contentExpression = BuildValueSerizationExpression(typeof(string), new EnumExpression(enumType, valueExpression).ToSerial());
+                    break;
```
ArcturusZhang, line 251: I think we can combine these two:
```suggestion
                case EnumType enumType:
                    contentExpression = BuildValueSerizationExpression(enumType.ValueType, new EnumExpression(enumType, valueExpression).ToSerial());
                    break;
```

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+                default:
+                    throw new NotSupportedException($"Cannot build serialization expression for type {valueSerialization.Type}, please add `CodeGenMemberSerializationHooks` to specify the serialization of this type with the customized property");
+            }
+            return Configuration.ApiTypes.GetMultipartFormDataRequestContentAddStatment(mulitpartContent, contentExpression, name, fileNameExpression, contentTypeExpression);
```
ArcturusZhang, line 255: These statements/expression related content should go to the `ExtensibleSnippets` class: https://github.com/Azure/autorest.csharp/blob/13aaeb8b85373ea1a9074fff8e54458cec0d7c3e/src/AutoRest.CSharp/Common/Output/Expressions/ExtensibleSnippets.cs#L9

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        private static ValueExpression BuildValueSerizationExpression(CSharpType valueType,  ValueExpression valueExpression) => valueType switch
+        {
+            _ when valueType.IsFrameworkType => valueExpression,
+            _ when valueType.Implementation is SerializableObjectType serializableObjectType => new InvokeStaticMethodExpression(typeof(ModelReaderWriter), nameof(ModelReaderWriter.Write), new[] { valueExpression, ModelReaderWriterOptionsExpression.Wire }, new[] { valueType }),
+            _ => new InvokeStaticMethodExpression(typeof(BinaryData), nameof(BinaryData.FromObjectAsJson), new[] { valueExpression }, new[] { valueType })
```
ArcturusZhang, line 262: we should use `BinaryDataExpression.FromObjectAsJson(valueExpression)`
if there is not one, create a static method.
This is the purpose of the whole expression system, we would like the code building expressions to look like we are writing the real generated code - if we are calling a static method on a specific type, we should create a static method on the corresponding expression.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+                default:
+                    throw new NotSupportedException($"Cannot build serialization expression for type {valueSerialization.Type}, please add `CodeGenMemberSerializationHooks` to specify the serialization of this type with the customized property");
+            }
+            return Configuration.ApiTypes.GetMultipartFormDataRequestContentAddStatment(mulitpartContent, contentExpression, name, fileNameExpression, contentTypeExpression);
+        }
+
+        private static ValueExpression BuildValueSerizationExpression(CSharpType valueType,  ValueExpression valueExpression) => valueType switch
+        {
+            _ when valueType.IsFrameworkType => valueExpression,
+            _ when valueType.Implementation is SerializableObjectType serializableObjectType => new InvokeStaticMethodExpression(typeof(ModelReaderWriter), nameof(ModelReaderWriter.Write), new[] { valueExpression, ModelReaderWriterOptionsExpression.Wire }, new[] { valueType }),
```
ArcturusZhang, line 261: we should use pattern matching:
```suggestion
            { IsFrameworkType: true } => valueExpression,
            { IsFrameworkType: false, Implementation: SerializableObjectType serializableObjectType } => ModelReaderWriterExpression.Write(valueExpression, ModelReaderWriterOptionsExpression.Wire),
```

ArcturusZhang: also why "wire" format here?

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -294,7 +296,6 @@ static bool IsSerializable(ObjectTypeProperty objectProperty, TypeFactory typeFa
 
         public BicepObjectSerialization? BuildBicepObjectSerialization(SerializableObjectType objectType, JsonObjectSerialization jsonObjectSerialization)
             => new BicepObjectSerialization(objectType, jsonObjectSerialization);
-
```
ArcturusZhang, line 297: Please keep the empty line between methods.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -617,5 +618,159 @@ private static PropertyBag<ObjectTypeProperty> PopulatePropertyBag(SerializableO
                     new CSharpType(typeof(Dictionary<,>), additionalPropertiesProperty.Declaration.Type.Arguments),
                     shouldExcludeInWireSerialization);
         }
+        public MultipartObjectSerialization BuildMultipartObjectSerialization(ObjectSchema objectSchema, SchemaObjectType objectType)
```
ArcturusZhang, line 621: do we really need to have this for swagger models?
I suggest to keep limit your change in DPG at first as your first step.
Considering the complexity and consolidation work, we should not change the schema models.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
chunyu3: @ArcturusZhang  The InputType will used when serialize it. So we need to update here. It is a workaround, when we adopt tcgc, we will do this transformation in emitter. Current we will keep to change here.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
chunyu3: When the property is a Model, we need to serialize it as serialize in wire.

src/AutoRest.CSharp/Common/Input/ApiTypes.cs
ArcturusZhang: every expression inherits `ValueExpression`. I am suggesting `ValueExpression` - because it is not always we pass in a variable, we could pass in an expression such as `something.Content` <- this is a value expression.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
ArcturusZhang: I think you did not get my point here.
My point is that we should do this type transformation when we determine the output type of the property.

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
chunyu3: remove `MediaTypes` and use `Usage`

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
chunyu3: keep it this time.

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+using AutoRest.CSharp.Output.Models;
+using AutoRest.CSharp.Output.Models.Serialization;
+using AutoRest.CSharp.Output.Models.Shared;
+using AutoRest.CSharp.Output.Models.Types;
+using static AutoRest.CSharp.Common.Output.Models.Snippets;
+using MultipartFormDataContent = System.Net.Http.MultipartFormDataContent;
+
+namespace AutoRest.CSharp.Common.Output.Models.Types.HelperTypeProviders
+{
+    internal class MultipartFormDataRequestContentProvider: ExpressionTypeProvider
```
ArcturusZhang, line 27: Please format this file

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+                name: "_random",
+                initializationValue: New.Instance(typeof(Random), Array.Empty<ValueExpression>()),
+                serializationFormat: SerializationFormat.Default);
+            _boundaryValuesFields = new FieldDeclaration(
+                modifiers: FieldModifiers.Private | FieldModifiers.Static | FieldModifiers.ReadOnly,
+                type: typeof(char[]),
+                name: "_boundaryValues",
+                initializationValue: Literal("0123456789=ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz").ToCharArray(),
+                serializationFormat: SerializationFormat.Default);
+            ValueExpression contentTypeHeaderExpression = new UnaryOperatorExpression("!", new MultipartFormDataContentExpression(_multipartContentField).Property("Headers").Property("ContentType"), true);
```
ArcturusZhang, line 52: Please add the `Headers` and `ContentType` property in the corresponding expression types, so that we could have:
```suggestion
            ValueExpression contentTypeHeaderExpression = new UnaryOperatorExpression("!", new MultipartFormDataContentExpression(_multipartContentField).Headers.ContentType, true);
```

ArcturusZhang: By checking your generated code, we are wanting `_multipartContent.Headers.ContentType!`.
The whole idea of expressions is to make everything to have the correct csharp semantic meaning.
Here, this `!` is the null forgiving operator which is not a `UnaryOperatorExpression`. We should not constructing this here.
Instead, since we did not enable nullable context in all our generated code, we do not need to add the `!` operator here.
Please just change this to:
```suggestion
            ValueExpression contentTypeHeaderExpression = new MultipartFormDataContentExpression(_multipartContentField).Headers.ContentType, true);
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+            ValueExpression contentTypeHeaderValueExpression = new InvokeInstanceMethodExpression(contentTypeHeaderExpression, nameof(MediaTypeHeaderValue.ToString), Array.Empty<ValueExpression>(), null, false);
```
ArcturusZhang, line 53: if this is the only place `contentTypeHeaderExpression` is used, we should inline it, since it is not so long right now.
And we should use this:
```suggestion
            ValueExpression contentTypeHeaderValueExpression = contentTypeHeaderExpression.InvokeToString();
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+            _contentTypeProperty = new PropertyDeclaration(
+                description: null,
+                modifiers: MethodSignatureModifiers.Public,
+                propertyType: typeof(string),
+                name: _contentTypePropertyName,
+                propertyBody: new MethodPropertyBody(new MethodBodyStatement[]
+                {
+                    Return(contentTypeHeaderValueExpression)
+                }));
```
ArcturusZhang, line 62: we do not need an array when there is only one statement:
```suggestion
                propertyBody: new MethodPropertyBody(Return(contentTypeHeaderValueExpression)));
```

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+            _httpContentProperty = new PropertyDeclaration(
+                description: null,
+                modifiers: MethodSignatureModifiers.Internal,
+                propertyType: typeof(HttpContent),
+                name: _httpContentPropertyName,
+                propertyBody: new ExpressionPropertyBody(_multipartContentField)
+                );
+            _multipartContentExpression = new MultipartFormDataContentExpression(_multipartContentField);
```
ArcturusZhang, line 70: move this up so that we could use this in `contentTypeHeaderValueExpression` (L53)

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+        }
+        private readonly FieldDeclaration _multipartContentField;
+        private readonly FieldDeclaration _randomField;
+        private readonly FieldDeclaration _boundaryValuesFields;
+
+        private readonly PropertyDeclaration _contentTypeProperty;
+        private const string _contentTypePropertyName = "ContentType";
+        private readonly PropertyDeclaration _httpContentProperty;
+        private const string _httpContentPropertyName = "HttpContent";
+        protected override string DefaultName { get; } = Configuration.ApiTypes.MultipartRequestContentTypeName;
```
ArcturusZhang, line 80: Please move this override property to the top of this class for clarity

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+        {
+            yield return _contentTypeProperty;
+            yield return _httpContentProperty;
+        }
+        protected override IEnumerable<Method> BuildConstructors()
+        {
+            var signature = new ConstructorSignature(
+                Type: Type,
+            Modifiers: MethodSignatureModifiers.Public,
+            Parameters: Array.Empty<Parameter>(),
```
ArcturusZhang, line 105: this needs formatting

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+                Summary: null, Description: null);
+            var body = new MethodBodyStatement[]
+            {
+                Assign(_multipartContentField, New.Instance(typeof(MultipartFormDataContent), new[]{CreateBoundary()})),
+            };
```
ArcturusZhang, line 110: if we only have one statement in the body, we do not need the array

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+                Parameters: new[]
+                {
+                    contentParam,
+                    nameParam,
+                    filenameParam,
+                    contentTypeParam
+                });
+            MethodBodyStatement? valueDelareStatement = null;
+            ValueExpression contentExpression;
+            contentExpression = New.Instance(typeof(ByteArrayContent), contentParam);
```
ArcturusZhang, line 193: merge these two lines

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+                case Type detype when detype == typeof(decimal):
+                    ValueExpression invariantCulture = new MemberExpression(typeof(CultureInfo), nameof(CultureInfo.InvariantCulture));
+                    var invariantCultureValue = new InvokeInstanceMethodExpression(contentParam, nameof(Int32.ToString), new[] {Literal("G"), invariantCulture }, null, false);
+                    valueDelareStatement = Declare(typeof(string), "value", invariantCultureValue, out TypedValueExpression variable);
+                    contentExpression = New.Instance(typeof(StringContent), variable);
+                    break;
+                case Type bdtype when bdtype == typeof(BinaryData):
+                    contentExpression = New.Instance(typeof(ByteArrayContent), new BinaryDataExpression(contentParam).ToArray());
+                    break;
+            }
```
ArcturusZhang, line 220: well it might be more clear if we make it a if-else if block
Also we need an escaping branch to throw exception so that we could know something is wrong if some incorrect change has been made.

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+            if (valueDelareStatement != null)
+            {
+                addContentStatements.Add(valueDelareStatement);
+            }
+            addContentStatements.Add(new InvokeInstanceMethodStatement(null, _addMethodName, new[] { contentExpression, nameParam, filenameParam, contentTypeParam }, false));
+            var body = new MethodBodyStatement[]
+            {
+                Argument.AssertNotNull(contentParam),
+                Argument.AssertNotNullOrEmpty(nameParam),
+                new EmptyLineStatement(),
```
ArcturusZhang, line 231: just use `EmptyLine` from `Snippets` static class

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
chunyu3: `ContentType` is property of `System.Net.Http.Headers.HttpContentHeaders`  and the contentType expression only used here, so Maybe we will not add a TypedValueExpression for `System.Net.Http.Headers.HttpContentHeaders`. 

`Headers` will update.

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+            var declaration = new CodeWriterDeclaration("contentLength");
+            VariableReference contentLengthVariable = new VariableReference(typeof(long), declaration);
+            var contentLengthDeclaration = new DeclarationExpression(contentLengthVariable,false);
+            var body = new MethodBodyStatement[]
+            {
+                new IfStatement(BoolExpression.Is(contentLengthHeaderExpression, contentLengthDeclaration))
+                {
+                    Assign(lengthParameter, contentLengthVariable),
+                    Return(Literal(true)),
+                },
```
ArcturusZhang, line 355: well let us simplify:
```suggestion
                new IfStatement(BoolExpression.Is(contentLengthHeaderExpression, new DeclarationExpression(typeof(long), "contentLength", out var contentLengthVariable)))
                {
                    Assign(lengthParameter, contentLengthVariable),
                    Return(Literal(true)),
                },
```

ArcturusZhang, line 355: let us simplify:
```suggestion
                new IfStatement(BoolExpression.Is(contentLengthHeaderExpression, new DeclarationExpression(typeof(bool), "contentLength", out var contentLengthVariable)))
                {
                    Assign(lengthParameter, contentLengthVariable),
                    Return(Literal(true)),
                },
```
And related declarations above this could be removed.

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -0,0 +1,495 @@
...
+        {
+            if (cancellationToken == null)
+            {
+                return new InvokeInstanceMethodStatement(multipartContent.Untyped, _writeToMethodName, new[] { stream }, false);
+            } else
+            {
+                return new InvokeInstanceMethodStatement(multipartContent.Untyped, _writeToMethodName, new[] { stream, cancellationToken }, false);
+            }
+        }
+        public MethodBodyStatement WriteToAsync(VariableReference multipartContent, ValueExpression stream, ValueExpression? cancellationToken)
```
ArcturusZhang, line 482:
```suggestion
        public MethodBodyStatement WriteTo(ValueExpression multipartContent, ValueExpression stream, ValueExpression? cancellationToken)
        {
            if (cancellationToken == null)
            {
                return new InvokeInstanceMethodStatement(multipartContent.Untyped, _writeToMethodName, new[] { stream }, false);
            } else
            {
                return new InvokeInstanceMethodStatement(multipartContent.Untyped, _writeToMethodName, new[] { stream, cancellationToken }, false);
            }
        }
        public MethodBodyStatement WriteToAsync(ValueExpression multipartContent, ValueExpression stream, ValueExpression? cancellationToken)
```

src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs
```diff
@@ -81,7 +91,13 @@ public IEnumerable<MethodBodyStatement> GetConvertStatements(LowLevelClientMetho
                     else
                     {
                         // if convenience parameter is null here, we just use the default value of the protocol parameter as value (we always do this even if the parameter is optional just in case there is an ordering issue)
-                        protocolInvocationExpressions.Add(DefaultOf(protocol.Type));
+                        if (isMultipartOperation && protocol.Name == "contentType" && content != null)
+                        {
+                            protocolInvocationExpressions.Add(MultipartFormDataRequestContentProvider.Instance.ContentTypeProperty(content));
+                        } else
```
ArcturusZhang, line 97: need formatting

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -606,11 +605,27 @@ public void AddRequestContext()
...
             var name = operationParameter.Name.ToVariableName();
-            var description = Parameter.CreateDescription(operationParameter, typeof(ContentType), requestMediaTypes);
+            if (requestMediaTypes.Count == 1 && requestMediaTypes.Contains("multipart/form-data"))
+            {
+                var description = Parameter.CreateDescription(operationParameter, typeof(string), requestMediaTypes);
+                var parameter = new Parameter(name, description, typeof(string), null, ValidationType.None, null, RequestLocation: RequestLocation.Header);
+                _orderedParameters.Add(new ParameterChain(null, parameter, parameter));
+                AddReference(operationParameter.NameInRequest, operationParameter, parameter, SerializationFormat.Default);
+                return;
+            } else
```
ArcturusZhang, line 615: need formatting here too

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -606,11 +605,27 @@ public void AddRequestContext()
...
+                _orderedParameters.Add(new ParameterChain(parameter, parameter, parameter));
+
+                AddReference(operationParameter.NameInRequest, operationParameter, parameter, SerializationFormat.Default);
+            }
+            /*
             var parameter = new Parameter(name, description, typeof(ContentType), null, ValidationType.None, null, RequestLocation: RequestLocation.Header);
             _orderedParameters.Add(new ParameterChain(parameter, parameter, parameter));
 
             AddReference(operationParameter.NameInRequest, operationParameter, parameter, SerializationFormat.Default);
+            */
```
ArcturusZhang, line 628: we need to remove the comments

src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs
```diff
@@ -670,6 +685,12 @@ private ReferenceOrConstant CreateReference(InputParameter operationParameter, P
 
             if (operationParameter.Kind is InputOperationParameterKind.Constant && parameter.DefaultValue is not null)
             {
+                /*
+                if (operationParameter.IsContentType && parameter.DefaultValue.HasValue && $"{parameter.DefaultValue.Value.Value:L}".StartsWith("multipart"))
+                {
+                    return new Reference($"content.{nameof(RequestContent.ContentType)}", typeof(string));
+                }
+                */
```
ArcturusZhang, line 693: remove here as well

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,236 @@
...
+                            MultipartFormDataRequestContentProvider.Instance.WriteTo(content, stream, null),
+                            new IfElseStatement(GreaterThan(stream.Property("Position"), new MemberExpression(typeof(int), nameof(int.MaxValue))),
+                            /*return BinaryData.FromStream(stream);*/
+                            Return(BinaryDataExpression.FromStream(stream, false)),
+                            /*return new BinaryData(stream.GetBuffer().AsMemory(0, (int)stream.Position));*/
+                            Return(New.Instance(typeof(BinaryData), new[]{ new InvokeInstanceMethodExpression((new StreamExpression(stream)).GetBuffer, "AsMemory", new[] { Literal(0), new CastExpression(stream.Property("Position"), typeof(int)) }, null, false) }))
+                            ),
+                        };
+        }
+        public static IEnumerable<MethodBodyStatement> BuildToMultipartRequestContentMethodBody(MultipartObjectSerialization? multipart)
```
lirenhe, line 79: Can we make the code consistent by adding a new line before each function and please apply this to all the functions in this file.

src/AutoRest.CSharp/LowLevel/Generation/DpgClientWriter.cs
```diff
@@ -292,10 +292,11 @@ private void WriteSubClientInternalConstructor(ConstructorSignature signature)
 
         private void WriteConvenienceMethod(LowLevelClientMethod clientMethod, ConvenienceMethod convenienceMethod, ClientFields fields, bool async)
         {
+            bool isMultipartOperation = convenienceMethod.RequestMediaTypes != null && convenienceMethod.RequestMediaTypes.Count == 1 && convenienceMethod.RequestMediaTypes.Contains("multipart/form-data");
             using (WriteConvenienceMethodDeclaration(_writer, convenienceMethod, fields, async))
             {
                 // This method now builds statements for normal operation and lro operations
-                MethodBodyStatement statement = convenienceMethod.GetConvertStatements(clientMethod, async, fields.ClientDiagnosticsProperty).ToArray();
+                MethodBodyStatement statement = convenienceMethod.GetConvertStatements(clientMethod, async, fields.ClientDiagnosticsProperty, isMultipartOperation).ToArray();
```
archerzz, line 299: Shall we move the logic of determining `isMultipartOperation` inside `convenienceMethod.GetConvertStatements()`? `isMultpartOperation` is all determined from properties of `convenienceMethod` itself. So maybe we don't need to calculate it outside `convenienceMethod`.

src/TypeSpec.Extension/Emitter.Csharp/src/type/usage.ts
```diff
@@ -5,5 +5,6 @@ export enum Usage {
     None = "None",
     Input = "Input",
     Output = "Output",
-    RoundTrip = "RoundTrip"
+    RoundTrip = "RoundTrip",
+    Multipart = "Multipart"
```
archerzz, line 9: I'm not sure if it's the right `usage`

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/MultipartFormDataRequestContentProvider.cs
```diff
@@ -189,34 +180,44 @@ private Method BuildAddMethod<T>()
...
+            {
+                contentExpression = New.Instance(typeof(ByteArrayContent), new BinaryDataExpression(contentParam).ToArray());
+            }
+            else if (contentParamType == typeof(byte[]))
+            {
+                contentExpression = New.Instance(typeof(ByteArrayContent), contentParam);
+            }
+            else
+            {
+                throw new Exception($"Not supported type {contentParamType}");
```
ArcturusZhang, line 220: if the message is "Not supported" how about we just throw an UnsupportedException?

src/TypeSpec.Extension/Emitter.Csharp/src/type/usage.ts
archerzz: It's actually a media type

ArcturusZhang: well because this kind of media type is contagious on models, Michael suggests that we put it on Usage as a flag of that model.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,243 @@
...
+        {
+            var serializeCallExpression = new InvokeInstanceMethodExpression(null, Configuration.ApiTypes.ToMultipartRequestContentName, Array.Empty<ValueExpression>(), null, false);
+            return new MethodBodyStatement[]
+                        {
+                            UsingDeclare("content", MultipartFormDataRequestContentProvider.Instance.Type, serializeCallExpression, out var content),
+                            /*using MemoryStream stream = new MemoryStream();*/
+                            UsingDeclare("stream", typeof(MemoryStream), New.Instance(typeof(MemoryStream), Array.Empty<ValueExpression>()), out var stream),
+                            /*content.WriteTo(stream, cancellationToken);*/
+                            MultipartFormDataRequestContentProvider.Instance.WriteTo(content, stream, null),
+                            new IfElseStatement(GreaterThan(stream.Property("Position"), new MemberExpression(typeof(int), nameof(int.MaxValue))),
```
ArcturusZhang, line 71: we have `IntExpression.MaxValue`
```suggestion
                            new IfElseStatement(GreaterThan(stream.Property("Position"), IntExpression.MaxValue),
```

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        public static IEnumerable<MethodBodyStatement> BuildMultipartSerializationMethodBody()
+        {
+            var serializeCallExpression = new InvokeInstanceMethodExpression(null, Configuration.ApiTypes.ToMultipartRequestContentName, Array.Empty<ValueExpression>(), null, false);
+            return new MethodBodyStatement[]
+                        {
+                            UsingDeclare("content", MultipartFormDataRequestContentProvider.Instance.Type, serializeCallExpression, out var content),
+                            /*using MemoryStream stream = new MemoryStream();*/
+                            UsingDeclare("stream", typeof(MemoryStream), New.Instance(typeof(MemoryStream), Array.Empty<ValueExpression>()), out var stream),
```
ArcturusZhang, line 68: there is an overload that we could have strong-typed expressions:
```
public static MethodBodyStatement UsingDeclare(string name, StreamExpression value, out StreamExpression variable)
```
therefore 
```suggestion
                            UsingDeclare("stream", new StreamExpression(New.Instance(typeof(MemoryStream), Array.Empty<ValueExpression>())), out var stream), // <- now stream is StreamExpression
```

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,243 @@
...
+                            /*content.WriteTo(stream, cancellationToken);*/
+                            MultipartFormDataRequestContentProvider.Instance.WriteTo(content, stream, null),
+                            new IfElseStatement(GreaterThan(stream.Property("Position"), IntExpression.MaxValue),
```
ArcturusZhang, line 71: and I believe `StreamExpression` has a property `Position`, we could just use it here.

src/AutoRest.CSharp/Common/Output/Builders/MultipartSerializationMethodsBuilder.cs
```diff
@@ -0,0 +1,243 @@
...
+                            /*return BinaryData.FromStream(stream);*/
+                            Return(BinaryDataExpression.FromStream(stream, false)),
+                            /*return new BinaryData(stream.GetBuffer().AsMemory(0, (int)stream.Position));*/
+                            Return(New.Instance(typeof(BinaryData), new[]{ new InvokeInstanceMethodExpression((new StreamExpression(stream)).GetBuffer, "AsMemory", new[] { Literal(0), new CastExpression(stream.Property("Position"), typeof(int)) }, null, false) }))
```
ArcturusZhang, line 75: `MemoryStream.GetBuffer` is a method, therefore we should make this expression as a method.
Also because `stream.GetBuffer()` returns `byte[]`, and `AsMemory` should be an extension method, we should use the `InvokeStaticMethodExpression` with `CallAsExtension: true` here to ensure the namespace of this extension is properly included.

ArcturusZhang: Also New.Isntance could take a params of ValueExpression therefore the array here is unnecessary and could make it more clear

```suggestion
                            Return(New.Instance(typeof(BinaryData), new InvokeStaticMethodExpression(typeof(MemoryExtensions), nameof(MemoryExtensions.AsMemory), new ValueExpression[] { stream.GetBuffer(), Literal(0), stream.Position.CastTo(typeof(int)) }, CallAsExtension: true)))
```

ArcturusZhang (approved):

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Models/CreateTranscriptionRequest.Serialization.cs
```diff
@@ -147,6 +155,45 @@ internal static CreateTranscriptionRequest DeserializeCreateTranscriptionRequest
...
+            else
+            {
+                return new BinaryData(stream.GetBuffer().AsMemory(0, (int)stream.Position));
+            }
+        }
+
+        internal virtual MultipartFormDataBinaryContent ToMultipartBinaryBody()
+        {
+            MultipartFormDataBinaryContent content = new MultipartFormDataBinaryContent();
+            content.Add(File, "file", "file", "application/octet-stream");
```
annelo-msft, line 176: We should not pass a value for filename here if there is no file name.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Models/CreateImageEditRequest.Serialization.cs
```diff
@@ -175,6 +190,52 @@ internal static CreateImageEditRequest DeserializeCreateImageEditRequest(JsonEle
...
+        }
+
+        internal virtual MultipartFormDataBinaryContent ToMultipartBinaryBody()
+        {
+            MultipartFormDataBinaryContent content = new MultipartFormDataBinaryContent();
+            content.Add(Prompt, "prompt");
+            content.Add(Image, "image", "image", "application/octet-stream");
+            if (Optional.IsDefined(Mask))
+            {
+                content.Add(Mask, "mask", "mask", "application/octet-stream");
```
annelo-msft, line 215: This will not work -- OpenAI needs to be able to pass the file name to this method to correctly format the header for the Mask property in the multipart/form-data content payload.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/AudioTranscriptions.cs
```diff
@@ -81,15 +81,16 @@ public virtual ClientResult<CreateTranscriptionResponse> Create(CreateTranscript
...
         /// <param name="content"> The content to send as the body of the request. </param>
+        /// <param name="contentType"> The <see cref="string"/> to use. Allowed values: "multipart/form-data". </param>
         /// <param name="options"> The request options, which can override default behaviors of the client pipeline on a per-call basis. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="content"/> is null. </exception>
         /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
         /// <returns> The response returned from the service. </returns>
-        public virtual async Task<ClientResult> CreateAsync(BinaryContent content, RequestOptions options = null)
+        public virtual async Task<ClientResult> CreateAsync(BinaryContent content, string contentType, RequestOptions options = null)
         {
             Argument.AssertNotNull(content, nameof(content));
```
annelo-msft, line 91: We should add `Assert.NotNullOrEmpty(contentType, nameof(contentType));` here -- I think I had left this out of the [original design](https://github.com/annelo-msft/openai-in-typespec/pull/3).  Thanks to @joseharriaga for catching this!

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Models/CreateFileRequest.Serialization.cs
```diff
@@ -88,6 +96,29 @@ internal static CreateFileRequest DeserializeCreateFileRequest(JsonElement eleme
...
+            else
+            {
+                return new BinaryData(stream.GetBuffer().AsMemory(0, (int)stream.Position));
+            }
+        }
+
+        internal virtual MultipartFormDataBinaryContent ToMultipartBinaryBody()
+        {
+            MultipartFormDataBinaryContent content = new MultipartFormDataBinaryContent();
+            content.Add(File, "file", "file", "application/octet-stream");
```
annelo-msft, line 117: We should also not be hard-coding the content-type here.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Internal/MultipartFormDataBinaryContent.cs
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        internal HttpContent HttpContent => _multipartContent;
+
+        private static string CreateBoundary()
+        {
+            Span<char> chars = new char[70];
+            byte[] random = new byte[70];
+            _random.NextBytes(random);
+            int mask = 255 >> 2;
```
annelo-msft, line 42: It looks like the comment explaining this is missing.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Internal/MultipartFormDataBinaryContent.cs
```diff
@@ -0,0 +1,197 @@
...
+            for (int i = 0; i < 70; i++)
```
annelo-msft, line 43: There should be a `Debug.Assert(_boundaryValues.Length - 1 == mask);` preceeding this line.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Internal/MultipartFormDataBinaryContent.cs
```diff
@@ -0,0 +1,197 @@
...
+            {
+                Argument.AssertNotNullOrEmpty(contentType, nameof(contentType));
+                AddContentTypeHeader(content, contentType);
+            }
+            _multipartContent.Add(content, name);
+        }
+
+        public static void AddFilenameHeader(HttpContent content, string name, string filename)
+        {
+            ContentDispositionHeaderValue header = new ContentDispositionHeaderValue("form-data") { Name = name, FileName = filename };
```
annelo-msft, line 153: The comment is missing from this method.

test/UnbrandedProjects/Platform-OpenAI-TypeSpec/src/Generated/Internal/MultipartFormDataBinaryContent.cs
```diff
@@ -0,0 +1,197 @@
...
+            content.Headers.ContentDisposition = header;
+        }
+
+        public static void AddContentTypeHeader(HttpContent content, string contentType)
+        {
+            MediaTypeHeaderValue header = new MediaTypeHeaderValue(contentType);
```
annelo-msft, line 159: Why are we adding this in a separate method?  I don't believe this is needed and should be removed.

chunyu3: Each part(not just file) of the multipart/form-data payload may set the `content-type` to indicate its content type(see following example). We need provide a way to set content-type of a part. 
image

annelo-msft: The BCL type will do this automatically for you.

chunyu3: BCL type will give default one. The request is customer can specify the `content-type` to each part.

PR #4625, opened 2024-04-23
fix the issue we have when we use enum as body in non-azure libraries

ArcturusZhang:
# Description

The issue this PR is fixing could be found here: https://github.com/Azure/autorest.csharp/pull/4623

In Azure version, we are doing this:
```csharp
using RequestContent content = BinaryData.FromObjectAsJson(stringValue);
```
this works because `RequestContent` has an implicit operator from `BinaryData`.
When we change to non-azure, this becomes:
```csharp
using BinaryContent content = BinaryData.FromObjectAsJson(stringValue);
```
But `BinaryContent` type from `System.ClientModel` does not define an implicit operator from `BinaryData` and we get compilation error.

This PR changes the implementation to be more explicit, like this:
```csharp
using RequestContent content = RequestContent.Create(BinaryData.FromObjectAsJson(stringValue));
using BinaryContent content = BinaryContent.Create(BinaryData.FromObjectAsJson(stringValue));
```
The implicit operation from `BinaryData` to `RequestContent` is also calling the same method `RequestContent.Create(BinaryData)` therefore this does not really change anything in azure libraries.

# Checklist

# Ready to Land?


test/CadlRanchProjects/client/naming/src/Generated/UnionEnum.cs
```diff
@@ -50,7 +50,7 @@ internal UnionEnum(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, U
         /// <include file="Docs/UnionEnum.xml" path="doc/members/member[@name='UnionEnumNameAsync(ClientExtensibleEnum,CancellationToken)']/*" />
         public virtual async Task<Response> UnionEnumNameAsync(ClientExtensibleEnum body, CancellationToken cancellationToken = default)
         {
-            using RequestContent content = BinaryData.FromObjectAsJson(body.ToString());
+            using RequestContent content = RequestContent.Create(BinaryData.FromObjectAsJson(body.ToString()));
```
ArcturusZhang, line 53: Our previous code works because we have an implicit operator from `BinaryData` to `RequestContent`.
But in our `System.ClientModel`, the equivalence of `RequestContent` is `BinaryContent`, but `BinaryContent` does not define an implicit operator from `BinaryData` to it, therefore in non-azure libraries, we have to call `BinaryContent.Create` explicitly.
To align the azure and non-azure libraries, I changed this to use the same way.

pshao25: Do we have non-azure libraries having enum body? In your PR, seems not.

pshao25: And just curious, seems RequestContent could be directly called from string as `RequestContent.Create(body.ToString());` not sure if there is any perf diff between these two statements.

ArcturusZhang: In the description: The issue this PR is fixing could be found here: https://github.com/Azure/autorest.csharp/pull/4623

ArcturusZhang: > And just curious, seems RequestContent could be directly called from string as `RequestContent.Create(body.ToString());` not sure if there is any perf diff between these two statements.

that would be a real change. My current change is not a real change.
And `RequestContent.Create(string)` is different. It is assuming the content is the string.
And `RequestContent.Create(BinaryData.FromObjectAsJson("value"))` will get the value with quotes

ArcturusZhang: And I believe our mockapi is implemented in this way (expecting the quotes): https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/enum/extensible/mockapi.ts

ArcturusZhang: FYI I put the cadl-ranch case in this PR as well, now we have a case of the issue this PR is trying to fix.

pshao25 (approved):

PR #4648, opened 2024-04-24
Fix relative paths for $root in powershell scripts

hallipr:
Several scripts were moved from /eng/scripts/preview to /eng/scripts, but internal relative paths weren't updated


hallipr:
fixes https://github.com/Azure/azure-sdk-tools/issues/8137

Issue #4590, opened 2024-04-17
[MPG][.Net] Unexpected diff for MgmtReferenceTypes test project


live1206:

live1206:
Planning to remove the logic of MgmtReferenceType in generator.
Then add the missing attributes, inherit types in the custom code.

PR #4663, opened 2024-04-26
Refine mgmt reference type

live1206:
Resolves https://github.com/Azure/autorest.csharp/issues/4590

- Maintain list of type names for `PropertyReferenceType` and `TypeReferenceType` and `ReferenceType`
  - For `ReferenceType`, it is `abstract` and there is no serialization for it since it is abstract
- Remove test project `MgmtReferenceTypes` because we can't have the same namespace and type names as `Azure.ResourceManager` is referenced in the test projects
- We can remove the custom code of `ResourceData` and `TrackedResourceData` after this


src/AutoRest.CSharp/Mgmt/Output/MgmtReferenceType.cs
```diff
@@ -14,23 +13,74 @@ namespace AutoRest.CSharp.Mgmt.Output
 {
     internal class MgmtReferenceType : MgmtObjectType
     {
+        private static HashSet<string> PropertyReferenceTypeModels = new HashSet<string>
```
ArthurMa1978, line 16: Can we put these into a config file?

ArthurMa1978 (approved):

src/AutoRest.CSharp/Mgmt/Output/MgmtReferenceType.cs
live1206: would prefer to keep them close to the actual logic

src/AutoRest.CSharp/Mgmt/Generation/ReferenceTypeWriter.cs
```diff
@@ -15,23 +15,30 @@ protected override void AddClassAttributes(CodeWriter writer, ObjectType objectT
...
+                writer.Line($"[{ReferenceClassFinder.TypeReferenceTypeAttribute}]");
+            }
+            else if (MgmtReferenceType.IsReferenceType(schema.ObjectSchema))
+            {
+                if (Configuration.IsBranded)
                 {
-                    writer.Line($"[{ReferenceClassFinder.TypeReferenceTypeAttribute}]");
+                    writer.UseNamespace("Azure.Core");
                 }
+                writer.Line($"[{ReferenceClassFinder.ReferenceTypeAttribute}(new string[]{{\"SystemData\"}})]");
```
ArcturusZhang, line 41: well we are hard coding a property name here which might not work if we would like to extend in the future.
But I do not think it is going to happen in any near or far future, so maybe we could add a TODO item here or add some comments here to note this hard coding because this only works for ResourceData and trackedResourceData case.

src/AutoRest.CSharp/Properties/launchSettings.json
```diff
@@ -432,10 +432,6 @@
       "commandName": "Project",
       "commandLineArgs": "--standalone $(SolutionDir)\\test\\TestProjects\\MgmtPropertyChooser\\src\\Generated -n"
     },
-    "MgmtReferenceTypes": {
-      "commandName": "Project",
-      "commandLineArgs": "--standalone $(SolutionDir)\\test\\TestProjects\\MgmtReferenceTypes\\src\\Generated -n"
-    },
```
ArcturusZhang, line 438: well looks like we have to delete this test project, which is unfortunate.
In the future we need to pay more attention on the regen PR for resourcemanager, because there are a few features that work differently when the IsArmCore configuration is true, and we now do not have test cases for them.

ArcturusZhang (approved):

src/AutoRest.CSharp/Properties/launchSettings.json
live1206: Yes, if we use `IsArmCore` instead of the namespace, we can keep them, but we prefer namespace for accuracy.

src/AutoRest.CSharp/Mgmt/Generation/ReferenceTypeWriter.cs
live1206: Added comments

PR #4682, opened 2024-05-07
Follow ups on adoption of `getAllModels`, and implement the new `additional-properties` cadl-ranch test cases

ArcturusZhang:
# Description

FYI the majority of the changed files in this PR comes from the changes in azure and non-azure versions of cadl-ranch case `type/property/additional-properties` because previously the `...Record<T>` syntax was not correctly supported, now it is generated correctly.
Changes in other test projects are either changes in `tspCodeModel.json` or in xml docs, no csharp file change, no API changes.

This PR includes:
1. Redirect the `getInputType` to use types from TCGC as well
2. Fix the `...Record<T>` syntax for additional properties (thanks to changes in 1)
3. Remove a lot of unreferenced code in our emitter
4. Update the generator so that we do not have to wrap unions into enums because TCGC did that for us

# Checklist

# Ready to Land?


archerzz (approved):

src/AutoRest.CSharp/Common/Input/InputTypes/Examples/ExampleMockValueBuilder.cs
```diff
@@ -63,11 +62,16 @@ private static InputParameterExample BuildParameterExample(InputParameter parame
                     // when it is constant, it could have DefaultValue
                     value = InputExampleValue.Value(parameter.Type, parameter.DefaultValue.Value);
                 }
-                else if (parameter.Type is InputUnionType unionType && unionType.UnionItemTypes.First() is InputLiteralType literalType)
+                else if (parameter.Type is InputUnionType unionType && unionType.UnionItemTypes[0] is InputLiteralType literalType)
                 {
                     // or it could be a union of literal types
                     value = InputExampleValue.Value(parameter.Type, literalType.Value);
                 }
+                else if (parameter.Type is InputEnumType enumType && enumType.AllowedValues[0].Value is { } enumValue)
```
chunyu3, line 70: what's the scenario does this handle?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/converter.ts
```diff
@@ -96,6 +96,31 @@ function fromSdkType(
     return {} as InputType;
 }
 
+// this function is only for the case when we get a type from a parameter, because in typespec, the parameters are
```
chunyu3, line 99:
```suggestion
// this function is only for the case when we get a type from a property, because in typespec, the properties are
```

ArcturusZhang: because this is handling an operation parameter, we have to add this comment to specify, this is not a typo.

test/TestProjects/MediaTypes-TypeSpec/src/Generated/tspCodeModel.json
```diff
@@ -116,38 +116,7 @@
...
-           "Kind": "Primitive",
-           "Name": "String",
-           "IsNullable": false
-          },
-          "Value": "application/octet-stream",
-          "IsNullable": false
-         }
-        ],
-        "IsNullable": false
+        "$ref": "2"
```
ArcturusZhang, line 119: "union to enum" is handled by TCGC now, therefore this changed from a union, to a enum (id 2 refers a enum)

src/AutoRest.CSharp/Common/Input/InputTypes/Examples/ExampleMockValueBuilder.cs
ArcturusZhang: This is fixing this case: https://github.com/Azure/autorest.csharp/pull/4682/files#r1593497407
And the purpose here is intuitive - a constant with a type of enum, we take its first value as example value.
If we do not handle this, the else branch returns a null value, and in generated examples, null or default will be used as the parameter value.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/converter.ts
chunyu3: The function called `fromSdkModelPropertyType` and it handle `SdKModelPropertyType`, so it is confused that it handle parameters in comments.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/converter.ts
```diff
@@ -316,7 +384,8 @@ function fromBytesType(bytesType: SdkBuiltInType): InputPrimitiveType {
 
 function fromStringType(
     program: Program,
-    stringType: SdkType
+    stringType: SdkType,
+    raw?: Type // we need the extra raw here because the format decorator is added to the property/parameter, but the raw in stringType is the type itself, and it does not have the format decorator we want.
```
chunyu3, line 388: What the raw? Is it `ModelProperty`? and it is a item related with `stringType`? If we set a parameter which need special relationship with another one, it may not good because customer may pass a wrong raw.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/converter.ts
ArcturusZhang: it is confused because TCGC named this type in this way. We cannot control it and this is the purpose we add this comment to emphasize that despite the name says "property", but we are handling parameters.
Also this should go away when we have getAllOperations

Issue #4683, opened 2024-05-07
`float` is emitted as `object`


joheredi:
**Describe the issue or request**
When a spec has a `float` type it gets generated as `object`

```ts
model Foo {
   TestNonNullableFloat: float;
   TestNullableFloat: float | null;
   TestNonNullableFloat32: float32;
   TestNullableFloat32: float32 | null;
   TestNonNullableFloat64: float64;
   TestNullableFloat64: float64 | null;
}
```

![image](https://github.com/Azure/autorest.csharp/assets/21109913/b5ebb1a6-bf50-489f-a397-89e1de6e0f59)

**Describe your ideas for solutions**
Maybe generate the max possible representation which would be `decimal`

/cc: @m-nash

ArcturusZhang:
Hi @joheredi per typespec document, `float` is the base type of `float32` and `float64`.
We did not handle this particular kind previously therefore it is emitted as `object`.
Since it is the base, from value's perspective, I think we should generate this type as `double`.
What do you think? @joheredi @m-nash

Issue #4131, opened 2024-01-14
Duplicate information in our input type and serialization system


ArcturusZhang:
We have quite a few kinds for primitive types: https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Input/InputTypes/InputTypeKind.cs#L6
it includes different variants of the same type, for instance:
```
    Date,
    DateTime,
    DateTimeISO8601,
    DateTimeRFC1123,
    DateTimeRFC3339,
    DateTimeRFC7231,
    DateTimeUnix,
```
these are all for the `DateTimeOffset` type.

In the meantime, we have a SerializationFormat: https://github.com/Azure/autorest.csharp/blob/8698fe440049c0adfe370c6286db9f248ea0d7e0/src/AutoRest.CSharp/Common/Output/Models/Serialization/SerializationFormat.cs#L8
it includes different ways of serializing/deserializing of the same type, for instance:
```
        DateTime_RFC1123,
        DateTime_RFC3339,
        DateTime_RFC7231,
        DateTime_ISO8601,
        DateTime_Unix,
        Date_ISO8601,
```
these are all for the `DateTimeOffset` type, and it overlaps with the defined kinds above.

We need to consolidate this so that we only keep one copy of this.

One way of doing this should be we simplify the InputTypeKind, and introduce the `encoding` property.

PR #4707, opened 2024-05-14
Consolidate `InputPrimitiveType`, `InputDateTimeType`, `InputDurationType`, `InputEnumType` and `InputIntrinsicType` with TCGC

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4683
Fixes https://github.com/Azure/autorest.csharp/issues/4712
Fixes https://github.com/Azure/autorest.csharp/issues/4713
Fixes https://github.com/Azure/autorest.csharp/issues/4714
Fixes https://github.com/Azure/autorest.csharp/issues/4131

# Description

This PR introduces the following types:

1. Use the `SdkTypeKinds` from TCGC as our kind for all input types (you should see related changes in the emitter and the generator)
	1. Related: since the kind is now from TCGC, the "encode" information is no longer bundled with the kind like we used to have such as `DateTimeISO8601`. This part fixes https://github.com/Azure/autorest.csharp/issues/4131
2. Move datetime type and duration type out of InputPrimitiveType to align current design of TCGC. This part fixes https://github.com/Azure/autorest.csharp/issues/4713
3. Remove `InputIntrinsicType`. Only `unknown` is meaningful to us and in TCGC it is a built-in type with the kind of `any`. Issue has been open for them to rename this to unknown. This part fixes https://github.com/Azure/autorest.csharp/issues/4712
4. Unify property names in enum type (EnumValueType => ValueType, AllowedValues => Values), minor change.
5. `ValueType` (formal `EnumValueType`)  is now a `InputPrimitiveType` instead of just a string.

# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/test/Unit/encode.test.ts
ArcturusZhang, line 1: removing the tests are not my intention.
Just because they have compilation errors which blocks me from generating code.
I will restore them and fix the compilation errors in them once I get everything in place.

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
```diff
@@ -1,44 +1,81 @@
...
+    Kind: "duration";
+    Encode: DurationKnownEncoding;
+    WireType: InputPrimitiveType;
+}
+
+export interface InputUnionType extends InputTypeBase {
     Kind: InputTypeKind.Union;
     Name: InputTypeKind.Union; // union type does not really have a name right now, we just use its kind
-    UnionItemTypes: InputType[];
+    UnionItemTypes: InputTypeBase[];
```
ArcturusZhang, line 71:
```suggestion
    UnionItemTypes: InputType[];
```

src/AutoRest.CSharp/Common/Input/InputTypes/InputDateTimeType.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AutoRest.CSharp.Common.Input;
+
+internal record InputDateTimeType(DateTimeKnownEncoding Encode, InputPrimitiveType WireType, bool IsNullable) : InputType("DateTime", IsNullable) // TODO -- name?
+{
+    public static InputDateTimeType Rfc3339 { get; } = new(DateTimeKnownEncoding.Rfc3339, InputPrimitiveType.String, false);
+    public static InputDateTimeType Rfc7231 { get; } = new(DateTimeKnownEncoding.Rfc7231, InputPrimitiveType.String, false);
+    public static InputDateTimeType UnixTimestamp { get; } = new(DateTimeKnownEncoding.UnixTimestamp, InputPrimitiveType.Int64, false);
```
chunyu3, line 10: UnixTimeStamp can use `int32` or `float` as `WireType`.  So it is not good enough to define a static instance `UnixTimestamp` with `Int64`

ArcturusZhang: This is no case that this is float - unix time stamp should always be integers
How about I rename this to `IntUnixTimestamp` and a new `LongUnixTimestamp`?

ArcturusZhang: And this is only a shortcut for us when we want to instantiate an instance of this. I think if we could make the name more descriptive to avoid a mistake when we use them, it should be fine.
The majority of the instance of this class should come from deserialization. We usually only use these static instances in the `CodeModelConverter` where we convert swaggers into input types.

ArcturusZhang: I changed my mind a little bit because we only handle the case with unix time stamp with long, therefore I will not add a static property here for IntUnixTimestamp

src/AutoRest.CSharp/Common/Input/InputTypes/BytesKnownEncoding.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AutoRest.CSharp.Common.Input;
+
+internal static class BytesKnownEncoding
+{
+    public const string Base64 = "base64";
+    public const string Base64Url = "base64url";
```
lirenhe, line 9: Shall we be consistent to use enum here?

archerzz: or maybe in TCGC this is defined as different type?

src/AutoRest.CSharp/Common/Input/InputTypes/DurationKnownEncoding.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AutoRest.CSharp.Common.Input;
+
+internal enum DurationKnownEncoding
+{
+    Iso8601, Seconds, Constant
```
lirenhe, line 8: Shall we use ISO8601?    And what does constant encoding mean?

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -48,46 +48,61 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
+                InputPrimitiveTypeKind.Int64 => new CSharpType(typeof(long), inputType.IsNullable),
+                InputPrimitiveTypeKind.SafeInt => new CSharpType(typeof(long), inputType.IsNullable),
+                InputPrimitiveTypeKind.Integer => new CSharpType(typeof(long), inputType.IsNullable), // in typespec, integer is the base type of int related types, see type relation: https://typespec.io/docs/language-basics/type-relations
+                InputPrimitiveTypeKind.Float => new CSharpType(typeof(double), inputType.IsNullable), // in typespec, float is the base type of float32 and float64, see type relation: https://typespec.io/docs/language-basics/type-relations
+                InputPrimitiveTypeKind.Numeric => new CSharpType(typeof(double), inputType.IsNullable), // in typespec, numeric is the base type of number types, see type relation: https://typespec.io/docs/language-basics/type-relations
+                InputPrimitiveTypeKind.IPAddress => new CSharpType(typeof(IPAddress), inputType.IsNullable),
+                InputPrimitiveTypeKind.ArmId => new CSharpType(typeof(ResourceIdentifier), inputType.IsNullable),
```
archerzz, line 77: Looks like we lost arm-id in old version of TypeFactory... Anyway, it's better now.

src/AutoRest.CSharp/Common/Input/InputTypes/DurationKnownEncoding.cs
ArcturusZhang: These values directly come from Typespec: https://github.com/archerzz/typespec/blob/c0376241326e5a8f413736a767dce1e5d4966fd9/packages/compiler/src/lib/decorators.ts#L667-L669

`Constant` is a backward compatible thing we introduced previously using directive.
For details of format in timespan, please refer: https://learn.microsoft.com/en-us/dotnet/api/system.timespan.parseexact?view=net-8.0#system-timespan-parseexact(system-string-system-string-system-iformatprovider-system-globalization-timespanstyles)
Our `constant` means `c` in this document, try search `"c"` (with quotes) and see how it looks like

ArcturusZhang: And usually C# uses camel cases when an abbr has 3 or more letters such as `Vmss` therefore I used `Iso` and I think we should change other places that we are using `ISO`.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
ArcturusZhang: No, old version of TypeFactory has arm-id, its previous name is `ResourceIdentifier`. Because we are now using the kinds from TCGC, we have to rename it.

src/AutoRest.CSharp/Common/Input/InputTypes/BytesKnownEncoding.cs
ArcturusZhang: No we cannot use enum here.
the type bytes is different from datetime and duration.
In TCGC, datetime and duration are standalone types, paralell with built-in types.
But the type bytes is a variant of built-in type. For this reason, the `encode` on built-in type (our InputPrimitiveType) must be `string` instead of enum.
Otherwise, please think, how should we name the enum? This encode should be shared with all types in the primitive types, it should be open to expand.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -48,46 +48,61 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
+                InputPrimitiveTypeKind.ResourceType => new CSharpType(typeof(ResourceType), inputType.IsNullable),
+                InputPrimitiveTypeKind.Stream => new CSharpType(typeof(Stream), inputType.IsNullable),
+                InputPrimitiveTypeKind.String => new CSharpType(typeof(string), inputType.IsNullable),
+                InputPrimitiveTypeKind.Uri or InputPrimitiveTypeKind.Url => new CSharpType(typeof(Uri), inputType.IsNullable),
+                InputPrimitiveTypeKind.Char => new CSharpType(typeof(char), inputType.IsNullable),
+                InputPrimitiveTypeKind.Any => UnknownType,
```
live1206, line 83: For MPG, I handled this as nullable to mitigate some regen diff.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -48,46 +48,61 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
             InputModelType { Namespace: "Azure.Core.Foundations", Name: "Error" } => SystemObjectType.Create(AzureResponseErrorType, AzureResponseErrorType.Namespace!, null).Type,
             // Handle DataFactoryElement, we are sure that the argument type is not null and contains only 1 element
-            InputModelType { Namespace: "Azure.Core.Resources", Name: "DataFactoryElement" } inputModel => new CSharpType(typeof(DataFactoryElement<>), inputType.IsNullable, CreateType(inputModel.ArgumentTypes![0])),
+            InputModelType { Namespace: "Azure.Core.Resources", Name: "DataFactoryElement" } inputModel => new CSharpType(typeof(DataFactoryElement<>), inputType.IsNullable, CreateTypeForDataFactoryElement(inputModel.ArgumentTypes![0])),
             InputModelType model => _library.ResolveModel(model).WithNullable(inputType.IsNullable),
             InputPrimitiveType primitiveType => primitiveType.Kind switch
             {
-                InputTypeKind.AzureLocation => new CSharpType(typeof(AzureLocation), inputType.IsNullable),
-                InputTypeKind.BinaryData => new CSharpType(typeof(BinaryData), inputType.IsNullable),
```
live1206, line 56: Is `BinarayData` handled `Any`?

src/AutoRest.CSharp/Common/Input/InputTypes/InputDurationType.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AutoRest.CSharp.Common.Input;
+
+internal record InputDurationType(DurationKnownEncoding Encode, InputPrimitiveType WireType, bool IsNullable) : InputType("Duration", IsNullable)
+{
+    public static InputDurationType Iso8601 { get; } = new(DurationKnownEncoding.Iso8601, InputPrimitiveType.String, false);
```
live1206, line 8: why are these always not nullable?

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: OK, I saw you handled nullable in CodeModelConverter instead.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
live1206: I found the answer in CodeModelConverter.

src/AutoRest.CSharp/Common/Input/InputTypes/InputDateTimeType.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AutoRest.CSharp.Common.Input;
+
+internal record InputDateTimeType(DateTimeKnownEncoding Encode, InputPrimitiveType WireType, bool IsNullable) : InputType("DateTime", IsNullable)
+{
+    public static InputDateTimeType Rfc3339 { get; } = new(DateTimeKnownEncoding.Rfc3339, InputPrimitiveType.String, false);
```
live1206, line 8: Same here, shall we allow the user to define nullability here?

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
```diff
@@ -1,34 +1,69 @@
...
+import {
+    DateTimeKnownEncoding,
+    DurationKnownEncoding
+} from "@typespec/compiler";
 
-export interface InputType {
-    Name: string;
-    Kind: InputTypeKind;
+interface InputTypeBase {
+    Kind: string;
```
live1206, line 14: why is `Kind` string instead of `enum`?

src/AutoRest.CSharp/Common/Input/InputTypes/InputDurationType.cs
ArcturusZhang: these are helpers for convenience usage. We do not have to have these.
I am just following the pattern we have in `InputPrimitiveType`.
If you want a nullable version, just call the ctor which is public as well

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
ArcturusZhang: this is not public type.
In public types, kind is either an enum or a literal value.
I am just copying how it is implemented in TCGC: https://github.com/Azure/typespec-azure/blob/9bd912d514d625a5106cfd7c83228c7f86467587/packages/typespec-client-generator-core/src/interfaces.ts#L82
Currently we are not ready to just simply use their types, therefore I mimic their implementation here.
But in ts code, it is not a big deal, we just need to ensure the json schema is the same which is enough

src/AutoRest.CSharp/Common/Input/InputTypes/InputDateTimeType.cs
ArcturusZhang: answered in the other comment: https://github.com/Azure/autorest.csharp/pull/4707#discussion_r1621603410

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
live1206: OK, since the contract is for public usage we should keep them `string`.
But when we assign the values, can we leverage `enum` to avoid typo? such as [this](https://github.com/Azure/autorest.csharp/blob/d2e20316508f3c013b4e88f9020f7e1476c7b088/src/TypeSpec.Extension/Emitter.Csharp/src/lib/converter.ts#L743)
And also, we are using const string for the mapping in generator [here](https://github.com/Azure/autorest.csharp/blob/d2e20316508f3c013b4e88f9020f7e1476c7b088/src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs#L66-L73). Would it be possible to put them into an `enum` instead?

Since this is not introduced by the current PR, maybe a TODO would be good enough to be handled later.

src/AutoRest.CSharp/Common/Input/InputTypes/InputDurationType.cs
live1206: I think it's OK to have the helpers, is it possible to turn them into methods to input more parameters such as nullable.
I am a bit concerned that with this in place, the consumer would simply ignore the nullability for them.

archerzz (approved):
LGTM

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs
```diff
@@ -63,46 +49,49 @@ private InputType CreateObject(ref Utf8JsonReader reader, JsonSerializerOptions
             return result ?? throw new JsonException("cannot deserialize InputType");
         }
 
-        private const string PrimitiveKind = "Primitive";
-        private const string LiteralKind = "Literal";
+        private const string LiteralKind = "constant";
         private const string UnionKind = "Union";
```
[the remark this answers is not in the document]
ArcturusZhang: We will gradually do this as well adopt more and more types from TCGC

src/AutoRest.CSharp/Common/Input/InputTypes/InputDateTimeType.cs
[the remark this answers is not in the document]
ArcturusZhang: This is not working as you said.
This is:
```csharp
public static InputDateTimeType Rfc3339 => new();
```

src/AutoRest.CSharp/Common/Input/InputTypes/InputDurationType.cs
ArcturusZhang: To avoid further confusions, I will remove these.

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
ArcturusZhang: Yeah it will be changed to lower case in following PRs. Let me add to dos.

ArcturusZhang: TODOs added

src/AutoRest.CSharp/Common/Input/InputTypes/InputDateTimeType.cs
ArcturusZhang: these helpers are removed.

live1206 (approved):
Need to sync shared core

lirenhe (approved):

Issue #4724, opened 2024-05-17
`float` based enums are generated as `int`


ArcturusZhang:
**Describe the issue or request**

See here: https://github.com/Azure/autorest.csharp/blob/d4442599b5f357d1ec2acb8ae27d4a10cb27cde4/test/UnbrandedProjects/Unbranded-TypeSpec/src/Generated/Models/FloatExtensibleEnum.cs#L17

Its corresponding tsp:
```typespec
@doc("Float based extensible enum")
union FloatExtensibleEnum {
    float32,
    One: 1.0,
    Two: 2.0,
    Four: 4.0,
}
```

PR #4725, opened 2024-05-17
fix the float based enum issue

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4724

# Description

Our float based enums are generated as based on int if all existing values are ints. For code examples please check the above issue.

This is incorrect and fixing it could be a breaking change - because the ctor changes for that enum:
```csharp
public FloatExtensibleEnum(float value) // <- this is `int value` before fixing
```
but we could get breaking changes if we do not fix it, and users cannot put float values into this enum which is just wrong.

# Checklist

# Ready to Land?


archerzz (approved):

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
```diff
@@ -76,32 +76,7 @@ public static InputEnumType CreateEnumType(ref Utf8JsonReader reader, string? id
                 throw new JsonException("Enum must have at least one value");
             }
 
-            InputPrimitiveType? currentType = null;
```
chunyu3, line 79: Why remove the value type creation and validation? we may still need check the enum value types consistence.

ArcturusZhang: I will introduce a fix for this in a new PR to make it throws a proper exception

PR #4732, opened 2024-05-20
now we throw a proper exception with correct information when enum value types are not consistent

ArcturusZhang:
# Description

# Checklist

# Ready to Land?


archerzz (approved):
Can we use a template function to verify value type?

ArcturusZhang:
> Can we use a template function to verify value type?

I think this is only a temporary solution, TCGC will update according to their discussion - but either way, in a short future, we no longer need this validation any more, because they are either forbidding it or allowing it by specify the value type to an union

PR #4743, opened 2024-05-21
feat(emitter): always make discriminator 1st property

archerzz:
# Description

# Checklist

# Ready to Land?


test/CadlRanchProjects/type/property/additional-properties/src/Generated/Docs/ExtendsUnknownDiscriminated.xml
```diff
@@ -141,8 +141,8 @@ using RequestContent content = RequestContent.Create(new
 {
     index = 1234,
     age = 123.45F,
-    name = "<name>",
     kind = "derived",
```
ArcturusZhang, line 144: so if we would like to put discriminator as the first property, we definitely need to do more, because here is an example of this fix is not fixing.
Here we have a case that the base model with discriminator, has another base, and we always write the base properties first, and those properties appear before your discriminator.
Therefore I think it needs more changes to fix this issue.

Issue #4151, opened 2024-01-23
[Dpg][Sample] System.ArgumentException: An item with the same key has already been added.


ArcturusZhang:
Exception will be thrown when there are parameters with the same name in the invocation chain of an operation.
For instance if we have a client with parameter `foo`, and the operation also has a parameter `foo`, we will get the exception.

Root cause:
Now in the sample generator, we are flattening all the parameters with their corresponding value by their name: https://github.com/Azure/autorest.csharp/blob/cfc58b14dd6a43a67bcc79dc3260caf1bf1b0ac4/src/AutoRest.CSharp/LowLevel/Output/Samples/DpgOperationSample.cs#L147

The issue first seen in cadl-ranch case `client/structure/default`.

ArcturusZhang:
The [fix](https://github.com/Azure/autorest.csharp/pull/4082) might solve this issue for the case where this issue is first found.

PR #4745, opened 2024-05-22
Fix client hierarchy issue

### LINKED ISSUE #4081: Parent class with no operations but subclass has wrong get subclient signature.
spec:
```
@server(
  "{endpoint}/client/structure/{client}",
  "",
  {
    @doc("Need to be set as 'http://localhost:3000' in client.")
    endpoint: url,

    @doc("Need to be set as 'default', 'multi-client', 'renamed-operation', 'two-operation-group' in client.")
    client: ClientType,
  }
)
@service({
  title: "MultiClient",
  version: "1.0.0",
})
namespace Client.Structure.Service;

namespace Baz {
  interface Foo {
    #suppress "@azure-tools/cadl-ranch-expect/missing-scenario" "This is by design those operations get defined as scenarios in the client"
    @route("/seven")
    @post
    seven(): void;
  }
}
```
In the `Baz`, there is no operations but there is a sub client `Foo`. And in the service there is a global parameter `client`.

Actual:
```C#
public virtual BazFoo GetBazFooClient(string client, string apiVersion = "1.0.0")
{
    Argument.AssertNotNullOrEmpty(client, nameof(client));
    Argument.AssertNotNull(apiVersion, nameof(apiVersion));

    return new BazFoo(ClientDiagnostics, _pipeline, _endpoint, client, apiVersion);
}
```

Expected:
```C#
private BazFoo _cachedBazFoo;

/// <summary> Initializes a new instance of BazFoo. </summary>
public virtual BazFoo GetBazFooClient()
{
    return Volatile.Read(ref _cachedBazFoo) ?? Interlocked.CompareExchange(ref _cachedBazFoo, new BazFoo(ClientDiagnostics, _pipeline, _endpoint, _client, _apiVersion), null) ?? _cachedBazFoo;
}
```

pshao25:
Fix #4081


src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
```diff
@@ -284,7 +284,7 @@ private IReadOnlyList<ClientInfo> SetHierarchy(IReadOnlyDictionary<string, Clien
                 var clientNamespace = Configuration.Namespace;
                 var infoForEndpoint = topLevelClients.FirstOrDefault(c => c.ClientParameters.Any(p => p.IsEndpoint));
                 var endpointParameter = infoForEndpoint?.ClientParameters.FirstOrDefault(p => p.IsEndpoint);
-                var clientParameters = endpointParameter != null ? new[] { endpointParameter } : Array.Empty<InputParameter>();
+                var clientParameters = topLevelClients.SelectMany(c => c.ClientParameters.Where(p => !p.IsRequired || p.IsApiVersion || p.IsEndpoint)).Distinct().ToArray();
```
archerzz, line 287: What about other optional client paramters?

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -63,12 +63,12 @@ public Parameter[] GetOrderedParametersByRequired()
...
 
-        public static IEnumerable<InputParameter> GetParametersFromOperations(IEnumerable<InputOperation> operations) =>
-            operations
-                .SelectMany(op => op.Parameters)
-                .Where(p => p.Kind == InputOperationParameterKind.Client)
-                .Distinct()
-                .ToList();
+        public static IEnumerable<InputParameter> GetParametersFromClient(InputClient client)
+        {
+            return (client.Parameters ?? new List<InputParameter>())
```
ArcturusZhang, line 68: I think we should use `Array.Empty<>()` here.
And also we should make `Parameters` non-nullable (in the deserialization) on the input types so that here you do not have to do this.

src/AutoRest.CSharp/LowLevel/Output/Samples/DpgOperationSample.cs
```diff
@@ -142,7 +142,7 @@ public IEnumerable<ValueExpression> GetValueExpressionsForParameters(IEnumerable
                     }
                     // if it is optional, we just do not put it in the map indicates that in the invocation we could omit it
                 }
-                else
+                else if (!result.ContainsKey(parameter.Name))
```
archerzz, line 145: under what case the parameter will be duplicated?

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -63,12 +63,12 @@ public Parameter[] GetOrderedParametersByRequired()
...
+                .Concat(client.Operations.SelectMany(op => op.Parameters).Where(p => p.Kind == InputOperationParameterKind.Client))
+                .Distinct();
```
ArcturusZhang, line 70: is it possible that two operations have parameters with the same name and kind, but they are different instances?
This could come from an issue in our emitter (for example they should be the same instance, but we incorrectly made them different) or come from a case they just have different parameter but by accident with same name and kind (different type for example)
If this could happen, here we may have trouble.

In summary that I sense this piece of code is fragile if we could find a way to make it more solid, it would be better.

archerzz (approved):

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
chunyu3: Current in out tspCodeModel, those client parameter in operations is the same instance (using ref to connect), so in most case it is not cause problem. 

But yes, we may need to find a way to make it more solid, especially when we adopt client.parameters from tcgc

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
chunyu3: Optional client parameter will be put in ClientOptions. So we still need optional client parameter, and may need to add int clientInfo.ClientParameters

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
pshao25: You are saying 
```
op method1(@global parameter: string): void;
op method2(@global parameter: string): void;
```
Customer wants these two parameters to be different instances. Then there must a way to let us/emitter know (no matter by decorators or other ways). We get this information and deal it with different `Parameter`s. 

I don't worry too much about it:
1. Customer writes bad spec, they should take responsibility themselves.
2. The same logic exists in M4.
3. Maybe in the new InputModel, the global parameters is a property of `InputClient` instead of `InputOperation`.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
pshao25: Normal topLevelClients will not go to this place. Only when user sets "single-top-level-client", we needing make up a new topLevelClient will go to this place. In this case, we need to bring the optional client parameters up to the new topLevelClient.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
ArcturusZhang: If we follow option 3, I agree there is no issues.

For 1, yes the customer should take responsibility if they wrote bad specs and bad API was generated. But if we generate compilation errors, they will just claim there is a bug in the generator.
Bad spec => bad API
but not
Bad spec /=> invalid code

src/AutoRest.CSharp/LowLevel/Output/Samples/DpgOperationSample.cs
ArcturusZhang: Is this fixing this issue? https://github.com/Azure/autorest.csharp/issues/4151

pshao25: Yes, but this happens before I did the current change. I'll revert this until next time we see this.

ArcturusZhang:
Please fix all the CI issues on the corresponding regen in order not to get surprises when we merge this PR and found out there is something we could not fix.

src/AutoRest.CSharp/LowLevel/Output/DpgOutputLibraryBuilder.cs
chunyu3: Even it is a made-up topLevelClient when apply `single-top-level-client` we still need make sure this client contains all the needed parameters.

PR #4749, opened 2024-05-22
Set Resourcename for InputOperation

live1206:
Set `InputOperation.ResourceName` from first part of OperationId


src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
```diff
@@ -81,6 +81,7 @@ public static InputOperation RemoveApiVersionParam(InputOperation operation)
     }
 
     public string CleanName => Name.IsNullOrEmpty() ? string.Empty : Name.ToCleanName();
+    internal string OperationId => $"{ResourceName}_{Name}";
```
live1206, line 84: This is not in use yet.

live1206:
Set `InputOperation.Resourcename` will cause brekaing changes for low-level DPG, will keep the behavior for them.

PR #4754, opened 2024-05-23
Remove CompositionModels from InputModelType

live1206:
Since [CompoistionModels](https://github.com/Azure/autorest.csharp/blob/568fab1cba38a3764a6b475856d61cd3aa43be07/src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs#L26) is only used to construct properties, we can deal with them in CodeModelConverter in advance. 
So that, it can be removed from InputModelType.


src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs
```diff
@@ -182,46 +180,33 @@ private static InputTypeSerialization UpdateInputSerialization(InputModelType in
             return serialization;
         }
 
-        private static IReadOnlyList<InputModelProperty> UpdateInputModelProperties(InputModelType inputModel, INamedTypeSymbol? existingBaseType, string ns, SourceInputModel? sourceInputModel)
+        private IReadOnlyList<InputModelProperty> UpdateInputModelProperties(InputModelType inputModel, INamedTypeSymbol? existingBaseType, string ns, SourceInputModel? sourceInputModel)
```
ArcturusZhang, line 183: do we have to remove the static?
If we remove the static, quite a few parameters could be removed

live1206: Remove static to be able to use instance fields, such as `typeFactory`, it does not need to be static

live1206: See your point now, will update it

ArcturusZhang (approved):

ArthurMa1978 (approved):

m-nash (approved):

Issue #4771, opened 2024-05-28
Hard coded repo name leads to test failure in other machine


pshao25:
In this test, we [hard code repo name as "autorest.csharp"](https://github.com/Azure/autorest.csharp/blob/b1eff796298fdec2f03e14f580167009d79ec689/test/AutoRest.TestServer.Tests/Mgmt/OutputLibrary/OutputLibraryTestBase.cs#L40).  However, when we clone the repo, the name doesn't have to be "autorest.csharp". It could be like "autorest", e.g. [in regen pipeline we call it autorest](https://github.com/Azure/azure-sdk-tools/blob/c59ab4e32d3e7decb59d6b5f9324b45c08d400d2/eng/pipelines/templates/stages/archetype-autorest-preview.yml#L242). This leads to all the tests based on `OutputLibraryTestBase` fail.

PR #4773, opened 2024-05-28
Remove hard-code segment from test base path

live1206:
Resolves https://github.com/Azure/autorest.csharp/issues/4771


ArthurMa1978 (approved):

pshao25:
Can we start a TypeSpec.next regen to check?

live1206:
The test failures are due to TypeSpec.next regen, which needs further investigation.
The fix in this PR is verified.

PR #4774, opened 2024-05-28
Call MGC emitter as plugin

chunyu3:
# Description
Problem to solve: 
Current we have two emitter: autorest.csharp emitter and mcg emitter. mcg emitter will is common library which will service both branding and non-branding service. Autorest.csharp emitter will service for azure service. 
There are some channels:

- We need to maintain two emitter and it is hard to maintain and keep consistent. 

- When we want to implement a feature, we need to update two emitters which is not convenient.

So we will update to use plug-in framework to resolve this.

- MCG emitter is the base emitter, and autorest.csharp will load MCG emitter as plugin and call it to emitter general inputModel: tspCodeModel, then autorest.csharp emitter will apply some special azure logic to enhance the tspCodeModel.

# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -1,203 +1,121 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-import { createSdkContext } from "@azure-tools/typespec-client-generator-core";
```
m-nash, line 4: I was expecting all other *.ts source files can be deleted?  Can we add that and see what it looks like in this pr?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -1,203 +1,121 @@
...
-import {
-    EmitContext,
-    Program,
-    logDiagnostics,
-    resolvePath
-} from "@typespec/compiler";
+import { EmitContext, Program, resolvePath } from "@typespec/compiler";
```
pshao25, line 4: How we are going to handle the Azure exclusive logic which would not go into http-client-csharp?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: all unused *.ts source files are deleted.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: As discussed stage 1: all decorators(include azure-specific decorator) will be handled in `http-client-csharp`, so we does not need specific Azure logic in emitter. 
in future, (stage 2 or later stage) tcgc will separate azure/non-azure libraries, and more complex emitter framework (under-discussing) designed, we then will handle this.

src/TypeSpec.Extension/Emitter.Csharp/package.json
```diff
@@ -27,46 +27,23 @@
...
-        "@typespec/compiler": ">=0.50.0 <1.0.0",
-        "@typespec/http": ">=0.50.0 <1.0.0",
-        "@typespec/openapi": ">=0.50.0 <1.0.0",
-        "@typespec/rest": ">=0.50.0 <1.0.0",
-        "@typespec/versioning": ">=0.50.0 <1.0.0"
+        "json-serialize-refs": "0.1.0-0",
+        "@typespec/http-client-csharp":"0.1.9-alpha.20240625.9"
     },
     "devDependencies": {
-        "@azure-tools/typespec-azure-core": "0.43.0",
```
pshao25, line 53: I think we could remove the peer dependency, but the devdependency is a must.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/load-module.ts
```diff
@@ -0,0 +1,174 @@
+import {
```
lirenhe, line 1: Please add license information at the top of this file.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -1,205 +1,135 @@
...
-    const outputFolder = resolveOutputFolder(context);
-
     /* set the loglevel. */
     Logger.initialize(program, options.logLevel ?? LoggerLevel.INFO);
 
-    if (!program.compilerOptions.noEmit && !program.hasError()) {
-        // Write out the dotnet model to the output path
-        const sdkContext = createSdkContext(context, packageName);
-        const root = createModel(sdkContext);
+    /* load microsft csharp emitter. */
```
lirenhe, line 26:
```suggestion
    /* load microsoft csharp emitter. */
```

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -1,205 +1,135 @@
...
-                "azure-arm":
-                    sdkContext.arm === false ? undefined : sdkContext.arm,
-                "disable-xml-docs":
-                    options["disable-xml-docs"] === false
-                        ? undefined
-                        : options["disable-xml-docs"]
-            };
+        await emitFunction(context);
+    } catch (error: unknown) {
+        throw new Error("emitter error");
```
lirenhe, line 60: Shall we show the details of the errors here?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -1,205 +1,135 @@
...
+
+    const emitFunction = entrypoint.esmExports.$onEmit;
+    try {
+        context.options.skipSDKGeneration = true;
         if (
-            context.program.diagnostics.length > 0 &&
-            context.program.diagnostics.filter(
-                (digs) => digs.severity === "error"
-            ).length > 0
+            context.emitterOutputDir &&
```
lirenhe, line 53: I think this would cause change in the test problem, why we don't use context directly here?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -1,205 +1,135 @@
...
-
-            if (!options["save-inputs"]) {
-                // delete
-                deleteFile(resolvePath(generatedFolder, tspOutputFileName));
-                deleteFile(resolvePath(generatedFolder, configurationFileName));
-            }
+        }
+        configurations["shared-source-folders"] = resolvedSharedFolders ?? [];
+        configurations["flavor"] =
+            options["flavor"] ??
```
lirenhe, line 91: Why we need the logic of handling flavor here?  Is this not handled in common emitter?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/load-module.ts
```diff
@@ -0,0 +1,174 @@
...
+    const metadata = computeLibraryMetadata(module, libDefinition);
+
+    return {
+        ...resolution,
+        metadata,
+        definition: libDefinition,
+        linter: entrypoint?.esmExports.$linter
+    };
+}
+function computeLibraryMetadata(
```
lirenhe, line 143: add new line here.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -1,205 +1,135 @@
...
     Logger.initialize(program, options.logLevel ?? LoggerLevel.INFO);
 
-    if (!program.compilerOptions.noEmit && !program.hasError()) {
-        // Write out the dotnet model to the output path
-        const sdkContext = createSdkContext(context, packageName);
-        const root = createModel(sdkContext);
+    /* load microsft csharp emitter. */
+    const basedir = url.fileURLToPath(new URL(".", import.meta.url));
+    Logger.getInstance().info(`basedir: ${basedir}.`);
+    const emitterNameOrPath = "@typespec/http-client-csharp";
```
lirenhe, line 29: As @typespec/http-client-csharp is added into package.json,  can we call the exported $onEmit directly here instead of writing code to load the module/entrypoint?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: This is azure-specific logic(set the flavor according to the sdk package name) which is not included in common emitter.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: `http-client-csharp` emitter require the output dir should be the sdk package root directory. But in Autorest.csharp emitter it allow src directory or root directory.

lirenhe (approved):

samples/AnomalyDetector/tspCodeModel.json
```diff
@@ -21,14 +20,14 @@
      "Value": "v1.1"
     }
    ],
+   "CrossLanguageDefinitionId": "AnomalyDetector.APIVersion",
```
ArcturusZhang, line 23: OK is this change from the emitter from microsoft/typespec side?
if so could you open a PR on that side to revert this? This is quite unnecessary and may spam other changes.
We should expect there was nothing changed here in this PR.

test/CadlRanchProjects/azure/core/model/Configuration.json
ArcturusZhang, line 1: why these two files are new files instead of moving to this location?

src/TypeSpec.Extension/Emitter.Csharp/package.json
```diff
@@ -27,15 +27,16 @@
...
     },
     "files": [
         "dist/src/**"
     ],
     "dependencies": {
-        "@autorest/csharp": "3.0.0-beta.20230309.1",
-        "json-serialize-refs": "0.1.0-0"
+        "@autorest/csharp": "3.0.0-beta.20240625.4",
+        "json-serialize-refs": "0.1.0-0",
+        "@typespec/http-client-csharp":"https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest/npm/registry/@typespec/http-client-csharp/-/http-client-csharp-0.1.9-alpha.20240712.1.tgz"
```
ArcturusZhang, line 39: so in the future if we want to bump the version of emitter from typespec, we should update this part, right?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -1,205 +1,117 @@
...
+            options["flavor"] ??
+            (configurations.namespace.toLowerCase().startsWith("azure.")
+                ? "azure"
+                : undefined);
+        /* set azure-arm */
+        const sdkContext = createSdkContext(
+            context,
+            "@azure-tools/typespec-csharp"
+        );
+        configurations["azure-arm"] =
```
ArcturusZhang, line 75: I heard that there was a TODO item about ARM to parse it from the decorators list?
If so, could you add a TODO here?

ArcturusZhang (approved):

src/TypeSpec.Extension/Emitter.Csharp/package.json
chunyu3: Yes. and also need to update the typespec library in following `dev-dependency` if we upgrade the typespec libraries in http-client-csharp.
Because all those typespec libraries and tcgc libraries in `@typespec/http-client-csharp` are defined as dev-dependency, in some case (such as in dev environment), dev-dependencies are not installed when we install `@typespec/http-client-csharp`. So we need to duplicate to add typespec libraries and tcgc libraries as dev-dependency in `autorest.csharp` emitter.  
In the future when the typespec libraries and tcgc libraries are stable enough, we need to change them as `dependency` in `http-client-csharp`. Then we can remove the duplication.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
chunyu3: Yes. Long-term goal we will parse this from decorator list or maybe other approach. TODO added

PR #4776, opened 2024-05-28
fix(emitter): filter out api versions not specified by `api-version`

### LINKED ISSUE #4775: bug(dpg): some api versions should be ignored when `api-version` is specified
## Problem

See the `api-version` config in `SpecificVersion-TypeSpec`: https://github.com/Azure/autorest.csharp/blob/b1eff796298fdec2f03e14f580167009d79ec689/test/TestProjects/SpecificVersion-TypeSpec/tspconfig.yaml#L4

After specifying `"api-version": "2022-09-01"`, the version `2022-12-01-preview` should be ignored in the generated code. However, in the generated `SpecificClientOptions.cs`, we still have it: https://github.com/Azure/autorest.csharp/blob/b1eff796298fdec2f03e14f580167009d79ec689/test/TestProjects/SpecificVersion-TypeSpec/src/Generated/SpecificClientOptions.cs#L26

archerzz:
# Description

- add logic to filter out api versions newer than specified `api-version`
- add unit tests
- regen

fix #4775

# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/src/lib/client-model-builder.ts
```diff
@@ -85,12 +85,30 @@ export function createModelForService(
 
     const apiVersions: Set<string> | undefined = new Set<string>();
     let defaultApiVersion: string | undefined = undefined;
-    const versions = getVersions(program, service.type)[1]?.getVersions();
+    let versions = getVersions(program, service.type)[1]
+        ?.getVersions()
+        .map((v) => v.value);
+    if (sdkContext.emitContext.options["api-version"] !== undefined) {
```
chunyu3, line 91: Need consider the scenario "api-version" = "all" where it should contain all the api-version enum value

src/TypeSpec.Extension/Emitter.Csharp/src/lib/client-model-builder.ts
```diff
@@ -85,12 +85,30 @@ export function createModelForService(
...
+        const targetApiVersion = sdkContext.emitContext.options["api-version"];
+        // azure swagger version doesn't follow SemVer spec so that we cannot use existing library to compare versions.
+        // We need to filter out the versions that are greater than the target version.
+        // Basically 3 cases regarding version v.s. targetApiVersion:
+        // - 2022-02-01-preview = 2022-02-01-preview: just compare equals
+        // - 2022-02-01-preview < 2022-03-01-preview: this can be compared by dictionary order.
+        //   - 2022-02-01 > 2022-02-01-preview: this need to check if the targetApiVersion contains the version.
+        // - 2022-02-01-preview < 2022-02-01: this need to check if the version contains the targetApiVersion.
+        versions = versions?.filter(
```
chunyu3, line 100: We may not use dictionary order to do the comparing. We need to use the order or the index of the api-version in the version array to identify if the api-version is newer one or older one.

archerzz: > We may not use dictionary order to do the comparing. We need to use the order or the index of the api-version in the version array to identify if the api-version is newer one or older one.

Then it relies on the upstream to always give the correct order.

archerzz: I think in near future, the available api-versions should be given by TCGC SdkPackage. So here we just make sure it can work for existing SDKs.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/client-model-builder.ts
archerzz: I remember Chenjie told me not to try `all` which is still pending.

archerzz:
Since we'll get available API versions from TCGC probably `SdkPackage`, I resolved the commentd by applying the suggested changes.

chunyu3 (approved):

Issue #4304, opened 2024-02-29
Model with discriminator that defined with 'is' operator - not generated.


asaflevi-ms:
(A)
**@discriminator("kind")**
model RadiologyInsightsInference **is** Fhir.R4.Extendible;

All models that extends from RadiologyInsightsInference not generated (only csharp emitter) 

(B)
**@discriminator("kind")**
model RadiologyInsightsInference {
  ...Fhir.R4.Extendible;
}

Models are generated!.

See long thread: https://teams.microsoft.com/l/message/19:acbd512e57bd475198ea6bf4564599e3@thread.skype/1707746708507?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1707412749502&teamName=Azure%20SDK&channelName=AutoRest%20-%20CSharp&createdTime=1707746708507

lirenhe:
@archerzz, I expect this problem will be auto resolved after we apply TCGC in emitter, could you help to confirm?

PR #4781, opened 2024-05-29
Consolidate model types from TCGC

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4784
Fixes https://github.com/Azure/autorest.csharp/issues/4304

# Description

Align with the model definition on TCGC.

Known issues:
1. The derived types with discriminator will still have the discriminator property defined on them. For example:
```
@discriminator("kind")
model Base {
	kind: string;
}

model Derived extends Base {
	kind: "a";
}
```
we will have the property `kind` on type `Derived` with a different type of literal. It is not causing anything changed in the generated code, but leading to a lot of extra enums because we are wrapping literals into enums. Made a workaround on the PR to fix this.

# Checklist

# Ready to Land?


ArcturusZhang:
`Models-TypeSpec` has some issue, need to investigate

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace AutoRest.CSharp.Common.Input
 {
-    internal record InputModelType(string Name, string? Namespace, string? Accessibility, string? Deprecated, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, string? DiscriminatorPropertyName, InputDictionaryType? InheritedDictionaryType, IReadOnlyList<InputType>? ArgumentTypes = null)
+    internal record InputModelType(string Name, string? Namespace, string? Access, string? Deprecation, string? Description, InputModelTypeUsage Usage, IReadOnlyList<InputModelProperty> Properties, InputModelType? BaseModel, IReadOnlyList<InputModelType> DerivedModels, string? DiscriminatorValue, InputModelProperty? DiscriminatorProperty, IReadOnlyDictionary<string, InputModelType> DiscriminatedSubtypes, InputType? AdditionalProperties, IReadOnlyList<InputType>? ArgumentTypes = null)
```
ArcturusZhang, line 10: it turns out that we have to keep the `DerivedModels` property here.
We need this to know whether this model has children, if so, we need to add `protected` to raw data field.

chunyu3: In that case, we still need to keep the field order in `InputModelType` that the `BaseModel` should keep `BaseModel` and `Properties` as the last two item in `InputModelType`

chunyu3 (approved):

src/AutoRest.CSharp/Common/Input/InputTypes/InputModelType.cs
ArcturusZhang: as you could see, everything works fine.
The order is "property deserialization order", not the order you are seeing here.

Issue #4701, opened 2024-05-13
Emitter needs to adopt TCGC's recent changes on `nullable`


chunyu3:
TCGC will label `Nullable` on property/parameter instead of type.
Codegen emitter will adapt this:

- update emitted model, apply `Nullable` on property/parameter
- update inputModel in autorest.csharp accordingly.
- Still label `Nullable` in CsharpType, update the TypeFactory.cs to pass `Nullable` parameter when create CsharpType.

ArcturusZhang:
we could take this opportunity to review and refine the design of our input types as well

ArcturusZhang:
Also this should fix https://github.com/Azure/autorest.csharp/issues/4314

archerzz:
Latest decision is to deprecate all `nullable`, and return union for nullable types in TCGC. There will be a helper function `isNullable` to help check if the type is nullable. For details, see https://github.com/Azure/typespec-azure/pull/870

Issue #4314, opened 2024-03-02
Type `type | null` overrides the behavior of `type`


pshao25:
See [this type](https://github.com/Azure/autorest.csharp/blob/68ce565d8056bff0e58952f4c2c9617bcc8f5847/test/TestProjects/FirstTest-TypeSpec/FirstTest-TypeSpec.tsp#L74) in our test.

It is [used as `|null`](https://github.com/Azure/autorest.csharp/blob/68ce565d8056bff0e58952f4c2c9617bcc8f5847/test/TestProjects/FirstTest-TypeSpec/FirstTest-TypeSpec.tsp#L128)
```
model ModelWithRequiredNullableProperties {
    requiredExtensibleEnum: StringExtensibleEnum | null;
}
```

It is also [used as itself](https://github.com/Azure/autorest.csharp/blob/68ce565d8056bff0e58952f4c2c9617bcc8f5847/test/TestProjects/FirstTest-TypeSpec/FirstTest-TypeSpec.tsp#L160).
```
model RoundTripModel {
    requiredDictionary: Record<StringExtensibleEnum>;
}
```

Actual behavior:
https://github.com/Azure/autorest.csharp/blob/68ce565d8056bff0e58952f4c2c9617bcc8f5847/test/TestProjects/FirstTest-TypeSpec/src/Generated/Models/RoundTripModel.cs#L169
```
public IDictionary<string, StringExtensibleEnum?> RequiredDictionary { get; }
```
The value of this dictionary becomes nullable, because our current behavior is " `type | null` overrides the behavior of `type`"

Expected behavior:
The value of this dictionary should be not nullable.

PR #4790, opened 2024-06-03
upgrade typespec library and remove Nullable flag from type and property

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/4840
Fix https://github.com/Azure/autorest.csharp/issues/4701
Fix https://github.com/Azure/autorest.csharp/issues/4314
- remove `IsNullable` from InputType, and define a new type `InputNullableType` to indicate a type which is nullable
- resolve serialization issue: when a property type is customized, `serializeType` should use the customized csharp type.
- upgrade typespec library to 0.57.0
- Current `Accessiblity` of each model/enum is `public` by default (previous is undefined if no `@access` decorated), so .NET will use `getAccessOverride` to get the overrided accessiblity via `@access` decorator, otherwise set `undefine` as before

# Checklist

# Ready to Land?


test/TestProjects/FirstTest-TypeSpec/src/Generated/tspCodeModel.json
```diff
@@ -386,8 +380,8 @@
     }
    ],
    "Namespace": "FirstTestTypeSpec",
+   "Accessibility": "public",
```
ArcturusZhang, line 383: this is the reason

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type-kind.ts
```diff
@@ -5,5 +5,6 @@ export enum InputTypeKind {
     Model = "Model",
     Array = "Array",
     Dictionary = "Dictionary",
-    Intrinsic = "Intrinsic"
+    Intrinsic = "Intrinsic",
+    Nullable = "Nullable",
```
ArcturusZhang, line 9: my consolidation PRs aim to remove this type by just simply using the kind from TCGC (even a literal value), therefore please do not add this value here.

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
```diff
@@ -105,6 +105,12 @@ export interface InputEnumType extends InputTypeBase {
     Usage: string;
 }
 
+export interface InputNullableType extends InputTypeBase {
+    Kind: InputTypeKind.Nullable;
+    Name: InputTypeKind.Nullable;
+    Type: InputType;
```
ArcturusZhang, line 111: we should align TCGC's definition when introducing a new type
```suggestion
    Kind: "nullable";
    Type: InputType;
```

test/TestProjects/OldestVersion-TypeSpec/src/Generated/Models/ExportedResource.cs
```diff
@@ -85,5 +90,7 @@ internal ExportedResource()
         public string ResourceUri { get; }
         /// <summary> Gets the name. </summary>
         public string Name { get; }
+        /// <summary> Gets the type. </summary>
+        public string Type { get; }
```
chunyu3, line 94: @typeSpec/Versioning resolve issue https://github.com/microsoft/typespec/issues/3239, so the `Type` Field should be generated.

test/TestProjects/OldestVersion-TypeSpec/src/Generated/VersioningOp.cs
```diff
@@ -257,16 +257,17 @@ public virtual Pageable<BinaryData> GetResources(IEnumerable<string> select, str
...
+        /// <param name="maxLines"> The <see cref="int"/>? to use. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="name"/> is null. </exception>
         /// <exception cref="ArgumentException"> <paramref name="name"/> is an empty string, and was expected to be non-empty. </exception>
-        /// <include file="Docs/VersioningOp.xml" path="doc/members/member[@name='ExportAsync(WaitUntil,string,string,string,CancellationToken)']/*" />
-        public virtual async Task<Operation<ExportedResource>> ExportAsync(WaitUntil waitUntil, string name, string projectFileVersion = null, string removedQueryParam = null, CancellationToken cancellationToken = default)
+        /// <include file="Docs/VersioningOp.xml" path="doc/members/member[@name='ExportAsync(WaitUntil,string,string,string,int?,CancellationToken)']/*" />
+        public virtual async Task<Operation<ExportedResource>> ExportAsync(WaitUntil waitUntil, string name, string projectFileVersion = null, string removedQueryParam = null, int? maxLines = null, CancellationToken cancellationToken = default)
```
chunyu3, line 265: same reason as above

src/TypeSpec.Extension/Emitter.Csharp/test/Unit/model-type.test.ts
```diff
@@ -69,13 +69,12 @@ op test(@body input: Pet): Pet;
                 SerializedName: "kind",
                 Type: {
                     Kind: "string",
-                    IsNullable: false,
                     Encode: undefined
                 },
                 IsRequired: true,
                 IsReadOnly: false,
                 IsDiscriminator: true,
-                Description: "Discriminator",
```
ArcturusZhang, line 78: why this changes?

chunyu3: TCGC update the description of `discriminator` in the latest version. So there are a lot of description changes in xml doc.

test/CadlRanchProjects/azure/client-generator-core/access/src/Generated/Models/AbstractModel.cs
```diff
@@ -60,7 +60,7 @@ protected AbstractModel(string name)
         }
 
         /// <summary> Initializes a new instance of <see cref="AbstractModel"/>. </summary>
-        /// <param name="kind"> Discriminator. </param>
+        /// <param name="kind"> Discriminator property for AbstractModel. </param>
```
chunyu3, line 63: There are such kind of description changes because TCGC updated the discriminator description in latest version.
Those changes are expected.

samples/AnomalyDetector/src/Generated/Models/ErrorResponse.cs
```diff
@@ -49,7 +49,7 @@ public partial class ErrorResponse
         /// <param name="code"> The error code. </param>
         /// <param name="message"> The message explaining the error reported by the service. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="code"/> or <paramref name="message"/> is null. </exception>
-        public ErrorResponse(string code, string message)
+        internal ErrorResponse(string code, string message)
```
ArcturusZhang, line 52: there are a lot of similar changes like this in the regen. We need to know the exact reason of this change.

ArcturusZhang: Looks like the reason should be usage changed from RoundTrip to Output. but we need to know why because the description is not mentioning a change of usage

samples/AnomalyDetector/src/Generated/tspCodeModel.json
```diff
@@ -1060,8 +988,7 @@
           "Name": "ErrorResponse",
           "Namespace": "AnomalyDetector.Multivariate",
           "Description": "ErrorResponse contains code and message that shows the error information.",
-          "IsNullable": false,
-          "Usage": "RoundTrip",
+          "Usage": "Output",
```
ArcturusZhang, line 991: the usage changed. Why? Is the new usage correct?

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -597,9 +597,9 @@ public static CSharpType FromLiteral(CSharpType type, object literalValue)
         /// <param name="unionItemTypes">The list of union item types.</param>
         /// <param name="isNullable">Flag used to determine if a type is nullable.</param>
         /// <returns>A <see cref="CSharpType"/> instance representing those unioned types.</returns>
-        public static CSharpType FromUnion(IReadOnlyList<CSharpType> unionItemTypes, bool isNullable)
+        public static CSharpType FromUnion(IReadOnlyList<CSharpType> unionItemTypes)
```
ArcturusZhang, line 600: I think this is irrelavent to your changes.
Like we never remove the `IsNullable` in CSharpType, we should remove it here.
But it makes more sense if it is optional here

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -40,55 +41,57 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
-            InputListType { IsEmbeddingsVector: true } listType => new CSharpType(typeof(ReadOnlyMemory<>), listType.IsNullable, CreateType(listType.ElementType)),
-            InputListType listType => new CSharpType(typeof(IList<>), listType.IsNullable, CreateType(listType.ElementType)),
-            InputDictionaryType dictionaryType => new CSharpType(typeof(IDictionary<,>), inputType.IsNullable, typeof(string), CreateType(dictionaryType.ValueType)),
-            InputEnumType enumType => _library.ResolveEnum(enumType).WithNullable(inputType.IsNullable),
+            InputUnionType unionType => CSharpType.FromUnion(unionType.VariantTypes.Select(CreateType).ToArray()),
+            InputListType { IsEmbeddingsVector: true } listType => new CSharpType(typeof(ReadOnlyMemory<>), false, CreateType(listType.ElementType)),
+            InputListType listType => new CSharpType(typeof(IList<>), false, CreateType(listType.ElementType)),
+            InputDictionaryType dictionaryType => new CSharpType(typeof(IDictionary<,>), typeof(string), CreateType(dictionaryType.ValueType)),
+            InputEnumType enumType => _library.ResolveEnum(enumType, false),
```
ArcturusZhang, line 48: why resolveEnum/resolveModel needs an extra parameter of boolean? We should not change it here.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -40,55 +41,57 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
-            InputUnionType unionType => CSharpType.FromUnion(unionType.VariantTypes.Select(CreateType).ToArray(), unionType.IsNullable),
-            InputListType { IsEmbeddingsVector: true } listType => new CSharpType(typeof(ReadOnlyMemory<>), listType.IsNullable, CreateType(listType.ElementType)),
-            InputListType listType => new CSharpType(typeof(IList<>), listType.IsNullable, CreateType(listType.ElementType)),
-            InputDictionaryType dictionaryType => new CSharpType(typeof(IDictionary<,>), inputType.IsNullable, typeof(string), CreateType(dictionaryType.ValueType)),
-            InputEnumType enumType => _library.ResolveEnum(enumType).WithNullable(inputType.IsNullable),
+            InputUnionType unionType => CSharpType.FromUnion(unionType.VariantTypes.Select(CreateType).ToArray()),
+            InputListType { IsEmbeddingsVector: true } listType => new CSharpType(typeof(ReadOnlyMemory<>), false, CreateType(listType.ElementType)),
+            InputListType listType => new CSharpType(typeof(IList<>), false, CreateType(listType.ElementType)),
```
ArcturusZhang, line 46: I think here we could just remove the `false`  to be clear, like what we are doing for dictionary

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -40,55 +41,57 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
+                InputPrimitiveTypeKind.Integer => new CSharpType(typeof(long)), // in typespec, integer is the base type of int related types, see type relation: https://typespec.io/docs/language-basics/type-relations
+                InputPrimitiveTypeKind.Float => new CSharpType(typeof(double)), // in typespec, float is the base type of float32 and float64, see type relation: https://typespec.io/docs/language-basics/type-relations
+                InputPrimitiveTypeKind.Numeric => new CSharpType(typeof(double)), // in typespec, numeric is the base type of number types, see type relation: https://typespec.io/docs/language-basics/type-relations
+                InputPrimitiveTypeKind.IPAddress => new CSharpType(typeof(IPAddress)),
+                InputPrimitiveTypeKind.ArmId => new CSharpType(typeof(ResourceIdentifier)),
+                InputPrimitiveTypeKind.ResourceType => new CSharpType(typeof(ResourceType)),
+                //InputPrimitiveTypeKind.RequestMethod => new CSharpType(typeof(RequestMethod)),
+                //InputPrimitiveTypeKind.ResourceIdentifier => new CSharpType(typeof(ResourceIdentifier)),
```
ArcturusZhang, line 81: we already have RequestMethod below and we should remove.
we should restore resource identifier here.

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
```diff
@@ -1,15 +1,17 @@
...
 
 internal abstract record InputType
 {
-    protected InputType(string name, bool isNullable)
+    protected InputType(string name)
     {
         Name = name;
         SpecName = name;
-        IsNullable = isNullable;
+        //IsNullable = isNullable;
```
ArcturusZhang, line 14: we should clean up

src/AutoRest.CSharp/Common/Output/Models/Types/OutputLibrary.cs
```diff
@@ -14,11 +14,11 @@ internal abstract class OutputLibrary
...
         /// Returns CSharpType that matches the enum definition provided in the input model, or a value type that enum can be converted to if enum wasn't defined in the input
         /// </summary>
-        public abstract CSharpType ResolveEnum(InputEnumType enumType);
+        public abstract CSharpType ResolveEnum(InputEnumType enumType, bool isNullable);
 
         /// <summary>
         /// Returns CSharpType that matches the model definition provided in the input model, or simple object type if model type wasn't defined in the input
         /// </summary>
-        public abstract CSharpType ResolveModel(InputModelType model);
+        public abstract CSharpType ResolveModel(InputModelType model, bool isNullable);
```
ArcturusZhang, line 22: we should not have this change, please revert

samples/AnomalyDetector/src/Generated/tspCodeModel.json
chunyu3: For this, I think it is. `ErrorResponse` only only in response, so it should be `output`

src/AutoRest.CSharp/Common/Input/CodeModelInputs.cs
```diff
@@ -6,5 +6,5 @@
 #pragma warning disable SA1649
 namespace AutoRest.CSharp.Common.Input
 {
-    internal record CodeModelType(Schema Schema, bool IsNullable) : InputType(Schema.Name, IsNullable);
+    internal record CodeModelType(Schema Schema) : InputType(Schema.Name);
```
ArcturusZhang, line 9: I wonder why we still have this type.
Should not we already clean up all the usage of this when we were doing the consolidation for MPG with input types?
@live1206

src/AutoRest.CSharp/Common/Output/Models/CmcRestClientBuilder.cs
```diff
@@ -302,7 +303,8 @@ protected virtual Parameter[] BuildMethodParameters(IReadOnlyDictionary<InputPar
                         // This method has a flattened body
                         if (bodyRequestParameter.Kind == InputOperationParameterKind.Flattened)
                         {
-                            var objectType = (SchemaObjectType)_library.ResolveModel((InputModelType)bodyRequestParameter.Type).Implementation;
+                            (InputType bodyType, bool isNullable) = bodyRequestParameter.Type is InputNullableType nullableType ? (nullableType.Type, true) : (bodyRequestParameter.Type, false);
+                            var objectType = (SchemaObjectType)_library.ResolveModel((InputModelType)bodyType, isNullable).Implementation;
```
ArcturusZhang, line 307: this is really unnecessary.
Our TypeFactory.CreateType could handle this perfectly.
Also since we only need its TypeProvider, and type provider does not have a concept of nullable, we do not need the extra parameter.

The best solution is that i think we should do this:
```csharp
var csharpType = _typeFactory.CreateType(bodyRequestParameter.Type); // this automatically handles nullable for you
Debug.Assert(csharpType is {IsFrameworkType: false, Implementation: SchemaObjectType objectType});
```

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -322,9 +323,10 @@ protected virtual Parameter[] BuildMethodParameters(IReadOnlyDictionary<InputPar
...
                         {
-                            var objectType = bodyRequestParameter.Type switch
+                            (InputType inputType, bool isNullable) = bodyRequestParameter.Type is InputNullableType nullableType ? (nullableType.Type, true) : (bodyRequestParameter.Type, false);
+                            var objectType = inputType switch
                             {
-                                InputModelType inputModelType => library.ResolveModel(inputModelType).Implementation as SerializableObjectType,
+                                InputModelType inputModelType => library.ResolveModel(inputModelType, isNullable).Implementation as SerializableObjectType,
                                 CodeModelType codeModelType => throw new InvalidOperationException("Expecting model"),
                                 _ => null
                             };
```
ArcturusZhang, line 332: So `CodeModelType` should have been removed (Wei is on it now)
Therefore the whole purpose of this piece of code is to assert the created typeprovider is SerializableObjectType, we should do it in this way:
```csharp
var csharpType = _typeFactory.CreateType(bodyRequestParameter.Type);
if (csharpType is not {IsFrameworkType: false, Implementation: SerializableObjectType objectType })
{
    throw new InvalidOperationException("Unexpected flattened type");
}
```

ArcturusZhang: I cannot left the comment on the part below - the above snippet already includes the below throw.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
chunyu3: I changed it to optional parameter

src/AutoRest.CSharp/Common/Input/CodeModelInputs.cs
live1206: This should be removed, updated it in https://github.com/Azure/autorest.csharp/pull/4834

live1206: removed

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -40,55 +41,55 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
-                InputPrimitiveTypeKind.Decimal => new CSharpType(typeof(decimal), inputType.IsNullable),
-                InputPrimitiveTypeKind.Decimal128 => new CSharpType(typeof(decimal), inputType.IsNullable),
-                InputPrimitiveTypeKind.PlainTime => new CSharpType(typeof(TimeSpan), inputType.IsNullable),
-                InputPrimitiveTypeKind.ETag => new CSharpType(typeof(ETag), inputType.IsNullable),
-                InputPrimitiveTypeKind.Float32 => new CSharpType(typeof(float), inputType.IsNullable),
-                InputPrimitiveTypeKind.Float64 => new CSharpType(typeof(double), inputType.IsNullable),
-                InputPrimitiveTypeKind.Float128 => new CSharpType(typeof(decimal), inputType.IsNullable),
-                InputPrimitiveTypeKind.Guid or InputPrimitiveTypeKind.Uuid => new CSharpType(typeof(Guid), inputType.IsNullable),
-                InputPrimitiveTypeKind.Int8 => new CSharpType(typeof(sbyte), inputType.IsNullable),
-                InputPrimitiveTypeKind.UInt8 => new CSharpType(typeof(byte), inputType.IsNullable),
```
ArcturusZhang, line 69: we removed this type. Is this expected?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs
```diff
@@ -46,6 +47,7 @@ private InputType CreateObject(ref Utf8JsonReader reader, JsonSerializerOptions
                 result = CreateDerivedType(ref reader, id, kind, name, options);
             }
 
+            result = result ?? CreateDerivedType(ref reader, id, kind, name, options);
```
ArcturusZhang, line 50: why do we have to add this line?

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -660,31 +659,38 @@ private InputType CreateTypeForObjectSchema(ObjectSchema objectSchema)
                 XMsFormat.DataFactoryElementOfInt => CreateDataFactoryElementIntputType(isNullable, InputPrimitiveType.Int32),
                 XMsFormat.DataFactoryElementOfDouble => CreateDataFactoryElementIntputType(isNullable, InputPrimitiveType.Float64),
```
ArcturusZhang, line 660: find some typos here, could we also fix them here?

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -660,31 +659,38 @@ private InputType CreateTypeForObjectSchema(ObjectSchema objectSchema)
...
+                XMsFormat.DataFactoryElementOfKeyValuePairs => CreateDataFactoryElementIntputType(isNullable, new InputDictionaryType(name, InputPrimitiveType.String, InputPrimitiveType.String)),
+                XMsFormat.DataFactoryElementOfDateTime => CreateDataFactoryElementIntputType(isNullable, new InputDateTimeType(DateTimeKnownEncoding.Rfc3339, InputPrimitiveType.String)),
                 XMsFormat.DataFactoryElementOfDuration => CreateDataFactoryElementIntputType(isNullable, InputPrimitiveType.PlainTime),
                 XMsFormat.DataFactoryElementOfUri => CreateDataFactoryElementIntputType(isNullable, InputPrimitiveType.Uri),
-                XMsFormat.DataFactoryElementOfKeyObjectValuePairs => CreateDataFactoryElementIntputType(isNullable, new InputDictionaryType(name, InputPrimitiveType.String, InputPrimitiveType.Any, false)),
+                XMsFormat.DataFactoryElementOfKeyObjectValuePairs => CreateDataFactoryElementIntputType(isNullable, new InputDictionaryType(name, InputPrimitiveType.String, InputPrimitiveType.Any)),
```
ArcturusZhang, line 668: well all of them have typos `Intput -> Input`

src/AutoRest.CSharp/Common/Input/CodeModelInputs.cs
ArcturusZhang: Please rmeove this type. This has been removed in a previous PR, and it should not be here.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -287,7 +292,7 @@ static bool IsSerializable(ObjectTypeProperty objectProperty, TypeFactory typeFa
                 name,
                 memberValueExpression,
                 serializedName,
-                serializedType,
+                propertyType.WithNullable(propertyType.IsNullable ? serializedType.IsNullable: false),
```
ArcturusZhang, line 295: need to format code
```suggestion
                propertyType.WithNullable(propertyType.IsNullable ? serializedType.IsNullable : false),
```

src/AutoRest.CSharp/Common/Output/Models/Shared/Parameter.cs
```diff
@@ -64,9 +65,16 @@ public static Parameter FromInputParameter(in InputParameter operationParameter,
                 : ValidationType.None;
 
             var inputType = type.InputType;
+            IEnumerable<string>? values = operationParameter.Type switch
+            {
+                InputEnumType enumType => enumType.Values.Select(c => c.GetValueString()),
+                InputNullableType { Type: InputEnumType e} => e.Values.Select(c => c.GetValueString()),
```
ArcturusZhang, line 71: please format the file

src/AutoRest.CSharp/Common/Input/InputTypes/InputNullableType.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AutoRest.CSharp.Common.Input.InputTypes
+{
+    internal record InputNullableType(InputType Type): InputType("nullable")
```
ArcturusZhang, line 6: let us make this sealed because we have so many places asserting `if (type is InputNullableType)`

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
```diff
@@ -322,9 +323,10 @@ protected virtual Parameter[] BuildMethodParameters(IReadOnlyDictionary<InputPar
                         // This method has a flattened body
                         if (bodyRequestParameter.Kind == InputOperationParameterKind.Flattened && library != null)
                         {
-                            var objectType = bodyRequestParameter.Type switch
+                            (InputType inputType, bool isNullable) = bodyRequestParameter.Type is InputNullableType nullableType ? (nullableType.Type, true) : (bodyRequestParameter.Type, false);
+                            var objectType = inputType switch
                             {
-                                InputModelType inputModelType => library.ResolveModel(inputModelType).Implementation as SerializableObjectType,
+                                InputModelType inputModelType => library.ResolveModel(inputModelType).WithNullable(isNullable).Implementation as SerializableObjectType,
```
ArcturusZhang, line 329: WithNullable here is actually redundant.
And if we care, we should do it through `TypeFactory.CreateType` which handles nullability automatically, you do not even need to unwrap the nullable type like we are doing above.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -139,15 +140,15 @@ private IEnumerable<InputType> UpdateBodyParameters()
                         if (parameter.Location != RequestLocation.Body)
                             continue;
 
-                        IncrementCount(usageCounts, parameter.Type.Name);
+                        IncrementCount(usageCounts, parameter.Type.GetImplementType().Name);
```
ArcturusZhang, line 143: OK so if we have this method, we have so many other places doing
```csharp
type = type is InputNullableType ? type.Type : type;
```
all of them should be changed to use this `GetImplementType` method.

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
chunyu3: @ArcturusZhang  it is not removed, it is edited.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs
chunyu3: some InputType only have `id, kind, name`, so previous `CreateDerivedType` is not called. So need to call it again after while read.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
chunyu3: No. This is static function, there is no `TypeFactory` instance. so we cannot use `TypeFactory.CreateType`. We may need to refactor this in future. maybe an static `TypeFactory` variable is needed.

Current we will keep this.

live1206: TypeFactory is available from the public static method, we can pass it along to fix this

src/AutoRest.CSharp/Mgmt/Models/MgmtRestOperation.cs
```diff
@@ -185,10 +186,10 @@ public MgmtRestOperation(MgmtRestOperation other, string nameOverride, CSharpTyp
             var finalSchema = Operation.LongRunning?.FinalResponse.BodyType;
             if (finalSchema is null)
                 return null;
-
+            (InputType finalType, bool isNullable) = finalSchema is InputNullableType nullableType ? (nullableType.Type, true) : (finalSchema, false);
             try
             {
-                return finalSchema is InputModelType inputModel ? MgmtContext.Library.ResolveModel(inputModel) : MgmtContext.TypeFactory.CreateType(finalSchema);
+                return finalType is InputModelType inputModel ? MgmtContext.Library.ResolveModel(inputModel).WithNullable(isNullable) : MgmtContext.TypeFactory.CreateType(finalSchema);
```
live1206, line 192: can we just return `MgmtContext.TypeFactory.CreateType(finalSchema)`?

src/TypeSpec.Extension/Emitter.Csharp/src/lib/converter.ts
```diff
@@ -277,24 +285,25 @@ export function fromSdkEnumType(
...
             Description: enumType.description,
             IsExtensible: enumType.isFixed ? false : true,
-            IsNullable: enumType.nullable,
             Usage: fromUsageFlags(enumType.usage)
         };
         if (addToCollection) enums.set(enumName, newInputEnumType);
         inputEnumType = newInputEnumType;
     }
-    inputEnumType.IsNullable = enumType.nullable; // TO-DO: https://github.com/Azure/autorest.csharp/issues/4314
+    //inputEnumType.IsNullable = false; // TO-DO: https://github.com/Azure/autorest.csharp/issues/4314
```
live1206, line 300: should this be removed?

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
ArcturusZhang: It could use library because we pass the library along.
Here we do not need the library, instead we should change it to TypeFactory.

src/AutoRest.CSharp/Mgmt/Models/MgmtRestOperation.cs
ArcturusZhang: yeah we should not do the same thing in multiple places.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs
ArcturusZhang: Well this is only explaining what is happening now why.
But I get it now.

- Previously all the types at least have another property `IsNullable` other than kind, id and name, now `IsNullable` is removed therefore now we have this case, therefore we have to fix it in this way.

ArcturusZhang: but this fix is quite redundant.
`CreateDerivedType` method never returns null, therefore we do not need to have the below throw, it now never happens.
We could change it in this way:
```suggestion
            return result ?? CreateDerivedType(ref reader, id, kind, name, options);
```
and we could remove the below line.

src/AutoRest.CSharp/Common/Generation/Types/CSharpType.cs
```diff
@@ -597,7 +597,7 @@ public static CSharpType FromLiteral(CSharpType type, object literalValue)
         /// <param name="unionItemTypes">The list of union item types.</param>
         /// <param name="isNullable">Flag used to determine if a type is nullable.</param>
         /// <returns>A <see cref="CSharpType"/> instance representing those unioned types.</returns>
-        public static CSharpType FromUnion(IReadOnlyList<CSharpType> unionItemTypes, bool isNullable)
+        public static CSharpType FromUnion(IReadOnlyList<CSharpType> unionItemTypes, bool isNullable = false)
```
live1206, line 600: it seems like we should remove `isNullable` from public constructor, we never set the nullability from public users

ArcturusZhang: well the idea of keeping this is for future usage to be parallel with the ctor of this type

samples/AnomalyDetector/src/Generated/AnomalyDetectorModelFactory.cs
```diff
@@ -233,6 +233,15 @@ public static MultivariateBatchDetectionResultSummary MultivariateBatchDetection
             return new MultivariateBatchDetectionResultSummary(status, errors?.ToList(), variableStates?.ToList(), setupInfo, serializedAdditionalRawData: null);
         }
 
+        /// <summary> Initializes a new instance of <see cref="Models.ErrorResponse"/>. </summary>
+        /// <param name="code"> The error code. </param>
+        /// <param name="message"> The message explaining the error reported by the service. </param>
+        /// <returns> A new <see cref="Models.ErrorResponse"/> instance for mocking. </returns>
+        public static ErrorResponse ErrorResponse(string code = null, string message = null)
```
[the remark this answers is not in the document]
ArcturusZhang: this shows up because of our rules on "who would have an entry in model factory"
the changes on `ErrorResponse` is that its usage changed from roundtrip to output, therefore the accessibility of the class did not change, but only its public ctor changed to internal and all its properties no longer have a setter now.

The rules of model factory is that when a model has a property without setter, it will show up in the model factory.
Previously every property in ResponseError is settable therefore it was not there, now it changed to output and all its properties do not have a setter now, therefore it shows up

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
ArcturusZhang: So if we do not plan to do the typefactory thing (despite I think it just harms nothing and make things clear), please remove the useless `WithNullable` here.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputTypeConverter.cs
chunyu3: Thanks, that make sense. removed the redundant line.

src/AutoRest.CSharp/Common/Output/Models/RestClientBuilder.cs
ArcturusZhang: we could fix this later

live1206 (approved):

PR #4791, opened 2024-06-03
Support TSP input for MPG

live1206:
Resolves https://github.com/Azure/autorest.csharp/issues/4416


src/AutoRest.CSharp/Mgmt/Output/ResourceCollection.cs
```diff
@@ -138,8 +138,6 @@ protected override IEnumerable<ContextualParameterMapping> EnsureExtraContextual
             var suppressListException = Configuration.MgmtConfiguration.ListException.Contains(RequestPath)
                 || Configuration.MgmtConfiguration.MgmtDebug.SuppressListException;
             var getAllOperation = ClientOperations.Where(operation => operation.Name == "GetAll").OrderBy(operation => ReferenceSegments(operation).Count()).FirstOrDefault();
-            if (!suppressListException && getAllOperation == null)
```
live1206, line 141: There wasn't much value for this check since the service team always add a suppression for it.
So, we decided to remove it.

live1206: keep this with TODO

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
```diff
@@ -48,6 +48,21 @@ public static async Task RunAsync(CommandLineOptions options)
             {
                 var json = await File.ReadAllTextAsync(tspInputFile);
                 var rootNamespace = TypeSpecSerialization.Deserialize(json) ?? throw new InvalidOperationException($"Deserializing {tspInputFile} has failed.");
+
+                // Update usage for Mgmt
+                // TODO: Consolidate InputTypeUsageProvider and SchemaUsageProvider
```
live1206, line 53: look into TCGC to resolve this there

src/AutoRest.CSharp/Mgmt/Output/ResourceCollection.cs
live1206: @m-nash Had another talk with Shanghai team, we still think it's not worth to keep this check and add suppression every time it breaks, it causes much more support effort than what gains. So, we would like to remove it for good.

src/AutoRest.CSharp/Common/AutoRest/Communication/StandaloneGeneratorRunner.cs
live1206: This has been addressed in https://github.com/Azure/typespec-azure/issues/916, waiting for a new release of TCGC.

live1206: I will revert the related to fix the usage in this PR, and updating TCGC will fix the usage.
Don't think this PR merge is pending on the TCGC upgrading since we are not officially supporting TSP input for MPG.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/client-model-builder.ts
```diff
@@ -230,16 +238,28 @@ export function createModelForService(
...
+        clients: InputClient[],
+        isAzureArm: boolean | undefined
     ) {
         const dpgOperationGroups = listOperationGroups(
             sdkContext,
             client as SdkClient
         );
         for (const dpgGroup of dpgOperationGroups) {
-            const subClient = emitClient(dpgGroup, client);
+            // For MPG, we skip the "Operations" operation group
```
live1206, line 249: Other languages don't exclude these operation groups, such as java: [Operations](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/deviceregistry/azure-resourcemanager-deviceregistry/src/main/java/com/azure/resourcemanager/deviceregistry/implementation/OperationsClientImpl.java) and [OperationStatusClientImpl](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/deviceregistry/azure-resourcemanager-deviceregistry/src/main/java/com/azure/resourcemanager/deviceregistry/implementation/OperationStatusClientImpl.java)
So, this is specific for .NET.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/client-model-builder.ts
```diff
@@ -270,9 +290,14 @@ export function createModelForService(
 
     function emitClient(
         client: SdkClient | SdkOperationGroup,
+        isAzureArm: boolean | undefined,
         parent?: SdkClient | SdkOperationGroup
-    ): InputClient {
+    ): InputClient | undefined {
         const operations = listOperationsInOperationGroup(sdkContext, client);
+        // For ARM, we skip the operation group without operations
```
live1206, line 297: This behavior is only needed for Mgmt, not DPG.

src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -18,30 +18,48 @@ namespace AutoRest.CSharp.Mgmt.Output
...
+            if (InputModel.IsUnknownDiscriminatorModel)
+            {
+                return InputModel.Properties.Select(CreateProperty).ToArray();
+            }
+
+            var result = new List<ObjectTypeProperty>();
+
+            // For TypeSpec input, we need to add all the properties from the base types
+            // For swagger input, the base type properties are already included in the properties list, we will just skip them
+            var propertyNames = InputModel.Properties.Select(p => p.Name).ToHashSet();
```
live1206, line 47: we should be able to remove this when https://github.com/Azure/autorest.csharp/pull/4797 is merged

package.json
```diff
@@ -2,12 +2,14 @@
...
-    "@azure-tools/typespec-client-generator-core": "0.42.3",
-    "@azure-tools/typespec-csharp": "file:src/TypeSpec.Extension/Emitter.Csharp",
-    "@azure-tools/unbranded-tests": "file:test/UnbrandedProjects",
-    "@microsoft.azure/autorest.testserver": "3.3.48",
-    "autorest": "3.7.1"
-  }
+    "devDependencies": {
+        "@azure-tools/cadl-ranch-mockapis": "file:test/CadlRanchMockApis",
+        "@azure-tools/typespec-azure-rulesets": "0.42.1",
+        "@azure-tools/typespec-azure-resource-manager": "^0.42.1",
```
pshao25, line 8: It should be a product dependency instead of devDependency.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/client-model-builder.ts
```diff
@@ -157,8 +157,16 @@ export function createModelForService(
     const clients: InputClient[] = [];
     const dpgClients = listClients(sdkContext);
     for (const client of dpgClients) {
-        clients.push(emitClient(client));
-        addChildClients(sdkContext.emitContext, client, clients);
+        const emittedClient = emitClient(client, sdkContext.arm);
```
pshao25, line 160: I assume it should not be in @typespec/http-client-csharp? How you are going to keep it after https://github.com/Azure/autorest.csharp/pull/4774?

ArthurMa1978 (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/client-model-builder.ts
live1206: Please check the latest commit.

live1206: Reverted the change in emitter, move the logic to generator instead.

package.json
live1206: reverted since they are not needed

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -926,8 +929,19 @@ private TypeProvider BuildResourceData(InputType inputType, MgmtObjectType? defa
         private Dictionary<string, OperationSet> CategorizeOperationGroups()
         {
             var rawRequestPathToOperationSets = new Dictionary<string, OperationSet>();
-            foreach (var inputClient in _input.Clients)
+            foreach (var inputClient in _inputClients)
             {
+                // If the client has no operations, we skip it
+                if (inputClient.Operations.Count == 0)
```
pshao25, line 935: Isn't `_inputClient` already filtered out the client without operations?

pshao25 (approved):

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
```diff
@@ -91,7 +92,17 @@ public async Task<GeneratedCodeWorkspace> ExecuteAsync(InputNamespace rootNamesp
             Directory.CreateDirectory(Configuration.OutputFolder);
             var project = await GeneratedCodeWorkspace.Create(Configuration.AbsoluteProjectFolder, Configuration.OutputFolder, Configuration.SharedSourceFolders);
             var sourceInputModel = new SourceInputModel(await project.GetCompilationAsync(), await ProtocolCompilationInput.TryCreate());
-            await LowLevelTarget.ExecuteAsync(project, rootNamespace, sourceInputModel, true);
+
+            if (Configuration.AzureArm)
+            {
+                InputTypeTransformer.Transform(rootNamespace);
```
ArcturusZhang, line 98: please add a TODO item here because in the future we should never need to transform anything
Everything we need to change should happen on the tsp instead of being done in the generator.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
```diff
@@ -926,8 +929,19 @@ private TypeProvider BuildResourceData(InputType inputType, MgmtObjectType? defa
...
+                {
+                    continue;
+                }
+
+                // Skip "Operations" operation group in the Operations client
+                if (inputClient.Name == "Operations")
+                {
+                    continue;
+                }
```
ArcturusZhang, line 944: here we need a TODO item as well because this is definitely not a smart way to remove the `Operations_List` method.

If possible I think we should remove this here, and add the internal or remove to those operations in tsp.
For swagger input, we should not include them, ie, remove them from the schema first, then convert, instead of convert then remove.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
live1206: yes, forgot to remove these

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
live1206: That has been done, this should be removed.

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
live1206: I am not 100% sure if we never need to transform anything.
This is a workaround to handle internal operations for now.
There might be cases in CodeModelTransformer, we need to handle for TSP as well, just we have not met yet.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
live1206: This has been removed.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtOutputLibrary.cs
live1206: removed

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
live1206: never mind, moved the logic to `MgmtOutputLibrary`, and we can add `InputTypeTransformer` until we need it.

live1206: Ok. We still need the transformer since we are manipulating the operations in client.

live1206:
Regen is https://github.com/Azure/azure-sdk-for-net/pull/44378, lost the connection due to force push.

src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs
live1206: Added TODO to remove this once we have a better way to remove operation, and we can discuss later whether we need the transformer.

ArcturusZhang (approved):

PR #4808, opened 2024-06-11
Skip "Type" from output properties of XXXResourceData

live1206:
For `XXXResourceData`, property `type` is same as `ResourceData.ResourceType` in inherits, we should skip it.

### DIFF
--- a/src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
+++ b/src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
@@ -56,8 +56,7 @@ namespace AutoRest.CSharp.Mgmt.Output
         protected override IEnumerable<ObjectTypeProperty> BuildProperties()
         {
             var parentProperties = GetParentPropertyNames();
-            var properties = base.BuildProperties().ToArray();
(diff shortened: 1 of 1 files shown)


src/AutoRest.CSharp/Mgmt/Output/MgmtObjectType.cs
```diff
@@ -49,8 +49,14 @@ private HashSet<string> GetParentPropertyNames()
...
                 .Skip(1)
                 .SelectMany(type => type.Properties)
-                .Select(p => p.Declaration.Name)
+                .Select(p => isResourceType(p) ? "Type" : p.Declaration.Name)
                 .ToHashSet();
+
+            // In common-types, "type" is a property to represent resource type
+            // In ResourceManager.ResourceData, this property is named as "ResourceType"
+            // Return "Type" as property name for it to skip it in the output properties
+            bool isResourceType(ObjectTypeProperty property)
```
pshao25, line 58: Seems more appropriate to call it `RenameResourceType` and put all the rename logic into it?

pshao25 (approved):
Let's do a regen before merge.

PR #4809, opened 2024-06-11
Support void type body

pshao25:
We cannot support such tsp whose body is a `void`
```
register is ArmResourceActionSync<ExtendedZone, void, ExtendedZone>;
```

### DIFF
--- a/src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
+++ b/src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
@@ -104,14 +104,14 @@ export function loadOperation(
 
     if (
         typespecParameters.body?.parameter &&
-        !isVoidBody(typespecParameters.body.type)
+        !isVoidType(typespecParameters.body.type)
     ) {
         parameters.push(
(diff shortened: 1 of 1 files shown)


src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -101,11 +102,17 @@ export function loadOperation(
...
+        typespecParameters.body?.parameter &&
+        !isVoidBody(typespecParameters.body.type)
+    ) {
         parameters.push(
             loadBodyParameter(sdkContext, typespecParameters.body?.parameter)
         );
-    } else if (typespecParameters.body?.type) {
+    } else if (
+        typespecParameters.body?.type &&
+        !isVoidBody(typespecParameters.body.type)
```
live1206, line 114: will this affect DPG?

pshao25: Yes, I think for DPG, a void body means no body too.

live1206 (approved):

PR #4814, opened 2024-06-11
Change no-route severity

JoshLove-msft:


src/TypeSpec.Extension/Emitter.Csharp/src/lib/lib.ts
```diff
@@ -15,7 +15,7 @@ const $lib = createTypeSpecLibrary({
             }
         },
         "no-route": {
-            severity: "error",
```
m-nash, line 18: I don't think dropping the severity here is the right thing to do.  This simply masks the problem which is that we are getting an empty service with no routes for a library that is supposed to be model only with no services.

JoshLove-msft: Added a TODO to revert this

m-nash: https://github.com/Azure/typespec-azure/issues/1018

m-nash (approved):

PR #4824, opened 2024-06-14
Load "x-ms-skip-url-encoding" from extension and fix type property skip

live1206:
# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
+++ b/src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
@@ -474,6 +474,13 @@ namespace AutoRest.CSharp.Output.Models.Types
                         continue;
                     }
                     var prop = CreateProperty(property);
+
+                    // If "Type" property is "String" and "ResourceType" exists in parents properties, skip it since they are the same.
+                    // This only applies for TypeSpec input, for swagger input we have renamed "type" property to "resourceType" in FrameworkTypeUpdater.ValidateAndUpdate
+                    if (property.CSharpName() == "Type" && prop.ValueType.Name == "String" && existingProperties.Contains("ReosurceType"))
+                    {
+                        continue;
+                    }
                     propertiesFromSpec.Add(prop.Declaration.Name);
                     yield return prop;
                 }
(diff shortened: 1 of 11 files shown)


ArthurMa1978 (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -306,7 +307,10 @@ export function loadOperation(
             IsResourceParameter: false,
             IsContentType: isContentType,
             IsEndpoint: false,
-            SkipUrlEncoding: false, //TODO: retrieve out value from extension
+            SkipUrlEncoding:
+                // TODO: update this when https://github.com/Azure/typespec-azure/issues/1022 is resolved
```
live1206, line 311: Update after https://github.com/Azure/typespec-azure/issues/1022 resolved

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -474,6 +474,13 @@ protected override IEnumerable<ObjectTypeProperty> BuildProperties()
                         continue;
                     }
                     var prop = CreateProperty(property);
+
+                    // If "Type" property is "String" and "ResourceType" exists in parents properties, skip it since they are the same.
+                    // This only applies for TypeSpec input, for swagger input we have renamed "type" property to "resourceType" in FrameworkTypeUpdater.ValidateAndUpdate
+                    if (property.CSharpName() == "Type" && prop.ValueType.Name == "String" && existingProperties.Contains("ResourceType"))
+                    {
+                        continue;
+                    }
```
ArcturusZhang, line 483: OK this is really like a workaround.
adding a client.tsp in the resourcemanager to change the client to ResourceType in all trackedresource and resource may be a better idea

live1206: Ideally, we should rename this for csharp for common-types, but for now we don't have an entry for common-types.
We will need to have an entry to be able to generate `Azure.ResourceManager` from common-types typespec.

live1206:
> Looks we only implements this workaround in dotnet for 'x-ms-skip-url-encoding', it would cause inconsistency around SDKs for the behavior of this extension. Considering there is no service impact by this now, shall we rollback this change and wait for the decision for [Azure/typespec-azure#1022](https://github.com/Azure/typespec-azure/issues/1022)? cc @m-nash @live1206

`Scvmm` is affected by this, .net generator failed without this fix, not like other languages could generate SDK with this part missing.
This time, we will release GA version for `Scvmm` so it generates from swagger, but I would prefer to keep the fix until we have a proper solution. And, we will talk about this in the coming meeting tomorrow.

PR #4828, opened 2024-06-18
refactor(generator): remove null check on client description

archerzz:
This is a prerequisite change for #3994 The null check for client description is meaningless, because `ClientBuilder` will add a default description anyway: https://github.com/Azure/autorest.csharp/blob/4d55b600ee877e68675297546e70a765fb679039/src/AutoRest.CSharp/Common/Output/Builders/ClientBuilder.cs#L28-L31 Right now in our emitter, we will set `description` to an empty string: https://github.com/Azure/autorest.csharp/blob/4d55b600ee877e68675297546e70a765fb679039/src/TypeSpec.Extension/Emitter.Csharp/src/lib/client-model-builder.ts#L281 But in TCGC, the `description` will be null.

So, let's remove this meaningless check.

# Description

# Checklist

# Ready to Land?


archerzz:
By the way, it is possible to have a client without description. See https://github.com/Azure/autorest.csharp/blob/4d55b600ee877e68675297546e70a765fb679039/test/TestProjects/MixAPIVersion-TypeSpec/MixAPIVersion-TypeSpec.tsp#L94 a namespace can have no `@summary`

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputClientConverter.cs
```diff
@@ -49,7 +49,7 @@ public override void Write(Utf8JsonWriter writer, InputClient value, JsonSeriali
             }
 
             name= name ?? throw new JsonException("InputClient must have name");
-            description = description ?? throw new JsonException("InputClient must have description");
+            description = description ?? string.Empty;
```
ArcturusZhang, line 52: a client does not have to have a description, therefore we should define the description as nullable then here just remove this line.

archerzz: Changing `string Description` to `string? Description` will cause a lot of chain reaction. Better to stick to a empty default value for now.

ArcturusZhang (approved):

PR #4838, opened 2024-06-19
Add default acronym mapping

live1206:
- Add default acronym mapping, autorest.md can overwrite them
- Add `InputAcronymTransformer.UpdateAcronyms`
- Add `InputUrlToUriTransformer.UpdateSuffix`
- Add `InputRenameTimeToOnTransformer.Update` to `InputNamespaceTransformer`


src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputRenameTimeToOnTransformer.cs
```diff
@@ -0,0 +1,68 @@
...
+                        {
+                            lengthToCut = 2;
+                        }
+
+                        if (lengthToCut > 0)
+                        {
+                            var prefix = propName.Substring(0, propName.Length - lengthToCut);
+                            var newName = (_nounToVerbDicts.TryGetValue(prefix, out var verb) ? verb : prefix) + "On";
+                            property.Name = newName;
+                        }
```
ArcturusZhang, line 62: since we are previously doing this in this way, this is not a comment that we have to change it right now.
But we have issues on the current implementation.
For example, we are checking `StartsWith("To", `, what if there was a word starting with `To` but it is actually not the word `To`? To be precise, this should be implemented in a way of "words" instead of just checking if the whole string starts with what or ends with what.

We might need to fix this when we migrate this logic to MGC.

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
```diff
@@ -108,10 +108,13 @@ public static string ToMgmtVariableName(this string name)
                 if (tempName != applyStep.NewName.VariableName)
                 {
                     var finalName = ToCleanName(applyStep.NewName.VariableName, isCamelCase: false);
-                    MgmtReport.Instance.TransformSection.AddTransformLogForApplyChange(
-                        TransformTypeName.AcronymMapping, applyStep.MappingKey, applyStep.MappingValue.RawValue,
-                        $"Variables.{name}",
-                        $"ApplyAcronymMappingOnVariable", tempName, $"{applyStep.NewName.VariableName}(ToCleanName={finalName})");
+                    if (applyStep.MappingValue.RawValue is not null)
```
ArcturusZhang, line 111: so this change only applies to the `ToMgmtVariableName` method.
Should we just remove the other `ToVariableName` method, and let this method to take it place since all the targets are having the acronym mapping?

ArcturusZhang (commented):
A few comments.
After this is merged, could you please also migrate this change (or the needed part of this change) to MGC? I believe they need to have this change as well

live1206:
> A few comments. After this is merged, could you please also migrate this change (or the needed part of this change) to MGC? I believe they need to have this change as well

Actually, I am not sure about the plan for MGC, because if we are planning to have naming checks for models and properties in typespec CI, and let the user to take over the renaming. We don't need these transformers for typespec input anymore. I think that's better and doable, but it takes time to get all rules implemented.

ArcturusZhang:
> > A few comments. After this is merged, could you please also migrate this change (or the needed part of this change) to MGC? I believe they need to have this change as well
> 
> Actually, I am not sure about the plan for MGC, because if we are planning to have naming checks for models and properties in typespec CI, and let the user to take over the renaming. We don't need these transformers for typespec input anymore. I think that's better and doable, but it takes time to get all rules implemented.

I am also not sure - if you have any thoughts, let us put them in this issue: https://github.com/microsoft/typespec/issues/3707

src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputRenameTimeToOnTransformer.cs
ArcturusZhang: Also we could also consider to change the implementation of these to using spans, which should improves the performance.

src/AutoRest.CSharp/Common/Utilities/StringExtensions.cs
live1206: `ToMgmtVariableName` is called via `ToVariableName`, makeing `ToMgmtVariableName` private instead.

src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputUrlToUriTransformer.cs
```diff
@@ -0,0 +1,32 @@
...
+
+                var schemaName = model.Name;
+                if (schemaName.EndsWith("Url", StringComparison.Ordinal))
+                    model.Name = schemaName.Substring(0, schemaName.Length - 1) + LowerCaseI;
+
+                foreach (var property in inputModel.Properties)
+                {
+                    var propertyName = property.Name;
+                    if (propertyName.EndsWith("Url", StringComparison.Ordinal))
+                        property.Name = propertyName.Substring(0, propertyName.Length - 1) + LowerCaseI;
```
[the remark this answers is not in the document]
live1206: updated

src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputRenameTimeToOnTransformer.cs
live1206: updated

src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputUrlToUriTransformer.cs
```diff
@@ -0,0 +1,32 @@
...
+                var schemaName = model.Name.AsSpan();
+                if (schemaName.EndsWith("Url", StringComparison.Ordinal))
+                    model.Name = schemaName.Slice(0, schemaName.Length - 1).ToString() + LowerCaseI;
+
+                foreach (var property in inputModel.Properties)
+                {
+                    var propertyName = property.Name.AsSpan();
+                    if (propertyName.EndsWith("Url", StringComparison.Ordinal))
+                        property.Name = propertyName.Slice(0, propertyName.Length - 1).ToString() + LowerCaseI;
+                }
```
ArcturusZhang, line 28: not only the `Substring` part has a performance issue, the `+` operator also has performance issue.
Here is a quick benchmark I did for the three approaches:

| Method              | Mean     | Error    | StdDev   | Median   |
|-------------------- |---------:|---------:|---------:|---------:|
| UseSubStringAndPlus | 25.24 ns | 1.929 ns | 5.689 ns | 25.90 ns |
| UseSpanAndPlus      | 22.87 ns | 1.389 ns | 3.962 ns | 21.14 ns |
| UseSpanOnly         | 17.44 ns | 0.410 ns | 0.519 ns | 17.33 ns |

Code:
```csharp

        [Benchmark]
        public string UseSubStringAndPlus()
        {
            return _stringToReplace.Substring(0, _stringToReplace.Length - 1) + _charToReplace;
        }

        [Benchmark]
        public string UseSpanAndPlus()
        {
            var span = _stringToReplace.AsSpan();
            return span.Slice(0, span.Length - 1).ToString() + _charToReplace;
        }

        [Benchmark]
        public string UseSpanOnly()
        {
            Span<char> span = _stringToReplace.ToCharArray().AsSpan();
            span[^1] = _charToReplace;
            return span.ToString();
        }
```

ArcturusZhang: I am not sure if I did it correctly because I saw no "allocation" column. Maybe because there is no allocation because all the arrays we constructed have been properly used in those strings.

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputUrlToUriTransformer.cs
live1206: For a single char replacement, I think it works great as you showed.
But still, we need to create another `Span<char>` along with `ReadonlySpan<char>`, which is needed for https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.endswith?view=net-8.0.

But for multiple-char appending, I doubt it gets better and it gets more complex and error-prune.
Therefore, I only updated it in `InputUrlToUriTransformer.cs`.

Issue #4833, opened 2024-06-18
Generated extensible enums use case-insensitive equality but case-sensitive hash codes


trrwilson:
As raised in the OpenAI .NET project: https://github.com/openai/openai-dotnet/issues/49

Extensible enums use `_value.GetHashCode()` for their `GetHashCode()` implementation while meanwhile using a case-insensitive `_value` comparison for `Equals()`. As @stephentoub points in the above issue, this is inconsistent with the conventions for `GetHashCode()` -- if two objects evaluate as equal, then they should have the same hash code.

Here's a minimal repro project to demonstrate; `npm install`, `npx tsp compile`, then `dotnet run` under "test" will demonstrate a manifestation of the problem via an unexpected `HashSet` lookup failure:

```
Hello, World!
NormalFoo == CapitalFoo     ? True
HashSet.Contains NormalFoo  ? True
HashSet.Contains CapitalFoo ? False
```

[eeHashTest.zip](https://github.com/user-attachments/files/15892873/eeHashTest.zip)

The fix may just be as simple as converting the `_value` into a matching lower invariant before calling `GetHashCode()` on it (though this could be further optimized with caching/etc.).

ArcturusZhang:
Hi @trrwilson thanks for reporting this.
The fix is ready in this PR: https://github.com/Azure/autorest.csharp/pull/4842
netstandard2.0 provides this handy API to let us get hashcode case-insensitively: https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparer.gethashcode?view=netstandard-2.0#system-stringcomparer-gethashcode(system-string)

ArcturusZhang:
Hi @trrwilson this should be resolved now in the latest version of the generator

PR #4842, opened 2024-06-20
fix for `GetHashCode` on extensible strings

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4833

# Description

Common convention for `GetHashCode` and `Equals`:
1. when `Equals` was overridden, `GetHashCode` must be overridden as well, and vice versa.
2. when `Equals` gives true, `GetHashCode` must give the same value on two instances, but when `GetHashCode` gives the same value on two instances, `Equals` does not have to return `true` (hashcode collision).

Our previous implementation compares string case insensitively, but returning hash code from the underlying string directly, which is violating rule no. 2.

# Checklist

# Ready to Land?


samples/AnomalyDetector/src/Generated/Models/FillNAMethod.cs
```diff
@@ -53,7 +53,7 @@ public FillNAMethod(string value)
 
         /// <inheritdoc />
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public override int GetHashCode() => _value?.GetHashCode() ?? 0;
+        public override int GetHashCode() => StringComparer.InvariantCultureIgnoreCase.GetHashCode(_value);
```
lirenhe, line 56: Shall we still do the 'null' check for _value?  
As this is a struct, people could create a new instance without calling the constructor.

ArcturusZhang: this method handles null values properly.

ArcturusZhang: well I thought this API handles null value properly but it turns out that it does not, it just simply throw when there is a null value.
I will update the code and add a test case

archerzz (approved):

lirenhe (approved):

PR #4858, opened 2024-06-21
Add pipeline for test typespec next

pshao25:
See this pipeline https://dev.azure.com/azure-sdk/internal/_build?definitionId=7035&_a=summary


eng/pipelines/preview-typespec.yml
```diff
@@ -0,0 +1,16 @@
...
+  default: true
+- name: RunGenerationChecks
+  displayName: Run code generation checks
+  type: boolean
+  default: true
+
+extends:
+- template: /eng/pipelines/preview.yml
+  parameters:
+    UseTypeSpecNext: true
```
ArcturusZhang, line 14: i wonder how this works - this uses the next version of typespec and overrides the version used in TCGC?

pshao25: You should review the latest commit.

But yes to your question, it will override all the versions in our package.json.

eng/pipelines/preview-typespec.yml
```diff
@@ -0,0 +1,83 @@
...
+
+extends:
+  template: eng/pipelines/templates/stages/archetype-autorest-preview.yml@azure-sdk-tools
+  parameters:
+    RegenerationJobCount: ${{ variables.RegenerationJobCount }}
+    BuildPrereleaseVersion: ${{ variables.BuildPrereleaseVersion }}
+    UseTypeSpecNext: ${{ variables.UseTypeSpecNext }}
+    EmitterPackageJsonPath: ${{ variables.EmitterPackageJsonPath }}
+    PublishInternal: ${{ variables.PublishInternal }}
+    Packages:
```
pshao25, line 43: My expectation is these three parameters could be read from variables too. But unfortunately complex type is not acceptable for a variable.

pshao25:
This solution not works.

Issue #4820, opened 2024-06-12
Check the usage of readonly property type in RoundTrip model for DPG


live1206:
TSP:
```ts
  model ResultModel {
    name: string;
  }

  model RoundTripModel {
    @visibility("read")
    result: ResultModel;
  }

  @route("/modelInReadOnlyProperty")
  @put
  op modelInReadOnlyProperty(@body body: RoundTripModel): {
    @body body: RoundTripModel;
  };
```

detailed case can be found in this pr: https://github.com/Azure/cadl-ranch/pull/587

In this case, usage for `ResultModel` should be `Output` instead of `RoundTrip`, because this model is readonly.

ArcturusZhang:
In current DPG's implementation, ResultModel is generated as a `RoundTrip` model. Therefore we need a fix

ArcturusZhang:
Also in today's offline sync up, we kind of have an agreement on the patch convenience methods, therefore we have a way to use the usage from TCGC directly therefore we could expect a final fix for this.

ArcturusZhang:
This should have fixed this issue despite we did not know quite well which part introduces the fix.
But from the result's perspective, this is fixed.

live1206:
> [This](https://github.com/Azure/autorest.csharp/pull/4790) should have fixed this issue despite we did not know quite well which part introduces the fix. But from the result's perspective, this is fixed.

Are you saying csharp emitter has not consumed TCGC usage yet, but the usage still got fixed?

ArcturusZhang:
> > [This](https://github.com/Azure/autorest.csharp/pull/4790) should have fixed this issue despite we did not know quite well which part introduces the fix. But from the result's perspective, this is fixed.
> 
> Are you saying csharp emitter has not consumed TCGC usage yet, but the usage still got fixed?

that statement is not fully correct. We converted the usage value from TCGC to ours (like [this](https://github.com/Azure/autorest.csharp/blob/c875f2eac3eaa0995ffdb0bb8f1ab77dd4d9719e/src/TypeSpec.Extension/Emitter.Csharp/src/lib/converter.ts#L114C20-L114C34)) - therefore when tcgc has the fix, we get the changes in DPG as well.

live1206:
> > > [This](https://github.com/Azure/autorest.csharp/pull/4790) should have fixed this issue despite we did not know quite well which part introduces the fix. But from the result's perspective, this is fixed.
> > 
> > 
[...]

Issue #4734, opened 2024-05-21
Use the usage flag from TCGC directly


ArcturusZhang:

ArcturusZhang:
Unfortunately this is currently blocked by our patch operations.
We did not include patch operation's convenience method, therefore those methods should not be considered during usage calculation, but currently TCGC does not have a way to let us have the filter.

PR #4859, opened 2024-06-21
Use the `UsageFlags` from TCGC directly in our emitter

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4820
Fixes https://github.com/Azure/autorest.csharp/issues/4734

# Description

This PR let our models to use the usageFlags from TCGC directly. Since the usage flags from TCGC is a bitmap which is a number, if we do nothing, the serialized `tspCodeModel.json` will have `usage` properties with integer values, which is not quite readable.
Therefore an additional change is that I added a `replacer` to replace the number value usage to string during serialization.
We could use similar way in the future to remove the `__raw` when we want to directly use the models from TCGC to serialize.

# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -204,6 +212,27 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
...
+                        result.push(UsageFlags[prop]);
+                    }
+                }
+            }
+            return result.join(",");
+        }
+    }
+
+    return value;
+}
```
ArcturusZhang, line 234: this function is to convert the usage flag from a number to a comprehensive string.
The usage from TCGC is defined as a enum flag to be extensible - which is just a bit map, a number.
Therefore if we directly serialize the usage into json, it will be a number like this:
```
{
"kind": "model",
"name": "Foo",
"usage": 6
}
```
I personally do not think letting them to be a number would be a big problem, but per others' requirement, I introduce this function to convert those numbers to a comprehensive string.

We have to do it in this way instead of doing it during the conversion because:
1. the flags are simple to handle with if we really have the requirement of checking if it has a flag or appending more flags. Strings like `Input,Output` is harder to do that.
2. I would like to make things like this as "plugin" of the things we already have. Eventually we would like to directly use the models from TCGC and remove the converter, then the current solution is the only option.

test/CadlRanchProjects/type/property/nullable/src/Generated/tspCodeModel.json
```diff
@@ -281,7 +281,7 @@
    "Name": "CollectionsStringProperty",
    "Namespace": "Type.Property.Nullable",
    "Description": "Model with collection string properties",
-   "Usage": "Output",
+   "Usage": "Input,Output,JsonMergePatch",
```
ArcturusZhang, line 284: this changed because previously we are excluding the patch operation.

src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -128,20 +128,21 @@ export function loadOperation(
...
-                bodyParameter.Type.Name === ""
-            ) {
-                // give body type a name
-                bodyParameter.Type.Name = `${capitalize(op.name)}Request`;
-                const bodyModelType = bodyParameter.Type as InputModelType;
-                bodyModelType.Usage = Usage.Input;
-                // update models cache
-                models.delete("");
-                models.set(bodyModelType.Name, bodyModelType);
+            if (isInputModelType(bodyParameter.Type)) {
```
lirenhe, line 131: As https://github.com/Azure/typespec-azure/issues/69 was resolved, do we still need the below logic?

ArcturusZhang:
Superseded by https://github.com/Azure/autorest.csharp/pull/4917

PR #4867, opened 2024-06-25
Add tag for typespec next

pshao25:
**Goal:**
Filter out all the typespec next run from https://dev.azure.com/azure-sdk/internal/_build?definitionId=5990

**Solution:**
For all the typespec next run , we add a tag `typespec_next` then we could choose from right top as below image shows. Or just this link: https://dev.azure.com/azure-sdk/internal/_build?definitionId=5990&_a=summary&tagFilter=typespec_next

image


eng/pipelines/preview.yml
```diff
@@ -68,6 +68,13 @@ extends:
         $newContent | Set-Content $path -NoNewline
       displayName: 'Update Packages.Data.props'
       workingDirectory: $(sdkRepositoryPath)
+    AutorestInitializationSteps:
+    - ${{ if eq(parameters.UseTypeSpecNext, 'true') }}:
+      - task: Bash@3
+        inputs:
+          targetType: 'inline'
+          script: 'echo "##vso[build.addbuildtag]typespec_next"'
```
archerzz, line 76: so the magic bash script can add a filter in the UI? Is this the formal way to do so?

pshao25: > so the magic bash script can add a filter in the UI?

Yes.

> Is this the formal way to do so?

To be honest I don't know the answer. But we don't have other ways.

pshao25: @hallipr Could you help review this PR?

lirenhe (approved):
This is a workable solution so I approved this one.  @m-nash, could you also help to review this PR?  thanks

eng/pipelines/preview.yml
hallipr: This is the standard way to add tags to a build mid-run.  See:

[Scheduled Builds](https://github.com/Azure/azure-sdk-for-net/blob/Azure.Core_1.40.0/eng/pipelines/templates/jobs/ci.yml#L72)

[Release Tags](
https://github.com/Azure/azure-sdk-for-net/blob/Azure.Core_1.40.0/eng/pipelines/templates/jobs/ci.mgmt.yml#L54)

hallipr, line 76: I would use powershell for xplat support

```suggestion
      - pwsh: |
          echo "##vso[build.addbuildtag]typespec_next"
```

hallipr (approved):

PR #4869, opened 2024-06-26
fix operation accessibility

live1206:


live1206:
Ideally, we should have a test case for typespec input of mgmt. So that, we should have seen this in the first place.

pshao25 (approved):

PR #4877, opened 2024-06-27
Onboard cadl-ranch tests for MPG

### LINKED ISSUE #4870: Add CadlRanch test for MPG with typespec in generator
We would like to have e2e CadlRanch test for MPG typespec in generator.

pshao25:
Fix #4870

Update the type of subscriptionId to string. Currently, compiler just returns guid type to us, which is not align with our current behavior.


test/CadlRanchProjects.Tests/resource-manager-models-resources.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Threading.Tasks;
+using AutoRest.TestServer.Tests.Infrastructure;
+using Azure.Core;
+using Azure.ResourceManager;
```
live1206, line 5: there are some unused references

test/CadlRanchProjects.Tests/MgmtTestHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+using static CadlRanchProjects.Tests.OAuth2TestHelper;
+
+namespace CadlRanchProjects.Tests
+{
+    public class MgmtTestHelper
+    {
+        public static ArmClient CreateArmClientWithMockAuth(Uri host, string defaultSubscriptionId = default)
+        {
+            var options = new ArmClientOptions();
+            options.AddPolicy(new MockBearerTokenAuthenticationPolicy(new MockCredential(), new string[] { "https://security.microsoft.com/.default" }, options.Transport), HttpPipelinePosition.PerCall);
```
live1206, line 13: since this is mock, do we still need a valid scope and audience?

live1206 (approved):

test/CadlRanchProjects.Tests/MgmtTestHelper.cs
pshao25: 1. We are not testing auth here, so no one will read these things.
2. Even we are testing auth, what we need to test is not "what is a valid scope", but "if the scope user sets is sent to server without change", and we already have that [test](https://github.com/Azure/autorest.csharp/blob/72659d42a5067035f0c2d969ed07dd523cbda4dc/test/CadlRanchProjects.Tests/authentication-oauth2.cs#L18).

live1206: I meant we can use test string for these things.

Issue #4868, opened 2024-06-25
Use `crossLanguageDefinitionId` for type replacement


ArcturusZhang:

ArcturusZhang:
Pending on https://github.com/Azure/typespec-azure/pull/1075

ArcturusZhang:
This has partially done by https://github.com/Azure/autorest.csharp/pull/4871 for models and enums

Issue #4801, opened 2024-06-07
We have a `IsEmbeddingsVector` in `InputListType`, TCGC does not


ArcturusZhang:
our definition of `InputListType` is:
```csharp
internal record InputListType(InputType ValueType, bool IsEmbeddingsVector, bool IsNullable) : InputType("Array", IsNullable);
```
which has a boolean `IsEmbeddingsVector`. TCGC does not have this property on its `SdkArrayType`.

ArcturusZhang:
We have this data structure in Azure Core template: https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/data-types#Azure.Core.EmbeddingVector
But currently TCGC and our emitter are doing nothing about it. We need to fix this.

ArcturusZhang:
Pending on https://github.com/Azure/typespec-azure/pull/1000

PR #4879, opened 2024-06-27
adopt the crossLanguageDefinitionId for array type

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4868
Fixes https://github.com/Azure/autorest.csharp/issues/4801

# Description

The new version of TCGC introduces name and `crossLanguageDefinitionId` for array types, which allows us to identify whether this is a regular array or a special array such as the embedding vector.
Now the embedding vector is no longer represented by a flag in the `InputArrayType`, it is distinguished by the generator by its `crossLanguageDefinitionId`.

# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/test/Unit/model-type.test.ts
```diff
@@ -556,34 +540,18 @@ op op5(@body body: IsFooArray): IsFooArray;
...
-        );
-        deepStrictEqual(fooModel, isFooModel.AdditionalProperties);
-        deepStrictEqual(
-            {
-                Kind: "array",
-                ValueType: fooModel
-            },
-            isFooArrayModel.AdditionalProperties
+            isFooArrayModel.AdditionalProperties.ValueType,
+            fooModel
```
ArcturusZhang, line 554: I changed these because of two reasons:
1. we are using the `strictEqual` and `deepStrictEqual` wrong, the `actual` is the first parameter but we are using it as `expected`.
2. we should be focusing on the properties we need to validate by this test case, instead of just simply asserting the model is the same which is hard to read and hard to understand.

PR #4882, opened 2024-06-28
Align ArmId kind with TCGC

live1206:


src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -76,7 +76,7 @@ public TypeFactory(OutputLibrary library, Type unknownType)
                 InputPrimitiveTypeKind.Float => new CSharpType(typeof(double)), // in typespec, float is the base type of float32 and float64, see type relation: https://typespec.io/docs/language-basics/type-relations
                 InputPrimitiveTypeKind.Numeric => new CSharpType(typeof(double)), // in typespec, numeric is the base type of number types, see type relation: https://typespec.io/docs/language-basics/type-relations
                 InputPrimitiveTypeKind.IPAddress => new CSharpType(typeof(IPAddress)),
-                InputPrimitiveTypeKind.ArmResourceIdentifier => new CSharpType(typeof(ResourceIdentifier)),
+                InputPrimitiveTypeKind.ArmId => new CSharpType(typeof(ResourceIdentifier)),
```
ArcturusZhang, line 79: why do we change it back?
Now our emitter just takes the kind from TCGC, and in TCGC, the kind value for arm id is this longer version

live1206: image
This is the output from emitter, so TCGC outputs the shorter name

ArthurMa1978 (approved):

ArcturusZhang (approved):

PR #4883, opened 2024-06-28
Adopt scalar type changes from TCGC

ArcturusZhang:
Adopt latest changes from TCGC.


src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
```diff
@@ -32,7 +32,10 @@ export type InputType =
 
 export interface InputPrimitiveType extends InputTypeBase {
     Kind: SdkBuiltInKinds;
+    Name: string;
     Encode?: string; // In TCGC this is required, and when there is no encoding, it just has the same value as kind
+    CrossLanguageDefinitionId: string;
+    BaseType?: InputPrimitiveType;
```
live1206, line 38: why is this added? it's never set from tcgc
And we don't need to handle this in generator, either.

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
```diff
@@ -62,8 +68,11 @@ export interface InputOffsetDateTimeType extends InputDateTimeTypeBase {
 
 export interface InputDurationType extends InputTypeBase {
     Kind: "duration";
+    Name: string;
     Encode: DurationKnownEncoding;
     WireType: InputPrimitiveType;
+    BaseType?: InputDurationType;
```
live1206, line 74: we are not handling the newly added `BaseType` in generator?

src/AutoRest.CSharp/Common/Generation/Types/TypeFactory.cs
```diff
@@ -75,23 +89,26 @@ public TypeFactory(OutputLibrary library, Type unknownType)
...
-#pragma warning restore CS0618 // Type or member is obsolete
                 _ => new CSharpType(typeof(object)),
-            },
-            InputDateTimeType dateTimeType => new CSharpType(typeof(DateTimeOffset)),
-            InputDurationType durationType => new CSharpType(typeof(TimeSpan)),
-            _ => throw new InvalidOperationException($"Unknown type: {inputType}")
+            };
+        }
+
+        private static readonly IReadOnlyDictionary<string, CSharpType> _knownPrimitiveTypes = new Dictionary<string, CSharpType>
```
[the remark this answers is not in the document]
ArcturusZhang: The key here is the `crossLanguageDefinitionId`.
And based on the principal that we should keep minimum logic in our emitter only - just taking everything TCGC gives us, the `InputPrimitiveTypeKind` should have exactly the same values as `SdkBuiltInTypeKinds`. And in my TCGC part, this `SdkBuiltInTypeKinds` is defined like this:
```typescript
export type SdkBuiltInKinds = Exclude<IntrinsicScalarName, SdkBuiltInKindsExcludes> | "any";

type SdkBuiltInKindsExcludes = "utcDateTime" | "offsetDateTime" | "duration";
```

Therefore, there should never be a situation that we should add a new type kind into this inputPrimitiveTypeKind, unless typespec is updated and decide to add a new kind into `IntrinsicScalarName` which means a new std type is added into typespec.

src/TypeSpec.Extension/Emitter.Csharp/src/type/input-type.ts
ArcturusZhang: We do not take it right now because we do not have a case yet.
We have it here, because the input types should be identical to the type system provided by TCGC

ArcturusZhang: For more details, please refer the description of this PR: https://github.com/Azure/typespec-azure/pull/1015

src/AutoRest.CSharp/Common/Input/InputTypes/InputPrimitiveType.cs
```diff
@@ -3,37 +3,59 @@
...
-    internal InputPrimitiveType(InputPrimitiveTypeKind kind, string? encode) : this(kind)
+    internal const string AzureLocationId = "Azure.Core.azureLocation";
+    internal const string ArmIdId = "Azure.Core.armResourceIdentifier";
+    internal const string UuidId = "Azure.Core.uuid";
+    internal const string ETagId = "Azure.Core.eTag";
+    internal const string CharId = "Azure.Core.char";
+    internal const string ContentTypeId = "Azure.Core.contentType";
+    internal const string ResourceTypeId = "Azure.Core.resourceType";
+
+    internal const string ObjectId = "Obsolete.object";
```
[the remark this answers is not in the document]
ArcturusZhang: [This comment](https://github.com/Azure/autorest.csharp/pull/4883#discussion_r1666387698) shares the same reply here.
Previously we are using some special `InputPrimitiveTypeKind` to handle some legacy types from swagger input.
In this PR since we need to keep the `InputPrimitiveTypeKind` identical as `SdkBuiltInTypeKinds` from TCGC, we no longer could add more kinds into the type (or in other words, we should not), therefore here we just construct a new scalar type here internally with a special `crossLanguageDefinitionId`, so that the `TypeFactory` could know which type to return.

src/AutoRest.CSharp/Common/Input/InputTypes/InputPrimitiveType.cs
```diff
@@ -3,37 +3,59 @@
...
+    internal const string RequestMethodId = "Azure.Core.requestMethod";
+
+    internal const string IPAddressId = "Temp.ipAddress";
```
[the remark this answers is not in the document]
ArcturusZhang: See https://github.com/Azure/autorest.csharp/pull/4883#discussion_r1683960316

For ipAddress, maybe we could change this to share with `ipv4Address` or `ipv6Address`.

These are not decided yet.

Issue #4898, opened 2024-07-08
[defect] codegen gen crash : null value in list is not allowed error


chunyu3:
Typespec  [tsp](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/schemaregistry/SchemaRegistry/main.tsp), and emitter `"2021-10" API Version`

Compiling tsp using @azure-tools/typespec-csharp...
Unhandled exception. System.Text.Json.JsonException: null value in list is not allowed
   at AutoRest.CSharp.Common.Input.Utf8JsonReaderExtensions.TryReadWithConverter[T](Utf8JsonReader& reader, String propertyName, JsonSerializerOptions options, IReadOnlyList`1& value) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Input\Utf8JsonReaderExtensions.cs:line 154
   at AutoRest.CSharp.Common.Input.TypeSpecInputOperationConverter.CreateOperation(Utf8JsonReader& reader, String id, String name, JsonSerializerOptions options, ReferenceResolver resolver) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Input\InputTypes\Serialization\TypeSpecInputOperationConverter.cs:line 54
   at AutoRest.CSharp.Common.Input.TypeSpecInputOperationConverter.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Input\InputTypes\Serialization\TypeSpecInputOperationConverter.cs:line 24
   at AutoRest.CSharp.Common.Input.Utf8JsonReaderExtensions.ReadWithConverter[T](Utf8JsonReader& reader, JsonSerializerOptions options) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Input\Utf8JsonReaderExtensions.cs:line 163
   at AutoRest.CSharp.Common.Input.Utf8JsonReaderExtensions.TryReadWithConverter[T](Utf8JsonReader& reader, String propertyName, JsonSerializerOptions options, IReadOnlyList`1& value) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Input\Utf8JsonReaderExtensions.cs:line 153
[...]

PR #4903, opened 2024-07-09
get the value from literal

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/4898

# Checklist

# Ready to Land?

### DIFF
--- a/src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
+++ b/src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
@@ -179,7 +179,8 @@ export function loadOperation(
     );
     if (contentTypeParameter) {
         if (isInputLiteralType(contentTypeParameter.Type)) {
-            mediaTypes.push(contentTypeParameter.DefaultValue?.Value);
+            if (contentTypeParameter.Type.Value as string)
+                mediaTypes.push(contentTypeParameter.Type.Value as string);
         } else if (isInputUnionType(contentTypeParameter.Type)) {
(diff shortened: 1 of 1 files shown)


src/TypeSpec.Extension/Emitter.Csharp/src/lib/operation.ts
```diff
@@ -179,7 +179,8 @@ export function loadOperation(
     );
     if (contentTypeParameter) {
         if (isInputLiteralType(contentTypeParameter.Type)) {
-            mediaTypes.push(contentTypeParameter.DefaultValue?.Value);
+            if (contentTypeParameter.Type.Value as string)
+                mediaTypes.push(contentTypeParameter.Type.Value as string);
```
ArcturusZhang, line 183: the `as` in typescript does not work in this way.
If this value is not undefined, even it is a number, this check is true.
We should do:
```
if (typeof contentTypeParameter.Type.Value === "string")
```
to check if the value is string.
Or if you just want to put in whatever value it has, we could just do `mediaTypes.push` without the if.

chunyu3: https://github.com/Azure/autorest.csharp/issues/4898 is just a trigger of the  issue. It passed Union as array with null value, but also when we handle `InputLiteralType`, we use the `DefaultValue` to get the value of an literal which is not correct, we should use `InputLiteralType.Value`.

I will close this PR, and file another PR in MGC emitter to resolve this.

PR #4916, opened 2024-07-15
Remove RenameGetList-TypeSpec

melina5656:
Added testing code to test whether the `getUser` and `getUsers` methods exist.
Remove code related to `RenameGetList-TypeSpec`.


test/CadlRanchProjects.Tests/azure-core-basic.cs
```diff
@@ -129,5 +129,23 @@ await foreach (Page<SecondItem> page in second.AsPages())
...
+
+        [Test]
+        public Task Azure_Core_basic_confirmGetUserName() => Test(async (host) =>
+        {
+            User response = await new BasicClient(host, null).GetUserAsync(1);
+            var method = typeof(BasicClient).GetMethod("GetUserAsync");
+            var attribute = method.GetCustomAttributes<OperationAttribute>();
+            Assert.AreEqual(attribute.name, "getUser")
+        });
+
```
pshao25, line 141: We just want to check the name is correct. So
1. We don't need to call cadl-ranch server.
2. We just check the method exits.

Something like
```
[Test]
        public Task Azure_Core_basic_confirmGetUserName()
        {
            var method = typeof(BasicClient).GetMethod("GetUserAsync"); // Get method whose parameter is int
            Assert.IsNotNull(method)
        });
```

test/CadlRanchProjects.Tests/azure-core-basic.cs
```diff
@@ -129,5 +129,23 @@ await foreach (Page<SecondItem> page in second.AsPages())
                 Assert.AreEqual("Madge", firstItem.Name);
             }
         });
+
```
pshao25, line 132: We are now safe to remove the whole RenameGetList-TypeSpec project with these two tests.

test/CadlRanchProjects.Tests/azure-core-basic.cs
```diff
@@ -129,5 +129,19 @@ await foreach (Page<SecondItem> page in second.AsPages())
...
+        [Test]
+        public Task Azure_Core_basic_confirmGetUserName() => Test(async (host) =>
```
pshao25, line 134:
```suggestion
        public Task Azure_Core_basic_RenameGetMethod()
```

test/CadlRanchProjects.Tests/azure-core-basic.cs
```diff
@@ -129,5 +129,19 @@ await foreach (Page<SecondItem> page in second.AsPages())
...
+        {
+            var method = typeof(BasicClient).GetMethod("GetUserAsync");
+            Assert.IsNotNull(method)
+        });
+
+        [Test]
+        public Task Azure_Core_basic_confirmGetUsersName() => Test(async (host) =>
```
pshao25, line 141:
```suggestion
        public Task Azure_Core_basic_RenameListMethod()
```

test/CadlRanchProjects.Tests/azure-core-basic.cs
```diff
@@ -129,5 +131,19 @@ await foreach (Page<SecondItem> page in second.AsPages())
                 Assert.AreEqual("Madge", firstItem.Name);
             }
         });
+
+        [Test]
+        public void Azure_Core_basic_RenameGetMethod()
+        {
+            var method = typeof(BasicClient).GetMethod("GetUserAsync", new[] { typeof(int), typeof(CancellationToken) });
```
pshao25, line 138: Can we add a line testing there is no method called `Get` ? Same as below, checking there is not method called `List`.

pshao25 (approved):

PR #4917, opened 2024-07-16
Use the `UsageFlags` from TCGC directly in our emitter

ArcturusZhang:
Takes the emitter changes from https://github.com/microsoft/typespec/pull/3842

# Description

This PR updates the version of emitter, which takes the usageFlags from TCGC directly to models and enums.
This means now our code model takes whatever the usage calculated by TCGC, we no longer do any patches or appendix to the usage any more.

We have a requirement that we do not generate convenience method for PATCH until it is ready. We still maintain this behavior, but the usage is controlled by TCGC which does not know the convenience method for PATCH should be excluded.
Therefore a diagnostic warning has been added to the emitter so that if the convenience method of PATCH is **NOT** disabled, a warning will be reported telling you that the operation should add `@convenienceAPI(false)`.

# Checklist

# Ready to Land?


test/CadlRanchProjects/parameters/body-optionality/src/Generated/BodyOptionalityClient.cs
```diff
@@ -165,9 +165,9 @@ public virtual async Task<Response> RequiredImplicitAsync(string name, Cancellat
         {
             Argument.AssertNotNull(name, nameof(name));
 
-            BodyModel bodyModel = new BodyModel(name, null);
```
archerzz, line 168: Is this change due to the TCGC update in MGC?

test/AutoRest.TestServer.Tests/Common/Utilities/NamedTypeSymbolExtensionsTests.cs
```diff
@@ -159,12 +159,12 @@ public void IsSameType_PrimitiveTypes()
         public void IsSameType_ModelTypes()
         {
             // Different namespace
-            var input = new InputModelType("MetadataModel", "", null, null, null, InputModelTypeUsage.RoundTrip, null, null, Array.Empty<InputModelType>(), null, null, null, null);
+            var input = new InputModelType("MetadataModel", "", null, null, null, InputModelTypeUsage.Input | InputModelTypeUsage.Output, null, null, Array.Empty<InputModelType>(), null, null, null, null);
```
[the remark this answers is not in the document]
ArcturusZhang: I do not like it because the input type csharp types should be reflecting the structure as the models defined in TCGC exactly.
We should not include extra public members to them since they are public and cannot be changed once added.
If you make the statistics, majority of the usage is in test cases.
Test cases could do something, but we should not add it into the InputModelTypeUsage API.

And since I have done all the mitigations of those usages, I do not see it is needed we have to introduce the convenience API for roundtrip in this PR right now.

test/CadlRanchProjects/parameters/body-optionality/src/Generated/BodyOptionalityClient.cs
ArcturusZhang: This is caused by cadl ranch update.

archerzz (approved):

src/TypeSpec.Extension/Emitter.Csharp/package.json
```diff
@@ -36,7 +36,7 @@
     "dependencies": {
         "@autorest/csharp": "3.0.0-beta.20240625.4",
         "json-serialize-refs": "0.1.0-0",
-        "@typespec/http-client-csharp":"0.1.9-alpha.20240718.1"
+        "@typespec/http-client-csharp": "https://artprodcus3.artifacts.visualstudio.com/A0fb41ef4-5012-48a9-bf39-4ee3de03ee35/29ec6040-b234-4e31-b139-33dc4287b756/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2F6dXJlLXNkay9wcm9qZWN0SWQvMjllYzYwNDAtYjIzNC00ZTMxLWIxMzktMzNkYzQyODdiNzU2L2J1aWxkSWQvMzk4OTA3OC9hcnRpZmFjdE5hbWUvYnVpbGRfYXJ0aWZhY3RzX2NzaGFycA2/content?format=file&subPath=%2fpackages%2ftypespec-http-client-csharp-0.1.9-alpha.20240724.7.tgz"
```
chunyu3, line 39: Please use the version released in public feed.

ArcturusZhang: I will update the version to the official version before I merge this.

ArcturusZhang: version updated

PR #4918, opened 2024-07-16
Enable ResponseBodyPolicy inside mock pipelines

### LINKED ISSUE #4878: Any operation except get just not working in cadl-ranch context
Let's see the difference between `get` and `create` first to understand the problem.

[**Get operation**](https://github.com/Azure/autorest.csharp/blob/5c661b2b2db11024160522fb4fa4cde3a91b29ae/test/CadlRanchProjects/azure/resource-manager/models/resources/src/Generated/RestOperations/TopLevelTrackedResourcesRestOperations.cs#L93-L96)
Before this method return, we already extract the data from `message.Response.ContentStream`
```C#
using var document = await JsonDocument.ParseAsync(message.Response.ContentStream, default, cancellationToken).ConfigureAwait(false);
```
So when we return as `return Response.FromValue(value, message.Response);`, we already get the value. Though we return `message.Response` which is now available in this method, it will be disposed when code goes out of this method. Since we already get the value, we will not read `message.Response` any more so that we are not facing problem.

[**Create operation**](https://github.com/Azure/autorest.csharp/blob/5c661b2b2db11024160522fb4fa4cde3a91b29ae/test/CadlRanchProjects/azure/resource-manager/models/resources/src/Generated/RestOperations/TopLevelTrackedResourcesRestOperations.cs#L191-L197)
This is a long running operation, which does not process the response like get operation. Instead it just returns `message.Response` to let its caller to process, at which point, the `message.Response` is disposed.

**To summarize which cases would have problem:** as long as `message.Response` is not processed after it is received and it is returned to let caller process, we will get `null` from this response.

**What operations have problem:** all the operations except `get`.

**It's a known issue. Related issues:**
https://github.com/Azure/autorest.csharp/issues/3707
https://github.com/Azure/autorest.csharp/issues/1571

pshao25:
Fix #4878 

The root cause is actually different from previous ones:

To bypass the HTTPS check in real BearerTokenAuthenticationPolicy, we have to short circuit to _transport inside MockBearerTokenAuthenticationPolicy, which also bypass the ResponseBodyPolicy. This policy is crucial to read response to memory so that when we disposing response the data won't lose. However, ResponseBodyPolicy is internal, so copy the logic directly here.


test/CadlRanchProjects.Tests/OAuth2TestHelper.cs
```diff
@@ -50,13 +55,55 @@ public override void Process(HttpMessage message, ReadOnlyMemory<HttpPipelinePol
...
                 await _transport.ProcessAsync(message).ConfigureAwait(false);
+                await ResponseBodyPolicyAsync(message, pipeline).ConfigureAwait(false);
 
                 var response = message.Response;
                 Type responseType = response.GetType();
                 PropertyInfo propInfo = responseType.GetProperty("IsError", BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
                 propInfo.SetValue(response, message.ResponseClassifier.IsErrorResponse(message));
             }
 
+            private async ValueTask ResponseBodyPolicyAsync(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
```
ArcturusZhang, line 66: I think this method name does not explain what this is doing, could we specify?

ArcturusZhang: Looks like this method is trying to copy out the content stream of this response to a new memory stream, and then set it back as the content stream of the response.

I think we need some more comment on the method or where we are doing this to explain why we need this.
I assume we must do this because we want the stream could always be seekable. Any way, please add some comment to explain what we are doing here.

Also does the same issue could happen in real world instead of mock tests? Maybe we need to fix our implementation to ensure that as well?

ArcturusZhang: Also since the pipeline argument is not used here, please remove it

test/CadlRanchProjects.Tests/resource-manager-models-resources.cs
```diff
@@ -32,7 +32,7 @@ public class ResourceManagerModelsResourcesTests: CadlRanchTestBase
         });
 
         [Test]
-        [Ignore("https://github.com/Azure/autorest.csharp/issues/4878")]
+        [Ignore("https://github.com/Azure/autorest.csharp/issues/4876")]
```
ArcturusZhang, line 35: so we changed this because we put the wrong issue link initially?

test/CadlRanchProjects.Tests/OAuth2TestHelper.cs
```diff
@@ -31,6 +33,9 @@ public override AccessToken GetToken(TokenRequestContext requestContext, Cancell
         public class MockBearerTokenAuthenticationPolicy : BearerTokenAuthenticationPolicy
         {
             private readonly HttpPipelineTransport _transport;
+            private readonly TimeSpan _networkTimeout = TimeSpan.FromSeconds(100);
+            // Same value as Stream.CopyTo uses by default
+            private const int DefaultCopyBufferSize = 81920;
```
live1206, line 38: consider to put this in `CopyToAsync` since it is the only usage.

test/CadlRanchProjects.Tests/OAuth2TestHelper.cs
live1206, line 66: consider to put this together with `CopyToAsync` and add comments where they are from.

test/CadlRanchProjects.Tests/resource-manager-models-resources.cs
live1206: I think both are needed in the first place, after this change, only the existing one left.

test/CadlRanchProjects.Tests/OAuth2TestHelper.cs
pshao25: Prepend "Process" to it. I'd like to keep the pipeline parameter to keep consistence with its source.

live1206 (approved):

Issue #4913, opened 2024-07-12
ProjectScaffolding create invalid project files for new Management-plane SDK cause compile failure


chunyu3:
`ProjectScaffolding` add un-expected project item in SDK project files (.csproj) for manage-plane SDK cause SDK compile failure: duplicate `compile` items

![image](https://github.com/user-attachments/assets/e6e73170-d244-4ca5-871f-26104d452b30)

expected:
in project file for management-plane SDK, no extra `projectReference` and `packageReference` added. Management-plane has add all those dependency/reference in azure-sdk-for-net engsys

PR #4923, opened 2024-07-17
remove packageReference from project file for management sdk

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/4913

Problem-tp-solve: `Duplicate compile items`
When generate a new service from typespec, the .NET codegen will generate scaffolding (include project file) for this new SDK. In the generated project file it includes several `packageReference` to add compile dependency. But in azure-sdk-for-net, the engsys already include all depended packages for management-plane SDK in `Directory.Build.Common.targets`. So the build will fail because of `duplicate compile items`.

Solution:
Since existing SDKs in `azure-sdk-for-net` contain define those `packageRefernce` in each SDK's project file, and we also need to   support management SDKs from swagger which use the dotnet templates to create scaffolding folder, we decide to change in codegen side to remove `packageReference` when create scaffolding which will keep consistent with current SDKs.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
```diff
@@ -315,18 +315,18 @@ private string GetBrandedSrcCSProj()
...
-            foreach (var packages in _brandedDependencyPackages)
+            if (!Configuration.AzureArm)
             {
-                builder.PackageReferences.Add(packages);
-            }
-            if (Configuration.AzureArm)
-            {
-                builder.PackageReferences.Add(new("Azure.ResourceManager"));
+                // only branded library will add these shared code compilation lines
+                builder.CompileIncludes.Add(new("$(AzureCoreSharedSources)AzureResourceProviderNamespaceAttribute.cs", "Shared/Core"));
```
live1206, line 323: Not quite follow the change for this file.
Here, we are only adding it for DPG. And, we are adding it in [Directory.Build.targets](https://github.com/Azure/autorest.csharp/pull/4923/files#diff-50e91c82311ea26f2a73202525dfdf2b0a89c178ee666b2bf3ad4c84ac4c06dc) for mgmt.
Does it mean we need this for both DPG and Mgmt? We can just remain as it was, adding it for branded library.

samples/Directory.Build.props
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <IsTestGenerationSrcProject Condition="!$(MSBuildProjectName.EndsWith('.Tests'))">true</IsTestGenerationSrcProject>
+    <IsMgmtLibrary Condition="!$(MSBuildProjectName.EndsWith('.Tests')) and ($(MSBuildProjectName.StartsWith('_Azure.ResourceManager')) or $(MSBuildProjectName.StartsWith('Mgmt')))">true</IsMgmtLibrary>
```
live1206, line 5: The CI is failing due to `AzureSample.ResourceManager.Storage` doesn't follow the naming check here.
Consider to update to `$(MSBuildProjectName.Contains('.ResourceManager')) or $(MSBuildProjectName.Contains('Mgmt'))`

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
chunyu3: hello @live1206  It is consistent with azure-sdk-for-net. SDKs in azure-sdk-for-net will not add this in project file, and it add this in azure-sdk-for-net eng for all SDKs.
So in autorest.csharp, we will do the same settings. then the generated project file will not include this compile item. Otherwise, when put this SDK in azure-sdk-for-net, the build will fail.

samples/Directory.Build.props
chunyu3: Yes. I will update. Thanks

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
live1206: OK, then we need to regen update the test csproj files, CI is failing due to the reference order change as below
image

samples/Directory.Build.props
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <IsTestGenerationSrcProject Condition="!$(MSBuildProjectName.EndsWith('.Tests'))">true</IsTestGenerationSrcProject>
+    <IsMgmtLibrary Condition="!$(MSBuildProjectName.EndsWith('.Tests')) and ($(MSBuildProjectName.StartsWith('AzureSample.ResourceManager')) or $(MSBuildProjectName.StartsWith('Mgmt')))">true</IsMgmtLibrary>
```
live1206, line 5: since this is for sample 
```suggestion
    <IsMgmtLibrary Condition="!$(MSBuildProjectName.EndsWith('.Tests')) and ($(MSBuildProjectName.StartsWith('AzureSample.ResourceManager')))">true</IsMgmtLibrary>
```

live1206 (approved):

Issue #4887, opened 2024-07-02
Add `decorator` fields in every Input types in tspCodeModel.json


chunyu3:
TCGC will emit raw decorator information `decorators` for each item(client, property, operation). InputModel should contain this information.

PR #4940, opened 2024-07-26
Emit decorator lists

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/4887

When we specify the allowed-emit-decorators in emitter, TCGC will help to emit out decoration information for each item (client, property, operation), e.g final-state-var, our .NET input-Model need to contain this information.

This PR will

- emit decorators to all kind of models, operations and operations
- add reference resolve strategy in TypeSpecInputDecoratorInfoConverter to support parse by reference

# Checklist

# Ready to Land?


src/TypeSpec.Extension/Emitter.Csharp/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+export const AzureDefaultAdditionalDecorators = [
+    "Azure\\.ClientGenerator\\.Core\\.@clientName"
```
chunyu3, line 5: This is for testing. We should add the real Decorators Azure emitter want to emit, e.g. "Azure\\.ResourceManager\\.$armProviderNamespace"

src/AutoRest.CSharp/Common/Input/InputTypes/InputDecoratedType.cs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace AutoRest.CSharp.Common.Input;
+internal record InputDecoratedType(IReadOnlyList<InputDecoratorInfo> Decorators);
```
ArcturusZhang, line 7: there is no point of adding this type here. We could just put this property in `InputType`.

src/TypeSpec.Extension/Emitter.Csharp/src/sdk-context-options.ts
```diff
@@ -6,5 +6,5 @@ import { CreateSdkContextOptions } from "@azure-tools/typespec-client-generator-
 export const azureSDKContextOptions: CreateSdkContextOptions = {
     versioning: {},
     // TODO: replace with `@hasJsonConverter` when it is available
-    additionalDecorators: []
+    additionalDecorators: ["Azure\\.ClientGenerator\\.Core\\.@clientName"]
```
chunyu3, line 9: This is for testing. We should add the real Decorators Azure emitter want to emit, e.g. "Azure\.ResourceManager\.$armProviderNamespace"

ArcturusZhang:
Please update the title to show the purpose of this PR. Sync MGC4031 is very ambiguous.
For instance yours could be `Implement decorator list` or just use the same name as the PR in MGC.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputUnionTypeConverter.cs
```diff
@@ -44,7 +47,7 @@ public static InputUnionType CreateInputUnionType(ref Utf8JsonReader reader, str
                 throw new JsonException("Union must have variant types.");
             }
 
-            var unionType = new InputUnionType(name, variantTypes);
+            var unionType = new InputUnionType(name, variantTypes) { Decorators = decorators ?? Array.Empty<InputDecoratorInfo>() };
```
ArcturusZhang, line 50: FYI since we have dotnet 8 here as well, we could also have
```suggestion
            var unionType = new InputUnionType(name, variantTypes) { Decorators = decorators ?? [] };
```

ArcturusZhang: but please do not consider this as a blocker, we do not need to change this, just keep in mind in the future we should use collection expression.

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputUnionTypeConverter.cs
chunyu3: Thanks.  This feature need c# 12 or higher, but autorest.csharp use c# 11, so we cannot use this.

PR #4945, opened 2024-07-30
Add WriteCore to serialization

### LINKED ISSUE #4884: Adopt the latest serialization code pattern
Similar to this:

https://github.com/microsoft/typespec/blob/1bb40322058ac5a0c33127446aa58c3c3be44da0/packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Models/ProjectedModel.Serialization.cs#L36

pshao25:
fix #4884 

Regen: https://github.com/Azure/azure-sdk-for-net/pull/45328


src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -43,6 +44,8 @@ namespace AutoRest.CSharp.Common.Output.Builders
 {
     internal static class JsonSerializationMethodsBuilder
     {
+        private static string _jsonModelWriteCoreMethodName = "JsonModelWriteCore";
```
live1206, line 47: should be `const string`

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -115,19 +118,29 @@ public static IEnumerable<Method> BuildJsonSerializationMethods(JsonObjectSerial
                 }
             }
 
+            if (iJsonModelInterface == null && iPersistableModelTInterface == null && Configuration.UseModelReaderWriter)
```
live1206, line 121: use `is null` to be consistent with others like L127

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -115,19 +118,29 @@ public static IEnumerable<Method> BuildJsonSerializationMethods(JsonObjectSerial
...
+            }
+
             if (iJsonModelInterface is not null && iPersistableModelTInterface is not null)
             {
                 var typeOfT = iJsonModelInterface.Arguments[0];
                 var model = typeOfT.Implementation as SerializableObjectType;
                 Debug.Assert(model != null, $"{typeOfT} should be a SerializableObjectType");
 
+                // void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options)
+                var jsonModelWriteCore = BuildJsonModelWriteCoreMethod(json, iPersistableModelTInterface, hasInherits, isSealed);
```
live1206, line 134: consider to put this line before L121, so that it can be reused for both cases.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -195,6 +208,61 @@ private static MethodBodyStatement[] WriteObject(JsonObjectSerialization seriali
...
+            var body = new List<MethodBodyStatement>()
+            {
+                Serializations.ValidateJsonFormat(options, iPersistableModelTInterface, Serializations.ValidationType.Write),
+                CallBaseJsonModelWriteCore(utf8JsonWriter, options, hasInherits),
+                WriteProperties(utf8JsonWriter, serialization.Properties, serialization.RawDataField?.Value, options).ToArray(),
+                SerializeAdditionalProperties(utf8JsonWriter, options, serialization.AdditionalProperties, false),
+            };
+            if (!hasInherits)
+            {
+                body.Add(SerializeAdditionalProperties(utf8JsonWriter, options, serialization.RawDataField, true));
```
live1206, line 253: how about adding this like `CallBaseJsonModelWriteCore`?

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -724,12 +725,22 @@ protected override bool EnsureIncludeSerializer()
...
 
         protected override JsonObjectSerialization? BuildJsonSerialization()
         {
-            return MgmtReferenceType.IsReferenceType(InputModel) || !InputModel.Serialization.Json ? null : _serializationBuilder.BuildJsonObjectSerialization(InputModel, this);
+            return !InputModel.Serialization.Json ? null : _serializationBuilder.BuildJsonObjectSerialization(InputModel, this);
+        }
+
+        protected override ObjectTypeSerialization BuildSerialization()
+        {
+            bool isReference = MgmtReferenceType.IsReferenceType(InputModel);
```
live1206, line 738: what are we doing here for refenrenceType?

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
pshao25: But it would be a waste if neither of these two conditions meets.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
pshao25: reference types don't have any interface. it is our existing behavior.

live1206: My question is why do we add a parameterless constructor of `SerializationInterfaces` and use it for referenceType only here?

pshao25: What we expect is:
Normal model: has interface, has writecore.
Reference model: doesn't have interface, but has writecore. 
My intention is to tell apart these two cases: to make interface of reference model null.

It's just when creating `ObjectTypeSerialization`, if we pass in `null` how could we tell of it is explicitly set or default `null`? https://github.com/Azure/autorest.csharp/blob/788cb700a0f236ac3adc8de6100bb848cb312d15/src/AutoRest.CSharp/Common/Output/Models/Serialization/ObjectTypeSerialization.cs#L29

live1206: sounds like we should have another flag `hasWriteCore` instead of dealing with `Interfaces` differently

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -724,12 +725,22 @@ protected override bool EnsureIncludeSerializer()
...
+            var json = BuildJsonSerialization();
+            var xml = isReference ? null : BuildXmlSerialization();
+            var bicep = isReference ? null : BuildBicepSerialization(json);
+            var multipart = isReference ? null : BuildMultipartSerialization();
+            return new ObjectTypeSerialization(this, json, xml, bicep, multipart, isReference);
```
live1206, line 743: how about we input interfaces instead like others? So that, isReference is kept within SchemaObjectType.

live1206 (approved):

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -195,6 +210,64 @@ private static MethodBodyStatement[] WriteObject(JsonObjectSerialization seriali
...
             };
 
+        private static MethodBodyStatement[] BuildJsonModelWriteMethodBody(Method jsonModelWriteCoreMethod, Utf8JsonWriterExpression utf8JsonWriter)
+        {
+            var coreMethodSignature = jsonModelWriteCoreMethod.Signature;
+
+            return new[]
+            {
+                utf8JsonWriter.WriteStartObject(),
+                This.Invoke(coreMethodSignature.Name, coreMethodSignature.Parameters.Select(p => (ValueExpression)p).ToList()).ToStatement(),
```
ArcturusZhang, line 220: we should be able just use
```csharp
This.Invoke(coreMethodSignature)
```
there is an overload taking a signature

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -195,6 +210,64 @@ private static MethodBodyStatement[] WriteObject(JsonObjectSerialization seriali
...
+
+        // void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options)
+        private static Method BuildJsonModelWriteCoreMethod(JsonObjectSerialization serialization, CSharpType? iPersistableModelTInterface, bool hasInherits, bool isSealed)
+        {
+            MethodSignatureModifiers modifiers = hasInherits ? MethodSignatureModifiers.Protected | MethodSignatureModifiers.Override : MethodSignatureModifiers.Protected | MethodSignatureModifiers.Virtual;
+            if (serialization.Type.IsValueType || isSealed)
+                modifiers = MethodSignatureModifiers.Private;
+
+            var writerParameter = new Parameter("writer", $"The JSON writer.", typeof(Utf8JsonWriter), null, ValidationType.None, null);
+            var optionsParameter = new Parameter("options", $"The client options for reading and writing models.", typeof(ModelReaderWriterOptions), null, ValidationType.None, null);
```
ArcturusZhang, line 233: I think the knownparameters.serializations type also defines parameters for writer and options, why we are not using them?

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -195,6 +210,64 @@ private static MethodBodyStatement[] WriteObject(JsonObjectSerialization seriali
...
+            var utf8JsonWriter = new Utf8JsonWriterExpression(KnownParameters.Serializations.Utf8JsonWriter);
+            var options = new ModelReaderWriterOptionsExpression(KnownParameters.Serializations.Options);
```
ArcturusZhang, line 235: the expressions here are built from the pre-defined parameters but the parameters are defined here.
We should avoid the inconsistency

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -195,6 +210,64 @@ private static MethodBodyStatement[] WriteObject(JsonObjectSerialization seriali
...
+
+        private static MethodBodyStatement[] BuildJsonModelWriteCoreMethodBody(JsonObjectSerialization serialization, Utf8JsonWriterExpression utf8JsonWriter, ModelReaderWriterOptionsExpression options, CSharpType? iPersistableModelTInterface, bool hasInherits)
+        {
+            return new[]
+            {
+                Serializations.ValidateJsonFormat(options, iPersistableModelTInterface, Serializations.ValidationType.Write),
+                CallBaseJsonModelWriteCore(utf8JsonWriter, options, hasInherits),
+                WriteProperties(utf8JsonWriter, serialization.SelfProperties, serialization.RawDataField?.Value, options).ToArray(),
+                SerializeAdditionalProperties(utf8JsonWriter, options, serialization.AdditionalProperties, false),
+                CallSerializeAdditionalPropertiesForRawdata(serialization, utf8JsonWriter, options, hasInherits)
```
ArcturusZhang, line 252:
```suggestion
                CallSerializeAdditionalPropertiesForRawData(serialization, utf8JsonWriter, options, hasInherits)
```

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -195,6 +210,64 @@ private static MethodBodyStatement[] WriteObject(JsonObjectSerialization seriali
...
+            return hasInherits ?
+                EmptyStatement
+                : SerializeAdditionalProperties(utf8JsonWriter, options, serialization.RawDataField, true);
+        }
+
+        private static MethodBodyStatement CallBaseJsonModelWriteCore(Utf8JsonWriterExpression utf8JsonWriter, ModelReaderWriterOptionsExpression options, bool hasInherits)
+        {
+            // base.<JsonModelWriteCore>()
+            return hasInherits ?
+                new KeywordExpression("base", null).Invoke(_jsonModelWriteCoreMethodName, utf8JsonWriter, options).ToStatement()
```
ArcturusZhang, line 267: Could we define a `Base` in the snippets class?

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -406,23 +407,32 @@ private IEnumerable<JsonPropertySerialization> GetPropertySerializationsFromBag(
...
+                    {
+                        selfPropertyBag.Properties.Add(objectTypeProperty, objectType.GetForMemberSerialization(objectTypeProperty.Declaration.Name));
+                    }
                 }
             }
             PopulatePropertyBag(propertyBag, 0);
+            PopulatePropertyBag(selfPropertyBag, 0);
+
             var properties = GetPropertySerializationsFromBag(propertyBag, objectType).ToArray();
+            var selfProperties = GetPropertySerializationsFromBag(selfPropertyBag, objectType).ToArray();
```
ArcturusZhang, line 429: Please add some comments here to explain why we have to have two "property bags" here. This looks so weird

src/AutoRest.CSharp/Common/Output/Builders/SerializationMethodsBuilder.cs
```diff
@@ -45,7 +45,9 @@ public static IEnumerable<Method> BuildSerializationMethods(SerializableObjectTy
             {
                 if (model.IncludeSerializer)
                 {
-                    foreach (var method in JsonSerializationMethodsBuilder.BuildJsonSerializationMethods(json, serialization.Interfaces))
+                    bool hasInherits = model.Inherits != null && model.Inherits is { IsFrameworkType: false };
```
ArcturusZhang, line 48: `model.Inherits is {IsFrameworkType: false}` implies `model.Inherits is not null`
therefore I think this could be simplified.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
pshao25: The overload is taking `MethodSignature`, not `MethodSignatureBase`.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
pshao25: That one doesn't have description, therefore the generate API won't have description.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -115,19 +117,32 @@ public static IEnumerable<Method> BuildJsonSerializationMethods(JsonObjectSerial
                 }
             }
 
+            if (interfaces is null && Configuration.UseModelReaderWriter)
```
chunyu3, line 120: I saw you defined an option `use-write-core` to indicate if we will use `write-core-strategy`, It seems we need to use this option to identify if we need to create core method.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: then cast it to MethodSignature.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: Then add parameters with descriptions in the knownparameters and use them here. Also below we should use these new parameters

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -425,6 +425,8 @@ public JsonObjectSerialization BuildJsonObjectSerialization(InputModelType input
             PopulatePropertyBag(propertyBag, 0);
             PopulatePropertyBag(selfPropertyBag, 0);
 
+            // properties: all the properties containing the properties from base needed to build the constructor
+            // selfProperties: the properties not containing properties from base just used to do the serialization
             var properties = GetPropertySerializationsFromBag(propertyBag, objectType).ToArray();
```
ArcturusZhang, line 430: I am more curious why we need both.
If this is a temporary solution, please add TODO and/or open an issue to track.
If this is permanent, I think we need comments to explain why we need both.
The serialization objects might not have the hierarchy and we have to do this, this is an acceptable explanation, but we need to write them somewhere so that in the future we will not be confused.

ArcturusZhang, line 430: I am more curious why we need both.
If this is a temporary solution, please add TODO and/or open an issue to track.
If this is permanent, I think we need comments to explain why we need both.
The serialization objects might not have the hierarchy and we have to do this, this is an acceptable explanation, but we need to write them somewhere so that in the future we will not be confused.

src/AutoRest.CSharp/Common/Output/Expressions/Snippets.cs
```diff
@@ -86,6 +86,7 @@ public static BoolExpression Is(ValueExpression left, ValueExpression right)
         public static KeywordStatement Return(ValueExpression expression) => new("return", expression);
         public static KeywordStatement Return() => new("return", null);
         public static KeywordStatement Throw(ValueExpression expression) => new("throw", expression);
+        public static KeywordExpression Base => new KeywordExpression("base", null);
```
ArcturusZhang, line 89: could we put this alongside with `This`?

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
pshao25: Yes, the flag is to control whether we have writecore. But the condition here is a special handling for ResourcedData and Resource.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
pshao25: Because from properties we don't know which is from base, which is from itself. Not sure if it is what you said "The serialization objects might not have the hierarchy".

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
pshao25: What benefit we could get from cast? I don't want to do casting. It makes me feel unsafe.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
pshao25: I updated the description to make it more clear.

src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
ArcturusZhang: In MGC, the signature from a "method" is `MethodSignature`.
This is a bad design on our side that the constructor method and normal method share the same type `Method` which requires that we should always "instantiate a Method with a MethodSignature, instantiate a constructor with a ConstructorSignature", therefore it is very safe we do the cast here because we know it should never be a constructor here.
In MGC, we changed it so that constructor has its own type `ConstructorProvider` and normal method has its own independent type `MethodProvider`.

Back to this comment, it would be cleaner if we use the cast here, this is the benefit. And the disadvantage you want to avoid is that you are afraid the signature we get here is not `MethodSignature`, but this will never happen, or if it happens, we should be happy that it blows up here because we need to know it goes wrong.

src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs
```diff
@@ -83,10 +83,14 @@ private void WriteObjectSerialization(CodeWriter writer, SerializableObjectType
                 }
 
                 writer.Append($"{declaration.Accessibility} partial {(model.IsStruct ? "struct" : "class")} {declaration.Name}")
-                    .AppendRawIf(" : ", model.IncludeSerializer);
-                foreach (var i in serialization.Interfaces)
+                    .AppendRawIf(" : ", model.IncludeSerializer && serialization.Interfaces?.Count() > 0);
```
ArcturusZhang, line 86: So we should no longer need the check of `IncludeSerializer`.
In our implementation, if this is false, the interface.Count should be zero, therefore we could simplify here:
```suggestion
                    .AppendRawIf(" : ", serialization.Interfaces?.Any() ?? false);
```
`Any` should be faster than Count.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
ArcturusZhang: I get the purpose now.
So I regard this as a temporary solution, we have to keep two list of properties because we do not know which property comes from myself and which comes from my base.
This leads to an immediate conclusion that the serialization should have hierarchy.
The serialization object on myself should only have my properties, and it has a reference like `Base` just like what we have in types.
And now if the flag of using write core is turned off, we could go through all the bases to collect all the properties, and if it is turned on, I believe the only thing we would care when writing this model is "my properties".
This should cover both scenarios.

But this change of implementation could be complicated to do right now. We should track it with an issue and refactor it in the future if possible, or at least keep in mind when we write the azure plugin, we should not do the things in this PR again.

pshao25: I created an issue https://github.com/Azure/autorest.csharp/issues/5024

ArcturusZhang (approved):

Issue #3469, opened 2023-06-05
[TypeSpec] Add an option to force System.Text.Json converter to be applied to models


JoshLove-msft:
For swagger, we can apply the value of `converter` to `x-csharp-usage`, e.g. https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventgrid/Azure.Messaging.EventGrid/src/autorest.md#apply-converters-and-update-namespace-for-system-event-data-models.

This is not possible in Typespec and is a blocker for migrating the .NET Event Grid system events from swagger to Typespec.

ArcturusZhang:
Well we could do it by using the `CodeGenModel` attribute to append usages to a model.
But this feature is not implemented for DPG yet. My point here is that to support this, we do not need anything updated on typespec side or TCGC side, we could do it all on the dotnet side.

ArcturusZhang:
We now believes that converter should not belong to `usage` therefore in the typespec world we no longer support the `converter` usage.
But in the meantime, we will create a decorator in TCGC's csharp namespace to let you mark a specific model needs a JSON converter.
On the other hand, since we are going to introduce the public serialization story, the converter should no longer be introduced for new models - if we are trying to solve a backward compatibility issue, we could always do that using customized code, I believe it is the case here.

ArcturusZhang:
Hi @JoshLove-msft considering the existence of this [new public serialization feature](https://github.com/Azure/autorest.csharp/pull/3434), new libraries never need the converters any more.
And in your case, if you would like to migrate old libraries with converters (because you cannot remove the attribute otherwise breaking change), we could use customization code to add back the converters.

Therefore the generator will not implement it any more, and we plan to remove the usage `converter` when it is possible.

Close this as a "wont fix"

live1206:
we need to implement this in both autorest.csharp and microsoft/autorest

JoshLove-msft:
Any update on this? This is needed for the Event Grid migration. /cc @m-nash @lmazuel

live1206:
I think https://github.com/Azure/autorest.csharp/issues/4887 will resolve this

jsquire:
@ArthurMa1978, @m-nash:   Can we please prioritize this for delivery no later than the second week of August?

live1206:
`@hasJsonConverter` in csharp namespace

live1206:
[...]

PR #4954, opened 2024-08-01
Add decorators to InputModelType

live1206:
https://github.com/Azure/autorest.csharp/issues/3469


live1206:
pending on https://github.com/Azure/typespec-azure/pull/1230

src/AutoRest.CSharp/Common/Input/InputTypes/InputType.cs
```diff
@@ -36,6 +37,9 @@ internal InputType GetCollectionEquivalent(InputType inputType)
 
     //public bool IsNullable { get; init; }
     public string Name { get; internal set; }
+
+    public IReadOnlyList<InputDecoratorInfo> Decorators { get; internal set; } = new List<InputDecoratorInfo>();
```
chunyu3, line 41: `Decorators` is required property, shall we add this in constructor?

live1206: It is not needed in constructor since it is default to an empty list.
We only need deserialization to parse it from TCGC output.

ArthurMa1978 (approved):

chunyu3 (approved):

Issue #4948, opened 2024-07-31
[MPG] ManagedServiceIdentity can't be replace by the common mode in Azure.ResourceManager


ArthurMa1978:
In this [example](https://github.com/Azure/azure-rest-api-specs/blob/4d8b76eb4c8bcfda28784bc429b9624e5788f564/specification/devopsinfrastructure/Microsoft.DevOpsInfrastructure.Management/main.tsp#L38), in the generated SDK the `ManagedServiceIdentity` is not the one in `Azure.ResourceManager` but a new class in current namespace:
![image](https://github.com/user-attachments/assets/7c5b9f8f-234f-4a82-995d-5714754d5af6)

PR #4957, opened 2024-08-02
Replace ManagedServiceIdentity for TypeSpec input

live1206:
Resolves https://github.com/Azure/autorest.csharp/issues/4948


src/AutoRest.CSharp/Mgmt/Decorator/ReferenceTypePropertyChooser.cs
```diff
@@ -26,6 +26,12 @@ internal static class ReferenceTypePropertyChooser
         private static readonly Type _resourceIdentifierType = typeof(Azure.Core.ResourceIdentifier);
         private static readonly Type _resourceTypeType = typeof(Azure.Core.ResourceType);
 
+        private static readonly IReadOnlyDictionary<string, string> _replacements = new Dictionary<string, string>
+        {
+            { "Azure.ResourceManager.CommonTypes.UserAssignedIdentity", "Azure.ResourceManager.Models.UserAssignedIdentity" },
+            { "Azure.ResourceManager.CommonTypes.ManagedServiceIdentity", "Azure.ResourceManager.Models.ManagedServiceIdentity" },
+        };
```
ArcturusZhang, line 33: I think we should put this in `TypeFactory` and let it returns an instance of `SystemObjectType` from this identity when it finds the crossLanguageDefinitionId of these
We should not let the typespec input part use these "replace types by shape" logic.

live1206: updated

ArcturusZhang (approved):

Issue #4929, opened 2024-07-19
Support rename operation in typespec input for MPG


ArcturusZhang:
This may only apply to the generated extension methods, or maybe instance methods as well.

The generator will do this when it sees an operation to be generated into an extension class:
it gets the operation Id, it should look like `A_B`
it use `{B}{A.ToSingular()}` as the name of the generated extension method.

The `@clientName` changes `B` in the operationId, therefore no matter what we put in `@clientName` for an operation as its name, such as
```
@@clientName(theOperation, "C", "csharp");
```
We will get `CA` as the new generated method name, we will never get `C` as the name.

We need to fix this.

ArcturusZhang:
There is indeed something we could do: 

operationId: FabricCapacities_checkNameAvailability
Name: CheckFabricCapacityNameAvailability

if (name != operationId.Name)
{
  return name;
}
else
{
  return GetNameFromOperationId(operationId);
}

PR #4958, opened 2024-08-05
Add `crossLanguageDefinitionId` for operations

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4929

# Description

This PR adds `CrossLanguageDefinitionId` to `InputOperation`, and adds a `IsChanged` property to `InputOperation`

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
+++ b/src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
@@ -105,14 +105,14 @@ namespace AutoRest.CSharp.Common.Input
 
             static bool IsNameChanged(string crossLanguageDefinitionId, string name)
             {
-                var span = crossLanguageDefinitionId.AsSpan();
-                int lastDotIndex = span.LastIndexOf('.');
+                int lastDotIndex = crossLanguageDefinitionId.LastIndexOf('.');
                 if (lastDotIndex < 0)
                 {
-                    return false;
+                    throw new JsonException($"The crossLanguageDefinitionId of {crossLanguageDefinitionId} does not contain a dot in it, this should be impossible to happen");
                 }
+                var span = crossLanguageDefinitionId.AsSpan(lastDotIndex + 1);
                 var nameSpan = name.AsSpan();
-                if (nameSpan.Equals(span.Slice(lastDotIndex + 1), StringComparison.Ordinal))
+                if (nameSpan.Equals(span, StringComparison.Ordinal))
                 {
                     return false;
                 }
--- a/src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
+++ b/src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
@@ -93,12 +93,32 @@ namespace AutoRest.CSharp.Common.Input
             parameters = parameters ?? throw new JsonException("InputOperation must have parameters");
             responses = responses ?? throw new JsonException("InputOperation must have responses");
 
-            var inputOperation = new InputOperation(name, resourceName, summary, deprecated, description, accessibility, parameters, responses, httpMethod, requestBodyMediaType, uri, path, externalDocsUri, requestMediaTypes, bufferResponse, longRunning, paging, generateProtocolMethod, generateConvenienceMethod, crossLanguageDefinitionId, keepClientDefaultValue);
+            var inputOperation = new InputOperation(name, resourceName, summary, deprecated, description, accessibility, parameters, responses, httpMethod, requestBodyMediaType, uri, path, externalDocsUri, requestMediaTypes, bufferResponse, longRunning, paging, generateProtocolMethod, generateConvenienceMethod, crossLanguageDefinitionId, keepClientDefaultValue)
+            {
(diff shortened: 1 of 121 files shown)


src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
```diff
@@ -85,17 +87,38 @@ public static InputOperation CreateOperation(ref Utf8JsonReader reader, string?
...
             {
                 resolver.AddReference(id, inputOperation);
             }
             return inputOperation;
+
+            static bool IsNameChanged(string crossLanguageDefinitionId, string name)
+            {
+                var span = crossLanguageDefinitionId.AsSpan();
+                int lastDotIndex = span.LastIndexOf('.');
+                if (lastDotIndex < 0)
```
chunyu3, line 110: crosssLanguageDefinitionId of an operation should be {service-name-space}.{operation-name}. So it seems that this condition should not exist. Right? If the scenario exists, and we filter out this, we may need to change the method name `IsNameChanged` because in this secario, even the name is changed, it will return `false`. Then the method name is not consistent with its real functionality.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
```diff
@@ -85,17 +87,38 @@ public static InputOperation CreateOperation(ref Utf8JsonReader reader, string?
...
+            };
             if (id != null)
             {
                 resolver.AddReference(id, inputOperation);
             }
             return inputOperation;
+
+            static bool IsNameChanged(string crossLanguageDefinitionId, string name)
+            {
+                var span = crossLanguageDefinitionId.AsSpan();
```
chunyu3, line 108: Maybe we can get the IndexOf (.) first, then create the span and then check equal
```
int lastDotIndex = crossLanguageDefinitionId.LastIndexOf('.');
if (lastDotIndex > 0) {
  var idSpan = crossLanguageDefinitionId.AsSpan(lastDotIndex + 1);
  var nameSpan = name.AsSpan();
 if (nameSpan.Equals(idSpan, StringComparison.Ordinal))
  {
           return false;
   } else {
          return true;
   }
   
   return false;
}
```

archerzz: I used to get comments in analyzers form azure core team. They ask for `Span` v.s. `String` for performance consideration. I guess here we use `Span` for similar reason.

chunyu3: yes. we should use span to get better performance. I write the code above may save a copy of string.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
```diff
@@ -85,17 +87,38 @@ public static InputOperation CreateOperation(ref Utf8JsonReader reader, string?
...
             uri = uri ?? throw new JsonException("InputOperation must have uri");
             path = path ?? throw new JsonException("InputOperation must have path");
             parameters = parameters ?? throw new JsonException("InputOperation must have parameters");
             responses = responses ?? throw new JsonException("InputOperation must have responses");
 
-            var inputOperation = new InputOperation(name, resourceName, summary, deprecated, description, accessibility, parameters, responses, httpMethod, requestBodyMediaType, uri, path, externalDocsUri, requestMediaTypes, bufferResponse, longRunning, paging, generateProtocolMethod, generateConvenienceMethod, keepClientDefaultValue);
+            var inputOperation = new InputOperation(name, resourceName, summary, deprecated, description, accessibility, parameters, responses, httpMethod, requestBodyMediaType, uri, path, externalDocsUri, requestMediaTypes, bufferResponse, longRunning, paging, generateProtocolMethod, generateConvenienceMethod, crossLanguageDefinitionId, keepClientDefaultValue)
+            {
+                IsNameChanged = IsNameChanged(crossLanguageDefinitionId, name)
```
archerzz, line 98: For Swagger input, `IsNameChanged` is always `false`. But we know we can change the name in other way. Are we diverse here?

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
ArcturusZhang: Let me add an assertion here to throw when it does not have the dot.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputOperationConverter.cs
ArcturusZhang: Yes and I think it is impossible to unify those because typespec will never have those configurations that allow us to override the operation name.

ArcturusZhang: I think we could merge these two ways of renaming operations when we fully moved to MGC by changing how it works on the swagger side

chunyu3 (approved):

PR #4959, opened 2024-08-05
Adopt examples from TCGC

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4980
Fixes https://github.com/Azure/autorest.csharp/issues/5006

# Description

To adopt the example information from TCGC, we are taking the following steps, and this PR is the second step (first step is done in the [typespec repo PR](https://github.com/microsoft/typespec/pull/4086)).

1. Update the emitter so that the emitter could spit out the examples from TCGC.
2. Refine our code in DPG so that the sample generator made for DPG is able to write out those examples, as if they are the mock examples.
3. Refine our code in MPG so that the examples for mgmt could be written when the input is typespec.
4. Possible follow ups: use the responses, now the responses in examples are not used.

# Checklist

# Ready to Land?


archerzz (approved):

samples/AnomalyDetector/src/Generated/Docs/Multivariate.xml
```diff
@@ -5,190 +5,82 @@
...
-]]></code>
-This sample shows how to call TrainMultivariateModelAsync with all parameters.
-<code><![CDATA[
-Uri endpoint = new Uri("<https://my-service.azure.com>");
+Uri endpoint = new Uri("<endpoint>");
 AzureKeyCredential credential = new AzureKeyCredential("<key>");
 Multivariate client = new AnomalyDetectorClient(endpoint, credential).GetMultivariateClient();
 
-ModelInfo modelInfo = new ModelInfo("<dataSource>", DateTimeOffset.Parse("2022-05-10T18:57:31.2311892Z"), DateTimeOffset.Parse("2022-05-10T18:57:31.2311892Z"))
+ModelInfo modelInfo = new ModelInfo("https://multiadsample.blob.core.windows.net/data/sample_data_2_1000.csv", DateTimeOffset.Parse("2019-04-01T00:00:00Z"), DateTimeOffset.Parse("2019-04-02T00:00:00Z"))
```
lirenhe, line 74: I expect there are a lot of changes of generated XML docs for protocol methods in regen, but I didn't find any change in in the regen PR, is this expected

ArcturusZhang: TCGC will not read the examples unless this configuration is set: https://github.com/Azure/azure-rest-api-specs/blob/ba268c5654f002bc17914e0bc157f19529f1fec1/specification/communication/Communication.JobRouter/tspconfig.yaml#L30

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -72,6 +72,14 @@ export async function $onEmit(context: EmitContext<AzureNetEmitterOptions>) {
                 : undefined);
         configurations["enable-internal-raw-data"] =
             options["enable-internal-raw-data"];
+        const examplesDir =
+            options["examples-dir"] ?? options["examples-directory"];
```
live1206, line 76: why is the fallback needed? due to legacy config?

src/AutoRest.CSharp/Common/Input/InputTypes/Examples/InputOperationExample.cs
```diff
@@ -6,5 +6,5 @@
 namespace AutoRest.CSharp.Common.Input.Examples
 {
     // TODO -- currently we do not need the responses. In the case that in the future we need to handle the responses in examples, we need to add a new class InputExampleResponse and a new property here
-    internal record InputOperationExample(InputOperation Operation, IReadOnlyList<InputParameterExample> Parameters, string? Name = null, string? ExampleFilePath = null);
+    internal record InputOperationExample(string Name, string? Description, string FilePath, IReadOnlyList<InputParameterExample> Parameters);
```
live1206, line 9: why does `InputOperationExample` not have an operation?
And why is it better to set it directly in `MockTestCase`?

live1206: I see why you are doing this now, align with TCGC output.

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
```diff
@@ -115,17 +117,24 @@ public static InputOperation RemoveApiVersionParam(InputOperation operation)
...
+    private IReadOnlyList<InputOperationExample>? _examples;
+    public IReadOnlyList<InputOperationExample> Examples => _examples ??= EnsureExamples();
+
+    private IReadOnlyList<InputOperationExample> EnsureExamples()
     {
-        examples[ExampleMockValueBuilder.ShortVersionMockExampleKey] = ExampleMockValueBuilder.BuildOperationExample(this, false);
-        examples[ExampleMockValueBuilder.MockExampleAllParameterKey] = ExampleMockValueBuilder.BuildOperationExample(this, true);
-        return examples;
+        // see if we need to generate the mock examples
+        if (Configuration.ExamplesDirectory != null)
```
live1206, line 127: Do we need the mock examples at all?
If example-dir is not specified in config, then no examples output from TCGC. Is that intended to not generate samples?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
ArcturusZhang: yeah TCGC now has two configs  - directory and dir, they want to keep the dir version to align some existing config names, and the directory version will be removed.
But since now we do have two configs, we have to read them both.

src/AutoRest.CSharp/Common/Input/InputTypes/Examples/InputOperationExample.cs
ArcturusZhang: yeah it is because TCGC does not have it.
It might be bad practice to have it from the json's perspective because the InputOperation holds examples and examples hold the operation itself.

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
ArcturusZhang: This is about:
when we generate the mock values?
For the operations with real example, we certainly do not want to generate the mock examples for it any more, I think everyone would agree this.
For those operations without real examples, do we need to generate mock values for it, or just leave it empty without an example?
My current implementation chooses the second solution.
Only when there is no real example configured, we will generate examples with mock values for all the operations.

Because we do not really have anything to indicate whether the example on an operation has mock value or real value, I do not really want to mix them up.

samples/AnomalyDetector/src/Generated/Docs/Multivariate.xml
ArcturusZhang: I enabled one of them to have the real examples: https://github.com/Azure/azure-sdk-for-net/pull/45614

src/AutoRest.CSharp/Common/Input/InputTypes/InputOperation.cs
live1206: I think there is another case: there is example in spec but without example configuration, shall we just add the config instead of generating mock examples?

ArcturusZhang: I am not sure how this would relate to the case I am describing because I am describing the case that we have example specs.
For the generator, it cannot know if there is example spec unless there is a configuration telling it where it is.

ArcturusZhang: Therefore this piece is saying:
if there is an example spec configured (no matter if there is really one, that is not our fault if there is actually none but still configured a path), we do not generate the mocking examples any more.

live1206: Yes, got your point.
The case I was describing is, the spec has examples in it but forgot to enable the examples configuration.
For this case, we should just enable the examples configuration instead of generating mock examples.
So, that's why I was wondering maybe we should get rid of the mock generation totally.
Anyhow, let's sort it out in https://github.com/Azure/autorest.csharp/issues/5011 later.

PR #4965, opened 2024-08-07
remove mockapi

mcgallan:
Remove CadlRanchMockApis folder in test. 

Fix https://github.com/Azure/autorest.csharp/issues/4865


pshao25:
Looks good. Make sure our regen pipeline won't break by https://github.com/Azure/autorest.csharp?tab=readme-ov-file#preview-changes-in-azure-sdk-for-net

pshao25 (approved):

Issue #4969, opened 2024-08-08
`TypeSpecAdditionalOptions` has been nullified since `tsp-client` update


ArcturusZhang:
We used to be able to add some extra options via `/p:TypeSpecAdditionalOptions` to inject some extra options to the generation script for debugging purpose, but this has been accidentally removed after the adoption of tsp-client: 
https://github.com/Azure/autorest.csharp/blob/61fa2f8d6e3784700a6b8ed1d0cf6d8c33e62c21/src/AutoRest.CSharp/build/CodeGeneration.targets#L65

This feature is documented here: 
https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#debugging-the-projects-in-the-azure-sdk-for-net-repo

PR #4970, opened 2024-08-08
add back the additional options

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4969
Fixes https://github.com/Azure/autorest.csharp/issues/5013

# Description

Last time when we adopt `tsp-client`, we accidentally removed the `TypeSpecAdditionalOptions` parameter. This PR adds it back.
But due to the existing issue in msbuild system, the semicolon `;` cannot be passed through the command line in msbuild, therefore it is impossible that you could pass two different options into `tsp-client`'s command line option to enable debug for the generator, and use a custom dll of the generator in the same time.
I think there is no way to resolve this, therefore I updated the readme to include a useful alternative to debug a typespec project.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/build/CodeGeneration.targets
```diff
@@ -36,7 +36,7 @@
     <!-- Here we append the generate-test-project configuration to TypespecAdditionalOptions if it is specified -->
     <TypespecAdditionalOptions Condition="'$(GenerateTestProject)' != '' AND '$(TypespecAdditionalOptions)' != ''">$(TypespecAdditionalOptions)%3Bgenerate-test-project=true</TypespecAdditionalOptions>
     <TypespecAdditionalOptions Condition="'$(GenerateTestProject)' != '' AND '$(TypespecAdditionalOptions)' == ''">generate-test-project=true</TypespecAdditionalOptions>
-    <_TypespecAdditionalOptions Condition="'$(TypespecAdditionalOptions)' != ''">-TypespecAdditionalOptions "$(TypespecAdditionalOptions)"</_TypespecAdditionalOptions>
+    <_TypespecAdditionalOptions Condition="'$(TypespecAdditionalOptions)' != ''">--emitter-options "$(TypespecAdditionalOptions)"</_TypespecAdditionalOptions>
     <_LocalSpecRepo Condition="'$(LocalSpecRepo)' != ''">--local-spec-repo $(LocalSpecRepo)</_LocalSpecRepo>
     <_Debug Condition="'$(Debug)' == 'true'">"--debug"</_Debug>
```
ArcturusZhang, line 41: So this debug flag is only debugging the tsp-client, tsp-client will not pass this through to our emitter.
Therefore this is a false advertisement - this is never trying to debugging our generator.

readme.md
```diff
@@ -241,15 +241,14 @@ and then run `dotnet build /t:GenerateCode` and it will attach a debugger to the
...
+dotnet build /t:GenerateCode /p:Debug=true
 ```
 to attach a debugger to the plugin process.
 
 If you are trying to use a locally built generator `dll` and debug it, you should run
 ```powershell
-dotnet build /t:GenerateCode /p:typespecAdditionalOptions="csharpGeneratorPath=/absolute/path/to/AutoRest.CSharp.dll%3Bdebug=true"
+dotnet build /t:GenerateCode /p:Debug=true /p:typespecAdditionalOptions="csharpGeneratorPath=/absolute/path/to/AutoRest.CSharp.dll"
 ```
-where you must use `%3B` as escape of the semicolon to append multiple options.
```
ArcturusZhang, line 252: Because of [this](https://github.com/Azure/autorest.csharp/pull/4970/files#r1708509245), we need to revert this because we still have to use the old way to debug the generator.

src/AutoRest.CSharp/build/CodeGeneration.targets
chunyu3: We may change it to debug emitter when $(Debug) is set.

<_Debug Condition="'$(Debug)' == 'true'">"--emitter-options debug=true"</_Debug>

chunyu3: We may need to merge this with `TypespecAdditionalOptions`
Or we remove this to not provide this flag, and ask use to set in `TypespecAdditionalOptions`

ArcturusZhang: > we remove this to not provide this flag, and ask use to set in TypespecAdditionalOptions

This is the behavior we used to have.

chunyu3 (approved):

archerzz (approved):

Issue #4949, opened 2024-07-31
[MPG] Generate wrong property name when the property name in TSP is `url`


ArthurMa1978:
In this [example](https://github.com/Azure/azure-rest-api-specs/blob/4d8b76eb4c8bcfda28784bc429b9624e5788f564/specification/devopsinfrastructure/Microsoft.DevOpsInfrastructure.Management/main.tsp#L103), the property name in the generated .Net SDK is weird:
![image](https://github.com/user-attachments/assets/7a292848-84e4-426b-ab14-0b65c3b9066e)

This can be fixed by renaming in client.tsp
![image](https://github.com/user-attachments/assets/cffaa3f8-793d-4006-8b72-6e8f53aab850)

So guessing this is a codegen bug.

ArcturusZhang:
This is showing up here as well: https://github.com/Azure/azure-sdk-for-net/pull/45242
![image](https://github.com/user-attachments/assets/3db4a015-b56d-4cdf-b3cc-d5148248aa99)

PR #4972, opened 2024-08-08
fix the property name issue when it is url

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4949

# Description

# Checklist

# Ready to Land?


live1206 (approved):
Could you update https://github.com/Azure/autorest.csharp/tree/feature/v3/test/TestProjects/MgmtTypeSpec to include this case?

src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputUrlToUriTransformer.cs
```diff
@@ -2,38 +2,58 @@
...
+            if (span[^3] == u && span[^2] == r && span[^1] == l)
+            {
+                Span<char> newSpan = span.ToArray();
+                newSpan[^1] = i;
+
+                newName = new string(newSpan);
+                return true;
+            }
+
+            return false;
```
ArcturusZhang, line 56: comparing with the previous implementation, this has better performance:
alt text

src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputUrlToUriTransformer.cs
```diff
@@ -2,38 +2,58 @@
...
+            const char l = 'l';
+            newName = null;
+            var span = name.AsSpan();
+            if (span.Length < 3)
+            {
+                return false;
+            }
+
+            // check if this ends with `Url`
+            if (span[^3] == u && span[^2] == r && span[^1] == l)
```
live1206, line 47: Can we still use the existing API `span.EndsWith(UrlSuffix)`? I would prefer not to rewrite the functionality if the framework already supports it. 
We can remove the `StringComparison` if not needed.

ArcturusZhang: I do not think we should remove `StringComparison`, I do not know how it is implemented but with this parameter, it is a specific overload for `ReadOnlySpan<char>` (others are taking a `T`) maybe there are optimizations only for strings performed, therefore this might be better.

ArcturusZhang: And in general, "generic code" has lower performance than "specific code".
The code I wrote here is only for checking "Url" endings which has a better performance as I showed above in the other comment.
But since the difference is minor, I changed it back.

Issue #4946, opened 2024-07-30
Compile failure caused by duplicated definitions created under same namespace


raych1:
**Describe the issue or request**
The generated SDK project fails to build with the following errors:
```powershell
cmdout 	[DotNet-Sdk-Automation]   Restored /mnt/vss/_work/1/s/azure-sdk-for-net/sdk/computefleet/Azure.ResourceManager.ComputeFleet/src/Azure.ResourceManager.ComputeFleet.csproj (in 1.91 sec).
07:44:01.648 cmdout 	[DotNet-Sdk-Automation] /mnt/vss/_work/1/s/azure-sdk-for-net/sdk/resourcemanager/Azure.ResourceManager/src/Shared/SharedExtensions.cs(17,27): error CS0101: The namespace 'Azure.ResourceManager' already contains a definition for 'SharedExtensions' [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/computefleet/Azure.ResourceManager.ComputeFleet/src/Azure.ResourceManager.ComputeFleet.csproj::TargetFramework=netstandard2.0]
07:44:01.649 cmdout 	[DotNet-Sdk-Automation] /mnt/vss/_work/1/s/azure-sdk-for-net/sdk/resourcemanager/Azure.ResourceManager/src/Shared/ManagedServiceIdentityTypeV3Converter.cs(13,20): error CS0101: The namespace 'Azure.ResourceManager.Models' already contains a definition for 'ManagedServiceIdentityTypeV3Converter' [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/computefleet/Azure.ResourceManager.ComputeFleet/src/Azure.ResourceManager.ComputeFleet.csproj::TargetFramework=netstandard2.0]
07:44:01.649 cmdout 	[DotNet-Sdk-Automation] /mnt/vss/_work/1/s/azure-sdk-for-net/sdk/resourcemanager/Azure.ResourceManager/src/Shared/ManagedServiceIdentityTypeV3Converter.cs(21,30): error CS0111: Type 'ManagedServiceIdentityTypeV3Converter' already defines a member called 'Write' with the same parameter types [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/computefleet/Azure.ResourceManager.ComputeFleet/src/Azure.ResourceManager.ComputeFleet.csproj::TargetFramework=netstandard2.0]
[...]

PR #4974, opened 2024-08-08
remove packageReference and shared code for mgmt sdk

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/4946

Problem-to-solve: build fail because there are duplicated definition
When generate a new service from typespec, the .NET codegen will generate scaffolding (include project file) for this new SDK. In the generated project file it includes several packageReference and shared code to add compile dependency. But in azure-sdk-for-net, the engsys already include all depended packages and shared code for management-plane SDK in Directory.Build.Common.targets. So the build will fail because of duplicate compile items.

Solution:
Since existing SDKs in azure-sdk-for-net contain define those packageRefernce and shared code in each SDK's project file, and we also need to support management SDKs from swagger which use the dotnet templates to create scaffolding folder, we decide to change in codegen side to remove packageReference when create scaffolding which will keep consistent with current SDKs.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
```diff
@@ -399,9 +398,12 @@ private string GetBrandedTestCSProj()
             }
             writer.ProjectReferences.Add(new($"..\\src\\{Configuration.Namespace}.csproj"));
             // add the package references
-            foreach (var package in _brandedTestDependencyPackages)
+            if (!Configuration.AzureArm)
```
live1206, line 401: If we take a look at https://github.com/Azure/azure-sdk-for-net/blob/f67a51ae48ae0628733b1cf207d6fec4b0db12ac/eng/Directory.Build.Common.targets#L262-L268, there is no `Azure.Identity`.
So, for a new branded test project will it be missing `Azure.Identity` if we don't add it to test csproj for mgmt?

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -674,6 +674,10 @@ public virtual ObjectTypeProperty GetPropertyForSchemaProperty(InputModelPropert
 
         public ObjectTypeProperty GetPropertyBySerializedName(string serializedName, bool includeParents = false)
         {
+            if (DefaultName == "SavingsPlanModelListResult")
```
live1206, line 677: what is this change for?

Directory.Build.targets
```diff
@@ -67,14 +67,24 @@
...
     <PackageReference Include="Azure.ResourceManager" />
 
     <!-- TODO: Review these file references-->
     <Compile Include="$(AzureCoreSharedSources)AzureResourceProviderNamespaceAttribute.cs" LinkBase="Shared" />
+    <!-- Include management shared code-->
+    <Compile Include="$(MSBuildThisFileDirectory)/src/assets/Management.Shared/**/*.cs" Link="Shared/Management/%(RecursiveDir)/%(Filename)%(Extension)" />
+  </ItemGroup>
+
+  <ItemGroup Condition ="'$(IsMgmtLibrary)' == 'true' and '$(IsTestGenerationTestProject)' == 'true'">
+    <PackageReference Include="Azure.Identity" />
```
live1206, line 83: As we can see in sdk-for-net, this is not needed for most test projects, and it is not included in the test needed sharing dependencies in https://github.com/Azure/azure-sdk-for-net/blob/f67a51ae48ae0628733b1cf207d6fec4b0db12ac/eng/Directory.Build.Common.targets#L262-L268.
And when this is needed, we can just add it manually like https://github.com/Azure/azure-sdk-for-net/blob/f67a51ae48ae0628733b1cf207d6fec4b0db12ac/sdk/compute/Azure.ResourceManager.Compute/tests/Azure.ResourceManager.Compute.Tests.csproj#L7
Shall we just remove it for mgmt test projects? it might just work.

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
live1206: since we have test project dependencies for mgmt in azure-sdk-for-net, we should remvoe them here.
Please check another comment regarding `Azure.Identity`.

src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
chunyu3: removed

Directory.Build.targets
chunyu3: This is for test SDKs in autorest.csharp. We will regen project files for them every time. So we cannot manually add `Azure.Identity`. It is only autoest.csharp test projects. it is ok to add more project dependency. But for sdks in azure-sdk-for-net, we will not add `Azure.Identity` by default, they need to manually add if needed.

src/AutoRest.CSharp/Common/AutoRest/Plugins/NewProjectScaffolding.cs
chunyu3: `Azure.Identity` is not always needed for SDK test projects. So in scaffolding, we will not add it by default. When a special SDK need this, it can manually added.

Directory.Build.targets
```diff
@@ -67,14 +67,24 @@
...
     <PackageReference Include="System.ClientModel" />
     <PackageReference Include="Azure.ResourceManager" />
 
     <!-- TODO: Review these file references-->
     <Compile Include="$(AzureCoreSharedSources)AzureResourceProviderNamespaceAttribute.cs" LinkBase="Shared" />
+    <!-- Include management shared code-->
+    <Compile Include="$(MSBuildThisFileDirectory)/src/assets/Management.Shared/**/*.cs" Link="Shared/Management/%(RecursiveDir)/%(Filename)%(Extension)" />
+  </ItemGroup>
+
+  <ItemGroup Condition ="'$(IsMgmtLibrary)' == 'true' and '$(IsTestGenerationTestProject)' == 'true'">
```
archerzz, line 82: Looks like for non-test projects of mgmt library, we don't need to add packages?

Directory.Build.targets
live1206: OK, sounds reasonable to me

live1206 (approved):

Directory.Build.targets
chunyu3: hello @archerzz above this, we will add dependent packages for src-project of mgmt SDKs.

ArcturusZhang (commented):
LGTM if the CIs could pass.

Issue #4975, opened 2024-08-08
[defect]System.InvalidOperationException: Unable to find object property with serialized name 'value' in schema SavingsPlanModelListResult


chunyu3:
When generate `billing\Azure.ResourceManager.Billing` sdk (target package: package-2024-04), the generate fail.
error:

```
fatal   | System.InvalidOperationException: Unable to find object property with serialized name 'value' in schema SavingsPlanModelListResult
     at AutoRest.CSharp.Output.Models.Types.SchemaObjectType.GetPropertyBySerializedName(String serializedName, Boolean includeParents) in D:\a\_work\1\s\autorest.
csharp\src\AutoRest.CSharp\Common\Output\Models\Types\SchemaObjectType.cs:line 679
```

ArcturusZhang:
why this is flagged with dpg when the issue happens to a mgmt library?

chunyu3:
> t set `includeParents=false`

Actually it is in common when handle PagingResponse, the `ModelTypeProvider` also ignore parents. DPG should also has this problem also. Need investigation.

PR #4984, opened 2024-08-09
include parents when search property

chunyu3:
# Description
Fix https://github.com/Azure/autorest.csharp/issues/4975

When get property of an paging response by serialize name, it does not search in parent model which cause the property cannot be found.
Solution: search property in parent model also.
# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Output/Models/Requests/PagingResponseInfo.cs
```diff
@@ -31,18 +31,18 @@ public PagingResponseInfo(string? nextLinkName, string? itemName, CSharpType typ
         public CSharpType PageType => new CSharpType(typeof(Page<>), ItemType);
         public CSharpType ItemType { get; }
 
-        private ObjectTypeProperty GetPropertyBySerializedName(CSharpType type, string name)
+        private ObjectTypeProperty GetPropertyBySerializedName(CSharpType type, string name, bool includeParents = false)
```
ArcturusZhang, line 34: since the only place we are using this **private** method is passing in a `true` for `includeParents`, and I think in this context, we should always include parents - so I think in the future there is never a chance that we will need this new flag.
Therefore I think we do not need to add this, just passing true below.

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Output/Models/Requests/PagingResponseInfo.cs
chunyu3: Thanks. Updated

PR #4985, opened 2024-08-12
Refactor how we get client parameters in examples

ArcturusZhang:
# Description

The examples from TCGC never have these models specially made for client parameters.
The operation always has all its necessary parameters including client parameters, therefore this `InputClientExample` is 100% redundant.

# Checklist

# Ready to Land?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "name": "@autorest/csharp",
       "devDependencies": {
-        "@azure-tools/cadl-ranch-mockapis": "file:test/CadlRanchMockApis",
+        "@azure-tools/cadl-ranch-specs": "0.34.9",
         "@azure-tools/typespec-client-generator-core": "0.44.3",
         "@azure-tools/typespec-csharp": "file:src/TypeSpec.Extension/Emitter.Csharp",
         "@azure-tools/unbranded-tests": "file:test/UnbrandedProjects",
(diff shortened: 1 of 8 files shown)


chunyu3:
How about uri parameter (e.g. {endpoint}\{clientName}, the uri has two parameters which is client parameter)? The generated sample/test, should include this parameter when constructor client.

ArcturusZhang:
> How about uri parameter (e.g. {endpoint}{clientName}, the uri has two parameters which is client parameter)? The generated sample/test, should include this parameter when constructor client.
> 
> The same as authentication parameter.
> 
> Seems current our generated sample code does not correct construct client.

The focus of this PR is not fixing any existing bug. If this is a bug, please open an issue for it and attach details so that we could fix it.
This is only a refactor, nothing changed is our goal. And nothing is changed in test project as well as the regen.
All things changed in regen comes from previous PRs, it exists there for a few days.

ArcturusZhang:
> How about uri parameter (e.g. {endpoint}{clientName}, the uri has two parameters which is client parameter)? The generated sample/test, should include this parameter when constructor client.
> 
> The same as authentication parameter.
> 
> Seems current our generated sample code does not correct construct client.

and also as I described in the description, TCGC does not have the categories of "client parameter" or "method parameter", an operation ALWAYS has all its parameters no matter they are client or method.

live1206 (approved):

archerzz (approved):

PR #4988, opened 2024-08-12
Update to .NET 8

m-redding:
Follow up to updating to .NET 8 in azure-sdk-for-net


Directory.Build.props
```diff
@@ -37,7 +37,7 @@
     <ArtifactsObjDir>$(ArtifactsDir)obj\</ArtifactsObjDir>
     <ArtifactsBinDir>$(ArtifactsDir)bin\</ArtifactsBinDir>
     <ArtifactsPackagesDir>$(ArtifactsDir)packages\$(Configuration)\</ArtifactsPackagesDir>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors> <!-- TEMPORARY -->
```
m-redding, line 40: I don't have enough context about the code to fix the new warnings in .NET 8. It should be straightforward for someone familiar with the repo to change this back and fix the warnings in the code.

src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs
```diff
@@ -547,7 +547,11 @@ private static PropertyBag<ObjectTypeProperty> PopulatePropertyBag(SerializableO
             ObjectTypeProperty? additionalPropertiesProperty = null;
             ObjectTypeProperty? rawDataField = null;
 
-            var inheritedDictionarySchema = inputModel.GetSelfAndBaseModels().OfType<InputDictionaryType>().FirstOrDefault();
+            // TODO - The following warning is being reported on L554:
```
m-redding, line 550: I'm not sure how to fix this one!

ArcturusZhang: Hi since we are more familiar with the code in this repo, how about we take over the task of updating this to dotnet 8?

ArcturusZhang: Passing a null might not be our initial intention here, but this is truly how this code would produce.
I am trying to fix this to be correct in my PR.

PR #5004, opened 2024-08-16
bump the langver so that we could use the fancy stuffs from dotnet 8

ArcturusZhang:
# Description

I should have updated this in the PR that updates to dotnet 8.

# Checklist

# Ready to Land?

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -63,7 +63,7 @@
 
   <PropertyGroup>
     <ImportDefaultReferences>false</ImportDefaultReferences>
-    <LangVersion>11.0</LangVersion>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
 </Project>


ArthurMa1978 (approved):

Directory.Build.props
```diff
@@ -63,7 +63,7 @@
 
   <PropertyGroup>
     <ImportDefaultReferences>false</ImportDefaultReferences>
-    <LangVersion>11.0</LangVersion>
+    <LangVersion>latest</LangVersion>
```
live1206, line 66: shall we ping it to a specific version instead of `latest` like before? so that, there will not be surprise later when we bump the .net version and we can control when to bump the language version.

ArcturusZhang: in general we should always want the latest language version, and we will never get a new one unless we bump the version of dotnet SDK. I do not think this is a thing we want to control.
MGC is doing this already: https://github.com/microsoft/typespec/blob/71c221d4d91da2e0dc40d114e7455ea524dafa3a/packages/http-client-csharp/generator/Directory.Build.props#L46

ArcturusZhang: and language version update will always be incremental.

live1206 (approved):

archerzz (approved):

PR #5031, opened 2024-09-02
Support allowReserved

### LINKED ISSUE #4966: Support AllowReserved in Path Implementation

pshao25:
Fix #4966 

Generator already support urlencoding, we just need to pass the flag from spec.


archerzz (approved):

test/CadlRanchProjects/azure/client-generator-core/flatten-property/Configuration.json
ArcturusZhang, line 1: why we have this project now? I did not see any change in your PR that changes the list of test projects

ArcturusZhang:
Also I see this PR only contains `tspCodeModel.json` changes.
So did our generator already support this feature (I assume so)? Therefore the only work left is to let the emitter could pass this flag through so that our generator could handle it?
Please add some more description about this in this PR

test/CadlRanchProjects/azure/client-generator-core/flatten-property/Configuration.json
pshao25: Some mistake coming from cadl-ranch breaking change. Deleted.

ArcturusZhang (approved):

Issue #4810, opened 2024-06-11
emit client parameters in InputClient.Parameters, not progagate those parameters to each operatio


chunyu3:
Some client parameters such as `host` is propagated into each operation, we need to move those into `InputClient.Paramters` property. Don't propagate those to operation.

PR #5044, opened 2024-09-04
don't propagate client parameter

chunyu3:
# Description
Sync https://github.com/microsoft/typespec/pull/4295
Fix https://github.com/Azure/autorest.csharp/issues/4810

We will not propagate client parameters (such as endpoint, or the parameter in the uri) to each operation, so

- When create client fields we need to merge the client parameters with the parameters in each operation whose kind is `Client` (means the operation parameter is bumped to client level).
- When generate operation method, merge client parameters with operation parameters, we need client parameters to construe uri.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/LowLevel/Output/Samples/DpgOperationSample.cs
```diff
@@ -138,7 +153,28 @@ private Dictionary<string, InputExampleParameterValue> EnsureParameterValueMappi
                     // if this is a required parameter and we did not find the corresponding parameter in the examples, we put the null
                     if (!parameter.IsOptionalInSignature)
                     {
-                        result.Add(parameter.Name, new InputExampleParameterValue(parameter, Null.CastTo(parameter.Type)));
+                        /*
+                         * TODO: remove this when example can contain client parameter value. Following workaround will not needed.
+                         */
+                        ValueExpression parameterExpression = parameter.Type.IsValueType && !parameter.Type.IsNullable ? Default.CastTo(parameter.Type) : Null.CastTo(parameter.Type);
```
ArcturusZhang, line 159: how about we simplify this just write the `default`?
`default` also works for reference types.

src/AutoRest.CSharp/LowLevel/Output/Samples/DpgOperationSample.cs
```diff
@@ -138,7 +153,28 @@ private Dictionary<string, InputExampleParameterValue> EnsureParameterValueMappi
...
+                                }
+                                else
+                                {
+                                    result.Add(parameter.Name, new InputExampleParameterValue(parameter, parameterExpression));
+                                }
+                                break;
+                            default:
+                                result.Add(parameter.Name, new InputExampleParameterValue(parameter, parameterExpression));
+                                break;
+                        }
```
ArcturusZhang, line 177: here is about "this is a required parameter, but we did not find its example value.
Why we have to need these extra stuff for different type kind of the parameters?
I saw we are making up some values, I think we should not do that even if this is a workaround - the example value should have those.

src/AutoRest.CSharp/LowLevel/Output/Samples/DpgOperationSample.cs
```diff
@@ -302,7 +338,7 @@ public InputExampleValue GetEndpointValue(string parameterName)
                 return operationParameterValue;
 
             // sometimes, especially in swagger projects, the parameter used as endpoint in our client, does not have the `IsEndpoint` flag, we have to fallback here so that we could at least have a value for it.
-            return InputExampleValue.Value(InputPrimitiveType.String, $"<{parameterName}>");
+            return InputExampleValue.Value(InputPrimitiveType.Url, Configuration.ApiTypes.EndPointSampleValue);
```
ArcturusZhang, line 341: this change is introducing the changes in samples.
Could we revert this in this PR to reduce the changes?
If we find this is a good change, we could do it in the next PR so that the changes in regen contents are clear.

src/AutoRest.CSharp/LowLevel/Output/Samples/DpgOperationSample.cs
ArcturusZhang: I am also curious how this snippet will take effect in the generated samples, but the samples have too many changes introduced by https://github.com/Azure/autorest.csharp/pull/5044#discussion_r1825585494
I think it should be better if we revert the change in [this comment](https://github.com/Azure/autorest.csharp/pull/5044#discussion_r1825585494) and introduce it back later if necessary, then we could see how it would change if we do not have this snippet.

chunyu3: Add this is to handle `Enum` type, otherwise the build will fail when just set `null`.

ArcturusZhang: This conflicts with the responsibility of this class - this class should be making things up any more.
I think this might be something related with TCGC, it does not include the client parameters in the example's parameters, and this may cause a change in the client structure cadl ranch cases.
But there is a dedicated effort to achieve "data and logic separation" in sample generation. We should not mix things up again.
I would suggest: 
either we do not introduce this workaround, let those sample to have that incorrect value `null` or `default`, creating an issue to track this, and if the root cause is in TCGC, we could make a fix there,
or if we insist to introduce the workaround, we should introduce it in a place where the data was prepared, not in this class.

chunyu3: remove the workaround, and now set `null` or `default` to those parameters which has no example value

test/CadlRanchProjects/client/structure/renamed-operation/tests/Generated/Samples/Samples_Group.cs
```diff
@@ -32,8 +32,8 @@ public void Example_Group_RenamedTwo_ShortVersion()
         [Ignore("Only validating compilation of examples")]
         public async Task Example_Group_RenamedTwo_ShortVersion_Async()
         {
-            Uri endpoint = new Uri("<https://my-service.azure.com>");
-            Group client = new RenamedOperationClient(endpoint, ClientType.Default).GetGroupClient();
+            Uri endpoint = new Uri("<endpoint>");
+            Group client = new RenamedOperationClient(endpoint, default).GetGroupClient();
```
chunyu3, line 36: This change because the clientType is client parameter, and it is propagate to operation, so there is no mocked value for it.

src/AutoRest.CSharp/LowLevel/Output/Samples/DpgOperationSample.cs
```diff
@@ -81,6 +81,21 @@ private IReadOnlyList<MethodSignatureBase> GetClientInvocationChain()
 
             return callChain.ToArray();
         }
+        /// <summary>
+        /// Get all the parameters to get the client.
+        /// </summary>
+        /// <returns></returns>
+        private IReadOnlyList<InputParameter> GetAllClientInvocatioParameters()
```
ArcturusZhang, line 88: now looks like nothing is using this method, could we remove it?

ArcturusZhang (approved):

PR #5056, opened 2024-09-09
Update Cadl-Ranch-Spec Version 0.37.1

mcgallan:
In the version 0.37.1 update, there was a change to the payload/multipart section. However, during code generation, the multipart section causes errors. Therefore, we need to temporarily remove this part and wait for the issue to be fixed before proceeding. 
Related Issue: https://github.com/Azure/autorest.csharp/issues/5059


ArcturusZhang:
Please add some details of the content of your PR on your title and description.

PR #5068, opened 2024-09-14
Fix for BinaryData as InputExampleObjectValue during sample generation

live1206:
Fix for https://teams.microsoft.com/l/message/19:7b87fb348f224b37b6206fa9d89a105b@thread.skype/1726280284865?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1726058282966&teamName=Azure%20SDK&channelName=Language%20-%20DotNet&createdTime=1726280284865

Since we are working on the implementation of Azure plugin, I just proved locally that it fixes the sample generation for azurestackhci, I am not adding a new test case for swagger input in autorest.csharp.

### DIFF
--- a/src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
+++ b/src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
@@ -107,15 +107,15 @@ namespace AutoRest.CSharp.MgmtTest.Extensions
 
             if (type.FrameworkType == typeof(BinaryData))
             {
-                if (exampleValue is InputExampleObjectValue objectValue)
+                switch (exampleValue)
                 {
-                    return writer.AppendBinaryData(objectValue);
+                    case InputExampleObjectValue objectValue:
+                        return writer.AppendBinaryData(objectValue);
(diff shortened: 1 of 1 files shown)


src/AutoRest.CSharp/MgmtTest/Extensions/CodeWriterExtensions.cs
```diff
@@ -106,7 +106,17 @@ private static CodeWriter AppendFrameworkTypeValue(this CodeWriter writer, CShar
...
 
             if (type.FrameworkType == typeof(BinaryData))
-                return writer.AppendBinaryData((InputExampleRawValue)exampleValue);
+            {
+                switch (exampleValue)
+                {
+                    case InputExampleObjectValue objectValue:
+                        return writer.AppendBinaryData(objectValue);
+                    case InputExampleRawValue rawValue:
+                        return writer.AppendBinaryData((InputExampleRawValue)exampleValue);
```
ArcturusZhang, line 115:
```suggestion
                        return writer.AppendBinaryData(rawValue);
```

ArcturusZhang (approved):

PR #5077, opened 2024-09-19
Skip operations_list by CrossLanguageDefinitionId

live1206:


src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputClientTransformer.cs
```diff
@@ -22,9 +22,8 @@ private static void RemoveInternalOperationFromInputClient(InputClient inputClie
             var operationsToKeep = new List<InputOperation>();
             foreach (var operation in inputClient.Operations)
             {
-                // "internal" is set only we set access decorator in typespec, default null represents public
-                // TODO: Skip internal operations for Mgmt, we might need a better way to remove operations, tracking in https://github.com/Azure/typespec-azure/issues/964
-                if (operation.Accessibility == "public")
+                // operations_list has been covered in Azure.ResourceManager already, we don't need to generate it in the client
+                if (operation.CrossLanguageDefinitionId != "Azure.ResourceManager.Operations.list")
```
ArcturusZhang, line 26: do you think we need to keep both the mechanisms?
Also do our generation honor the accessibility?
If the `operations_list` operation is marked as internal, this operation should be generated as internal and all its related models should also be internal - and linters should not report those.

ArcturusZhang (approved):

src/AutoRest.CSharp/Common/Input/InputTypes/Transformation/InputClientTransformer.cs
live1206: I don't think we need both.
We can remove the internal accessibility from the spec, I modified the MgmtTypeSpec test to reflect the change.
You can see that, the operation and corresponding models are not generated in SDK code.

live1206: And once this is merged, I don't think we need https://github.com/Azure/typespec-azure/issues/964 anymore.

PR #5088, opened 2024-09-25
Bump version for TCGC and typespec-arm template

live1206:


ArcturusZhang (approved):
Be sure to bump to the official emitter version after https://github.com/microsoft/typespec/pull/4522 is merged

PR #5094, opened 2024-09-27
feat(generator): flatten model properties

### LINKED ISSUE #4449: [cadl-ranch] add test for cadl-ranch project
Add test for cadl-ranch project:

```[tasklist]
### Tasks
- [x] azure/core/basic
- [x] client/name
- [ ] client/structure
- [x] encode/bytes
- [ ] type/model/inheritance
```

### LINKED ISSUE #4788: Move the flatten handling logic to the generator
Now in the property we get from TCGC in the emitter, we only get a flag `flatten` to show if this has been flattened by adding the decorator `@flatten`
We need to flatten the property somewhere and put those flattened properties inside this model.
Currently this happens in the emitter.

With the principal that we need to keep the emitter as simple as possible, we should remove it from the emitter.
Two approaches:
1. TCGC can do this so that every emitter does not have to do this again.
2. The emitter just emit out whatever it gets from TCGC, and the generator flattens the property when deserializing the output of emitter.

archerzz:
# Description

This commit will flatten the model property in generator, instead of in the emitter.

- Modify `InputModelProperty`
  - add an `IsFlatten` property for original emitter input
  - remove `FlattenNames` property from constructor
- change model json converter:
  - convert `TypeSpecInputModelProperty` to `InputModelProperty`
  - flatten model property if necessary
- refactor

resolve #4788 

# Checklist

# Ready to Land?


archerzz:
Regen preview: https://github.com/Azure/azure-sdk-for-net/pull/46340

No code change

src/AutoRest.CSharp/Common/Input/CodeModelConverter.cs
```diff
@@ -436,8 +436,10 @@ private InputModelProperty GetOrCreateProperty(Property property)
                 ConstantValue: property.Schema is ConstantSchema constantSchema ? CreateConstant(constantSchema, constantSchema.Extensions?.Format, property.IsNullable) : null,
                 IsRequired: property.IsRequired,
                 IsReadOnly: property.IsReadOnly,
-                IsDiscriminator: property.IsDiscriminator ?? false,
-                FlattenedNames: property.FlattenedNames.ToList());
+                IsDiscriminator: property.IsDiscriminator ?? false)
+            {
```
chunyu3, line 440: set the `IsFlattened` also. otherwise the property value may not consistent. e.g. FlattenedNames is not empty, but the `IsFlattened` is false.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -135,58 +135,65 @@ private static InputModelType CreateInputModelTypeInstance(string? id, string? n
...
-                                                    property.ConstantValue,
-                                                    property.IsRequired,
-                                                    property.IsReadOnly,
-                                                    property.IsDiscriminator,
-                                                    property.FlattenedNames);
+                var flattenedProperties = FlattenPropertyIfNecessary(rawProperty);
+
+                foreach (var property in flattenedProperties)
+                {
+                    properties.Add(isMultipartType ? ConvertMultipartProperty(property) : property);
```
chunyu3, line 158: MultipartConvertor is only apply on the first-level property, it should not apply on flattened property of a model.

chunyu3, line 158: Convert-Bytes-To-Stream-For-Multipart is only apply on the first-level property, it should not apply on flattened property of a model.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -135,58 +135,65 @@ private static InputModelType CreateInputModelTypeInstance(string? id, string? n
             return model;
         }
 
-        private static void CreateProperties(ref Utf8JsonReader reader, ICollection<InputModelProperty> properties, JsonSerializerOptions options, bool isMultipartType)
+        private static void CreateProperties(ref Utf8JsonReader reader, ICollection<InputModelProperty> properties, JsonSerializerOptions options, string modelName, bool isMultipartType)
```
chunyu3, line 138: Is parameter `modelName` only used in a log? If so, shall we add this parameter?

archerzz: It's used to compose exception message to make things easier to debug.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
archerzz: changed the logic to only flatten non-multipart model properties

chunyu3: I mean that we only apply Convert-Bytes-To-Stream-For-Multipart to the property of the model itself, not apply to the properties which is flattened out from other property.

archerzz: When we flatten property, it only starts from the model's properties.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -135,58 +135,74 @@ private static InputModelType CreateInputModelTypeInstance(string? id, string? n
...
+                    var flattenedProperties = FlattenProperty(rawProperty);
+
+                    foreach (var property in flattenedProperties)
+                    {
+                        properties.Add(property);
+                    }
+                }
+                else
+                {
+                    properties.Add(isMultipartType ? ConvertMultipartProperty(rawProperty) : rawProperty);
```
ArcturusZhang, line 170: here we have an issue about the optionality of the properties.
In modeler v4, if the `properties` is optional, all the flattened properties will be optional, but here we are not doing anything with their optionality.
This issue exists before in the old flatten properties code in the emitter, but we did not notice until today.

ArcturusZhang: it does not have impacts right now (or maybe but we did not notice), because in almost all typespec input MPG specs, the flatten is not enabled. But we need to fix this.

pshao25: We discussed offline. If the flattened property is optional, we prefer to make all the properties in that property optional.

archerzz: It's easy to fix the property definition, but it's not easy to fix our serialization logic. I think it would be better to resolve it in another issue.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
archerzz: after discussion, we decided to remove the check on the property of both flattened and request is multi-part.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
archerzz: filed https://github.com/Azure/autorest.csharp/issues/5101 for follow-up discussion

ArcturusZhang (approved):

Issue #5097, opened 2024-10-08
Generator should generate the method reference with full signature when there is customization code for the `CreateResourceIdentifier` method


ArcturusZhang:
When we need to mitigate breaking changes, we might need to add `CreateResourceIdentifier` with different signature in customized code.
But some other methods may reference this method by `XXXResource.CreateResourceIdentifier`, this cannot compile when there are customized methods.
The generator should generate the full signature when this happens.

PR #5100, opened 2024-10-09
Fix the `CreateResourceIdentifier` issue

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/5097

# Description

We have reference of the `CreateResourceIdentifier` method in xml doc in some of our generated classes. This could lead to compilation error if there are overloads written in customization code to mitigate breaking changes (the code will not compile because the reference `XXXResource.CreateResourceIdentifier` is ambiguous).

This PR fixes this by including the reference with its full signature when there is a method with the same name in its corresponding customization code.

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Mgmt/Output/Resource.cs
+++ b/src/AutoRest.CSharp/Mgmt/Output/Resource.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Text;
 using AutoRest.CSharp.Common.Input;
 using AutoRest.CSharp.Common.Output.Expressions.KnownValueExpressions;
@@ -601,7 +602,25 @@ namespace AutoRest.CSharp.Mgmt.Output
                 return $"{Type}.{CreateResourceIdentifierMethod.Signature.Name}";
             }
 
-            return $"";
+            var parameters = CreateResourceIdentifierMethod.Signature.Parameters;
+            var formatBuilder = new StringBuilder("{0}.")
+                .Append(CreateResourceIdentifierMethod.Signature.Name)
+                .Append("(");
+            var arguments = new List<object?>(parameters.Count + 1)
+            {
+                Type
+            };
+            for (int i = 0; i < parameters.Count; i++)
+            {
+                formatBuilder.Append($"{{{i + 1}}}"); // +1 because we should start from 1
+                if (i != parameters.Count - 1)
+                {
+                    formatBuilder.Append(",");
+                }
+                arguments.Add(parameters[i].Type);
+            }
+            formatBuilder.Append(")");
+            return FormattableStringFactory.Create(formatBuilder.ToString(), [.. arguments]);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -585,5 +587,40 @@ public ValueExpression ResourceDataIdExpression(ValueExpression dataExpression)
...
+            var formatBuilder = new StringBuilder("{0}.")
+                .Append(CreateResourceIdentifierMethod.Signature.Name)
+                .Append("(");
+            var arguments = new List<object?>(parameters.Count + 1)
+            {
+                Type
+            };
+            for (int i = 0; i < parameters.Count; i++)
+            {
+                formatBuilder.Append($"{{{i + 1}}}"); // +1 because we should start from 1
```
live1206, line 615:
```suggestion
                formatBuilder.Append($"{{{i + 1}}}"); // +1 because the first argument is set to Resource Type
```

src/AutoRest.CSharp/Mgmt/Output/Resource.cs
```diff
@@ -585,5 +587,40 @@ public ValueExpression ResourceDataIdExpression(ValueExpression dataExpression)
 
         public Parameter ResourceParameter => new(Name: "resource", Description: $"The client parameters to use in these operations.", Type: typeof(ArmResource), DefaultValue: null, ValidationType.None, null);
         public Parameter ResourceDataParameter => new(Name: "data", Description: $"The resource that is the target of operations.", Type: ResourceData.Type, DefaultValue: null, ValidationType.None, null);
+
+        private FormattableString? _createResourceIdentifierReference;
+        public FormattableString CreateResourceIdentifierReference => _createResourceIdentifierReference ??= BuildCreateResourceIdentifierMethodReference();
+
+        private FormattableString BuildCreateResourceIdentifierMethodReference()
+        {
+            // see if there is method with same name in customization code
```
live1206, line 596:
```suggestion
            // see if there is method naming "CreateResourceIdentifier" in customization code
```

live1206 (approved):
Wondering if we should have a general solution for this kind of method invoke in the xml doc? Not just for this case.

ArcturusZhang:
> Wondering if we should have a general solution for this kind of method invoke in the xml doc? Not just for this case.

We could have it in the MGC maybe? It is possible, but here we did not have a system for it, therefore we have to implement them individually.

live1206:
> > Wondering if we should have a general solution for this kind of method invoke in the xml doc? Not just for this case.
> 
> We could have it in the MGC maybe? It is possible, but here we did not have a system for it, therefore we have to implement them individually.

Yes, I was thinking about Azure Plugin, not here.
Could you create an issue in MGC for this? Thanks.

ArcturusZhang:
> > > Wondering if we should have a general solution for this kind of method invoke in the xml doc? Not just for this case.
> > 
> > 
> > We could have it in the MGC maybe? It is possible, but here we did not have a system for it, therefore we have to implement them individually.
> 
> Yes, I was thinking about Azure Plugin, not here. Could you create an issue in MGC for this? Thanks.

Created one: https://github.com/Azure/azure-sdk-for-net/issues/46529

archerzz (approved):

PR #5154, opened 2024-11-06
Remove `use-write-core` configuration

welovej:
Resolves https://github.com/Azure/autorest.csharp/issues/5155


samples/AnomalyDetector/src/Generated/Models/AlignPolicy.Serialization.cs
```diff
@@ -19,14 +19,22 @@ public partial class AlignPolicy : IUtf8JsonSerializable, IJsonModel<AlignPolicy
         void IUtf8JsonSerializable.Write(Utf8JsonWriter writer) => ((IJsonModel<AlignPolicy>)this).Write(writer, ModelSerializationExtensions.WireOptions);
 
         void IJsonModel<AlignPolicy>.Write(Utf8JsonWriter writer, ModelReaderWriterOptions options)
```
pshao25, line 21: Remember delete all the configuration i, e.g., test\CadlRanchProjects\type\model\inheritance\enum-discriminator\tspconfig.yaml

welovej: I did a global research in vscode and there is only one 'use-write-core: true' left.

live1206 (approved):

live1206:
The change LGTM, let's verify the regen in azure-sdk-for-net before merging.

ArcturusZhang (approved):

PR #5157, opened 2024-11-11
Try to remove `example-dir` config in cadl-ranch case

ArcturusZhang:
# Description

https://github.com/Azure/autorest.csharp/pull/5153/files#r1835912319

I think we should be able to omit this config, and TCGC should find the location of those example files.
But somehow when we remove this config, it does not work.

# Checklist

# Ready to Land?


ArcturusZhang:
So it turns out, tsp always use the location of `tspconfig.yaml` as `project-root` therefore TCGC cannot find the examples.
Maybe there is no way of omitting that config. Closing this PR.

Issue #5161, opened 2024-11-14
Optional operation parameter Content-Type with default constant value is not handled properly


live1206:
https://teams.microsoft.com/l/message/19:7b87fb348f224b37b6206fa9d89a105b@thread.skype/1731574353651?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1729143576400&teamName=Azure%20SDK&channelName=Language%20-%20DotNet&createdTime=1731574353651

There an optional operation parameter Content-Type with constant value "application/json", the generator doesn't parse the constant value and caused compile error.

PR #5162, opened 2024-11-14
Parse constant for optional operation paramter as well

live1206:
Resolves https://github.com/Azure/autorest.csharp/issues/5161


test/TestProjects/MgmtTypeSpec/PrivateEndpointConnection.tsp
```diff
@@ -18,4 +18,27 @@ interface PrivateLinks {
...
+}
+
+/**
+ * Start SAP instance(s) request body.
+ */
+model StartRequest {
+  /**
+   * The boolean value indicates whether to start the virtual machines before starting the SAP instances.
+   */
+  startVm?: boolean = false;
```
ArcturusZhang, line 43: i am fine with the fixes, but our principal is decided not to support client default values.
To double confirm, would this `= false` contribute anything to the fix?

ArcturusZhang (approved):

test/TestProjects/MgmtTypeSpec/PrivateEndpointConnection.tsp
live1206: This has nothing to do with the fix, copied from the user spec. Removed now.

PR #5204, opened 2024-12-20
Migrate cadl-ranch specs to http-specs and azure-http-specs

pshao25:
This is the azure part of https://github.com/Azure/azure-sdk-for-net/issues/47325, and fix https://github.com/Azure/azure-sdk-for-net/issues/47325

We split previous cadl-ranch tests into two parts: pure Typespec: [http-specs](https://www.npmjs.com/package/@typespec/http-specs), and Azure Typespec: [azure-http-specs](https://www.npmjs.com/package/@azure-tools/azure-http-specs).

Now, 
1. `CadlRanchProjects` still means the code is generated under azure mode, no matter the spec is from http-specs or azure-http-specs. 
2. `CadlRanchProjectsNonAzure` still means the code is generated under non-azure mode, no matter the spec is from http-specs or azure-http-specs.
3. Run tests under `CadlRanchProjects.Tests` will generate reports for both azure and non-azure.
4. Run tests under `CadlRanchProjectsNonAzure.Tests` will generate reports for both azure and non-azure.
5. In the end, only azure report under `CadlRanchProjects.Tests` will be published to dashboard.


eng/testProjects.json
```diff
@@ -42,7 +42,6 @@
         "payload/content-negotiation",
         "payload/media-type",
         "payload/multipart",
-        "payload/pageable",
         "serialization/encoded-name/json",
```
pshao25, line 45: This spec is non-azure paging, should not be in azure tests.

eng/Generate.ps1
```diff
@@ -70,14 +71,42 @@ function Add-CadlRanch-TypeSpec([string]$testName, [string]$projectPrefix, [stri
...
+        $entryFile = $clientTsp
+        if (!(Test-Path $entryFile)) {
+            $entryFile = $mainTsp
+        }
+        if (!(Test-Path $entryFile)) {
+            $entryFile = $azureClientTsp
+        }
+        if (!(Test-Path $entryFile)) {
+            $entryFile = $azureMainTsp
+        }
```
live1206, line 90: Why needs to fall back to azure tsp for non-azure projects?

eng/Generate.ps1
```diff
@@ -70,14 +71,42 @@ function Add-CadlRanch-TypeSpec([string]$testName, [string]$projectPrefix, [stri
...
+        $entryFile = $azureClientTsp
+        if (!(Test-Path $entryFile)) {
+            $entryFile = $azureMainTsp
+        }
+        if (!(Test-Path $entryFile)) {
+            $entryFile = $clientTsp
+        }
+        if (!(Test-Path $entryFile)) {
+            $entryFile = $mainTsp
+        }
```
live1206, line 102: why needs to fall back to non-azure tsp for azure projects?

eng/testProjects.json
live1206: is this relocated to somewhere else?

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/BytesProperty.cs
```diff
@@ -74,7 +74,7 @@ internal BytesProperty()
         }
 
         /// <summary> Required property. </summary>
-        public string RequiredProperty { get; }
+        public string RequiredProperty { get; set; }
```
live1206, line 77: what is causing the generation change?

eng/testProjects.json
pshao25: Not sure. It's out of this RP's scope. If any tests missing, dashboard will report then.

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/BytesProperty.cs
pshao25: The client.tsp is removed. https://github.com/microsoft/typespec/issues/5399

This change also causes some tests fail. I added the ignore in those tests.

eng/Generate.ps1
pshao25: https://github.com/microsoft/typespec/issues/5400

Previous test specs are split into azure tsp and non-azure tsp. Therefore, some existing tests are generated from non-azure tsp. I don't want to delete these tests, so add the fall back logic here.

test/CadlRanchProjects/type/property/nullable/src/Generated/Models/BytesProperty.cs
ArcturusZhang: could we create an issue to track these failures? later might our vendors could take a look

pshao25: https://github.com/Azure/autorest.csharp/issues/5205

PR #5212, opened 2025-01-06
feat(generator): use InputOperation.OperationId

### LINKED ISSUE #5117: Operation ID changed after tsp 0.61
## Description

In tsp 0.60 and prior, [getResourceOperation](https://github.com/microsoft/typespec/blob/64f1dbae9240010cf3797357c16f14fd6fb04164/packages/http-client-csharp/emitter/src/lib/operation-converter.ts#L75) will return undefined for some mgmt cases, like https://github.com/Azure/autorest.csharp/blob/ab56c7d7cf3ac1b60bab56523e653f929e34e757/test/TestProjects/MgmtTypeSpec/foo.tsp#L10-L15

Image

So, previous the `ResourceName` is `Foos` which is the interface name: https://github.com/Azure/autorest.csharp/blob/ab56c7d7cf3ac1b60bab56523e653f929e34e757/test/TestProjects/MgmtTypeSpec/foo.tsp#L36

But in 0.61, `getResourceOperation` can return the resource operation, which seems the correct behavior. However, that will return different `ResourceName`. Check this PR: https://github.com/Azure/autorest.csharp/pull/5110/files#diff-81dc15d9aacc476e7f1fc431451c064ac2727e5c09cde422e3ddc362b7b57981L2714

I think the upstream actually has fixed an issue, but unfortunately cause regression in some cases.

archerzz:
# Description

Previously we use resource name to guess operation id, which doesn't work well in some cases.
Instead of gussing operation id in generator, we try best to get the id in emitter and pass it to generator.

resolve #5117

# Checklist

# Ready to Land?


archerzz:
this is just a duplicate of #5110 due to pipeline error when pushing changes to github.

PR #5227, opened 2025-01-30
add autorest emitter options

jorgerangel-msft:
Follow up to https://github.com/microsoft/typespec/pull/5803, which removed these autorest specific options from the unbranded emitter. This PR adds those options into the autorest emitter.

### DIFF
--- a/src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
+++ b/src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
@@ -67,12 +67,77 @@ export async function $onEmit(context: EmitContext<AzureNetEmitterOptions>) {
                     .replaceAll("\\", "/")
             );
         }
-        configurations["shared-source-folders"] = resolvedSharedFolders ?? [];
+
+        if ("head-as-boolean" in options) {
+            configurations["head-as-boolean"] = options["head-as-boolean"];
+        }
+
+        configurations["deserialize-null-collection-as-null-value"] =
+            options["deserialize-null-collection-as-null-value"];
         configurations["flavor"] =
             options["flavor"] ??
             (configurations.namespace.toLowerCase().startsWith("azure.")
                 ? "azure"
                 : undefined);
+
+        //only emit these if they are not the default values
+        if ("generate-sample-project" in options) {
+            configurations["generate-sample-project"] =
+                options["generate-sample-project"] === true
+                    ? undefined
+                    : options["generate-sample-project"];
+        }
+
+        if ("generate-test-project" in options) {
+            configurations["generate-test-project"] =
+                options["generate-test-project"] === false
+                    ? undefined
+                    : options["generate-test-project"];
+        }
+
+        configurations["use-model-reader-writer"] = true;
+
+        if ("use-model-reader-writer" in options) {
+            configurations["use-model-reader-writer"] =
+                options["use-model-reader-writer"];
+        }
+
+        if ("single-top-level-client" in options) {
+            configurations["single-top-level-client"] =
+                options["single-top-level-client"];
+        }
+
+        if ("keep-non-overloadable-protocol-signature" in options) {
+            configurations["keep-non-overloadable-protocol-signature"] =
+                options["keep-non-overloadable-protocol-signature"];
+        }
+
+        configurations["models-to-treat-empty-string-as-null"] =
+            options["models-to-treat-empty-string-as-null"];
+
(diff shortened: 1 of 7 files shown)


JoshLove-msft (commented):
Do we have a azure-sdk-for-net regen PR against these changes?

jorgerangel-msft:
> Do we have a azure-sdk-for-net regen PR against these changes?

@JoshLove-msft yes - https://github.com/Azure/azure-sdk-for-net/pull/48041

JoshLove-msft:
> > Do we have a azure-sdk-for-net regen PR against these changes?
> 
> @JoshLove-msft yes - [Azure/azure-sdk-for-net#48041](https://github.com/Azure/azure-sdk-for-net/pull/48041)

It looks like the regen failed for some libraries https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4525285&view=logs&j=d731b843-2b38-5e88-bb49-012fa10bc295

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -67,12 +67,77 @@ export async function $onEmit(context: EmitContext<AzureNetEmitterOptions>) {
...
+            configurations["use-model-reader-writer"] =
+                options["use-model-reader-writer"];
+        }
+
+        if ("single-top-level-client" in options) {
+            configurations["single-top-level-client"] =
+                options["single-top-level-client"];
+        }
+
+        if ("keep-non-overloadable-protocol-signature" in options) {
```
JoshLove-msft, line 110: Why do we sometimes check for the presence of the key, but other times don't? Does `configurations` have a default value that we don't want to override?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -67,12 +67,77 @@ export async function $onEmit(context: EmitContext<AzureNetEmitterOptions>) {
                     .replaceAll("\\", "/")
             );
         }
-        configurations["shared-source-folders"] = resolvedSharedFolders ?? [];
+
+        if ("head-as-boolean" in options) {
```
JoshLove-msft, line 71: Could we do something like this instead:

```ts
configurations["head-as-boolean"] = options["head-as-boolean"] ?? configurations["head-as-boolean"];
```

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -67,12 +67,77 @@ export async function $onEmit(context: EmitContext<AzureNetEmitterOptions>) {
...
+                      ].filter(
+                          (item) =>
+                              options[
+                                  "additional-intrinsic-types-to-treat-empty-string-as-null"
+                              ].indexOf(item) < 0
+                      )
+                  )
+                : undefined;
+
+        configurations["methods-to-keep-client-default-value"] =
```
JoshLove-msft, line 137: Do we need to respect the configurations value here?

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
jorgerangel-msft: `options` is just an object of `any` type so typescript complains since some of these properties didn't have explicit default values and it can't validate if they exist in the object or not. As a workaround, I'll give these options a default value (undefined) so we can just assign them without doing this extra check.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
jorgerangel-msft: I don't think so since configurations is the parsed JSON of the emitted file produced by the mgc emitter, and that option was removed.

JoshLove-msft (approved):

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
jorgerangel-msft: Discussed offline, the options is of `any` type and strict usage is configured so this is currently not possible. We will opt to explicitly defining the possible options in the schema.

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
jorgerangel-msft: Discussed offline, the options is of `any` type and strict usage is configured so this is currently not possible. We will opt to explicitly defining the possible options in the schema.

PR #5236, opened 2025-02-13
Add namespace/library-name options and decouple from MGC onEmit function

JoshLove-msft:
- Add namespace option into autorest emitter
- Add library-name option into autorest emitter
- Default library-name config to CodelModel.Name if not specified in option.library-name.
- Export CodeModel and CSharpEmitterContext from http-client-csharp
- Write the CodeModel in Autorest emitter
- No longer call into http-client-csharp OnEmit function.
- Remove skipSDKGeneration as we no longer need it


src/AutoRest.CSharp/package.json
```diff
@@ -6,11 +6,11 @@
   "scripts": {
     "start": "dotnet --roll-forward Major ./AutoRest.CSharp.dll --server"
   },
-  "devDependencies": {
+  "dependencies": {
+    "@autorest/csharp": "file:"
```
ArcturusZhang, line 10: this might be the issue, we should remove this change.

package.json
```diff
@@ -5,20 +5,23 @@
...
     "@azure-tools/typespec-azure-resource-manager": "0.50.0",
+    "@azure-tools/typespec-client-generator-core": "0.50.2",
     "@azure-tools/typespec-csharp": "file:src/TypeSpec.Extension/Emitter.Csharp",
     "@azure-tools/unbranded-tests": "file:test/UnbrandedProjects",
     "@microsoft.azure/autorest.testserver": "3.3.50",
+    "@typespec/http-specs": "0.1.0-alpha.7",
     "autorest": "3.7.1"
+  },
+  "dependencies": {
+    "@autorest/csharp": "file:"
```
ArcturusZhang, line 25: same here

test/UnbrandedProjects/package.json
```diff
@@ -3,7 +3,9 @@
     "private": true,
     "description": "Preserve the dependencies for unbranded tests.",
     "devDependencies": {
-      "@typespec/openapi": "0.64.0"
+        "@typespec/openapi": "0.64.0"
+    },
+    "dependencies": {
+        "@azure-tools/unbranded-tests": "file:"
```
ArcturusZhang, line 9: this is also unexpected

src/TypeSpec.Extension/Emitter.Csharp/src/emitter.ts
```diff
@@ -38,22 +40,36 @@ export async function $onEmit(context: EmitContext<AzureNetEmitterOptions>) {
...
     );
     const isSrcFolder = path.basename(outputFolder) === "src";
     const generatedFolder = isSrcFolder
         ? resolvePath(outputFolder, "Generated")
         : resolvePath(outputFolder, "src", "Generated");
-    if (options.skipSDKGeneration !== true) {
+    if (root) {
+        if (!fs.existsSync(generatedFolder)) {
+            fs.mkdirSync(generatedFolder, { recursive: true });
+          }
```
ArcturusZhang, line 59: since we are not using the `fs` from the system to write the file, we called typespec compiler's hook, do we still need to create those directories? I think typespec compiler should be able to handle it

JoshLove-msft: not sure I follow this is the generated folder directory. TypeSpec compiler wouldn't create this.

m-nash (approved):
LGTM remember to remove the temp change before merging here https://github.com/Azure/autorest.csharp/pull/5236/files#diff-c405fe2c565a2a5a238c4f06b415cff96bcc6e6d38070fa0a64bbf963edf835bR38

PR #5242, opened 2025-02-21
Set MaxDepth to 256 for JsonDocument.Parse and regen tests

live1206:
We did some performance test that setting MaxDepth to a higher number for JsonDocumentOption doesn't affect the performance of the deserialization. 
The memory was allocated dynamically starting with default value, so with the data not exceeding the default MaxDepth, the performance remains the same.
Hence, we decided to update the default value of MaxDepth to 256 for JsonDocument.Parse for SDK generation.

Resolves https://github.com/Azure/azure-sdk-for-net/issues/48260


samples/AnomalyDetector/src/Generated/Docs/Multivariate.xml
```diff
@@ -33,7 +33,7 @@ Multivariate client = new AnomalyDetectorClient(endpoint, credential).GetMultiva
 
 Response response = await client.GetMultivariateBatchDetectionResultAsync(Guid.Parse("663884e6-b117-11ea-b3de-0242ac130004"), null);
 
-JsonElement result = JsonDocument.Parse(response.ContentStream).RootElement;
+JsonElement result = JsonDocument.Parse(response.ContentStream, new JsonDocumentOptions { MaxDepth = 256 }).RootElement;
```
ArcturusZhang, line 36: I think we do not really to update all the samples - samples are just showing how we could parse the result and get those properties, we do not have to show every detail

samples/AnomalyDetector/src/Generated/Models/AlignPolicy.Serialization.cs
```diff
@@ -58,7 +58,7 @@ protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWrit
 #if NET6_0_OR_GREATER
 				writer.WriteRawValue(item.Value);
 #else
-                    using (JsonDocument document = JsonDocument.Parse(item.Value))
+                    using (JsonDocument document = JsonDocument.Parse(item.Value, new JsonDocumentOptions { MaxDepth = 256 }))
```
ArcturusZhang, line 61: could we add a `static readonly` member of this JsonDocumentOptions in the `ModelSerializationExtensions` class so that here we could use it as a reference?

live1206: sounds good

samples/AnomalyDetector/src/Generated/Docs/Multivariate.xml
live1206: Let's discuss this with a wider group to decide what should the sample behave.

samples/AnomalyDetector/src/Generated/Models/AlignPolicy.Serialization.cs
live1206: updated

samples/AnomalyDetector/src/Generated/Docs/Multivariate.xml
live1206: removed

ArcturusZhang (approved):
LGTM

m-nash (approved):

Issue #5243, opened 2025-02-25
Generate incorrect SubResource property


ArthurMa1978:
The issue was happened on TypeSpec, which there is model contains a sub-resource as bellow  (spec is [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/servicenetworking/ServiceNetworking.Management/main.tsp#L89))

``` JavaScript
/** The updatable properties of the Association. */
model AssociationUpdateProperties {
  /** Association Type */
  associationType?: AssociationType;

  /** Association Subnet */
  subnet?: AssociationSubnetUpdate;
}
```

The `subnet` is a property in a patch content which the access attribute is `input`, then the generate code is:
``` C#
/// <summary> Association Subnet. </summary>
internal SubResource Subnet { get; set; }
/// <summary> Gets Id. </summary>
public ResourceIdentifier SubnetId
{
    get => Subnet is null ? default : Subnet.Id;
}
```
Unfortunately the `Id` in `SubResource` is protected which can't be public access, so developer here can't modify the `subnet` property in the patch operation which is incorrect.

BTW, it generates the code from the swagger spec of this RP, the code is:
``` C#
/// <summary> Association Subnet. </summary>
internal WritableSubResource Subnet { get; set; }
/// <summary> Gets or sets Id. </summary>
public ResourceIdentifier SubnetId
{
    get => Subnet is null ? default : Subnet.Id;
    set
    {
        if (Subnet is null)
            Subnet = new WritableSubResource();
        Subnet.Id = value;
    }
}
```

live1206:
The root cause if `isRequired` changed for (AssociationSubnetUpdate .Id)[https://github.com/Azure/azure-rest-api-specs/blob/8607e2d9583bc85039fc50bf46539e98b0f75658/specification/servicenetworking/ServiceNetworking.Management/main.tsp#L95], we need to update tsp to 
```
/** Association Subnet. */
model AssociationSubnetUpdate {
  /** Association ID. */
  id?: string;
}
```

Verified locally.

PR #5246, opened 2025-02-26
Fix for property setter omit

live1206:
Resolves https://github.com/Azure/autorest.csharp/issues/5243

Right now, the property setters will be omitted if the usage doesn't have Input but output.
It should be the other way around, if the usage only has output not input, the setter should be omitted.


src/AutoRest.CSharp/Common/Output/Models/Types/SchemaObjectType.cs
```diff
@@ -565,8 +565,8 @@ property.Type is not InputNullableType &&
             {
                 // In mixed models required properties are not readonly
                 propertyShouldOmitSetter |= property.IsRequired &&
-                              _usage.HasFlag(InputModelTypeUsage.Input) &&
-                              !_usage.HasFlag(InputModelTypeUsage.Output);
+                              !_usage.HasFlag(InputModelTypeUsage.Input) &&
+                              _usage.HasFlag(InputModelTypeUsage.Output);
```
ArcturusZhang, line 569: in ModelTypeProvider, did we do the same?
Or let me ask in this way:
Are the behaviors of ModelTypeProvider and SchemaObjectType different in the first place? And now we are just fixing them to be the same

live1206: There is no such logic in ModelTypeProvider, it just takes the readOnly from InputProperty https://github.com/Azure/autorest.csharp/blob/7a6528a102390ace593306bb82025a08ff6465b7/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs#L525

This tweaking is only for mgmt.

ArcturusZhang: well this ctor is taking the `readonly` flag from the `field`, and whether the field has readonly flag comes from here:
https://github.com/Azure/autorest.csharp/blob/7a6528a102390ace593306bb82025a08ff6465b7/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProviderFields.cs#L204
We should ensure these two pieces generate the same thing or if not, we should know the gap.

ArcturusZhang: Also according to Arthur, when generating from swagger, the setter is fine. The issue comes when we change the input to typespec.
As far as I know, when the input is swagger, we use `SchemaObjectType` to generate models, and when the input is typespec, we use `ModelTypeProvider` to generate models.
According to the generated code, swagger-generated code is correct, therefore we should change `ModelTypeProvider` (the code above)

The "safe flatten" part is done in the `ObjectTypeProperty` class therefore it applies to both swagger input and typespec input - it is only enabled by the arm flag.

live1206: >As far as I know, when the input is swagger, we use SchemaObjectType to generate models, and when the input is typespec, we use ModelTypeProvider to generate models.

Not really. For tsp input, we also use `SchemaObjectType`
https://github.com/Azure/autorest.csharp/blob/7a6528a102390ace593306bb82025a08ff6465b7/src/AutoRest.CSharp/Common/AutoRest/Plugins/CSharpGen.cs#L84-L89

live1206: The reason the swagger input was correct is because the usage was input | output.
And the usage of tsp input is input only, without output.

live1206:
Found the root cause here: https://github.com/Azure/autorest.csharp/issues/5243#issuecomment-2690087009

PR #5265, opened 2025-03-17
Honor @armResourceOperations as the flag for resource operation

### LINKED ISSUE #5260: Support request-path-is-non-resource for TypeSpec input
For swagger input we have request-path-is-non-resource config to turn off a resource. We should have equivalent way for TypeSpec input.

pshao25:
Fix #5260

When swagger input we have configuration request-path-is-non-resource to make a resource become a non-resource. In the TypeSpec world, we use decorator `@armResourceOperations` to mark a resource operation. Therefore if an operation doesn't have this operation, it means it is not resource operation.


ArcturusZhang:
I think your current implementation could work in some ways but it does not provide the full functionalities of this configuration.

One of the functionalities for this configuration is that, we could use this configuration to make a "resource" into a non resource, even if it is recognized as resources.
Your current implementation is that we make those interfaces with `@armResourceOperations` decorators to be resource, and those without to be non-resource, looks like a workaround to our resource detection.
If this is expected, we should change the description accordingly.

live1206 (approved):

src/AutoRest.CSharp/Mgmt/Models/OperationSet.cs
```diff
@@ -40,6 +40,18 @@ public OperationSet(string requestPath, InputClient? inputClient)
             Operations = new HashSet<InputOperation>();
         }
 
+        public bool IsNonResourceOperations(InputNamespace? inputNamespace)
+        {
+            // We use decorator @service to tell if it is a TypeSpec input
+            bool isTsp = inputNamespace?.Clients.FirstOrDefault(c => c.Decorators.Any(d => d.Name == "TypeSpec.@service")) != null;
+            // If swagger input, we don't use this operation to check if it is a non resource operation
+            if (!isTsp)
+                return false;
```
ArcturusZhang, line 49: we could add a `IsTypeSpec` flag on `MgmtContext`, which is static so that we could get the result and cache it there.

ArcturusZhang: well we might have a concurrent issue when running test cases if we do this

src/AutoRest.CSharp/Mgmt/Models/OperationSet.cs
```diff
@@ -40,6 +40,18 @@ public OperationSet(string requestPath, InputClient? inputClient)
...
+
+            // If TypeSpec input, and there is no @armResourceOperations, we think it is a non-resource client
+            return _inputClient != null && _inputClient.Decorators?.FirstOrDefault(d => d.Name == "Azure.ResourceManager.@armResourceOperations") == null;
```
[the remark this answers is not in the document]
ArcturusZhang: Emm in swagger we never have this, and it causes everything to be a non resource which is not expected at all

PR #5267, opened 2025-03-18
Adopt the property name changes in `tspCodeModel.json`

ArcturusZhang:
# Description

Adopting the changes from https://github.com/microsoft/typespec/pull/6400

Since the property names of those input types are always different from their json property names, we always need a converter for them to properly deserialize.
This could be inconvenient but I think it might be fine, we should never need to "maintain" them, they would never change once we introduced them.

# Checklist

# Ready to Land?


src/AutoRest.CSharp/Common/Input/Utf8JsonReaderExtensions.cs
```diff
@@ -129,7 +164,7 @@ public static bool TryReadWithConverter<T>(this ref Utf8JsonReader reader, strin
             return true;
         }
 
-        public static bool TryReadWithConverter<T>(this ref Utf8JsonReader reader, string propertyName, JsonSerializerOptions options, ref IReadOnlyList<T>? value)
+        public static bool TryReadComplexType<T>(this ref Utf8JsonReader reader, string propertyName, JsonSerializerOptions options, ref IReadOnlyList<T>? value)
```
live1206, line 167: how about `TryReadComplextArrayType`?

src/AutoRest.CSharp/Common/Input/Utf8JsonReaderExtensions.cs
```diff
@@ -3,14 +3,31 @@
...
 using System.Collections.Generic;
-using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
 namespace AutoRest.CSharp.Common.Input
 {
     internal static class Utf8JsonReaderExtensions
     {
+        public static bool TryReadReferenceId(this ref Utf8JsonReader reader, ref string? id)
```
live1206, line 13: why is this needed? can it be replaced with the below method of `isFirstProperty = true`?

ArcturusZhang: this is only a result of copying code from MTG.
I think it is the other version of this method that we do not need.
In our current usage, we did not really try to verify if `$id` is the first property we met, we just assume this way.

src/AutoRest.CSharp/Common/Input/Utf8JsonReaderExtensions.cs
ArcturusZhang: MTG has it in the same way: https://github.com/microsoft/typespec/blob/586a120eb425cf8aeb383ba76becc8a8fd6efa1d/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/Utf8JsonReaderExtensions.cs#L145
I would say arrays are also complex type, the overload actually should help us to make the code simpler because they just accept the same parameter list.

src/AutoRest.CSharp/Common/Input/Utf8JsonReaderExtensions.cs
live1206: ok, then the duplication is in MTG in the first place. We can clean it up later there instead.

src/AutoRest.CSharp/Common/Input/Utf8JsonReaderExtensions.cs
live1206: well, I think it's kind of confusing for user to choose which method to use based on the ref parameter.
Anyway, this is just a copy from MTG.

live1206 (approved):

src/AutoRest.CSharp/Common/Input/Utf8JsonReaderExtensions.cs
ArcturusZhang: yeah I am planning to remove the one with `isFirstProperty` in a near future

PR #5269, opened 2025-03-18
Support duration-constant

### LINKED ISSUE #5261: Support duration-constant
Refer to https://github.com/Azure/typespec-azure/issues/1901. We don't support duration-constant for now.

pshao25:
Fix #5261


src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputDurationTypeConverter.cs
```diff
@@ -51,7 +51,7 @@ public static InputDurationType CreateDurationType(ref Utf8JsonReader reader, st
             crossLanguageDefinitionId = crossLanguageDefinitionId ?? throw new JsonException("Duration type must have crossLanguageDefinitionId");
             encode = encode ?? throw new JsonException("Duration type must have encode");
             wireType = wireType ?? throw new JsonException("Duration type must have wireType");
-
+            encode = encode == "duration-constant" ? "constant" : encode;
```
live1206, line 54: consider to combine this to L52

ArcturusZhang: we should not do it in this way, it would be hard to track these magic values.

Deserialization should just simply do "deserialization" without modification unless we could prove it is absolutely necessary

Therefore here we should deserialize this as is, and if other parts in our generator is expecting `constant` instead of `duration-constant`, those parts should change accordingly to properly handle `duration-constant`.

pshao25: I try to add something like `TryReadEncoding`, but that cannot keep `throw new JsonException($"Encoding of Duration type {encode} is unknown.");`. So changed by adding a `CreateDurationKnownEncoding`

ArcturusZhang (approved):
@live1206 we might need to consider how this extends in plugins of MTG.
If we represent the encoding in MTG by an enum, it could never be extensible.

src/AutoRest.CSharp/Common/Input/InputTypes/Serialization/TypeSpecInputDurationTypeConverter.cs
live1206: In Azure plugin, we don't have the ability to manipulate deserialization.
So, encoding is a decorator we can get from the model. Can we deal with this after the deserialization with `InputModelType.Decorators`?

pshao25: Created issue here: https://github.com/microsoft/typespec/issues/6621

PR #5274, opened 2025-03-21
add the step that sets node version to 20 in publish pipeline

ArcturusZhang:
# Description

The publish pipeline now cannot properly generate the typespec projects because it does not install node therefore default 18.x version is using and the latest typespec only support >=20 version for node.

# Checklist

# Ready to Land?


eng/pipelines/update-azure-sdk-for-net-codes.yml
```diff
@@ -1,6 +1,9 @@
 # Generate SDK codes
 
 parameters:
+  - name: NodeVersion
```
pshao25, line 4: I feel like we pass the value from outside, so that when it comes to 22.x, we don't need to change two places.

ArcturusZhang: actually we have to change 5 places now if you search `20.x` in this repo.
This is not the priority right now. We could fix it later.

live1206 (approved):

PR #5276, opened 2025-03-24
Refactor the pipeline files to use a global value of node version

ArcturusZhang:
# Description

Now in our pipeline definitions, the value `node version: 20.x` was written in 6 different places.
This PR introduces a global variable to hold it and use the global variable in those places for node version.
Next time when we need to bump the version, we just need to change in one place.

# Checklist

# Ready to Land?


eng/pipelines/variables/globals.yml
```diff
@@ -22,3 +22,7 @@ variables:
     value: windows
   - name: MACOS
     value: macOS
+  
+  # Indicates the node version to use
+  - name: NodeVersion
```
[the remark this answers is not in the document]
live1206: Or, we can let this as it is, and add a variable for this in https://github.com/Azure/autorest.csharp/blob/main/eng/pipelines/1es-redirect.yml

eng/pipelines/variables/globals.yml
```diff
@@ -22,3 +22,7 @@ variables:
     value: windows
   - name: MACOS
     value: macOS
+
```
hallipr, line 25: image.yml is meant to contain only 1es agent image information and should be separate from globals.yml

hallipr: see https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/variables/globals.yml
vs https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/variables/image.yml

eng/pipelines/variables/globals.yml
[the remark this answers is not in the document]
hallipr: per the norm from other repos, we'd put this in a separate file "globals.yml" using proper case and reference "globals.yml" and "image.yml" in each stage that needs them

eng/pipelines/build.yml
```diff
@@ -42,10 +42,10 @@ extends:
                 inputs:
                   useGlobalJson: true
                   performMultiLevelLookup: true
-              - task: NodeTool@0
-                displayName: "Install Node 20.x"
+              - task: UseNode@1
+                displayName: 'Install Node.js'
                 inputs:
-                  versionSpec: '20.x'
+                  version: $(NodeVersion)
```
pshao25, line 48: Personally I like what we did in the specs repo https://github.com/Azure/azure-rest-api-specs/blob/e278de5e90b0fee9cddd99ac3a5b30e1f3f6f137/eng/pipelines/typespec-apiview.yml#L27

eng/pipelines/variables/globals.yml
ArcturusZhang: I separated the node version into a standalone file `globals.yml` and expands it in 1es-redirect.yml like we have in `image.yml`.

live1206 (approved):

hallipr (approved):

PR #5283, opened 2025-04-01
consume decorator deserializeEmptyStringAsNull.

haiyuazhang:
# Description

# Checklist

# Ready to Land?

### DIFF
--- a/src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
+++ b/src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
@@ -885,7 +885,7 @@ namespace AutoRest.CSharp.Common.Output.Builders
         private static BoolExpression GetCheckEmptyPropertyValueExpression(JsonPropertyExpression jsonProperty, JsonPropertySerialization jsonPropertySerialization, bool shouldTreatEmptyStringAsNull)
         {
             var jsonElement = jsonProperty.Value;
-            if (!shouldTreatEmptyStringAsNull)
+            if (!shouldTreatEmptyStringAsNull && !jsonPropertySerialization.DeserializeEmptyStringAsNull)
             {
                 return jsonElement.ValueKindEqualsNull();
             }
--- a/src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonPropertySerialization.cs
+++ b/src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonPropertySerialization.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Linq;
 using AutoRest.CSharp.Common.Output.Expressions.ValueExpressions;
 using AutoRest.CSharp.Generation.Types;
 using AutoRest.CSharp.Output.Models.Types;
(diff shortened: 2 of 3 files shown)


src/AutoRest.CSharp/Common/Output/Builders/JsonSerializationMethodsBuilder.cs
```diff
@@ -885,7 +885,7 @@ private static MethodBodyStatement CreatePropertyNullCheckStatement(JsonProperty
         private static BoolExpression GetCheckEmptyPropertyValueExpression(JsonPropertyExpression jsonProperty, JsonPropertySerialization jsonPropertySerialization, bool shouldTreatEmptyStringAsNull)
         {
             var jsonElement = jsonProperty.Value;
-            if (!shouldTreatEmptyStringAsNull)
+            if (!shouldTreatEmptyStringAsNull && !jsonPropertySerialization.DeserializeEmptyStringAsNull)
```
live1206, line 888: So, `jsonPropertySerialization.DeserializeEmptyStringAsNull` is based on the property decorator, which only deal with TypeSpec inputs.
And `shouldTreatEmptyStringAsNull` is the result from configuration of swagger inputs.
I tend to separate the usage here. If `jsonPropertySerialization.DeserializeEmptyStringAsNull` is true, we should just return L903 regardless the other conditions.
We can add a test case which doesn't meet L898.

In L898,

src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonPropertySerialization.cs
```diff
@@ -40,6 +41,7 @@ public JsonPropertySerialization(
             ValueSerialization = valueSerialization;
             CustomSerializationMethodName = serializationHooks?.JsonSerializationMethodName;
             CustomDeserializationMethodName = serializationHooks?.JsonDeserializationMethodName;
+            DeserializeEmptyStringAsNull = property.InputModelProperty?.Decorators.Any(d => d.Name == "Azure.ClientGenerator.Core.@deserializeEmptyStringAsNull") ?? false;
```
ArcturusZhang, line 44: Could we add an extension method to `InputModelProperty` type in `Common/Input` like:
```
internal static class InputTypeExtensions
{
	public static bool ShouldDeserializeEmptyStringAsNull(this InputModelProperty property)
		=> property.Decorators.Any(d => d.Name == "Azure.ClientGenerator.Core.@deserializeEmptyStringAsNull");
}
```
and here it should be more readable

src/TypeSpec.Extension/Emitter.Csharp/src/sdk-context-options.ts
```diff
@@ -7,6 +7,7 @@ export const azureSDKContextOptions: CreateSdkContextOptions = {
     versioning: {},
     additionalDecorators: [
         "Azure\\.ClientGenerator\\.Core\\.@useSystemTextJsonConverter",
+        "Azure\\.ClientGenerator\\.Core\\.@deserializeEmptyStringAsNull",
         "Azure\\.ResourceManager\\.@armResourceOperations",
```
pshao25, line 11: I feel like we should delete the existing "models-to-treat-empty-string-as-null" in options.ts.

live1206 (approved):

src/AutoRest.CSharp/Common/Output/Models/Serialization/Json/JsonPropertySerialization.cs
haiyuazhang: @ArcturusZhang is it ok to resolve this?

src/TypeSpec.Extension/Emitter.Csharp/src/sdk-context-options.ts
ArcturusZhang: We could do it and know if it impacts any real RPs in our regen.
If it impacts anything, we could do it in a separate PR.

Issue #5237, opened 2025-02-13
Update generator to use new ModelReaderWriter Read/Write overloads


m-nash:
Update the serialization generation to include the new ModelReaderWriterContext instance for the library and use the new overload for all calls to ModelReaderWriter.Read

PR #5297, opened 2025-04-17
Use new mrw overloads

m-nash:
Fixes https://github.com/Azure/autorest.csharp/issues/5237


src/AutoRest.CSharp/Mgmt/AutoRest/MgmtTarget.cs
```diff
@@ -235,7 +236,17 @@ public static async Task ExecuteAsync(GeneratedCodeWorkspace project)
             //if (_overriddenProjectFilenames.TryGetValue(project, out var overriddenFilenames))
             //    throw new InvalidOperationException($"At least one file was overridden during the generation process. Filenames are: {string.Join(", ", overriddenFilenames)}");
 
-            var modelsToKeep = Configuration.MgmtConfiguration.KeepOrphanedModels.ToImmutableHashSet();
+            var emitContext = Configuration.MgmtConfiguration.IsArmCore ? Configuration.Namespace == "Azure.ResourceManager" ? true : false : true;
```
live1206, line 239: curious about the false case here.
Checked in azure-sdk-for-net repo, `arm-core` only occurs in 2 place: 
https://github.com/Azure/azure-sdk-for-net/blob/351f0401db53b6709159a04139cd41af796407df/sdk/resourcemanager/Azure.ResourceManager/src/autorest.md?plain=1#L7
https://github.com/Azure/azure-sdk-for-net/blob/351f0401db53b6709159a04139cd41af796407df/sdk/resourcemanager/Azure.ResourceManager/eng/autorest-resources.md?plain=1#L6
seems like arm-core always comes within `Azure.ResourceManager`

m-nash: arm-core also happens with Azure.ResourceManager.Resources and Azure.ResourceManager.ManagementGroup namespaces inside the Azure.ResourceManager library.  We do a batch generation of 3 libraries into one.

src/AutoRest.CSharp/Mgmt/AutoRest/MgmtTarget.cs
```diff
@@ -235,7 +236,17 @@ public static async Task ExecuteAsync(GeneratedCodeWorkspace project)
...
+            if (emitContext)
```
ArcturusZhang, line 240: well the above trenary expression is quite confusing.
Would it be possible that we change that to a helper method to make it clearer?

src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/ModelSerializationExtensionsProvider.cs
```diff
@@ -51,21 +51,18 @@ public ModelSerializationExtensionsProvider() : base(Configuration.HelperNamespa
                 InitializationValue = New.Instance(typeof(JsonDocumentOptions), new Dictionary<string, ValueExpression>() { { "MaxDepth", Literal(256) } })
             };
 
-            if (Configuration.EnableInternalRawData)
```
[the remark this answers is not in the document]
m-nash: Just for consistency.  In the test projects we need to pull in one projects "internal helpers" so everything can compile and because this property isn't there always it made that more complicated.  Have an extra property doesn't really have a downside so we always add the property now.

This also enables custom code to take advantage of it even if they have the internal raw data feature flag off

live1206 (approved):

PR #5347, opened 2025-06-13
Apply IJsonmodel<> changes from Azure.Core

live1206:
- upgrade Azure.Core and Azure.Core.Expressions.DataFactory to the latest version to apply IJsonModel<> changes from Azure.Core
- get rid of `ManagedServiceIdentityTypeV3Converter.cs` and remove it from shared file download
- remove serialization helper methods from generated `ModelSerializationExtensions.cs`
-  make the generated code AOT compatible, the only exceptions are
   - use JsonSerializer.Serialize for BinaryData when .net version is under .net 6, which is not needed for AOT
    - use JsonSerializer.(De)serialize for DataFactory related types


live1206:
pending on https://github.com/Azure/azure-sdk-for-net/issues/50703

agocke:
Is this ready for review?

live1206:
> Is this ready for review?

The SDK regen kept failing, there is no easy quick fix for now. Considering the priority of this, we will merge this and regen the SDK manually after the merge.

PR #5374, opened 2025-07-25
Simplify generation for ArmOperation.GetOperationId

live1206:


JoshLove-msft (changes requested):
Please close this PR - we don't need to take fixes like this in autorest.

Issue #5375, opened 2025-07-25
Exception thrown by ModelReaderContextWriter


sbomer:
Using the ModelReaderWriter generation mode for Azure.Search.Documents hits a generator bug. This is blocking fixing some AOT warnings in azure-mcp. @JoshLove-msft @m-nash 

```
System.ArgumentException: An item with the same key has already been added. Key: global::Azure.Search.Documents.Indexes.Models.KeywordTokenizer

   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable1 source, Func2 keySelector, Func2 elementSelector, IEqualityComparer1 comparer)
   at AutoRest.CSharp.Common.Generation.Writers.ModelReaderWriterContextWriter.CollectBuildableTypes(IEnumerable`1 models) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Generation\Writers\ModelReaderWriterContextWriter.cs:line 91
   at AutoRest.CSharp.Common.Generation.Writers.ModelReaderWriterContextWriter.Write(CodeWriter writer, IEnumerable`1 models) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\Generation\Writers\ModelReaderWriterContextWriter.cs:line 28
   at AutoRest.CSharp.AutoRest.Plugins.DataPlaneTarget.Execute(GeneratedCodeWorkspace project, InputNamespace inputNamespace, SourceInputModel sourceInputModel) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\DataPlane\AutoRest\DataPlaneTarget.cs:line 95
   at AutoRest.CSharp.AutoRest.Plugins.CSharpGen.ExecuteAsync(CodeModel codeModel) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\AutoRest\Plugins\CSharpGen.cs:line 39
   at AutoRest.CSharp.AutoRest.Plugins.CSharpGen.Execute(IPluginCommunication autoRest) in D:\a\_work\1\s\autorest.csharp\src\AutoRest.CSharp\Common\AutoRest\Plugins\CSharpGen.cs:line 136
[...]

PR #5376, opened 2025-07-25
Don't use C# name in model dictionary

sbomer:
Attempting to fix https://github.com/Azure/autorest.csharp/issues/5375 @m-nash

### DIFF
--- a/eng/pipelines/preview.yml
+++ b/eng/pipelines/preview.yml
@@ -23,7 +23,7 @@ resources:
       type: github
       endpoint: Azure
       name: Azure/azure-sdk-tools
-      ref: refs/pull/11373/head
+      ref: refs/pull/11383/head
 
 pool:
   name: azsdk-pool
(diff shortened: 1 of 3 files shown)


eng/pipelines/preview.yml
```diff
@@ -23,7 +23,7 @@ resources:
       type: github
       endpoint: Azure
       name: Azure/azure-sdk-tools
-      ref: refs/heads/main
+      ref: refs/pull/11383/head
```
JoshLove-msft, line 26:
```suggestion
      ref: refs/heads/main
```

JoshLove-msft (approved):

live1206:
> should we expect some MRWContext change in the regen PR [Azure/azure-sdk-for-net#51536](https://github.com/Azure/azure-sdk-for-net/pull/51536)?

Never mind, the regen is supposed to be the same, I verified the fix with local regen as well.

